PR #113, opened 2018-02-23
Fix an intermittent problem when scale down multiple times back to back, and typos in wercker.yml

doxiao:
Fix an intermittent issue in scale down. 

Problem: When two scale down operations come together, the first is cancelled, and the second one is suspended waiting for the cancelation of the first one forever.

Cause:  Fiber's resume method only handles NOT_COMPLETED state, but a cancelled fiber is in CANCALLED state.

Solution: Add code to handle CANCELLED state in resume.

In addition, fix string substitutions in wercker.yml to correctly handle "/" in branch names.

### DIFF
--- a/src/main/java/oracle/kubernetes/operator/work/Fiber.java
+++ b/src/main/java/oracle/kubernetes/operator/work/Fiber.java
@@ -231,6 +231,10 @@ public final class Fiber implements Runnable, Future<Void>, ComponentRegistry {
       } finally {
         lock.unlock();
       }
+    } else if (status.get() == CANCELLED) {
+      if (--suspendedCount == 0) {
+        completionCheck();
+      }
     }
   }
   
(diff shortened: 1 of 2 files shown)


src/main/java/oracle/kubernetes/operator/work/Fiber.java
```diff
@@ -231,6 +231,15 @@ public void resume(Packet resumePacket, CompletionCallback callback) {
       } finally {
         lock.unlock();
       }
+    } else if (status.get() == CANCELLED) {
```
rjeberhard: Let me propose a different fix here -- The problem with this fix is that the second fiber is hung until resume() is called on the first fiber, which might never happen.  Instead, let's move the exitCallback outside of completionCheck().

doxiao: The testing showed that the resume did happen on the first fiber for the use cases in the QA test cases. But I can try the solution that you suggested.

PR #118, opened 2018-02-27
Simplify ingress processing flow by having per-cluster service

rjeberhard:
My proposed experiment worked!

- Still creating per-WebLogic server Services
- However, I'm creating these Services up front and not deleting them as Pods are deleted
- Newly creating per-WebLogic cluster Services
- Ingress now references the per-WebLogic cluster Service

This means that as we scale up and down we only need to create or delete Pods.  There is no more creating / deletion of Services and no more update to the Ingress.

I've found that the load balancing is very responsive to Pods terminating or coming ready.

@anpanigr, I'm looping you in because any scale up/down tests should no longer look for Services coming and going, but only for Pods.


src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -979,6 +983,113 @@ else if (StartupControlConstants.AUTO_STARTUPCONTROL.equals(sc)) {
...
+    for (WlsClusterConfig wlsClusterConfig : scan.getClusterConfigs().values()) {
+      for (WlsServerConfig wlsServerConfig : wlsClusterConfig.getServerConfigs()) {
+        String serverName = wlsServerConfig.getListenAddress();
+        if (!allServers.contains(serverName)) {
+          allServers.add(serverName);
+          ssic.add(new ServerStartupInfo(wlsServerConfig, wlsClusterConfig, null, null));
+        }
+      }
+    }
+    for (Map.Entry<String, WlsServerConfig> wlsServerConfig : scan.getServerConfigs().entrySet()) {
```
doxiao: It seems to me that the ServerConfigs here may include admin server too. If so, do we need to exclude it in the "ssic" list that we generate? The "ssic" list is later used in ManagedServerServicesStep, which I assume should not impact admin server. Or maybe the admin server is excluded later somewhere?

rjeberhard: Yes, this includes admin server here so that we validate that the admin server specific Service is created and correct.

This list of ServerStartupInfo ("ssic") is different than the other list.  I used similar data because we need details about port/nodePort to correctly create the services.

PR #129, opened 2018-03-01
Remove AS node port

doxiao:


src/integration-tests/bash/run.sh
```diff
@@ -1105,15 +1105,15 @@ function verify_admin_server_ext_service {
...
-    local get_configured_nodePort="kubectl get domains -n $NAMESPACE -o jsonpath='{.items[?(@.metadata.name == \"$DOMAIN_UID\")].spec.asNodePort}'"
+    local get_configured_nodePort="kubectl get domains -n $NAMESPACE -o jsonpath='{.items[?(@.metadata.name == \"$DOMAIN_UID\")].spec.serverStartup[?(@.serverName == \"admin-server\")].nodePort}'"
 
     local configuredNodePort=`eval $get_configured_nodePort`
 
-    trace "configured asNodePort in domain $DOMAIN_UID is ${configuredNodePort}"
+    trace "configured NodePort for the admin server in domain $DOMAIN_UID is ${configuredNodePort}"
 
     if [ -z ${configuredNodePort} ]; then
       kubectl describe domain $DOMAIN_UID -n $NAMESPACE
```
tbarnes-us: Please check whether this if clause is reached during testing (before & after your change) -- I assume it shouldn't be, and wonder why it's even there...

doxiao: This was tested as follows:
1) I made the json path wrong, the "if" resulted true, and the trace message inside the "if" appeared.  
2) I also verified that the current json path does return the configured value, which showed up in the trace message before the "if" statement.

PR #130, opened 2018-03-01
Use WatchBuilder to create Watch objects

russgold:
Uses the builder pattern at the Watch level rather than the Call level, in preparation for providing unit test alternatives

### DIFF
--- a/src/main/java/oracle/kubernetes/operator/DomainWatcher.java
+++ b/src/main/java/oracle/kubernetes/operator/DomainWatcher.java
@@ -3,17 +3,17 @@
 
 package oracle.kubernetes.operator;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
-import com.google.gson.reflect.TypeToken;
 import io.kubernetes.client.ApiException;
 import io.kubernetes.client.util.Watch;
 import oracle.kubernetes.operator.domain.model.oracle.kubernetes.weblogic.domain.v1.Domain;
 import oracle.kubernetes.operator.helpers.ClientHelper;
 import oracle.kubernetes.operator.helpers.ClientHolder;
+import oracle.kubernetes.operator.builders.WatchBuilder;
 import oracle.kubernetes.operator.watcher.Watcher;
 import oracle.kubernetes.operator.watcher.Watching;
 import oracle.kubernetes.operator.watcher.WatchingEventDestination;
+
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * This class handles Domain watching. It receives domain events and sends
@@ -73,14 +73,9 @@ public class DomainWatcher implements Runnable {
        */
       @Override
       public Watch<Domain> initiateWatch(Object context, String resourceVersion) throws ApiException {
-        return Watch.createWatch(client.getApiClient(),
-            client.callBuilder().with($ -> {
-              $.resourceVersion = resourceVersion;
-              $.timeoutSeconds = 30;
-              $.watch = true;
-            }).listDomainCall(ns),
-            new TypeToken<Watch.Response<Domain>>() {
-            }.getType());
+        return new WatchBuilder(client)
+                  .withResourceVersion(resourceVersion)
+                .createDomainsInNamespaceWatch(ns);
       }
 
       @Override
--- a/src/main/java/oracle/kubernetes/operator/PodWatcher.java
+++ b/src/main/java/oracle/kubernetes/operator/PodWatcher.java
@@ -2,14 +2,6 @@
 // Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 package oracle.kubernetes.operator;
 
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicBoolean;
-
-import com.google.gson.reflect.TypeToken;
-
 import io.kubernetes.client.ApiException;
 import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.models.V1Pod;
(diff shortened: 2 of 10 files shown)


src/main/java/oracle/kubernetes/operator/DomainWatcher.java
```diff
@@ -73,14 +73,9 @@ public void run() {
...
-            client.callBuilder().with($ -> {
-              $.resourceVersion = resourceVersion;
-              $.timeoutSeconds = 30;
-              $.watch = true;
-            }).listDomainCall(ns),
-            new TypeToken<Watch.Response<Domain>>() {
-            }.getType());
+        return new WatchBuilder(client)
+                  .withResourceVersion(resourceVersion)
+                .createDomainsInNamespaceWatch(ns);
```
rjeberhard: You call this createDomainsInNamespaceWatch, but the Pod, Ingress, and Service versions are just createXXXWatch().  Why is this?  I'm guessing it's because you have an *all* namespaces watch variant, correct?  That variant would never be interesting to the runtime so we should probably refactor the test code to not use it either.

src/main/java/oracle/kubernetes/operator/PodWatcher.java
```diff
@@ -49,7 +48,6 @@
    * Factory for PodWatcher
    * @param ns Namespace
    * @param initialResourceVersion Initial resource version or empty string
-   * @param destination Callback for watch events
```
rjeberhard: I'm suspicious here...  Need to verify logic with you.  For Ingress, Service, and Domain, we have a single callback method, but for Pod we have two use cases: 1) general watch for add, change, delete as for other resource types, and 2) watch for Pod to come to its "Ready" state.  Obviously, these could be refactored into one method called by the PodWatcher, but I'm going to want to verify the behavior has been refactored correctly.

src/main/java/oracle/kubernetes/operator/builders/CallParams.java
```diff
@@ -0,0 +1,91 @@
+package oracle.kubernetes.operator.builders;
+
+import io.kubernetes.client.ProgressRequestBody;
+import io.kubernetes.client.ProgressResponseBody;
+
+public class CallParams {
```
rjeberhard: Two comments:

1) This reproduces behavior in CallBuilder.  Is your plan to refactor that as well?  (I'm okay with this if we preserve the client pooling behavior)

2) Need Javadoc for all public classes/methods

src/main/java/oracle/kubernetes/operator/builders/WatchBuilder.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+    }
+
+    /**
+     * Creates a web hook object to track changes to weblogic domains in all namespaces
+     * @return the active web hook
+     * @throws ApiException if there is an error on the call that sets up the web hook.
+     */
+    public Watch<Domain> createDomainsInAllNamespacesWatch() throws ApiException {
```
rjeberhard: These "All" namespace variants will only be used by test code.  Runtime code will always work with some set of namespaces that will require separate watches for each namespace.

src/main/java/oracle/kubernetes/operator/builders/WatchBuilder.java
```diff
@@ -0,0 +1,320 @@
...
+    /**
+     * Creates a web hook object to track changes to custom objects
+     * @param namespace the namespace in which to track custom objects.
+     * @param responseBodyType the type of objects returned in the events
+     * @param group the custom resource group name
+     * @param version the custom resource version
+     * @param plural the custom resource plural name    @return the active web hook
+     * @throws ApiException if there is an error on the call that sets up the web hook.
+     */
+    public <T> Watch<T> createCustomObjectWatch(String namespace, Class<?> responseBodyType, String group, String version, String plural) throws ApiException {
```
rjeberhard: Similar...  I think the test that uses this is Bill's very first test.  We only have one "custom" type, which is Domain.  We would never watch or list for generic custom types.

src/test/java/oracle/kubernetes/custom/CustomObjectApisTest.java
```diff
@@ -1,8 +1,7 @@
-// Copyright 2017, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2017-2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
rjeberhard: Format is "2017, 2018"

src/test/java/oracle/kubernetes/operator/WatchTest.java
```diff
@@ -1,9 +1,8 @@
-// Copyright 2017, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2017-2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
rjeberhard: Format is "2017, 2018"

src/main/java/oracle/kubernetes/operator/DomainWatcher.java
russgold: I largely modeled the createXXX methods after the ones in CallBuilder. I am certainly open to changing them.

src/main/java/oracle/kubernetes/operator/PodWatcher.java
rjeberhard: Actually, you didn't change the functionality...  Just please restore the Javadoc.

src/main/java/oracle/kubernetes/operator/DomainWatcher.java
rjeberhard: Yes, I see that...  My only comment is about watching domains -- you have two variants: one for a single namespace and the other for all namespaces.  I'm guessing you did this because the test watches for domains in all namespaces, but I'd prefer to change the test since there is no runtime use case for watching domains in *all* namespaces.

src/main/java/oracle/kubernetes/operator/builders/CallParams.java
russgold: Yes, I was planning on eventually adding a StepBuilder which would do similar things for steps, with an eye toward replacing CallBuilder altogether. I didn't initially add the client pooling behavior because the unit tests seem to invoke the client to do set up, and I wanted to use the same client in both the setup and the builder. Do you have an idea on how to do that well? Or does it not matter?

PR #133, opened 2018-03-05
Add unit tests for Watcher, some code cleanup

russgold:
This also fixes a bug that occurs when Kubernetes returns an Error response from Watch. Bill had fixed it with a workaround to a bug in Watch.next(), but the current version of the Java API fixes that -- and therefore breaks his workaround.


russgold:
Next step will be some simplification of the four existing watchers, which contain a lot of duplicate code.

PR #145, opened 2018-03-09
Create shared domain scripts config map and use for liveness and readiness probes

rjeberhard:
This functionality is moved from the domain creation job to this config map created by the operator.

The readiness probe is modified from being an HTTP request to the ready app. to being a check of the server's state in the NM's state file.  This was done because a server in ADMIN mode should not be included in load balancing, which means that we need this server to show as not ready.


src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -1,143 +1,165 @@
...
+
+      cm.setMetadata(metadata);
+
+      Map<String, String> data = new HashMap<>();
+      
+      data.put("livenessProbe.sh", 
+          "#!/bin/bash\n" + 
+          "# Kubernetes periodically calls this liveness probe script to determine whether\n" + 
+          "# the pod should be restarted. The script checks a WebLogic Server state file which\n" + 
+          "# is updated by the node manager.\n" +
```
tbarnes-us: It'd be nice to add a description of $1 and $2, or make it self documenting by adding a DOMAIN_UID=$1 and WL_SERVER_NAME=$2, and referencing the new names .  (domain-uid & wl-server-name respectively).

Ditto for readinessProbe.sh

src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -1,143 +1,165 @@
...
+          "#!/bin/bash\n" + 
+          "\n" + 
+          "# Kubernetes periodically calls this readiness probe script to determine whether\n" + 
+          "# the pod should be included in load balancing. The script checks a WebLogic Server state\n" + 
+          "# file which is updated by the node manager.\n" + 
+          "\n" + 
+          "STATEFILE=/shared/domain/$1/servers/$2/data/nodemanager/$2.state\n" + 
+          "\n" + 
+          "if [ `jps -l | grep -c \" weblogic.NodeManager\"` -eq 0 ]; then\n" + 
+          "  exit 1\n" +
```
tbarnes-us: The original liveness probe shell script has echos that perhaps should be put back in:

if [ `jps -l | grep -c " weblogic.NodeManager"` -eq 0 ]; then
  echo "Error: WebLogic NodeManager process not found."
  exit 1
fi

if [ -f ${STATEFILE} ] && [ `grep -c "FAILED_NOT_RESTARTABLE" ${STATEFILE}` -eq 1 ]; then
  echo "Error: WebLogic Server FAILED_NOT_RESTARTABLE."
  exit 1
fi

The echo's are actually helpful for debugging as kubernetes records them and reports them when you kubectl describe events...

Ditto for the new readinessProbe script.

src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -150,26 +149,38 @@ public NextAction apply(Packet packet) {
...
-      livenessProbe.exec(livenessExecAction);
+      V1ExecAction livenessAction = new V1ExecAction();
+      livenessAction.addCommandItem("/weblogic-operator/scripts/livenessProbe.sh");
+      livenessAction.addCommandItem(weblogicDomainName);
+      livenessAction.addCommandItem(spec.getAsName());
+      livenessProbe.exec(livenessAction);
       livenessProbe.setInitialDelaySeconds(10);
-      livenessProbe.setPeriodSeconds(1);
+      livenessProbe.setTimeoutSeconds(5);
+      livenessProbe.setPeriodSeconds(10);
```
tbarnes-us: Maybe Period seconds can be even lower?   This will speedup the lifecycles.

src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -150,26 +149,38 @@ public NextAction apply(Packet packet) {
...
-      readinessProbe.setInitialDelaySeconds(15);
+      V1ExecAction readinessAction = new V1ExecAction();
+      readinessAction.addCommandItem("/weblogic-operator/scripts/readinessProbe.sh");
+      readinessAction.addCommandItem(weblogicDomainName);
+      readinessAction.addCommandItem(spec.getAsName());
+      readinessProbe.exec(readinessAction);
+      readinessProbe.setInitialDelaySeconds(2);
       readinessProbe.setTimeoutSeconds(5);
-      readinessProbe.setPeriodSeconds(15);
+      readinessProbe.setPeriodSeconds(10);
```
tbarnes-us: Would an even lower period seconds be possible?   The lower the interval, the faster the lifecycle...

Issue #100, opened 2018-02-15
Setup Containers to mount NFS instead HostPath


antonionappi88:
Hello,
I tried to run the weblogic domain using as shared volume a NFS volume. So the containers where attaching a NFS volume instead of using HostPath. [Here](https://db-blog.web.cern.ch/blog/antonio-nappi/2018-02-setup-nfs-storage-oracle-cloud-infrastructure-using-linux-storage) there is how I setup a NFS server in OCI. On the NFS server I created a shared volume under `/shares/kube-nfs`. 

To let it to work I had to modify the `persistent-volume-template.yaml` from :

```yaml
apiVersion: v1                          
kind: PersistentVolume                  
metadata:                               
  name: %PERSISTENT_VOLUME%             
  labels:                               
    weblogic.domainUID: %DOMAIN_UID%    
spec:                                   
  storageClassName: %STORAGE_CLASS_NAME%
  capacity:                             
    storage: %PERSISTENT_VOLUME_SIZE%   
  accessModes:                          
    - ReadWriteMany                     
  persistentVolumeReclaimPolicy: Retain 
  hostPath:                             
    path: "%PERSISTENT_VOLUME_PATH%"    
```

 to :

``` yaml
apiVersion: v1                          
kind: PersistentVolume                  
metadata:                               
  name: %PERSISTENT_VOLUME%             
  labels:                               
    weblogic.domainUID: %DOMAIN_UID%    
spec:                                   
  storageClassName: %STORAGE_CLASS_NAME%
  capacity:                             
    storage: %PERSISTENT_VOLUME_SIZE%   
  accessModes:                          
    - ReadWriteMany                     
  persistentVolumeReclaimPolicy: Retain 
  nfs:                                  
     server: NFS_SERVER_IP               
     path: "%PERSISTENT_VOLUME_PATH%"   
```
And in the create-domain-job-inputs.yaml I had to set `/shares/kube-nfs` as persistencePath.

Hope it helps.

simon-meng-cn:
I think your change is reasonable.  What is your requirement for this issue?

simon-meng-cn:
Hi @antonionappi88,
I am not sure what the requirement of this issue is. Can you clarify it?
Thanks

rjeberhard:
Simon,

[...]

PR #146, opened 2018-03-12
Add persistent volume configuration to support NFS

simon-meng-cn:
The fix contains following content:
1. Add NFS related configuration in kubernetes/internal/persistent-volume-template.yaml
2. Add two configurable parameters (persistenceType and nfsServer) in kubernetes/create-domain-job-inputs.yaml
3. Update kubernetes/create-domain-job.sh
3.1 Validate the two input parameters. The value of persistenceType must be hostPath or nfs. If using nfs, nfsServer must be specified.
3.2 Generate the correct persistent volume yam file based on input parameters.
4. Fix a typo.  validatePersistenVolumeName --> validatePersistentVolumeName

Jenkins passed http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator/86/

### DIFF
--- a/kubernetes/create-domain-job-inputs.yaml
+++ b/kubernetes/create-domain-job-inputs.yaml
@@ -36,6 +36,13 @@ managedServerNameBase: managed-server
 # Port number for each managed server
 managedServerPort: 8001
 
+# Persistent volume type, the value must be hostPath or nfs.
+# If using nfs, nfsServer must be specified.
+persistenceType: hostPath
+
+# NFS server name or ip
+nfsServer: nfsServer
+
 # Physical path of the persistent volume storage
 persistencePath: /scratch/k8s_dir/persistentVolume001
 
--- a/kubernetes/create-domain-job.sh
+++ b/kubernetes/create-domain-job.sh
@@ -86,13 +86,24 @@ function validatePersistentVolumeClaimName {
 #
 # Function to validate the persistent volume name
 #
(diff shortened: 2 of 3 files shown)


doxiao:
Looks good to me.
It would be nice if we could add a test case in the integration test suite for this feature. We'll need a NFS server that is accessible from different environments, such as hosted box, wercker pipeline, and Jenkins machines.

tbarnes-us:
I agree with Dongbo that a run.sh test case would be nice - perhaps by modifying one of the existing domains that are already tested (domain3 or domain4?).    I suggest testing it only when QUICKTEST is not true and JENKINS=true with Brian Moyers help to setup a permanent NFS location.   You can design the test on hosted linux, but IIRC, getting NFS up on running on arbitrary hardware is problematic --  we don't want to assume there's root access on hosted linux or Jenkins.

rjeberhard:
Fixes issue #100

simon-meng-cn:
Since the commit has been merged, I suggest close this pull request.
I have created a new JIRA task for testing NFS in  integration test suite. See https://jira.oraclecorp.com/jira/browse/OWLS-63719

doxiao:
Can you add the new parameters to the table in weblogic-kubernetes-operator/site/creating-domain.md?

simon-meng-cn:
A new pull request is created for it. Thanks.

PR #163, opened 2018-03-27
disable leasing while we try debug why it stopped working

tbarnes-us:
Something strange has started to happen with a kubectl create cm 'dry-run' piping to stdin 'kubectl replace -f'.   Disabling leasing for now while we investigate.

Wercker test results for this change pending.


tbarnes-us:
Closing pull request - have alternate solution in progress.

PR #174, opened 2018-04-02
Add NFS integration test case for OWLS-63719

simon-meng-cn:
JENKINS passed. http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator/172/

### DIFF
--- a/src/integration-tests/bash/run.sh
+++ b/src/integration-tests/bash/run.sh
@@ -805,6 +805,10 @@ function run_create_domain_job {
     sed -i -e "s/^exposeAdminT3Channel:.*/exposeAdminT3Channel: true/" $inputs
 
     # Customize more configuration 
+    if [ "$DOMAIN_UID" == "domain7" ] ; then
+      sed -i -e "s/^weblogicDomainStorageType:.*/weblogicDomainStorageType: NFS/" $inputs
+      sed -i -e "s/^#weblogicDomainStorageNFSServer:.*/weblogicDomainStorageNFSServer: $NODEPORT_HOST/" $inputs
+    fi
     sed -i -e "s;^#weblogicDomainStoragePath:.*;weblogicDomainStoragePath: $PV_ROOT/acceptance_test_pv/$DOMAIN_STORAGE_DIR;" $inputs
     sed -i -e "s/^#domainUID:.*/domainUID: $DOMAIN_UID/" $inputs
     sed -i -e "s/^clusterName:.*/clusterName: $WL_CLUSTER_NAME/" $inputs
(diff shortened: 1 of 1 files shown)


src/integration-tests/bash/run.sh
```diff
@@ -2698,6 +2703,9 @@ function test_suite {
       # create domain6 in the default namespace with pvReclaimPolicy="Recycle", and verify that the PV is deleted once the domain and PVC are deleted
       test_create_domain_pv_reclaim_policy_recycle domain6
 
+      # create domain7 in the default namespace with weblogicDomainStorageType="NFS", and verify it
+      test_domain_creation domain7
```
rjeberhard: What does this test do?  I see that it creates a domain with the "NFS" storage type, but does that NFS share exist.  If not, what have we proved or verified?

simon-meng-cn: Configure NFS as a persistent volume and test the domain creation.
Test cases use the local host's NFS server. $PV_ROOT/acceptance_test_pv/$DOMAIN_STORAGE_DIR is the shared nfspath in domain7

rjeberhard: Does this mean that we could no longer run the integration tests locally?  We need to either create the NFS share from within the test infrastructure or document the pre-requisites and allow the person running the tests to inject details about NFS shares to use.

simon-meng-cn: Yes. Our initial goal was to configure a dedicated remote NFS server, and test cases could only run on JENKINS machines.  Since we don't have any extra NFS shares. We have to use the local NFS server running on the JENKINS machine.

src/integration-tests/bash/run.sh
```diff
@@ -2698,6 +2703,11 @@ function test_suite {
       # create domain6 in the default namespace with pvReclaimPolicy="Recycle", and verify that the PV is deleted once the domain and PVC are deleted
       test_create_domain_pv_reclaim_policy_recycle domain6
 
+      if [ "$JENKINS" = "true" ]; then
+        # create domain7 in the default namespace with weblogicDomainStorageType="NFS", and verify it
+        test_domain_creation domain7
```
tbarnes-us: Adding a new domain lengthens the test time.   Instead, please piggyback on an existing domain - domain5 or domain6 look like they can be used.   

Also, add a note such as "# on Jenkins, this domain will also test NFS instead of HOSTPATH PV storage (search for [ "$DOMAIN_UID" == "domain5" ]).

Finally, modify the DOMAIN_UID check  to also check if on JENKINS...  && [ "$JENKINS" = "true" ]

tbarnes-us:
This looks good to me.   I don't think the wercker build failure is related - it doesn't test domain5 or even set JENKINS to true...

PR #176, opened 2018-04-02
Standardize names phase2

moreaut:
Standardize the kubernetes artifact names and labels.  Also, standardize the names of the yaml files for configuring a domain's persistent volume and persistent volume claim.


operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -47,16 +48,92 @@ public ScriptConfigMapStep(String namespace, Step next) {
...
       
       AnnotationHelper.annotateWithFormat(metadata);
       
       Map<String, String> labels = new HashMap<>();
+      // This config map is a singleton that is shared by all the domains
+      // We need to add a domain uid label so that it can be located as
+      // related to the operator.  However, we don't have a specific domain uid
+      // to set as the value.  So, just set it to an empty string.  That way,
+      // someone seleting on just the weblogic.domainUID label will find it.
+      labels.put(LabelConstants.DOMAINUID_LABEL, "");
```
rjeberhard: Why is it necessary that this cm be found by a search for domainUID?

moreaut: It's part of the labeling effort - all operator related artifacts should have a
weblogic.operatorName label, and all domain related artifacts should have
weblogic.domainUID label.  That way, a customer can find all of operator / domain
related kubernetes artifacts using a selector that checks for the presence of either
of these labels.

moreaut: For all of the operator artifacts, which are essentially singletons, we set the weblogic.operatorName value to the operator's namespace.  For all of the other domain
 elated artifacts, we set weblogic.domainUID to that domain's uid.  However, since this config
map is shared by all domains, and is still domain related, I needed to add the
weblogic.domainUID tag to indicate that the config map is domain related, but I don't have
a specific domainUID (since all domains used it), so I just set it to an empty string.

This seemed better than inventing yet another tag to indicate that config map was domain related.

(Tom Barnes & I worked out this strategy)

PR #186, opened 2018-04-05
Replace the domainUID="" label on the per-domain operator config map with an operatorName="<operatornamespace>" label

moreaut:
The delete script for removing all domains does it by looking for kubernetes artifacts with a domainUID tag.  Currently, the operator creates a config map per domain namespace containing scripts that should be shared by all domains in that namespace, and tags it with a domainUID="" label.

This causes the delete script for removing all domains to remove this config map too, and, the next time the operator needs to create a domain in that namespace, it has to recreate the config map for the scripts in that namespace.

To prevent this, change the label on the config map from 'domainUid=""' to 'operatorName="<operatorNamespace>"' - that is, tag it as operator related instead of domain related.

Note: eventually we should add a script to remove the operator.  That script can use the
operatorLabel to find and delete these config maps (as well as the other operator-related
k8s artifacts).

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Main.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Main.java
@@ -169,10 +169,7 @@ public class Main {
 
   private static void begin() {
     // read the operator configuration
-    String namespace = System.getenv("OPERATOR_NAMESPACE");
-    if (namespace == null) {
-      namespace = "default";
-    }
+    String namespace = getOperatorNamespace();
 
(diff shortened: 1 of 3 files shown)


tbarnes-us:
LGTM

One minor comment on your pull description:   'This will cause the config map to remain there until the script to remove the operator is invoked (since that script uses the operatorName label to find the k8s artifacts to remove)'.   I agree - except, as far as I know there's no script to remove the operator (yet).

PR #198, opened 2018-04-12
OWLS-64528 - Integrate Voyager to WLS Operator

lilyhe123:
integration with voyager/HAProxy load balancer


doxiao:
Lily, it seems that your branch was created from the master.  We need to use "develop" branch for new PRs. 

See Ryan's announcement in the slack channel on 4/9.
"I've branched master to develop at the v0.2 tag.  All PR's should now follow the "git flow" branching model: http://nvie.com/posts/a-successful-git-branching-model/"

tbarnes-us:
Two comments:

Comment 1:  It looks like the integration tests only test TRAEFIK.   I guess they should also test VOYAGER?  I suppose one way to do this would be to add a run.sh global env var that controls the choice of LB, and then set this env var differently as needed, plus make sure the 'cleanup.sh' script finds and deletes voyager k8s resources.   For example, maybe we could always run TRAEFIK by default, modify 'wercker.yml' to set VOYAGER for 1.7 runs and and TRAEFIK for 1.8 runs (Werker runs two run.sh runs in parallel - one for k8s v1.7 and another for v1.8).

Comment 2: Please verify that any new voyager related resources are fully labeled.    They should have have a weblogic.domainUID and weblogic.domainName labels, plus if they are scoped to a cluster, they should also have a weblogic.clusterName label.   "kubectl get all,domains,crd,ingress,cm,pv,pvc,ns --show-labels=true --all-namespaces=true"

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -321,6 +323,11 @@ function initialize {
     validationError "The template file ${traefikInput} for generating the traefik deployment was not found"
   fi
 
+  voyagerInput="${scriptDir}/voyager-ingress-template.yaml"
+  if [ ! -f ${voyagerInput} ]; then
+    validationError "The template file ${voyagerInput} for generating the voyager ingress was not found"
```
rjeberhard: Voyager Ingress

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -473,25 +481,38 @@ function createYamlFiles {
...
+    echo Generating ${traefikSecurityOutput}
+    sed -i -e "s:%NAMESPACE%:$namespace:g" ${traefikSecurityOutput}
+    sed -i -e "s:%DOMAIN_UID%:${domainUID}:g" ${traefikSecurityOutput}
+    sed -i -e "s:%DOMAIN_NAME%:${domainName}:g" ${traefikSecurityOutput}
+    sed -i -e "s:%CLUSTER_NAME%:${clusterName}:g" ${traefikSecurityOutput}
+    sed -i -e "s:%CLUSTER_NAME_LC%:${clusterNameLC}:g" ${traefikSecurityOutput}
+  fi
+
+  if [ "${loadBalancer}" = "VOYAGER" ]; then
+    # Voyager ingress file
```
rjeberhard: Ingress

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
 
 }
 
+#
+# Deploy voyager/HAProxy load balancer
+#
+function setupVoyagerLoadBalancer {
+  # deploy voyager ingress controller
```
rjeberhard: Voyager Ingress

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  kubectl create namespace voyager
```
rjeberhard: Should this namespace be configurable?

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  curl -fsSL https://raw.githubusercontent.com/appscode/voyager/6.0.0/hack/deploy/voyager.sh \
+  | bash -s -- --provider=baremetal --namespace=voyager
+
+  # deploy voyager ingress resource
+  kubectl create -f ${voyagerOutput}
+
+  echo Checking voyager deploy
```
rjeberhard: Voyager

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
 
 }
 
+#
+# Deploy voyager/HAProxy load balancer
+#
+function setupVoyagerLoadBalancer {
+  # deploy voyager ingress controller
+  kubectl create namespace voyager
+  curl -fsSL https://raw.githubusercontent.com/appscode/voyager/6.0.0/hack/deploy/voyager.sh \
```
rjeberhard: Might need proxies?

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  | bash -s -- --provider=baremetal --namespace=voyager
+
+  # deploy voyager ingress resource
+  kubectl create -f ${voyagerOutput}
+
+  echo Checking voyager deploy
+  vdep=`kubectl get deploy | grep voyager | wc | awk ' { print $1; } '`
+  if [ "$vdep" != "1" ]; then
+    fail "The deployment of voyager ingress was not created"
```
rjeberhard: Voyager Ingress

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  # deploy voyager ingress controller
+  kubectl create namespace voyager
+  curl -fsSL https://raw.githubusercontent.com/appscode/voyager/6.0.0/hack/deploy/voyager.sh \
+  | bash -s -- --provider=baremetal --namespace=voyager
+
+  # deploy voyager ingress resource
+  kubectl create -f ${voyagerOutput}
+
+  echo Checking voyager deploy
+  vdep=`kubectl get deploy | grep voyager | wc | awk ' { print $1; } '`
```
rjeberhard: This is a pretty weak check.  It only validates that the deployment exists, but not that its running properly.

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  if [ "$vdep" != "1" ]; then
+    fail "The deployment of voyager ingress was not created"
+  fi
+
+echo Checking voyager service
```
rjeberhard: Voyager

kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -584,6 +605,30 @@ function createDomain {
...
+  vscv=`kubectl get service voyager-stats | grep voyager-stats | wc | awk ' { print $1; } '`
```
rjeberhard: Similar comment...

kubernetes/internal/voyager-ingress-template.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: voyager.appscode.com/v1beta1
+kind: Ingress
+metadata:
+  name: voyager-ingress
+  namespace: default
```
rjeberhard: I thought this was in the voyager namespace?  Also, does this have to go in the namespace of the domain, which might not be default.

rjeberhard:
@lilyhe123 I'm surprised you don't have any Java changes...  The domain presence logic creates an Ingress per cluster that has Traefik related annotations.

doxiao:
Two comments:
1)  We need to clean up Voyager related K8S resources that are associated with a domain in delete-weblogic-domain-resources.sh.
2) We need to clean up Voyager related resources in cleanup.sh when Voyager is enabled for run.sh.

kubernetes/delete-weblogic-domain-resources.sh
```diff
@@ -53,6 +53,46 @@ cat << EOF
...
+  local voyagerIngressName="ingress.voyager.appscode.com"
+  local domainName=$1
+  local ns=`kubectl get ingress.voyager.appscode.com --all-namespaces | grep $domainName | awk '{ print $1 }'`
+  if [ -n "$ns" ]; then
+    echo $voyagerIngressName $domainName-voyager -n $ns >> $2
+  fi
+}
+
+#
+# get voyager related resources
```
tbarnes-us: Would it be practical to use the 'kubectl label' command to label the voyager generated resources that we don't create directly ourselves?

That would eliminate the need to put voyager-specific code in delete-domain-resources, plus helpfully make the resources show up in global domain label queries... 

      kubectl -n ${namespace} label \
              $RESOURCE_TYPE $RESOURCE_NAME \
              domainUID=$DOMAIN_UID

src/integration-tests/bash/cleanup.sh
```diff
@@ -63,248 +50,74 @@ USER_PROJECTS_DIR="$RESULT_DIR/user-projects"
...
-  return 1
-}
-
-function kubectlDeleteF {
-   if [ -f "$1" ]; then
-      kubectl delete -f "$1" --ignore-not-found
-   else
-      echo @@ File \"$1\" not found.  Skipping kubectl delete -f.
-   fi
+  # now, get all non-namespaced types with -l $LABEL_SELECTOR
```
tbarnes-us: I think it's good to replace 'orderly delete' with a label based deletes, but I think the 'genericDelete' step should not be removed and should still be the final step.   IMO, it's still needed to help guarantee a full cleanup.   Cleanup.sh is used by wercker and on an ad-hoc basis after test runs, so it seems important to have it cleanup even if someone forgets to label a new resource properly.

Also, it's important IMO to do generic deletes in a timer loop, since sometimes 'out-of-order' deletes aren't helpful  -- for example,. a resource can be automatically created by another resource that hasn't been deleted yet.   See the original version of cleanup.sh or see the delete-domain script for the pattern.

tbarnes-us: It's equally important to do orderly deletes in a retry loop for similar reasons.   The delete-domain script already does this internally for domains, but the same needs to be done for the operator label based delete...

tbarnes-us: Finally, the delete-domain script tries to wait for a clean shutdown, but this wait isn't needed for 'cleanup.sh' which has no need for one.   To keep the running time short, maybe it'd be good to add a new option to the delete-domain script '-f' to cause it to skip directly to its 'Phase 3'.

tbarnes-us: One more comment (sorry).  It looks like the code for deleting namespaces was removed from cleanup.sh -- this code should be restored as well.   

To summarize:  it's good that cleanup.sh now calls delete-domain, and that it now deletes operator stuff via labels, but (A) it should also retain its old logic for cleaning up anything (the old 'generic delete') in case label deletes fail, (B) the new operator label delete needs to be retried up to some number of seconds  (similar to the logic in delete-domain), and (C) it needs to retain its old logic for deleting namespaces.

lilyhe123: all your comments are addressed except (A) it should also retain its old logic for cleaning up anything (the old 'generic delete') in case label deletes fail.
If all the resources are labeled properly, deleting with label can always succeed. If some resources don't have the proper label and cause wercker fail, what need to be fixed is the resource.

Now the cleanup.sh doesn't  call the delete-domain script. Seems to me the logic in delete-domain script is more complicated than it should be. 
One thing I'm not sure is the differences between the two approaches below to delete domain:
1. setting domain's startupControl to NONE to let wls operator shutdown the domain
2. delete domain related resources directly
The delete-domain script can be cleanup after understanding this better.

tbarnes-us:
I think the cleanup script still needs work.

(1) Wait for a max time instead of a hard coded number of iterations when doing label deletes.   15 seconds would be adequate IMO.
(2) Wait for all resources to delete, not just pods.
(3) Add back in generic delete to ensure we have a complete cleanup _even in the presence of bugs_.   This is necessary for environments like Wercker, which depend on cleanup.sh to work even if everything isn't labeled propertly, etc.   Otherwise a failed Wercker run could leave behind resources that prevent future Wercker runs from running. 
(4) It would be fine in my opinion to implement  2 by adding back in the generic delete in 3 (as it waits for all resources).
(5) Does voyager leave behind any 'per domain' resources after delete-weblogic-domain is called?   If so, then delete-weblogic-domain should be updated to handle this.    For example, via the 'kubectl label' command I mentioned earlier.

FYI:   Since voyager leaves behind resources that are shared between domains, it looks like we'll eventually need to have a generic delete script similar delete-weblogic-domain that does a cleanup if there are no domains depending on it -- or make the delete-weblogic-domain script smart about detecting when the last domain is deleted.    (We'll eventually need delete-weblogic-operator, delete-voyager, etc).
~                      



FYI:   Since voyager leaves behind resources that are shared between domains, it looks like we'll eventually need to have a generic delete script similar delete-weblogic-domain that does a cleanup if there are no domains depending on it -- or make the delete-weblogic-domain script smart about detecting when the last domain is deleted.    (We'll eventually need delete-weblogic-operator, delete-voyager, etc).
~

tbarnes-us:
Sorry - didn't meant to close the request.  Re-opening.

By the way, I'm fine with merging this pull request -- as long as the issues I pointed out today are corrected in a new pull request.    (Dongbo has dependencies on this change, and it'd be helpful to get in her changes by EOW.)

kubernetes/internal/create-weblogic-domain.sh
lilyhe123: yes, depending on proxy setting on environment variables

kubernetes/internal/create-weblogic-domain.sh
lilyhe123: This is the namespace of Voyager Controller, not Voyager Ingress resource. We only need one Voyager Controller. I think it's fine to hardcoded it.

kubernetes/internal/voyager-ingress-template.yaml
lilyhe123: fixed. this is the namespace of Voyager Ingress resource which is the same namespace of domain.

lilyhe123:
<blockquote>I'm surprised you don't have any Java changes... The domain presence logic creates an Ingress per cluster that has Traefik related annotations.</blockquote>

This is non-related to Voyager. Voyager creates its own crd ingresses.voyager.appscode.com.
Seems to me the logic to create Ingress need to be moved out to create-domain script, instead of in the operator java code. This can be covered by another task.

tbarnes-us:
The latest delete scripts LGTM, but we'll probably eventually want to have a future pull that updates the 'cleanup.sh' script to handle the download from github of a voyager cleanup script differently.   

The current cleanup.sh script now has:
```
 curl -fsSL https://raw.githubusercontent.com/appscode/voyager/6.0.0/hack/deploy/voyager.sh \
 +      | bash -s -- --provider=baremetal --namespace=voyager --uninstall --purge
```

This download potentially has three issues, if not more: (1) if https_proxy is needed, but not set, the script won't download, (2) I'm fairly sure we generally don't want to depend on third party downloads at runtime, and (3) it looks like it hard codes a version# which may not continue to work if voyager is updated to a newer version.   I think the reasons for (2) might be security related and also that the script won't work in environments where the public internet isn't available...   

Please track this issue somewhere for future discussion and resolution - perhaps via a new JIRA.

PR #199, opened 2018-04-12
Feature/dynamic cluster

alai8:


operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -99,10 +99,17 @@ public NextAction apply(Packet packet) {
       List<V1ServicePort> ports = new ArrayList<>();
       V1ServicePort servicePort = new V1ServicePort();
       servicePort.setPort(port);
+      servicePort.setName("weblogic");
       if (nodePort != null) {
         servicePort.setNodePort(nodePort);
       }
       ports.add(servicePort);
+
```
rjeberhard: Why are you exposing the node manager port here if we didn't before?

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsDomainConfig.java
```diff
@@ -16,16 +16,77 @@
 import java.util.Map;
 
 /**
- * Contains configuration for a WebLogid Domain
+ * Contains a snapsot of configuration for a WebLogic Domain
```
rjeberhard: snapshot

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsRetriever.java
```diff
@@ -319,46 +401,182 @@ public WlsDomainConfig readConfig() {
...
+  /**
+   * Update the dynamic cluster size of the WLS cluster configuration.
+   *
+   * @param wlsClusterConfig The WlsClusterConfig object of the WLS cluster whose cluster size needs to be updated
+   * @param machineNamePrefix Prefix of names of new machines to be created
+   * @param targetClusterSize The target dynamic cluster size
+   * @return true if the request to update the cluster size is successful, false if it was not successful within the
+   *         time period, or the cluster is not a dynamic cluster
+   */
+  public boolean updateDynamicClusterSize(final WlsClusterConfig wlsClusterConfig,
```
rjeberhard: Who calls this?  We should discuss...  Other places use a pattern where this is wrapped in a Step which suspends until the asynchronous activity finishes.  This allows for timeout and retry, etc.

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
alai8: Sorry, this was leftover code when we were trying to address the WLDF delay issue by exposing nm port that I forgot to remove.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsRetriever.java
alai8: There is no usages. It was added some time ago when we thought the RestBackEndImpl class may need to adjust the dynamic cluster size. I will remove this method.

kubernetes/internal/create-weblogic-domain-job-template.yaml
```diff
@@ -416,28 +186,59 @@ data:
...
+      print('Done setting attributes for Server Template: %s' % templateName);
+
+
+      cd('/Clusters/%s' % cluster_name)
+      create(cluster_name, 'DynamicServers')
+      cd('DynamicServers/%s' % cluster_name)
+      set('ServerTemplate', st1)
+      set('ServerNamePrefix', "%MANAGED_SERVER_NAME_BASE%")
+      set('DynamicClusterSize', number_of_ms)
+      set('MaxDynamicClusterSize', 800)
```
lennyphan: Remove commented out set calls.  Should we remove setting MaxDynamicClusterSize?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -633,12 +639,23 @@ protected V1Pod computeManagedPodConfig(TuningParameters configMapHelper, Packet
       V1ContainerPort containerPort = new V1ContainerPort();
       containerPort.setContainerPort(scan.getListenPort());
       containerPort.setProtocol("TCP");
+//      containerPort.setName("weblogic"); // commented out as we are not exposing node manager port
```
lennyphan: Cleanup commented out section for exposing NM port

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsRetriever.java
```diff
@@ -319,46 +401,138 @@ public WlsDomainConfig readConfig() {
...
+  private static boolean updateDynamicClusterSizeWithServiceURL(final WlsClusterConfig wlsClusterConfig,
+                                                                final String machineNamePrefix,
+                                                                final int targetClusterSize,
+                                                                final HttpClient httpClient,
+                                                                final String serviceURL) {
+    LOGGER.entering();
+
+    boolean result = false;
+    // Create machine(s)
+    // Commented out as we are not configuring machines for servers
```
lennyphan: Remove since we're not creating machines implicitly anymore?

kubernetes/internal/create-weblogic-domain-job-template.yaml
tbarnes-us: In effect, D-C-S sets the number of ms that show up in the bean tree (virtual config beans of some kind), which in effect designates the servers that can be booted.   For example, when DCS is 10, you will see 10 managed servers in the config tree on the console even when they aren't booted.    It also can do 'double duty' of letting the wlst-scaling/external frameworks know the number of servers you'd like to be running.    Max DCS, as far as I know, simply prevents DCS from being configured higher than the specified value - I personally don't think it has much utility.   JMS and JTA usually require that DCS never be lowered during a scale down (instead just shutdown the servers you don't want) -- this lets these services iterate over the aforementioned 'virtual' beans to locate data that may be trapped in servers that are no longer going to be booted.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsRetriever.java
alai8: Right. Will also be removing some method arguments and utility class related to machine creation.

kubernetes/internal/create-weblogic-domain-job-template.yaml
alai8: This is being replaced with set('MaxDynamicClusterSize', number_of_ms). Also will be adding some validation logic to compare replicas and max dynamic cluster size in WlsClusterConfig class and make sure the max dynamic cluster size is honored and we won't scale above it.

PR #205, opened 2018-04-23
simplify operator script maintenance

simon-meng-cn:
The change includes following content
1. Move the shell and python scripts out of operator java helper source code to operator/src/main/resource/scripts
2. Modify the helper to load scripts from /scripts directory that in the operator jar file


tbarnes-us:
LGTM
+1

rjeberhard:
I rebased this to a local branch so that I can resolve the unit-test failure

PR #209, opened 2018-04-26
Support apache load balancer

doxiao:
The PR covers the following.
1) add two additional properties in domain inputs: loadBalancerVolumePath and loadBalancerAppPrepath.
2) add templates for apache deployment yaml file and RBAC yaml file
3) enhance create domain job to create apache yaml files and k8s artifacts. Right now for domains that are generated using create domain job, there is one apache instance per domain.  But customers can use one apache instance per K8S cluster if they manually setup and start an apache instance.
4) add an environment variable LB_TYPE in integration test run.sh to enable apache load balancer
5) update delete domain job
6) update cleanup.sh
7) minimum unit test. Need to complete the unit test cases, probably in a different PR.


moreaut:
I did a quick scan of the code and it looks good.

When will you write the unit tests?  Is there a reason why you need to merge this work before
the tests are written?  Normally I don't consider a feature ready to merge until there are
tests for it.  Also, writing these tests will be a fairly large amount of work.

doxiao:
Thanks for reviewing the change!

This feature is for 1.0 operator release, which is due by the end of next week. I am trying to get the product code reviewed and merged as soon as possible to give QA time to test it. Meanwhile I need to focus on the doc part of this feature to meet the deadline.

We might be able to get the unit test completed sooner if we could get help from others in the team. The doc part of the work is probably harder for others to help, especially the initial draft, in my opinion.

Note that the change has been tested manually and using integration test cases. I know it is different from unit test, but I believe we have reasonable coverage. The sooner we make the feature available to QA, the more test coverage we may get.

kubernetes/create-weblogic-domain-inputs.yaml
```diff
@@ -89,9 +89,16 @@ exposeAdminNodePort: false
 # Name of the domain namespace
 namespace: default
 
-# Load balancer to deploy.  Supported values are:TRAEFIK, NONE
+# Load balancer to deploy.  Supported values are:TRAEFIK, APACHE, NONE
 loadBalancer: TRAEFIK
 
+# Load balancer app prepath 
+loadBalancerAppPrepath: /
```
tbarnes-us: It looks like Prepath is only used by the Apache LB -- if so, the comment above should mention this.  Also, I wonder if the the Prepath and VolumePath fields comments should have a little more detail to help explain their purpose.

tbarnes-us:
Changes LGTM, except for the minor issue of fleshing out comments.

Followup Issue:   I suspect we need to come up with an integration test plan for Apache.   The plan for Voyager is to use voyager in one of the Wercker pipeline's and traefik in the other -- but this doesn't scale two three LBs since we're currently only running two parallel pipelines in Wercker.    Beyond Wercker, maybe we should consider having automated periodic Jenkins runs for each LBs.   If you agree this merits follow-up, please file a follow up JIRA to track.

PR #212, opened 2018-04-26
first pass edit on architecture doc

rosemarymarano:
The architecture doc refers to ELK which is now referred to as the Elastic Stack. Do you want to keep the older term (throughout) or update it?

Under, "This diagram shows the following details" please review the bulleted item regarding, "An Ingress is created for each WebLogic cluster." In particular, this sentence, "In the technology preview release, there is one instance of the load balancer running for each WebLogic cluster, and the load balancers are configured with the root URL path (“/”). More flexible load balancer configuration is planned for a future release."


mriccell:
Yes ELK should be changed to the Elastic Stack.
I believe we still have the limitation of One Ingress LB or one Apache - to - 1 WebLogic cluster/domain.  We should remove technology preview,

rjeberhard:
Changes look good here.  Are you going to remove tech preview description in this PR or another.  I need some way to know when the PR switches from a work-in-progress to ready to commit 😀

rosemarymarano:
I had already removed the reference to the Technology Preview before Monica made her comment. But, I understand your need to know when I am done incorporating edits. I will make a note of my status after each review comment. So, unless others have further input, from my point of view, this doc is done.

PR #228, opened 2018-04-30
editing pass

rosemarymarano:
Basic editing pass. Two things to note:
1. I changed the capitalization scheme to line up with the other docs (we are using sentence capitalization versus title capitalization.
2. Because its official name is the Apache HTTP Server, if we refer to it as the Apache web server, we don't need to capitalize any word other than Apache.

### DIFF
--- a/site/apache.md
+++ b/site/apache.md
@@ -1,15 +1,15 @@
 
-# Load Balancing with Apache Web Server
+# Load balancing with the Apache web server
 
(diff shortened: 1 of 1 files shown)


rosemarymarano:
I'll make that change (Apache web server to Apache HTTP Server) throughout this doc. However, I will have to go back into the README and install docs to change it there, too.

PR #234, opened 2018-05-01
April cleanup corrected

russgold:
Added unit tests, started refactoring Domain presence logic


operator/src/main/java/oracle/kubernetes/operator/helpers/DomainHomeHelper.java
```diff
@@ -0,0 +1,200 @@
...
+          // TODO: wait for the job to finish or fail
+          //       update domain status condition
+          return null;
+        }
+      });
+      
+      return doNext(create, packet);
+    }
+    
+    protected V1Job computeDomainHomeJob(DomainPresenceInfo info) {
```
moreaut: Need a unit test for this method.

moreaut: Currently, create-weblogic-domain.sh makes a copy of create-weblogic-domain-job-template.yaml and customizes it, then loads it into kubernetes to create the job.

It looks you're changing this - i.e. having code in the operator directly create the job.

But it looks like you've only partially done the work - I don't see changes to create-weblogic-domain.sh.

Perhaps this code snuck in by mistake from the WDT branch?

(p.s. it's also missing javadoc)

rjeberhard:
@russgold @moreaut Russ, in the original April cleanup review, I noted several changes for WDT that were included with your refactored files.  Tom has noted at least one of those here, but I suspect the others are still present in these changes.  Can you please review and remove changes that are specific to WDT.

PR #238, opened 2018-05-01
minor edits

rosemarymarano:
Minor text edits, and I removed this line, "Please take a look at our wish list to get an idea of the load balancing improvements we would like to do."


mriccell:
Rosemary the doc makes reference to a "Traefik admin UI" the name of the UI is "Clean AngularJS Web UI" or "Traefik Web UI"

PR #262, opened 2018-05-08
Update k8s_setup.md

Rajakavitha:
Fixed the numbered list for the installation steps. Also we may need to add installing kubectl as a prerequisite for Install Docker for Mac with Kubernetes because step 5 assumes that the user has installed kubectl on the Mac.

### DIFF
--- a/site/k8s_setup.md
+++ b/site/k8s_setup.md
@@ -108,7 +108,7 @@ scripts/cluster-check.sh
 $ rm -f generated/instances_id_rsa && terraform output ssh_private_key > generated/instances_id_rsa && chmod 600 generated/instances_id_rsa
 ```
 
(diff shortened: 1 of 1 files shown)


rjeberhard:
Hi Rajakavitha, thanks for the review.  Just to clarify your comment, are you seeing that installing Docker for Mac with Kubernetes does not install kubectl?  My experience has been that kubectl is included when you install this version of Docker.

Rajakavitha:
My suggestion was to rephrase Step 5. If you have previously used `kubectl` on your Mac.
Because this implies that you may have to do the step 5 only if you have previously used 'kubectl' on your Mac. May be: Ensure that kubectl points to the docker-for-desktop-cluster that you created.

Rajakavitha:
Yes, I agree.

PR #268, opened 2018-05-09
Fix race condition in Fiber

rjeberhard:
Fixes issue #267 

There was a race condition in Fiber's calling of the completion callback and when Future.get() would report that the Fiber was done.


russgold:
I could wish that the logic were easier to follow, but that's a carp. 

In the unit tests, the preferred way to silence loggers now is:

mementos.add(TestUtils.silenceOperatorLogger())

and to restore them:

for (Memento memento : mementos) memento.revert()

It's obviously not clear enough; you chose the legacy way to do it. I should probably clean out the old code.

PR #280, opened 2018-05-14
Feature lifecycle runtime phase1

moreaut:
Add some classes that will isolate the code in the operator runtime that starts and stops servers (and creates their pods) from the specifics of the corresponding configuration on the domain CRD.  It adds the concept of set of Config classes that tell the runtime the exact desired state of each server and cluster.  It adds the concept of a DomainConfigBuilder that, given a domain CRD, creates the
corresponding Config instances that the runtime needs.  It also adds a DomainConfigBuilderV1
implementation, that creates the config from the V1 domain CRD (i.e. startupControl,
clusterStartup and serverStartup instances, as well as the domain level image and image pull policy.).
It also adds a LifeCycleHelper class, which the operator runtime will use to get the Config instances
it needs (v.s. dealing with the domain CRD directly).

This is a first step in adding more fine grained properties on the domain CRD that let the user
control the state of each server and cluster (i.e. so that the user can control whether specific
servers are started or stopped, and so that the user can control the order of restarting servers
when rolling restarts need to occur).

Note: this pull request only adds the new Config support.  A subsequent pull request will change the
operator runtime to use this new Config support.


russgold:
In general, the test methods are hard to read. Please use blank lines to separate initialization, action, and validation sections. If possible, use separate methods to make it clear what is happening in the code. Any time you have a method (include a test method) with more than a few lines, the next developer has to reverse engineer your code, and that can easily lead to misunderstanding and errors.

Don't both converting int constants to Integer; the compiler will do that for you.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainConfigBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+  protected static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
+
+  protected static Integer DEFAULT_NODE_PORT = new Integer(0);
+  protected static String DEFAULT_STARTED_SERVER_STATE = ServerConfig.STARTED_SERVER_STATE_RUNNING;
+  protected static String DEFAULT_SHUTDOWN_POLICY = ServerConfig.SHUTDOWN_POLICY_FORCED_SHUTDOWN;
+  protected static Integer DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT = new Integer(0);
+  protected static Boolean DEFAULT_GRACEFUL_SHUTDOWN_IGNORE_SESSIONS = Boolean.FALSE;
+  protected static Boolean DEFAULT_GRACEFUL_SHUTDOWN_WAIT_FOR_SESSIONS = Boolean.FALSE;
+  protected static Integer DEFAULT_REPLICAS = new Integer(0); // TBD - is this correct?
```
rjeberhard: I'm fine with this.  Let's remove the TBD

PR #289, opened 2018-05-15
added note about restrictions to NFS3.0.0, please commit to the devel…

marinakog:
added note about restrictions to NFS3.0.0, please commit to the developer line but not to master until May 30


site/k8s_setup.md
```diff
@@ -110,6 +110,10 @@ $ rm -f generated/instances_id_rsa && terraform output ssh_private_key > generat
 
 8. If you need shared storage between your Kubernetes worker nodes, enable and configure NFS:
 
+In the current GA version, OCI Container Engine for Kubernetes support network Block storage that can be shared across nodes with access permission RWOnce (meaning that only one can write the rest only read). The WebLogic on Kubernetes domain created by the WebLogic Server Kubernetes Operator at this time requires a shared file system to store the WebLogic domain configuration, which MUST be accessible from all the pods across the nodes. As a workaround, we need to install an NFS server in one node and share the file system across all nodes.
```
rosemarymarano: +In the current GA version, the OCI Container Engine for Kubernetes supports network block storage that can be shared across nodes with access permission RWOnce (meaning that only one can write, others can read only). At this time, the WebLogic on Kubernetes domain created by the WebLogic Server Kubernetes Operator, requires a shared file system to store the WebLogic domain configuration, which MUST be accessible from all the pods across the nodes. As a workaround, you need to install an NFS server on one node and share the file system across all the nodes.

site/k8s_setup.md
```diff
@@ -110,6 +110,10 @@ $ rm -f generated/instances_id_rsa && terraform output ssh_private_key > generat
...
+Note: Currently we advise to use NFS version 3.0 for running WebLogic Server on OCI Container Engine for Kubernetes. We found during certification that when using NFS 4.0, the Servers in the WebLogic Domain went into failed state intermittently. Since multiple threads use NFS (default store, diag store, node manager, logging, domain_home), there are issues to access the file store. These issues disappear by changing NFS to version 3.0.
```
rosemarymarano: +Note: Currently, we recommend that you use NFS version 3.0 for running WebLogic Server on OCI Container Engine for Kubernetes. During certification, we found that when using NFS 4.0, the servers in the WebLogic domain went into a failed state intermittently. Because multiple threads use NFS (default store, diagnostics store, Node Manager, logging, and domain_home), there are issues when accessing the file store. These issues are removed by changing the NFS to version 3.0.

site/k8s_setup.md
rosemarymarano: Did a slight re-write (clean-up) on both paragraphs.

rjeberhard:
@marinakog and @rosemarymarano, I can't tell if current text is ready to be committed or if there is still editing.  Please approve the pull request when you both are happy with the text.

rosemarymarano:
There are outstanding edits that Marina should make to the two paragraphs she had me review. After she incorporates them, she can approve the request.

Issue #293, opened 2018-05-16
use predefined yaml files to install and uninstall voyager controller


lilyhe123:
Currently we depend on https://raw.githubusercontent.com/appscode/voyager/6.0.0/hack/deploy/voyager.sh to do this which is undesirable.

PR #294, opened 2018-05-16
Voyager update

lilyhe123:
use predefined yaml files to install and uninstall voyager controller.
#293


moreaut:
1) does this create backwards incompatibility issues with existing inputs files? i.e. do we need to create a version of the inputs files for this?

2) all of the kubernetes resources the yaml files create should have a weblogic.resourceVersion label (like all our others do)

3) where are the unit tests?

anpanigr:
while bringing up the domain I see the following unresolved variable as follows ...
[exec] The service ${domainUID}-voyager-stats is created successfully.
I know it is harmless, but should be taken care in internal/create-weblogic-domain.sh

lilyhe123:
<blockquote>1) all of the kubernetes resources in the yaml files create should have a weblogic.resourceVersion label (like all our others do)</blockquote>
Fixed
<blockquote>2) where are the unit tests?</blockquote>
UT for what? verify the generated yaml file? I'm not familiar with the UT framework. And the two new yaml files are subject to change. I'm thinking move them out of domain creation. So I'd prefer hold the UT.
<blockquote>3) are there any backwards compatibility issues here? i.e. GA domains that had older versions of these resources needing to catch up?</blockquote>
No, only rename one and add two new.

moreaut:
We have unit tests for all of the other generated yaml files and all the other kubernetes objects that are created at runtime by the operator (e.g. the server pods).  We need to have unit tests for all the voyager artifacts too.

It will be a fair amount of work, and I can sort of see waiting until you create them from the
operator (instead of from the create scripts) if this will be done very soon.  Otherwise, I think
you should hold off on submitting this until you write the unit tests.

As a starting point, look at the tests under operator/src/test/java/oracle/kubernetes/operator/create and the utilities under operator/src/main/java/oracle/kubernetes/operator/utils.

When you get to the point of writing the unit tests, we can have a zoom conf so I help get you pointed in the right direction.

Regardless, long term, this feature is not complete until it has unit tests.

moreaut:
On backwards compatibility - what happens if a customer installs GA and sets up a domain with voyager, then installs the next version of the operator with your changes and tries to run the old domain?  Will it just work?  Will/should the customer upgrade the old voyager artifacts to your new ones?  If so, how is this done?

lilyhe123:
About UT:
I agree that UT is always good thing to have. But for now I don't think it's high priority for voyager related function given that:
- The voyager yaml files are subject to change.
- The voyager yaml files are generated by create-domain script, not by operator code. It's not efficient to me to test a shell script by java classes and the tests are only to validate the generated yaml files.

About backwards compatibility:
I didn't think about cases like that before and I don't think it's a critical case need to support to voyager related function. Voyager is a 3th-party component. What we do in create-domain script can be considered as a POC with a simple scenario to show that WLS operator can work well with voyager. What's the best way to integrate WLS operator with Voyager is a subject to change also.

LorenKonkus:
In  my opinion, we could consider a releasing a feature as a POC only if it's entirely independent. In this case, we're  making changes to supported scripts and code,  and so it's not a POC, we're supporting it, and the level of quality of those changes and their support by tests and  an upgrade strategy needs to be  consistent with the rest of the operator.

tbarnes-us:
FYI:   One 'domain model' Ryan and I are looking at is to assume, just as Lily states, that LB yaml needs to be freely customizable.  The control of flow will be something like:  (1) we supply client side sample yaml with 'macros' that's suitable for most customers, (2) customers can modify if needed, (3) we supply a very simple client side script for deploying all yaml & scripts into a config map in k8s, (4) the operator uses the contents of this config map to generate a new domain (and does macro substitution and verification as needed).

mriccell:
Yes but I agree with Loren, we have a GA version of the Operator and we have announced support for Voyager, meaning that any script we provide or yaml no matter how simple can not be POC quality.  It needs to follow the naming standards, format standards, and needs to pass tests.

lilyhe123:
Maybe POC is not an accurate word. We need to do thorough testing before we claim whatever we support. And I agree with what you said mostly. 

My point is that writing UT with java code to test a shell script is not a good approach, specially given that the function is within the shell script and have nothing to do with operator code. Say if anything is broken in the shell script, what need to be fixed is the script itself, not the operator code. Do we want to fail building the operator image? I assume the preference is no. But if we including the scripts UT in the operator, it will fail building the image.

moreaut:
We already have similar unit tests for all of the other k8s resources (including load balancer related ones) that are created by shell scripts.  I don't see why your case is different.

Yet, if you change a script or yaml template, the operator build will fail.  That's what we want since
it's our signal that we might have just introduced a backwards compatibility issue.

PR #331, opened 2018-06-13
Allow WebLogic Docker image to be specified in inputs file for genera…

lennyphan:
…ting CRD

Added a new 'weblogicImage' value to the  create-weblogic-domain-inputs.yaml  file to allow the ability to explicitly specify different version of a WebLogic Docker image to be used.  This is a usability enhancement as requested in JIRA WLS-66115.

### DIFF
--- a/kubernetes/create-weblogic-domain-inputs.yaml
+++ b/kubernetes/create-weblogic-domain-inputs.yaml
@@ -40,6 +40,9 @@ managedServerNameBase: managed-server
 # Port number for each managed server
 managedServerPort: 8001
 
+# WebLogic Docker image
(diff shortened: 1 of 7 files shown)


moreaut:
What will happen if a customer uses an old inputs file that doesn't specify the image name?
i.e. what it the backwards compatibility story?

lennyphan:
Changes:
1. Handle backward compatability by initializing weblogic image value to 'store/oracle/weblogic:12.2.1.3", if not defined.  This is the default value we fall back to if 'image' value is not specified in CRD.
2. Update the create/delete domain jobs with the specified weblogic image.

PR #332, opened 2018-06-13
Issue 330:  Fix an issue in create domain script for one Apache use case 

doxiao:
The create domain script generates an invalid apache yaml file when 
loadBalancerVolumePath is set and loadBalancerExposeAdminPort is false.

### DIFF
--- a/kubernetes/internal/create-weblogic-domain.sh
+++ b/kubernetes/internal/create-weblogic-domain.sh
@@ -583,6 +583,16 @@ function createYamlFiles {
       sed -i -e "s|#   value: '%ADMIN_PORT%'|    value: '%ADMIN_PORT%'|g" ${apacheOutput}
     fi
 
+    if [ ! -z "${loadBalancerVolumePath}" ]; then
+      sed -i -e "s:%LOAD_BALANCER_VOLUME_PATH%:${loadBalancerVolumePath}:g" ${apacheOutput}
+      sed -i -e "s:# volumes:volumes:g" ${apacheOutput}
+      sed -i -e "s|# - name: %DOMAIN_UID%-apache-webtier|- name: %DOMAIN_UID%-apache-webtier|g" ${apacheOutput}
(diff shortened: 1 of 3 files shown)


kubernetes/internal/create-weblogic-domain.sh
```diff
@@ -583,6 +583,16 @@ function createYamlFiles {
...
     fi
 
+    if [ ! -z "${loadBalancerVolumePath}" ]; then
+      sed -i -e "s:%LOAD_BALANCER_VOLUME_PATH%:${loadBalancerVolumePath}:g" ${apacheOutput}
+      sed -i -e "s:# volumes:volumes:g" ${apacheOutput}
+      sed -i -e "s|# - name: %DOMAIN_UID%-apache-webtier|- name: %DOMAIN_UID%-apache-webtier|g" ${apacheOutput}
+      sed -i -e "s:#   hostPath:  hostPath:g" ${apacheOutput}
+      sed -i -e "s:#     path:    path:g" ${apacheOutput}
+      sed -i -e "s:# volumeMounts:volumeMounts:g" ${apacheOutput}
+      sed -i -e "s:#   mountPath:  mountPath:g" ${apacheOutput}
```
tbarnes-us: A pattern I've seen in other templates for commenting/uncommenting lines of yaml code is to make the comment symbol itself a parameter, rather than search and replace the contents of each line.   I think this approach makes converting the template to a helm template much easier.    For example, see "EXPOSE_T3_CHANNEL_PREFIX" in the domain-custom-resource-template.yaml file.

tbarnes-us: This is not a show-stopper of course, as I know this pattern existed before the change as well.

tbarnes-us:
Aside from the my comment which refers to an issue that's independent of this fix, the code LGTM.

kubernetes/internal/create-weblogic-domain.sh
tbarnes-us: The fix itself LGTM.

PR #360, opened 2018-07-25
integration tests with helm charts

alai8:
updated integration test src/integration-tests/bash/run.sh and cleanup.sh to use helm charts - 1.0 style domain charts and 2.0 style operator helm charts, instead of scripts.

wercker test would still be using install scripts with this change. Will continue to get wercker to work with helm charts in OWLS-67340

### DIFF
--- a/src/integration-tests/bash/run.sh
+++ b/src/integration-tests/bash/run.sh
@@ -671,7 +671,7 @@ function create_image_pull_secret_wercker {
 
 function op_define {
     if [ "$#" != 5 ] ; then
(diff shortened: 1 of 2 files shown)


src/integration-tests/bash/run.sh
```diff
@@ -642,13 +670,14 @@ function create_image_pull_secret_wercker {
 #
 
 function op_define {
-    if [ "$#" != 4 ] ; then
-      fail "requires 4 parameters: OP_KEY NAMESPACE TARGET_NAMESPACES EXTERNAL_REST_HTTPSPORT"
+    if [ "$#" != 5 ] ; then
+      fail "requires 4 parameters: OP_KEY NAMESPACE TARGET_NAMESPACES EXTERNAL_REST_HTTPSPORT SETUP_KUBERNETES_CLUSTER"
```
moreaut: Typo - should say 'requires 5 parameters'

PR #361, opened 2018-07-25
Add 1 0 domain helm charts

alai8:
Merge the 1.0 style domain helm charts from feature/helm_operator_1_1 branch to develop


alai8:
Regarding testing, I have done manual testing, as well as passing integration testing both locally and on jenkins, after modifying src/integration-tests/bash/run.sh to use helm charts instead of scripts. The changes to run.sh and cleanup.sh are not being checked into the develop branch, but they were just merged into the operator 1.1 branch.

The README.md was brought over from the old helm charts as well, with some modifications to get them up to date with the values.yaml in the develop branch. It most probably needs more work but it is probably ok as a start.

PR #363, opened 2018-07-27
Add doc for updating a domain

doxiao:
Add a new section to cover the steps for updating a domain.

@rjeberhard @rosemarymarano @mriccell


README.md
```diff
@@ -195,12 +195,12 @@ This can be achieved by following the steps below. For convenience of discussion
...
 * Recreate the domain using the [Create WebLogic domain](kubectles/create-weblogic-domain.sh) script with the updated create WebLogic domain inputs file.
 
-If for some reason the last step fails, repeat the steps above after deleting the failed create WebLogic domain job using the following command. The `<output-dir>` is the directory that is passed to the create WebLogic domain script via `-o`, which is `/path/to/weblogic-operator-output-directory` in our example if the same command is used to recreate the domain.
+If for some reason the last step fails, repeat the steps above after deleting the failed create WebLogic domain job using the following command. The `<output-dir>` is the directory that is specified to the create WebLogic domain script via `-o`, which is `/path/to/weblogic-operator-output-directory` in our example if the same command is used to recreate the domain.
```
rosemarymarano: If for some reason the last step fails, then repeat the steps above after deleting the failed create WebLogic domain job, using the following command. The `<output-dir>` is the directory that is specified to the create WebLogic domain script via the `-o` option, which is `/path/to/weblogic-operator-output-directory` in our example if the same command is used to recreate the domain.

PR #374, opened 2018-08-07
Fix helm chart dictionary merging, add support for the operator image…

moreaut:
Enable and fix all of the commented out create operator unit tests.

Start adding test support for customer supplied internal operator REST cert/key

Add operator image pull secrets to helm (forgot to do it in the 1st cut of the chart)

Rework how the operator helm chart merges dictionaries so that we can support null values

Not sure if we should wait on pulling this until after the 1.1 branch is cut.

http://wls-jenkins/job/weblogic-kubernetes-operator/482/


kubernetes/src/test/java/oracle/kubernetes/operator/helm/HelmOperatorValues.java
```diff
@@ -100,6 +119,17 @@ private void addDomainsNamespaces(HashMap<String, Object> map) {
...
 
+  private void addOperatorImagePullSecrets(HashMap<String, Object> map) {
+    String secretName = getWeblogicOperatorImagePullSecretName();
+    if (secretName != null && (!secretName.isEmpty())) {
+      Map<String, String> secret = new HashMap();
+      secret.put("name", secretName);
+      List<Map<String, String>> secrets = new ArrayList();
+      secrets.add(secret);
+      map.put("operatorImagePullSecrets", secrets);
+    }
```
russgold: Consider using the built-ins to handle singleton lists like this. 

map.put("operatorImagePullSecrets", 
           Collections.singletonList(
                       Collections.singletonMap("name", secretName)));

is easier to understand at a glance, especially for something like a map of lists of maps.

kubernetes/src/test/java/oracle/kubernetes/operator/helm/HelmOperatorValues.java
```diff
@@ -100,6 +119,17 @@ private void addDomainsNamespaces(HashMap<String, Object> map) {
     }
   }
 
+  private void addOperatorImagePullSecrets(HashMap<String, Object> map) {
+    String secretName = getWeblogicOperatorImagePullSecretName();
+    if (secretName != null && (!secretName.isEmpty())) {
```
russgold: Prefer a function from Guava: 

Strings.isNullOrEmpty(secretName)

returns true for both null and ""

PR #376, opened 2018-08-07
Fixes #375,  no namespace passed to Volume template

### LINKED ISSUE #375: Operator fails when changing namespace from default
In the input section of the `create-weblogic-domain-inputs.yaml` their is a section for namespace.  If I change the value to something other than default the scripts fail.  This is because the template for the persistent volume does not contain a namespace.  The Job API assumes that the specified volume and claim are all in the same namespace. 

Fix is to add namespace to the PersistentVolume template

I will attach PR

jeffellin:


rjeberhard:
@jeffellin Sorry I've been unavailable for a few days...  My reading of the K8s doc says that PV's are non-namespaced.  PVC's do live in a namespace though.  Why do you think that PV's require a namespace?

jeffellin:
@rjeberhard because it fixed the issue I was having,   but now that I tried a clean K cluster I am not able to reproduce, so we can close this for now.

PR #380, opened 2018-08-08
OWLS-67486: Expose internal env to pod

doxiao:
Cherry-pick a fix in master to develop.

PR #382, opened 2018-08-08
Podhelper refinement

russgold:
Includes Dongbo's symbol substitution fix
Adds unit test for the PodHelper
Refactors the code to eliminate duplication, simplify code analysis, and prepare for templating of pods.


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -0,0 +1,605 @@
...
+  private static void hideAdminUserCredentials(List<V1EnvVar> vars) {
+    addEnvVar(vars, "ADMIN_USERNAME", null);
+    addEnvVar(vars, "ADMIN_PASSWORD", null);
+  }
+
+  static void addEnvVar(List<V1EnvVar> vars, String name, String value) {
+    vars.add(new V1EnvVar().name(name).value(value));
+  }
+
+  void doSubstitution(List<V1EnvVar> vars) {
```
doxiao: Two comments to the env token substitution part (refinement to PR #380).

1) Generic substitution is powerful, but one iteration of substitution may not be sufficient to cover all use cases. 
For example, there are three environment variables in order (A, B, C).
A references $(C)
C references $(B)
After one iteration, A still has an un-substituted reference to $(B).

2) It may not safe to allow an Pod env to reference all internal environment variables. For example, do we allow an environment variable to reference ADMIN_PASSWORD?

russgold: Point (2) is very well taken, and I need to fix that.

Point (1) is more complex. My thinking was that we would not (yet) guarantee behavior in such cases; however, I think it can be done by simply iterating until no more changes are made. I will write some tests to verify that.

russgold: On further examination, there is no risk with regard to ADMIN_PASSWORD, as we override it and set it to null.

moreaut: Where did the requirement for doSubstition come from?  I don't think it aligns with the 2.0 domain helm chart wiki - 
http://aseng-wiki.us.oracle.com/asengwiki/display/ASDevWLS/2.0+Domain+Helm+Chart
See 4.4.9 Template Parameters

doxiao: The requirement is from Verizon, see OWLS-67486: Expose internal env to pod.

russgold: I have added a new test and modified the code to address your case (1)

doxiao: The modified substitution code looks good to me.

moreaut: The 2.0 domain helm chart wiki supports by letting you set javaOptions:"...%SERVER_NAME%..."
Still don't see the need for general env var substition.
i.e. 2.0 takes what the customer configured, which can use a fixed set of variables from the table in 4.4.9, and makes a pod template.  Then the operator runtime looks up the pod template, and knows what the variable values should be for that server, and substitutes them.
i.e. it isn't based on the list of env vars.

doxiao: The original change in PR #380 only allows a number of internal environment variables to be referenced by external pod environment variables. 
I don't strongly object the general substitution approach that is proposed here as long as we don't accidentally expose things that we don't want the external to access.

We need this fix in 1.1, where the new helm charts are not available.

russgold: I agree that using predefined variables rather than env vars makes sense. My new change addresses that, and thus will be more in line with the 2.0 direction.

doxiao: The part for exposing a predefined set of variables to the Pod env looks good to me.

PR #383, opened 2018-08-09
Feature/helm operator namespace

moreaut:
Change the operator helm chart to get the namespace from helm install --namespace (instead of operatorNamespace in values.yaml), default the operator service account name to 'default'.

Change the helm testing infra to accept the --name parameter too since we'll use it as the domain uid in the domain helm chart.

Fix a few minor helm issues at the same time.

Passed jenkins: http://wls-jenkins/job/weblogic-kubernetes-operator/490


moreaut:
Includes the https://github.com/oracle/weblogic-kubernetes-operator/pull/374 changes too.

src/integration-tests/bash/run.sh
```diff
@@ -3015,11 +3036,11 @@ function test_suite {
     kubectl create namespace test1 2>&1 | sed 's/^/+/g' 
     kubectl create namespace test2 2>&1 | sed 's/^/+/g' 
 
-    if ! [ "$USE_HELM" = "true" ]; then
-      # do not create ooperator namespace when using helm charts
-      kubectl create namespace weblogic-operator-1 2>&1 | sed 's/^/+/g' 
-      kubectl create namespace weblogic-operator-2 2>&1 | sed 's/^/+/g' 
-    fi
+    # do not create operator namespace when using helm charts
```
alai8: Should we remove this comment since we need to create operator namespace now?

src/integration-tests/bash/run.sh
```diff
@@ -2502,9 +2500,31 @@ function wait_for_operator_shutdown {
...
+}
+
+function wait_for_operator_deployment_deleted {
+  namespace=$1
+  deployment="weblogic-operator"
+  deleted=false
+  iter=1
+  trace "waiting for the operator deployment ${deployment} in the namespace ${namespace} to longer exist"
+  while [ ${deleted} == false -a $iter -lt 101 ]; do
+    helm get deployment -namespace ${namespace} ${deployment}
```
alai8: i am not familiar with helm get deployment, but could you double check the syntax? I usually see namespace option with 2 dashes instead of 1. Thanks

alai8:
Do you mean "--namespace" instead of "--name"?

src/integration-tests/bash/run.sh
moreaut: Good catch - I meant 'kubectl delete ...'

src/integration-tests/bash/run.sh
moreaut: yes

moreaut:
Addressed Anthony's issues, jenkins still passes http://wls-jenkins/job/weblogic-kubernetes-operator/491/

PR #384, opened 2018-08-13
Better conflict flow resolution so that Operator resumes

rjeberhard:


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -670,6 +676,37 @@ public NextAction onSuccess(
...
   }
 
+  private static Step getRereadDomainConflictStep(
+      DomainPresenceInfo info, V1ObjectMeta meta, Step next) {
+    return new CallBuilder()
+        .readDomainAsync(
+            meta.getName(),
+            meta.getNamespace(),
+            new ResponseStep<Domain>(next) {
+              @Override
```
russgold: If you use DefaultResponseStep, you will get much of this behavior automatically. That will improve readability by avoiding needless duplication.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -444,6 +515,77 @@ public NextAction onSuccess(Packet packet, CallResponse<V1Pod> callResponse) {
...
+    @Override
+    public NextAction apply(Packet packet) {
+      String domainUID = getDomainUID();
+      Step list =
+          new CallBuilder()
+              .withLabelSelectors(forDomainUid(domainUID))
+              .listPersistentVolumeAsync(
+                  new ResponseStep<V1PersistentVolumeList>(getNext()) {
+                    @Override
+                    public NextAction onFailure(
```
russgold: Use DefaultResponseStep to get this onFailure behavior without adding code.

Also, consider pulling anonymous inner classes out into named classes, to keep methods short and easier for the next person to read and modify, and please use the newer version of the onSuccess method that includes CallResponse as a replacement for the deprecated version that takes result plus statusCode plus responseHeaders.

operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
```diff
@@ -196,7 +198,7 @@ V1Container getCreatedPodSpecContainer() {
 
   @Test
   public void whenNoPod_createIt() {
-    expectReadPod(getPodName()).failingWithStatus(HttpURLConnection.HTTP_NOT_FOUND);
+    onAdminExpectListPersistentVolume();
```
russgold: It looks as though you are adding onAdminExpectListPersistentValue to each and every test method. If so, avoid duplication by adding it to a setUp method.

rjeberhard: Thanks, I'll look.  That may have happened because I worked on each test case somewhat independently until I got them to work.

PR #388, opened 2018-08-17
OWLS-67486: Update doc for expose predefined env variables

doxiao:


doxiao:
This is covered in PR #391

PR #389, opened 2018-08-21
updated fix for OWLS-67651: startupControl=ALL instability

lennyphan:
Use WLS managed server name, instead of listenAddress, when handling startupControl=ALL.  This resolves instability of managed servers due to changed env properties when startupControl=ALL. This bug fix is much more simple than the previous fix for this issue.


rjeberhard:
@lennyphan  Is this in develop as well?

PR #396, opened 2018-08-27
Op20scriptenhdev

tbarnes-us:
WL Pod liveness, readiness, start, stop script update: 

* use env vars instead of command line args
* parameterize domain-home + domain-uid + nodemgr-home
* reduce boot time by about 15 seconds (replace hard coded sleep and instead monitor node mgr output to find when it's ready)
* modify run.sh to provide finer-grained control over when to enable helm

https://jira.oraclecorp.com/jira/browse/OWLS-67684


tbarnes-us:
@moreaut The run.sh changes are intentional.   They tighten up the logic around its existing "USE_HELM" env var a little bit to provide better tracing, error reporting, and to properly honor it when it's set externally prior to calling run.sh.   USE_HELM will go away completely once more of 2.0 is in place and the old ways get put out to pasture.

moreaut:
The changes look good to me.

PR #400, opened 2018-08-29
OWLS-67274: make generated WLS domain artifacts consistent with Kubernetes resource names

doxiao:
1) Added the code to check and fail create domain script if the domainUID contains "_". (it already fails if domainUID contains upper case letters). Add unit test case for this.
2) Changed the listenaddress in config.xml to match the server service's DNS name on the K8S side.
3) Changed URLs in the node manager properties to match  the server services' DNS name on the K8S side.
4) Modified the integration test run.sh to cover use cases where clusterName or/and managedServerNameBase contain invalid characters.  The test suite passed with invalid names and all load balancer types.
5) Updated the doc
Note when use helm charts to create a domain, the invalid characters are not still supported with the changes in this PR, and run.sh only test invalid characters when helm charts are not used.  We need to implement the similar logic in the helm chart use case once we decide how to do pre-installation processing in helm charts. 

@rjeberhard @tbarnes-us @moreaut @rosemarymarano


moreaut:
RE:  We need to implement the similar logic in the helm chart use case once we decide how to do pre-installation processing in helm charts.

The 2.0 plan is to remove the domain helm chart (as well as the create domain script), and tell the customer to create a domain resource by hand.  The operator runtime will be responsible for validating it (i.e. the logic that checks for invalid domain uids will need to move into the operator runtime, not
the domain helm chart).

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -26,6 +26,8 @@ for varname in domain_uid \
                domain_home \
                log_home \
                nodemgr_home \
+               service_name \
+               admion_hostname \
```
tbarnes-us: Typo - admion --> admin

tbarnes-us:
Change LGTM overall, only saw the minor issue of the 'admion' typo.

operator/src/main/resources/scripts/startServer.sh
doxiao: Good eye. It is in a debug/info statement so testing did not catch it. fixed.

kubernetes/create-weblogic-domain-inputs.yaml
```diff
@@ -14,7 +14,7 @@ adminServerName: admin-server
 domainName: base_domain
 
 # Unique id identifying a domain.
-# This id must be lowercase and unique across all domains in a Kubernetes cluster.
+# This id must not contain "_", and must be lowercase and unique across all domains in a Kubernetes cluster.
```
rosemarymarano: must not contain an underscore ("_"), and must be ...

Also, "id" should be uppercase, ID.

site/creating-domain.md
```diff
@@ -154,7 +154,7 @@ Finally, run the create script, pointing it at your inputs file and output direc
 The script will perform the following steps:
 
 *	Create a directory for the generated Kubernetes YAML files for this domain.  The pathname is `/path/to/weblogic-operator-output-directory/weblogic-domains/<domainUID>`.
-*	Create Kubernetes YAML files based on the provided inputs.
+*	Create Kubernetes YAML files based on the provided inputs. Note that the names of the Kubernetes resources in the generated YAML files may be formed with the value of some of the properties specified in the create-domain-job-inputs file. Those properties include the adminServerName, clusterName and managedServerNameBase. If those values contains any charaters that are nvalid in a Kubernetes service name, those charaters are converted to valid values in those YAML files. For example, a upper case letter is converted to lower case and "_" is converted to "-".
```
rosemarymarano: A few typos to fix:
**"If those values contains any charaters that are nvalid in a Kubernetes service name," change to"If those values contain any characters that are invalid in a Kubernetes service name,"
**"those charaters are converted to valid values in those YAML files. For example, a upper case letter is converted to lower case and "_" is converted to "-"." change to "those characters are converted to valid values in those YAML files. For example, an uppercase letter is converted to a lowercase letter and an underscore ("_") is converted to a hyphen ("-").

PR #403, opened 2018-08-30
Isolate operator from domain details

russgold:
With this change, there is a new API (DomainConfigurator) for the operator runtime and tests to modify the domain model, that should (with some exceptions) be resilient to the changes we're going to make in the model itself. This should make it easier to make the changes we want, and have them quickly integrated with the runtime.


model/src/main/java/oracle/kubernetes/weblogic/domain/v1/WlsDomain.java
```diff
@@ -0,0 +1,30 @@
...
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.weblogic.domain.v1;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Represents information from a WLS domain that is needed in order to validate an operator domain.
+ */
+public interface WlsDomain {
```
alai8: Who are the users of this interface? I see that WlsDomainConfig implements it but I can't seem to find anywhere else that references the WlsDomain interface.

russgold: No, that's part of the plan to move the validation into the domain itself, but I didn't move it just yet

alai8: I see. The changes related to wlsconfig look fine to me.

PR #409, opened 2018-09-07
implement initial domain 2 model

russgold:
Initial implementation of the new domain model
- includes scoping rules
- implements the classes used by the runtime
- can read either v1 or v2 domain models

Still needs:
- domain watcher / domain job enhancements
- updated schema
- new server startup behavior
- additional fields for configuring pods and services

### DIFF
--- a/model/src/main/java/oracle/kubernetes/weblogic/domain/ConfigurationNotSupportedException.java
+++ b/model/src/main/java/oracle/kubernetes/weblogic/domain/ConfigurationNotSupportedException.java
@@ -1,0 +1,24 @@
+// Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Licensed under the Universal Permissive License v 1.0 as shown at
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.weblogic.domain;
+
+/**
+ * An exception to be thrown on an attempt to configure a field not supported by the domain model.
+ * Only used in unit testing.
+ */
+public class ConfigurationNotSupportedException extends RuntimeException {
+  private final String context;
+  private final String field;
+
+  public ConfigurationNotSupportedException(String context, String field) {
+    this.context = context;
+    this.field = field;
+  }
+
+  @Override
+  public String getMessage() {
+    return String.format("Configuring field %s is not supported for a %s", field, context);
+  }
+}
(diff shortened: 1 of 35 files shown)


model/src/main/java/oracle/kubernetes/weblogic/domain/ConfigurationNotSupportedException.java
```diff
@@ -0,0 +1,24 @@
...
+  private final String field;
+
+  public ConfigurationNotSupportedException(String context, String field) {
+    this.context = context;
+    this.field = field;
+  }
+
+  @Override
+  public String getMessage() {
+    return String.format("Configuring field %s is not supported for a %s", field, context);
```
rjeberhard: I18n

model/src/main/java/oracle/kubernetes/weblogic/domain/ServerConfigurator.java
```diff
@@ -4,10 +4,20 @@
 
 package oracle.kubernetes.weblogic.domain;
 
+/** An interface for an object to configure a server in a test. */
+@SuppressWarnings("UnusedReturnValue")
```
rjeberhard: Just curious...  I've seen you add lots of these statements to suppress warnings, which I'm guessing are specific to IntelliJ.  Since I use Eclipse (and the team is mixed), I have no way to know which warnings I should suppress, or no longer need to suppress.  Do you know of any standard way to configure the IDE's to respect the same set of warnings?

model/src/main/java/oracle/kubernetes/weblogic/domain/ConfigurationNotSupportedException.java
russgold: The configuration classes are intended for unit tests only. As noted in oracle.kubernetes.weblogic.domain.DomainConfigurator:

/**
 * Configures a domain, adding settings independently of the version of the domain representation.
 * Note that the configurator uses a predefined domain schema, and should only be used for testing.
 * Using it in the runtime runs the risk of corrupting the domain.
 */

The code is sitting in "main" rather than "test" because it is needed for testing, not only the model itself, but also the operator runtime and the installation (and ideally) integration tests. Because it depends on the model and is also used by model tests, I see no other place to put it.

model/src/main/java/oracle/kubernetes/weblogic/domain/ServerConfigurator.java
russgold: Unfortunately, the Java team hasn't published a standard list of warnings, as far as I know. In Intellij, I can suppress a warning to acknowledge that I am aware of a problem and consider it acceptable. I do that, though, by clicking on the warning and asking the editor to add the annotation.

I think that Eclipse often uses the same annotations, but not always, which is really annoying. I suspect it should be OK to add both, when they use different ones, but I don't know how to do that in Eclipse.

PR #416, opened 2018-09-17
Use additional tests to upgrade helm charts to match scripts

russgold:
In preparation for removing the domain chart piecemeal, ensure that it has the same functionality as the scripts, so that nothing is lost.

### DIFF
--- a/kubernetes/charts/weblogic-domain/templates/_domain-custom-resource.tpl
+++ b/kubernetes/charts/weblogic-domain/templates/_domain-custom-resource.tpl
@@ -27,6 +27,9 @@ spec:
   image: "{{ .weblogicImage }}"
   # imagePullPolicy defaults to "Always" if image version is :latest
   imagePullPolicy: "IfNotPresent"
+  # Identify which Secret containst the credentials to access the docker repository
+  imagePullSecret:
+    name: {{ .weblogicImagePullSecretName }}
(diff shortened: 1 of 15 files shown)


kubernetes/charts/weblogic-domain/templates/_domain-custom-resource.tpl
```diff
@@ -27,6 +27,9 @@ spec:
   image: "{{ .weblogicImage }}"
   # imagePullPolicy defaults to "Always" if image version is :latest
   imagePullPolicy: "IfNotPresent"
+  # Identify which Secret containst the credentials to access the docker repository
```
moreaut: typo - containst

Might not be important to fix since we're planning on removing the domain helm chart.

kubernetes/charts/weblogic-domain/templates/_traefik-security.tpl
```diff
@@ -6,7 +6,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  name: {{ .Release.Name }}-{{ .clusterName | lower }}-traefik
+  name: {{ .Release.Name }}-{{ .clusterName | lower | replace "_" "-" }}-traefik
```
moreaut: It would be better to write a helm function that converts a string to a valid k8s resource name, but I don't think it's worth holding up the review for it.

PR #419, opened 2018-09-25
OWLS-66100: Add sample scripts to create domain on pv including pv/pvc and load balancer

doxiao:
1) Break up the current create domain scripts into separate scripts for pv/pvc, domain home and resources, and load balancer resources. 
2) Create domain script takes PVC name and PV root as additional input properties
3) By default, the scripts only generate Kubernetes resources and do not create them. 
4) Optionally the scripts run kubectl command to create the resources as well.
5) Updated run.sh to use the new scripts as an example.

To do: add an option to customize the tooling (for example WDT) and the scripts to generate a WebLogic domain home. Currently the scripts use offline wlst and a simple domain script.

@moreaut @rjeberhard


tbarnes-us:
LGTM overall.

Minor suggestion:   shorten long names by removing 'weblogic-sample' from them?  For example, create-weblogic-sample-domain-load-balancer-inputs.yaml becomes create-weblogic-sample-domain-load-balancer-inputs.yaml.   It seems OK to do this IMO since the files are already located in the samples directory.

Long term suggestions (as part of some future JIRA - not this one):   The scripts and templates are lengthy, which detracts from the goal of keeping the samples simple.   Longer term, it'd help to simplify, shorten, and componentize them in various ways.  For example, move validation logic into the operator/introspection-job/etc, use a single helper script for all macro substitutions instead of bunches of seds (I have sample code for this), and break out scripts such as the off-line WLST domain creation script that are embedded in cm yaml and instead generate the cm by pulling in the individual files.

kubernetes/samples/scripts/common/validate.sh
```diff
@@ -0,0 +1,335 @@
...
+        validationError "Invalid value for clusterType: ${clusterType}. Valid values are 'CONFIGURED' and 'DYNAMIC'."
+      ;;
+    esac
+  fi
+}
+
+#
+# Function to validate the weblogic domain storage reclaim policy
+#
+function validateWeblogicDomainStorageReclaimPolicy {
```
moreaut: I thought the customer was going to configure the PV & PVC, not the sample, therefore the inputs file wouldn't contain the properties needed to configure them.

doxiao:
@moreaut I have extended the task to do end-to-end samples. As we discussed during the scrum meeting, the development of the samples will take multiple iterations as the 2.0 project moves forward. The initial version uses scripts for load balancer while the load balancer helm charts are still being developed. We'll replace the scripts with charts once the charts are available. Having scripts for pv/pvc provides a convenient way to do a quick testing or kick-the-tires, and it helped me in testing the new create domain sample scripts from end-to-end.

moreaut:
The changes look good to me.

PR #423, opened 2018-09-26
Clean up run.sh for operator and fix a merge error

doxiao:
1) After recent merges, run.sh fails if USE_HELM is set to false. We always use Helm charts for the operator now, so we need to clean up the test so that it does not care about USE_HELM for the operator part. 
2) Some of the code in run.sh that was added in PR#419 was lost in the merges of other RPs to develop. Putting those code back.

Signed-off-by: doxiao <dongbo.xiao@oracle.com>


src/integration-tests/bash/run.sh
```diff
@@ -1299,7 +1289,23 @@ function create_domain_pv_pvc_load_balancer {
       trace "helm install output:"
       cat $outfile
     else
-      fail "create-weblogic-domain.sh is longer supported"
```
moreaut: Are you sure that the code to call these sample scripts should be in this 'else' clause?

moreaut: So, USE_HELM now is used to say whether to setup the domain using the domain helm chart, or whether to setup the domain using the sample scripts?

I think it used to indicate whether to setup both the operator and domain with helm, v.s. the old create scripts.

Maybe it should be renamed to USE_DOMAIN_HELM_CHART?

moreaut: On the other hand, run.sh is supposed to be deleted soon so it's probably not worth the effort to rename it.

doxiao: We could rename the property, but I don't think USE_HELM is causing any confusion because we don't support setting up operator using scripts any more. So USE_HELM only means use helm for the domain creation or not. It would make more sense to rename the property (and add another property USE_OPERATOR_HELM_CHART) if we support both script and helm for operator, and support mixed cases between operator and domain.

moreaut: There are no plans to configure the operator with scripts - we only support helm.

Anyway, I wouldn't bother renaming the property since run.sh will go away soon.

rjeberhard:
@moreaut @doxiao Is this discussion resolved?

moreaut:
Dongbo, could you make sure that run.sh has execute permissions (chmod +x) on your branch?

Somehow the permission recently got removed, and I just fixed it in develop earlier this afternoon.

Thanks, -Tom

moreaut:
FYI Jenkins passed:

http://wls-jenkins/job/weblogic-kubernetes-operator/571/
http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/136/

PR #425, opened 2018-09-27
document how to update the plugin WL module configuration

simon-meng-cn:


site/apache.md
```diff
@@ -446,6 +446,20 @@ The generated YAML files will look similar except with un-commented entries like
 
 ```
 
+### Update the plugin WL module configuration
+
+
+Sometimes users need to add new WebLogic domains/clusters to a running system. Users can update or add new Apache plugin configuration first. Then run the following commands to restart Apache HTTP Server gracefully:
```
rosemarymarano: Re-write as: "Users can update or add the new Apache plugin configuration first, then run the following commands to restart the Apache HTTP Server gracefully:"

PR #447, opened 2018-10-19
Enable Domain 2.0 model

russgold:
This commit adds support for exportedNetworkAccessPoints and verifies the new startup policy behavior. The unit tests are updated to use domain 2.0 definitions.

models/src/test/resources contains some example domain yaml files, demonstrating the supported domain 2 features. Domain 1 is still supported for backwards compatibility.


rjeberhard:
Merged through command line

PR #465, opened 2018-10-29
Feature/java integration tests

vanajamukkara:
A domain with mostly default values for the attributes(only few customized) from sample domain inputs  mainly default exposeAdminT3Channel, exposeAdminNodePort values is added.

Basically a domain should be started and scaleup/scaledown using operator rest endpoint, domain destroy/create, operator destroy/create should all work fine even with default values for exposeAdminT3Channel and exposeAdminNodePort which is false.

### DIFF
--- a/integration-tests/README.md
+++ b/integration-tests/README.md
@@ -14,14 +14,14 @@ The tests currently run in three modes: "Wercker", "Jenkins", and "standalone" O
 
 Java integration tests cover all use cases from run.sh. The below are the use cases:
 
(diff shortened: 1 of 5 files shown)


vanajamukkara:
Both wercker and jenkins passed with these changes.
http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/233/consoleFull

anpanigr:
I am OK with the Change

PR #469, opened 2018-10-31
WIP: Add apache load balancer sample

simon-meng-cn:
The change contains following content:
1. Update the apache-webtier helm chart to support custom SSL cert/key.
2. Set environment variables to adapt current default configuration
3. Provide two samples for apache load balancer.


doxiao:
How do we handle custom SSL cert/key?

simon-meng-cn:
Currently, the custom mod_ssl configuration file is located in the apache container. We have to rebuild the apache plugin docker image to handle the custom SSL key/certificate.
One idea is to put that configuration file in the volume path. But I am not sure if it is feasible. Let me experiment again and then come to a conclusion.

doxiao:
> Currently, the custom mod_ssl configuration file is located in the apache container. We have to rebuild the apache plugin docker image to handle the custom SSL key/certificate.
> One idea is to put that configuration file in the volume path. But I am not sure if it is feasible. Let me experiment again and then come to a conclusion.

One could provide the SSL cert/key at the startup time of the Apache webtier docker container. Please refer to "Provide Your Own Certificates" section of the doc at https://github.com/oracle/docker-images/blob/master/OracleWebLogic/samples/12213-webtier-apache/README.md.
One approach to handle this in the Operator is to allow user to configure a secret that holds the certificates, and the Apache pod can access the certificates at startup time.

simon-meng-cn:
We need change the docker file to build a new apache plugin docker image. Then we can provide custom  configuration file for mod_ssl similar as mod_weblogic. The custom configuration can be located in volume path. It also support generate the SSL cert/key at the startup time.

doxiao:
I already added the mod_ssl to the Apache plugin webtier docker image a while ago. See https://github.com/oracle/docker-images/blob/master/OracleWebLogic/samples/12213-webtier-apache/custom_mod_ssl_apache.conf.sample.
Monica has been working on getting the docker image to the docker hub. 
If you are talking about the docker image in our local repository, Brian has updated it with the latest. 
Can you double check? Or I have missed something here.

simon-meng-cn:
The configuration file custom_mod_ssl_apache.conf.sample is really in the apache docker image. But we need enhance it. Currently, this file is located in /etc/httpd/conf.d. We need update it to the similar style as custom_mod_wl_apache.conf.sample. We need the following change:
1. Create a new configuration file named mod_ssl_apache.conf in /etc/httpd/conf.d. The content is as below

LoadModule ssl_module modules/mod_ssl.so

Listen 4433 https

<IfModule mod_ssl.c>
  Include /config/custom_mod_ssl_apache.conf 
</IfModule>

2. Move the  custom_mod_ssl_apache.conf to /config.  It contains SSL parameters.

The /config path can be mount to external volume path. So we can provide custom SSL key/cert easily with this enhancement.

doxiao:
I believe what we have is already sufficient for what we need to support in Operator. Can we have a quick chat on the phone tonight? But I am okay with making more changes to Apache docker image if more enhancement is really needed. I'll check Monica about the status of publishing the docker image.

doxiao:
The change looks good to me.

But I have a question for Ryan.  Shall we leave the cluster service name completely to customers or shall we do necessary formatting in the template?

For example, the service name for a cluster  is <domainUID>-cluster-<clusterName>, and it cannot contain upper case and "_".  The operator converts the names to make them valid DNS names. So the helm charts need to have the correct format and value. We could allow customers to provide domainUID and clusterName in the values yaml, and in the template we format the service name for them and also do the necessary conversion. This is easy to do.

Alternatively we leave everything to customers.

We need to document the format of a cluster service name and the restrictions either way.

I am leaning to doing it in the samples as an example for the customers.

doxiao:
I am okay with merging what we have and do the formatting and conversion enhancement in a separate PR if needed.

kubernetes/samples/charts/apache/default-sample.md
```diff
@@ -0,0 +1,38 @@
+# Apache Load Balancer default sample
+## Configure Apache Webtier as Load Balancer for WLS Domains
+In this section we will demonstrate how to use Apache webtier to handle traffic to backend WLS domains.
+
+### 1. Install WebLogic Server Domain
+Now we need to prepare some backends for Apache to do load balancing.
+
+Create a WebLogic Server domain:
```
doxiao: Can refer the create-weblogic-domain/domain-home-on-pv sample as an example.

kubernetes/samples/charts/apache/default-sample.md
```diff
@@ -0,0 +1,38 @@
...
+- Under default namespace.
+- Domain name is 'domain1'.
+- Deploy a webapp  with url context 'testwebapp'.
```
doxiao: Do we have the webapp jar in the samples?
Do we have the command to deploy the app?

kubernetes/samples/charts/apache/default-sample.md
```diff
@@ -0,0 +1,38 @@
...
+
+### 2. Pull Apache Webtier Docker Image
+Run the following commands to pull Apache webtier docker image from repositry manually.
```
doxiao: This is an internal repo. We need to at least mention that so that users from outside of Oracle is aware of this.

kubernetes/samples/charts/apache/default-sample.md
simon-meng-cn: What I used is the integration test script to create domain and deploy web application. It call WLST command to deploy app.  The testwebapp.war and python script can be found in /shared directory in the managed server pod.

doxiao: For each step in the README, we need to have the supporting files accessible, either as part of the sample itself, or pointers to the location where the files can be accessed.  We also need to document the command or points to another doc where the command is documented.

kubernetes/samples/charts/apache/default-sample.md
mriccell: Yes we need to standarize our samples to all point to other samples in our project.  Please point to sample of domain in pv.

kubernetes/samples/charts/apache/custom-sample.md
```diff
@@ -0,0 +1,86 @@
+# Apache Load Balancer custom sample
+## Configure Apache Webtier as Load Balancer for WLS Domains
+In this section we will demonstrate how to use Apache webtier to handle traffic to backend WLS domains.
```
mriccell: Please make reference to the domain in pv sample

kubernetes/samples/charts/apache/custom-sample.md
```diff
@@ -0,0 +1,86 @@
...
+In this sample, both Apache and WebLogic Server domain instances are located in the namespace `apache-sample`.
+```
+$ kubectl create namespace apache-sample
+```
+
+### 2. Install WLS Domains
+Now we need to prepare some backends for Apache to do load balancing.
+
+Create two WebLogic Server domains: 
+- One domain with name 'domain1' under namespace 'apache-sample'.
```
mriccell: Please base in domain sample in pv created by Dongbo

kubernetes/samples/charts/apache/custom-sample.md
```diff
@@ -0,0 +1,86 @@
...
+WebLogicCluster domain2-cluster-cluster-1:8021
+PathTrim /weblogic2
+</Location>
+```
+- Place the `custom_mod_wl_apache.conf` file in a local directory `<host-config-dir>` on the host machine.
+
+### 4. Pull Apache Webtier Docker Image
+Run the following commands to pull Apache webtier docker image from repositry manually.
+```
+$ docker pull wlsldi-v2.docker.oraclecorp.com/weblogic-webtier-apache-12.2.1.3.0:latest
```
mriccell: When we have this sample in master branch users will not have access to the image in this repository.  
Until we get approval to have an Apache image in the Docker Hub you should point to the GitHub project https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12213-webtier-apache for them to build their own Apache image.

kubernetes/samples/charts/apache/default-sample.md
mriccell: When we have this sample in master branch users will not have access to the image in this repository.
Until we get approval to have an Apache image in the Docker Hub you should point to the GitHub project https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12213-webtier-apache for them to build their own Apache image.

mriccell:
> The change looks good to me.
> 
> But I have a question for Ryan and Monica (@mriccell). Shall we leave the cluster service name completely to customers or shall we do necessary formatting in the template?
> 
> The operator generates the service name for a cluster as <domainUID>-cluster-<clusterName>, and converts the names to make them valid DNS names (for example it converts upper case to lower case and replace "_" with a"_"). So the LB helm charts need to have the correct format and value to connect to the cluster via the cluster service name.
> 
> If we leave the helm charts as is in the PR, we leave it to customers to get the format and conversion correct.
> 
> We could improve the helm charts to allow customers to provide domainUID and clusterName in the values yaml, and in the template we format the service name and also do the necessary conversion. This is easy to do, and we do it in the script samples.
> 
> I am leaning to adding the formatting and conversion in the samples as an example for the customers. This will help usability. What do you think?
> 
> BTW, We need to document the format of cluster service names and the restrictions either way.

I agree that we should provide as much help as possible to make it simple and resilient.  So, we should take inputs for domainUID and cluster name and generate the proper service name for them.  The LB helm charts need to have the correct format and value to connect to the cluster via the cluster service name.   We should validate and modify the cluster service name to be a valid K8S name.

kubernetes/samples/charts/apache-webtier/values.yaml
```diff
@@ -39,3 +39,39 @@ httpNodePort: 30305
...
 # NodePort to expose for https access
 httpsNodePort: 30443
+
+# The VirtualHostName of the Apache HTTP server. It is used to enable custom SSL configuration.
+# If it is set, the Apache HTTP Server is configured to listen to port 4433 for SSL traffic.
+virtualHostName:
+
+# The customer supplied certificate to use for Apache webtier SSL configuration.
+# The value must be a string containing a base64 encoded certificate.
+customCert:
```
doxiao: If the customer does not provide the SSL cert and key, does the sample use the default built-in auto-generated sample cert and key in the apache image? if so, we need to document this.

doxiao: We could also document the steps to generate the cert and key using openssl. 
In addition, I am wondering if it is easier to ask customers to provide the cert and key directly here, or to provide the paths to the cert and key files.

simon-meng-cn: Correct. If the customer enable custom SSL configuration and does not provide the cert and private key, the sample will use the default built-in auto-generated sample cert and key in the apache image.
The behavior is same as the weblogic-operator helm chart, both require that the customer provide base64 encoded string. It is not complex. Execute following command will get the base64 encoded certificate.
base64 -i ${cert_file} | tr -d '\n'

kubernetes/samples/charts/apache/custom-sample.md
```diff
@@ -0,0 +1,96 @@
...
+$ helm install --name my-release --set volumePath=<host-config-dir> --namespace apache-sample apache-webtier
+```
+
+## 6. Run the sample application
+Now you can send requests to different WLS domains with the unique entry point of Apache with different path. Alternatively, you can access the URLs in a web browser.
+```
+$ curl --silent http://${HOSTNAME}:30305/weblogic1/testwebapp/
+$ curl --silent http://${HOSTNAME}:30305/weblogic2/testwebapp/
+```
+You can also access SSL URL `https://${HOSTNAME}:30443/weblogic1/testwebapp/` and `https://${HOSTNAME}:30443/weblogic2/testwebapp/` in your web browser.
```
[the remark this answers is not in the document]
simon-meng-cn: No. The current Apache image is fine. We can use the input parameters of the helm chart to control the SSL configuration.
During helm chart installation, it sets environment variable to enable custom SSL configuration. If customer provides a custom certificate and key, the other two environment variables will be set . These environment variables will configure Apache correctly.
Our sample will show how to use custom SSL configuration.

PR #482, opened 2018-11-07
OWLS-69320: Automatically generate JSON schema from the model

russgold:
This change adds two new modules to the build - one is a schema generator, with annotations that can be applied to the model (the main reason I could not combined them), and one is a maven plugin that is run during the model build. When the build completes, the generated schema can be found under models/target/classes/schema.

We might consider merging the model module into the operator module, as it contains a fair amount of the operator logic, and there doesn't appear to be any value in keeping them separate now.


russgold:
Um.... the descriptions aren't showing up in the schema. Don't merge this yet.

russgold:
Looks good now; check it out!

russgold:
Sounds good. Can you create a JIRA to track this?

PR #485, opened 2018-11-07
Fix domain beans

doxiao:
Make sure all domain beans override equals and hashCode methods, and the calculation is up-to-date.


doxiao:
Passed Java integration test on Jenkins: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/286/.

Issue #49, opened 2018-02-02
docker login required for build/installation


amyles:
In https://github.com/oracle/weblogic-kubernetes-operator/blob/master/site/installation.md there should be a comment to run "docker login" before the docker build command so the images can be pulled from the docker store.

PR #489, opened 2018-11-08
WIP:  domain introspector

tbarnes-us:
**Introspector Related**

- Introspector Features in this Pull
  - Automatically configure listen addresses and log directories (eliminate need to configure these correctly in domain home config.xml)
  - Automatically setup NM (no need to configure via domain config)
  - Operator now gets domain home configuration information from a 'domain job'.  This eliminates need to get the information from a running admin server and/or to supply it in the Domain CR.
  - Perform some basic validation of domain configuration.
  - Encrypt secrets for use by wl-pod, so wl-pod doesn't need to directly mount plain-text secrets

- Operator Internals
  - runs domain home introspection job when a new Domain CR is discovered

- Introspection Job Internals
  - does basic domain validation
  - generates topology.yaml: topology information (server/cluster/domain names, cluster sizes, ports, etc)
  - generates situational-config.xml: situation config to override listen addresses and log home
  - generates boot.properties: encrypted boot properties from admin secret
  - generates userConfigNodeManager.secure and userKeyNodeManager.secure:  encrypted properties for node manager login/password

- Operator Internals
  - No longer depends on getting topology information via Admin REST 'WLSRetriever" calls
  - Gets topology from topology.yaml in introspector job stdout
  - Puts the other job output files into a config-map named \<domain-uid\>-weblogic-domain-introspect-cm
  - Removed ability to modify max-cluster-size setting on domain home (users expected to use sit-config for this purpose)
  - Future:  Admin is still started first before any managed server, it can now be started in parallel instead (post-2.0 JIRA OWLS-70101)

- WL-Pod Internals
  - Uses userConfigNodeManager.secure and userKeyNodeManager.secure from the introspect cm to start its node manager
  - Copies situation config from the introspect cm (for overriding network addresses and log locations)
  - Copies boot.properties from introspect cm for server boot
  - Uses a refactored start script (factors out node manager start)
  - Future: WL Pod still mounts admin secret, but it doesn't need to anymore (OWLS-70100)

**WL Server Log: make configurable, and redirect stdout to pod log**

- Implements 'IncludeServerOutInPodLog' JIRA (true by default)
- Implements configurable logHome

**WIP:  Test Status**

- Passing QUICKTEST locally and on Wercker

- WIP:  Failing frequently on Jenkins during 'lifecycle' tests


kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain.py
```diff
@@ -62,11 +59,6 @@ def getEnvVar(var):
 set('ListenAddress', '%s-%s' % (domain_uid, admin_server_name_svc))
 set('ListenPort', t3_channel_port)
```
tbarnes-us: - We should be able to remove the ListenAddress  for admin server, etc
- Need to check if sit-config is overriding ListenAddress for T3Channel

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -36,6 +36,9 @@ spec:
           value: "false"
         - name: "JAVA_LOGGING_LEVEL"
           value: {{ .javaLoggingLevel | quote }}
+        # TBDTEB Remove this env var
+        - name: "ENABLE_DOMAIN_INTROSPECTOR_JOB"
```
tbarnes-us: No longer needed.  Remove.

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -85,6 +85,8 @@
 /** A Kubernetes Operator for WebLogic. */
 public class Main {
 
+  private static final String ENABLE_DOMAIN_INTROSPECTOR_JOB_ENV = "ENABLE_DOMAIN_INTROSPECTOR_JOB";
+
```
tbarnes-us: We can remove this env var now.

operator/src/main/resources/scripts/start-server.py
```diff
@@ -2,71 +2,56 @@
...
+import base64
+
 def getEnvVar(var):
   val=os.environ.get(var)
   if val==None:
     print "ERROR: Env var ",var, " not set."
     sys.exit(1)
   return val
 
+admin_username = 'TBDRemoveNotUsed'
```
tbarnes-us: Address the TBD

src/scripts/operator.sh
```diff
@@ -26,6 +26,11 @@ else
   DEBUG=""
 fi
 
+if [[ ! -z "$ENABLE_DOMAIN_INTROSPECTOR_JOB" ]]; then
```
tbarnes-us: Remove this (always enable)

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -0,0 +1,722 @@
...
+#   This code partly depends on a node manager so that we can use it to encrypt 
+#   the username and password and put them into files that can be used to connect
+#   to the node manager later in the server pods (so that the server pods don't
+#   have to mount the secret containing the username and password).
+#
+# TBD:
+#
+#   It's not clear if this script supports mixed clusters. It seems to assume
+#   a cluster with a dynamic server means the cluster is purely dynamic.
+#
```
tbarnes-us: This TBD needs to be officially tracked.

model/src/main/java/oracle/kubernetes/weblogic/domain/v1/DomainSpec.java
```diff
@@ -847,6 +930,7 @@ public String toString() {
           .append("imagePullSecrets", imagePullSecrets)
```
doxiao: Need to add logHome to toString(), hashCode() and equals() as well.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -0,0 +1,722 @@
...
+      self.close()
+
+  def addSitCfg(self):
+    self.addSitCfgXml()
+
+  def addSitCfgXml(self):
+    self.writeln("<?xml version='1.0' encoding='UTF-8'?>")
+    self.writeln("<d:domain xmlns:d=\"http://xmlns.oracle.com/weblogic/domain\" xmlns:f=\"http://xmlns.oracle.com/weblogic/domain-fragment\" xmlns:s=\"http://xmlns.oracle.com/weblogic/situational-config\">")
+    self.indent()
+    self.writeln("<s:expiration> 2999-07-16T19:20+01:00 </s:expiration>")
```
moreaut: In src191000, you can just skip the expiration element.  Not sure about earlier branches.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -55,6 +55,19 @@ productionModeEnabled: true
 # The name must be lowercase
 weblogicCredentialsSecretName: domain1-weblogic-credentials
 
+# Name of the Kubernetes secret to access the Docker Store to pull the WebLogic Server Docker image
+# The presence of the secret will be validated when this parameter is enabled.
+#imagePullSecretName:
```
doxiao: imagePullSecretName already exists in the inputs file, see line #49.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -38,7 +38,7 @@ managedServerNameBase: managed-server
 managedServerPort: 8001
 
 # WebLogic Docker image
-image: store/oracle/weblogic:12.2.1.3
+image: store/oracle/weblogic:19.1.0.0
```
rjeberhard: We can't do this until 19.1.0.0 is released.  You're tests will just have to customize this value.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain.sh
```diff
@@ -285,10 +288,11 @@ function createYamlFiles {
   enabledPrefix=""     # uncomment the feature
   disabledPrefix="# "  # comment out the feature
 
-  # For backward compatability, default to "store/oracle/weblogic:12.2.1.3" if not defined in
+  # Operator 2.0 requires using 19.1.0.0 or later in Domain CR.
```
rjeberhard: This is not correct.  We will have a patch for 12.2.1.3 with the situational config enhancements.  These instructions will have to be documented before we release 2.0.

kubernetes/src/test/java/oracle/kubernetes/operator/create/CreateOperatorGeneratedFilesTestBase.java
```diff
@@ -208,10 +208,18 @@ protected ExtensionsV1beta1Deployment getExpectedWeblogicOperatorDeployment() {
                                             newEnvVar()
                                                 .name("JAVA_LOGGING_LEVEL")
                                                 .value(getInputs().getJavaLoggingLevel()))
+                                        .addEnvItem(
+                                            newEnvVar()
+                                                .name("ENABLE_DOMAIN_INTROSPECTOR_JOB")
+                                                .value("true"))
```
rjeberhard: Can we remove this now?

model/src/main/java/oracle/kubernetes/operator/KubernetesConstants.java
```diff
@@ -6,7 +6,7 @@
 
 /** Kubernetes constants. */
 public interface KubernetesConstants {
-  String DEFAULT_IMAGE = "store/oracle/weblogic:12.2.1.3";
+  String DEFAULT_IMAGE = "store/oracle/weblogic:19.1.0.0";
```
rjeberhard: Same comment... Can't do this.

model/src/main/java/oracle/kubernetes/weblogic/domain/v1/DomainSpec.java
```diff
@@ -85,13 +86,25 @@
...
+  /** Whether to include the server .out file to the pod's stdout. Default is true. */
+  @SerializedName("includeServerOutInPodLog")
+  @Expose
+  private String includeServerOutInPodLog;
+
+  /*
    * The WebLogic Docker image.
    *
-   * <p>Defaults to store/oracle/weblogic:12.2.1.3.
+   * <p>Defaults to store/oracle/weblogic:19.1.0.0
```
rjeberhard: And, again :)

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -0,0 +1,232 @@
+// Copyright 2017, 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
rjeberhard: Not 2017

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -0,0 +1,232 @@
...
+import oracle.kubernetes.operator.work.Step;
+
+/** Watches for Jobs to become Ready or leave Ready state */
+public class JobWatcher extends Watcher<V1Job> implements WatchListener<V1Job> {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
+
+  private final String ns;
+
+  // Map of Pod name to OnReady
+  private final ConcurrentMap<String, OnReady> readyCallbackRegistrations =
```
rjeberhard: Is OnReady the right concept?  I'd like to understand.  Would think that you would want to know when the Job is complete or failed?

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -0,0 +1,232 @@
...
+
+  public void receivedResponse(Watch.Response<V1Job> item) {
+    LOGGER.entering();
+
+    LOGGER.fine("JobWatcher.receivedResponse response item: " + item);
+    switch (item.type) {
+      case "ADDED":
+      case "MODIFIED":
+        V1Job job = item.object;
+        Boolean isComplete = isComplete(job); // isReady(job);
```
rjeberhard: I was right!

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -0,0 +1,232 @@
...
+        LOGGER.severe(MessageKeys.JOB_IS_FAILED, job.getMetadata().getName());
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public static boolean isComplete(V1Job job) {
+    V1JobStatus status = job.getStatus();
+    LOGGER.info("++++ JobWatcher.isCpmplete status: " + status);
```
rjeberhard: i18n

rjeberhard: And misspelled.

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -172,6 +172,205 @@ Step updateConfigMap(Step next, V1ConfigMap existingConfigMap) {
...
+                this,
+                domain.getDomainUID(),
+                getOperatorNamespace(),
+                domain.getMetadata().getNamespace(),
+                data);
+
+        return doNext(context.verifyConfigMap(getNext()), packet);
+      }
+
+      // TODO: How do we handle no topology?
```
rjeberhard: What is the answer?

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -172,6 +172,205 @@ Step updateConfigMap(Step next, V1ConfigMap existingConfigMap) {
...
+      super(conflictStep, operatorNamespace, domainNamespace);
+
+      this.cmName = getConfigMapName(domainUID);
+      this.data = data;
+      this.model = createModel(data);
+    }
+
+    private V1ConfigMap createModel(Map<String, String> data) {
+      return new V1ConfigMap()
+          .apiVersion("v1")
```
rjeberhard: apiVersion and kind are defaults

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -190,25 +389,159 @@ public NextAction onFailure(Packet packet, CallResponse<V1ConfigMap> callRespons
...
+      while (line != null) {
+        if (line.startsWith(">>>") && !line.endsWith("EOF")) {
+          // Beginning of file, extract file name
+          String filename = extractFilename(line);
+          readFile(reader, filename, map);
+        }
+        line = reader.readLine();
+      }
+    } catch (IOException exc) {
+      // quit
```
rjeberhard: log?

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -190,25 +389,159 @@ public NextAction onFailure(Packet packet, CallResponse<V1ConfigMap> callRespons
...
+          return;
+        } else {
+          // add line to StringBuilder
+          stringBuilder.append(line);
+          stringBuilder.append(System.getProperty("line.separator"));
+        }
+        line = reader.readLine();
+      }
+    } catch (IOException ioe) {
+
```
rjeberhard: log?

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -190,25 +389,159 @@ public NextAction onFailure(Packet packet, CallResponse<V1ConfigMap> callRespons
...
+    try {
+      DomainTopology domainTopology = mapper.readValue(topologyYaml, DomainTopology.class);
+
+      LOGGER.fine(
+          ReflectionToStringBuilder.toString(domainTopology, ToStringStyle.MULTI_LINE_STYLE));
+
+      return domainTopology;
+
+    } catch (Exception e) {
+      LOGGER.warning("Failed to parse WebLogic Domain topology", e);
```
rjeberhard: i18n

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -0,0 +1,310 @@
...
+      cleanupJobArtifacts(packet);
+      return super.onFailure(packet, callResponse);
+    }
+
+    @Override
+    public NextAction onSuccess(Packet packet, CallResponse<String> callResponse) {
+      String result = callResponse.getResult();
+
+      // Log output to Operator log
+      LOGGER.info("+++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
```
rjeberhard: i18n... Actually, this should be FINE

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -0,0 +1,279 @@
...
+    V1PodSpec podSpec =
+        new V1PodSpec()
+            .activeDeadlineSeconds(60L)
+            .restartPolicy("Never")
+            .addContainersItem(createContainer(tuningParameters))
+            .addVolumesItem(new V1Volume().name(SECRETS_VOLUME).secret(getSecretsVolume()))
+            .addVolumesItem(
+                new V1Volume().name(SCRIPTS_VOLUME).configMap(getConfigMapVolumeSource()));
+
+    // TODO: where do we get image pull secret for job pod?
```
rjeberhard: Resolve this...  but, doesn't Job run as the operator's image?  I was going to say that we already know that the operator's image has been pulled, but... that's NOT true.  The Job/Pod could run on a different node then where the operator is running and the operator's image might not be there.  So, we need the imagePullSecrets from the operator's own Deployment.

operator/src/main/java/oracle/kubernetes/operator/steps/DeleteDomainStep.java
```diff
@@ -34,14 +36,32 @@ public DeleteDomainStep(String namespace, String domainUID) {
...
-            deleteServices(),
-            deletePods(),
-            deletePersistentVolumes(),
-            deletePersistentVolumeClaims()),
-        packet);
+    return doNext(Step.chain(deleteResources()), packet);
+  }
+
+  private Step[] deleteResources() {
+    ArrayList<Step> resources = new ArrayList<>();
```
rjeberhard: Have type be the generic List<Step> resources = new ArrayList<>();

operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
```diff
@@ -0,0 +1,77 @@
...
+
+    return doNext(packet);
+  }
+
+  private static void dispatchJobWatch(Watch.Response<V1Job> item) {
+    V1Job j = item.object;
+    if (j != null) {
+      V1ObjectMeta metadata = j.getMetadata();
+      String domainUID = metadata.getLabels().get(LabelConstants.DOMAINUID_LABEL);
+      if (domainUID != null) {
```
rjeberhard: This does nothing but log on delete.  Why do we have this?

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -0,0 +1,722 @@
+# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
```
rjeberhard: Not 2017

operator/src/main/resources/scripts/introspectDomain.py
rjeberhard: Yes, let's test without this.  We're going to support 12.2.1.3, but only with patches for sit config.

operator/src/main/resources/scripts/traceUtils.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
```
rjeberhard: Not 2107

src/integration-tests/bash/run.sh
```diff
@@ -3122,6 +3126,18 @@ function test_suite_init {
     declare_new_test 1 "$@"
 
     cd $PROJECT_ROOT || fail "Could not cd to $PROJECT_ROOT"
+ 
+    # Handle '--show-all' deprecation for 'kubectl get jobs' and 'kubectl get pods'.
+    # --show-all yields a deprecation warning in stderr in 1.10 in later, since
+    # it isn't needed in 1.10 and later.
+    # TBD It's not clear if we should check the Client version or the Server version.
```
rjeberhard: I suspect we just care about the server version.

src/integration-tests/introspector/createDomain.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
```
rjeberhard: Not 2017

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
tbarnes-us: @rjeberhard 

Unpatched 12.2.1.3 is invalid for any dynamic cluster used in combination with the introspector, but the introspector always needs to run.   This means that since we don't have patched versions of older WL versions, we must have 19.1 for any test that involves a dynamic cluster (not just new tests - all tests).   

So if we fully parameterize the image-name to avoid hard coding 19.1 into github, we’ll still ultimately need to have a ‘19.1’ reference somewhere to get the ball rolling.  I suppose this means we need a non-github approach to setting the value externally in each environment:   e.g. an internal script on Jenkins that sets up an env var, a Wercker env var, and maybe a reference to /net/some-machine/scratch/image-name.txt for local runs.   

Sound about right?

rjeberhard: We have a Wercker env var now: IMAGE_NAME_WEBLOGIC and IMAGE_TAG_WEBLOGIC.  I'm open to whatever makes sense for local testing.

kubernetes/src/test/java/oracle/kubernetes/operator/create/CreateOperatorGeneratedFilesTestBase.java
tbarnes-us: Yes.  Working on it.

model/src/main/java/oracle/kubernetes/weblogic/domain/v1/DomainSpec.java
tbarnes-us: @alai8 Anthony, has this been done?

alai8: yes, they are added to v2/DomainSpec.java

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain.py
tbarnes-us: These bullets already have follow-on JIRAs.

operator/src/main/resources/scripts/introspectDomain.py
tbarnes-us: Good points.  Thanks.  I've made the change and tests still pass.

src/integration-tests/bash/run.sh
tbarnes-us: Looks like its the client version, so the code can stay unchanged.   The Wercker runs use a 1.10 client and a 1.9 server, and tests are passing.  This indicates the 1.10 client gets 'show all' behavior by default even though the server itself is older.  I've pushed a change with the TBD comment removed.

operator/src/main/resources/scripts/start-server.py
tbarnes-us: Done.

operator/src/main/resources/scripts/introspectDomain.py
tbarnes-us: Done.  We will at least validate for this in 2.0, and potentially address this at some point.

model/src/main/java/oracle/kubernetes/weblogic/domain/v1/DomainSpec.java
tbarnes-us: thanks

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
tbarnes-us: Huh.  Maybe this came from a bad merge?  Fixed.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
tbarnes-us: Resolved via 
https://github.com/oracle/weblogic-kubernetes-operator/pull/489/commits/4de0cf306c42f6d40fdbd24f1ea082930b83ec20

model/src/main/java/oracle/kubernetes/operator/KubernetesConstants.java
tbarnes-us: https://github.com/oracle/weblogic-kubernetes-operator/pull/489/commits/4de0cf306c42f6d40fdbd24f1ea082930b83ec20

model/src/main/java/oracle/kubernetes/weblogic/domain/v1/DomainSpec.java
tbarnes-us: https://github.com/oracle/weblogic-kubernetes-operator/pull/489/commits/4de0cf306c42f6d40fdbd24f1ea082930b83ec20

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain.sh
tbarnes-us: Resolved via https://github.com/oracle/weblogic-kubernetes-operator/pull/489/commits/4de0cf306c42f6d40fdbd24f1ea082930b83ec20

operator/src/main/java/oracle/kubernetes/operator/steps/DeleteDomainStep.java
alai8: Do you mean using 
List resources = new ArrayList<>();
?
We use ArrayList<Step> because the result of this method is used as argument to Step.chain() which expects an array of Step, but I might be misunderstanding what you are suggesting though.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
alai8: Right now we’re running the job in a WebLogic image (similar to managed server pod), not the operator image.  Do we need to run the job as the operator’s image?  If so, we’ll need to spend time reconfiguring the job to the operator’s Deployment and access to the appropriate k8s resources.
With WebLogic image we have access to WLST for introspection so not sure running in Operator image will work unless it has WLST deployed and configured to it.
If it makes sense to use WLS image, we may be able to set the image pull secret using something like podSpec.setImagePullSecrets(info.getDomain().getSpec().getImagePullSecrets())

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
alai8: I checked with Lenny and we need to think more about this. Will file a new JIRA to track it.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
tbarnes-us: We definitely need a WL install since we need WLST & the node manager.  How is this passing on Wercker?  I thought Wercker runs pull WL directly and need the secret.   Could it be in the cache since the create-domain sample job already pulls it?

rjeberhard: You may be getting "lucky".  If your job gets scheduled on a node where the WebLogic image has already been pulled then you don't need the secret (default imagePullPolicy=IfNotPresent).  However, this will blow up if you are scheduled on a node where the image has not yet been pulled.

operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
tbarnes-us: Closing as resolved, as tracked via OWLS-70161

operator/src/main/java/oracle/kubernetes/operator/steps/DeleteDomainStep.java
tbarnes-us: @rjeberhard ?

PR #493, opened 2018-11-09
Domain CRD to v2

rjeberhard:
This change updates the Domain CRD to version "v2".

Given that we are moving to v2 and not supporting (even as deprecated) v1, this change removes support for the pre-existing V1 schema and code paths.

All tests and samples are updated to use v2.

I noticed and corrected some schema mismatches between our proposed v2 schema wiki and the Java code, such as that DomainSpec.clusters and .managedServers should have been maps.

There are other inconsistencies, but this change only modified those required to full move to v2 schema.


vanajamukkara:
I am working on making the java tests pass with V2 domain changes. Do you want to cover them in this PR? 
src/scripts/scaling/scalingAction.sh needs to be changed to replace clusterStartup with clusters to make WLDF scaling work.

PR #501, opened 2018-11-14
Support more options for schema generation

russgold:
With this change, it is now possible to configure the plugin to:

- disable the "additionalProperties" field (required for K8s CRD validation)
- generate object fields in-line rather than using the $ref field (required for K8s CRD validation)
- specify the name and location of the file to generate (needed to generate one for validation and one for the website)

### DIFF
--- a/model/src/main/java/oracle/kubernetes/weblogic/domain/v2/ProbeTuning.java
+++ b/model/src/main/java/oracle/kubernetes/weblogic/domain/v2/ProbeTuning.java
@@ -24,6 +24,12 @@ public class ProbeTuning {
 
   public ProbeTuning() {}
 
+  void copyValues(ProbeTuning fromProbe) {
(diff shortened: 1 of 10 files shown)


model/src/main/java/oracle/kubernetes/weblogic/domain/v2/ProbeTuning.java
```diff
@@ -24,6 +24,12 @@
 
   public ProbeTuning() {}
 
+  void copyValues(ProbeTuning fromProbe) {
```
rjeberhard: Any idea why these changes are here?  Otherwise, everything looks good.

russgold:
That’s was a minor refactoring that I forgot to put into the previous change. I had in mind collecting a bunch of such cleanups, but didn’t wind up doing any others.

russgold:
There is, BTW, a corresponding change in ServerPod

PR #544, opened 2018-11-30
OWLS-68701 -- Support for additional service labels and annotations 

vkraemer:
This PR is just to get initial feedback... Not sure I am going in the right direction.

Will add unit and integration tests as necessary and appropriate after some initial review comments... unless I have completely missed the mark on this task.


rjeberhard:
Vince, this looks really good.  The only missing work is the unit-tests, which should be very similar to the unit tests for adding pod specific labels.

model/src/test/java/oracle/kubernetes/weblogic/domain/v2/BaseConfigurationTestBase.java
```diff
@@ -84,4 +84,16 @@ public void whenServerStartPoliciesDiffer_objectsAreNotEqual() {
   private V1EnvVar env(String name, String value) {
     return new V1EnvVar().name(name).value(value);
   }
+
+  @Test
```
russgold: The names of tests don't affect their functionality, but they are very useful for explaining what is going on to the next developer. Note that most test names have statements that tell you what is being tested.

Also, each case should test a single functionality. That makes naming the tests easy. See the other tests in this class.

model/src/test/resources/oracle/kubernetes/weblogic/domain/v2/domain-sample-2.yaml
```diff
@@ -77,6 +77,9 @@ spec:
           periodSeconds: 20
         nodeSelector:
           os: linux
+        serviceLabels:
```
russgold: Yes, single letter names work, but they require the next developer to read a lot more than necessary. Things like "label1" or "name1" and "value1" are a lot easier to understand.

model/src/test/java/oracle/kubernetes/weblogic/domain/v2/BaseConfigurationTestBase.java
russgold: Looks good

rjeberhard:
@vkraemer @russgold Hurray, unit tests!  Is this ready to go?

vkraemer:
Hi @rjeberhard, i will defer to @russgold on that question...

I just updated to a recent commit-id, so we will need to verify that the integration-test-java is still green

model/src/test/resources/oracle/kubernetes/weblogic/domain/v2/domain-sample-2.yaml
vkraemer: i have replaced the single character label and annotation names as requested

PR #550, opened 2018-12-03
Owls 69919 domain resource properties

rakillen:
The code changes are complete, but I'm concerned they have affected integration tests, and I have not had a clean Jenkins run due to other problems. Changing the default for domainHomeInImage to true changes the default value for domainHome, which affected some of the unit tests and at least one integration test - ITOperator domain8.


model/src/main/java/oracle/kubernetes/weblogic/domain/v2/DomainSpec.java
```diff
@@ -95,6 +98,19 @@
...
+   *
+   * @since 2.0
+   */
+  @SerializedName("logHomeEnabled")
+  @Expose
+  @Description(
+      "Specified whether the log home folder is enabled (Not required). "
+          + "Defaults to true if domainHomeInImage is false. "
+          + "Defaults to false if domainHomeInImage is true. ")
+  private Boolean logHomeEnabled; // Boolean object, null if unspecified
```
rjeberhard: Need to add new fields to equals(), hashCode(), and toString().

rakillen: Added those, ready for merge

PR #552, opened 2018-12-03
OWLS-70382: sample for WLS domain credential secret

doxiao:
1) add a sample script to create and label the secret given the username, password and optionally domainUID and namespace.
2) set the domainHomeInImage to false in the domain-home-on-pv sample script
3) label the generated domain custom resource with domainUID.


kubernetes/samples/scripts/create-weblogic-domain/create-weblogic-credentials.sh
```diff
@@ -0,0 +1,96 @@
...
+  echo "${script}: -p must be specified."
+  missingRequiredOption="true"
+fi
+
+if [ "${missingRequiredOption}" == "true" ]; then
+  usage 1
+fi
+
+# check and see if the secret already exists
+result=`kubectl get secret ${secretName} -n ${namespace} --ignore-not-found=true | grep ${secretName} | wc | awk ' { print $1; }'`
```
tbarnes-us: Current grep could grab a secret that has a name where the new secret name is embedded within the name.
Fix by putting '^name$' in the grep, or better yet, using a -o=jsonpath expression instead of grep.

tbarnes-us: I have no preference.   We might also want to look around to see if there are examples of the same pattern in other scripts.

rjeberhard: Okay, then I'm going to punt on requiring a fix now.

PR #554, opened 2018-12-03
OWLS-70324: make delete-weblogic-domain-resources.sh a sample, remove delete all domains option

chunshi-ora:
- Moved the sh script from kubernetes/ to kubernetes/samples/scripts/.
- Removed support of special domain UID value "all", which would remove all domain resources in a cluster, regardless of domain UIDs.

### DIFF
--- a/kubernetes/samples/scripts/delete-weblogic-domain-resources.sh
+++ b/kubernetes/samples/scripts/delete-weblogic-domain-resources.sh
@@ -29,8 +29,8 @@ cat << EOF
   or all resources were deleted (default $default_maxwaitsecs seconds).
 
   The domains can be specified as a comma-separated list of 
-  domain-uids (no spaces), or the keyword 'all'.  The domains can be
-  located in any kubernetes namespace.
+  domain-uids (no spaces).  The domains can be located in any
+  kubernetes namespace.
 
   Specify '-t' to run the script in a test mode which will
   show kubernetes commands but not actually perform them.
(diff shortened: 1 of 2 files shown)


kubernetes/samples/scripts/delete-weblogic-domain-resources.sh
```diff
@@ -73,18 +73,14 @@ EOF
...
-      domain_regex="$domain_regex^Domain ${UIDS[$i]} "
-    done
-  fi
+  LABEL_SELECTOR="weblogic.domainUID in ($1)"
+  IFS=',' read -ra UIDS <<< "$1"
+  for i in "${!UIDS[@]}"; do
+    if [ $i -gt 0 ]; then
+      domain_regex="$domain_regex|"
+    fi
+    domain_regex="$domain_regex^Domain ${UIDS[$i]} "
```
moreaut: I believe that Dongbo has updated (or is updating) the sample create domain scripts to add the weblogic.domainUID label to the domain resource too, so I don't think the delete script needs to have a special rule to remove the domain resource.

chunshi-ora: Ok. @doxiao, are your changes in develop already?

doxiao: Not yet. See https://github.com/oracle/weblogic-kubernetes-operator/pull/552.

PR #561, opened 2018-12-04
OWLS-69878: Have the introspector return info about all of the enabled channels of all the servers

swapnanitin:
Added support in:

**Introspector**
- Return the administration port and default secure channel.
- Reject the domain if a server has a network access point named 'default, 'defaultSecure', or 'defaultAdmin'.
- <WIP>Reject a cluster if its servers don't have exactly the same set of channels enabled.

**Operator** 
- Parse the administration port and ValidationErrors.


tbarnes-us:
@swapnanitin @lennyphan @alai8 It's my understanding the Operator needs to communicate with the admin server, so, if the admin port is configured, it needs to know the fact that it now requires t3s/https instead of t3/http.    Maybe I'm missing something since I'm not very familiar with the Operator java, but I expected to see corresponding changes to the Operator to handle this change.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -477,8 +488,19 @@ def addClusteredServer(self, cluster, server):
     self.writeln("- name: " + name)
     self.writeln("  listenPort: " + str(server.getListenPort()))
     self.writeln("  listenAddress: " + self.quote(self.env.toDNS1123Legal(self.env.getDomainUID() + "-" + server.getName())))
+    self.writeln("  adminPort: " + str(server.getAdministrationPort()))
```
tbarnes-us: perhaps just don't include the adminPort in the output if adminPort is disabled?   ditto for SSL, or the default channel.     (Even the default channel can be disabled.)

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -475,10 +484,23 @@ def getClusteredServers(self, cluster):
...
+      self.writeln("  adminPortEnabled: " + self.booleanToString(server.isAdministrationPortEnabled()))
+    self.addSSL(server)
     self.addNetworkAccessPoints(server)
 
+  def addSSL(self, server):
+    ssl = server.getSSL()
+    if ssl is not None and ssl.isEnabled():
+      self.indent()
+      self.writeln("sslListenPort: " + str(ssl.getListenPort()))
+      self.writeln("sslPortEnabled: " + self.booleanToString(ssl.isEnabled()))
```
tbarnes-us: Why pass 'sslPortEnabled'?  Isn't the presence of 'sslListenPort' enough to show that the port is enable?
Ditto for 'adminPortEnabled'.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -280,6 +280,7 @@ def __init__(self, env):
   def validate(self):
     self.validateAdminServer()
     self.validateClusters()
+    self.validateServerCustomChannelName()
```
tbarnes-us: I may be missing something, but I don't see a check to make sure that:
-- All defaultChannel in the same configured cluster are either enabled xor all are disabled
-- All SSL Channel in the same configured cluster are either enabled xor all are disabled, and, if enabled, have the same port
-- All admin port in the same configure cluster are either enabled xore all are disabled, and, if enabled, have the same port
-- All nap with same name in configured cluster have same protocol

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -120,6 +120,7 @@ export ADMIN_NAME=${ADMIN_NAME:-"admin-server"}
 export ADMIN_PORT=${ADMIN_PROT:-7001}
 export MANAGED_SERVER_NAME_BASE=${MANAGED_SERVER_NAME_BASE:-"managed-server"}
 export DOMAIN_NAME=${DOMAIN_NAME:-"base_domain"}
+export ADMINISTRATION_PORT=${ADMINISTRATION_PORT:-7099}
```
tbarnes-us: No file in this pull uses 'ADMINISTRATION_PORT'.  I suspect this is dead code.

swapnanitin: createDomain script uses it to set the AdministrationPort..

lennyphan:
> ?

On my last conversation with Swapna last Friday, she was working on how the Operator runtime is to handle any validation errors returned from the introspection.

PR #563, opened 2018-12-05
Rolling Restart Model changes

AdrianPadilla:


model/src/main/java/oracle/kubernetes/weblogic/domain/v2/BaseConfiguration.java
```diff
@@ -53,6 +54,21 @@
           + "Legal values are NEVER, ALWAYS, or IF_NEEDED.")
   private String serverStartPolicy;
 
+  /**
```
moreaut: The operator will restart the pods whenever the value is changed and there is no code enforcing that when the value changes, it must be incremented by one.

i.e. these comments do not accurately reflect how this field works.

model/src/main/java/oracle/kubernetes/weblogic/domain/v2/BaseConfiguration.java
```diff
@@ -194,12 +210,23 @@ void addPodAnnotations(String name, String value) {
     serverPod.addPodAnnotations(name, value);
   }
 
+  public Integer getRestartVersion() {
+    return restartVersion;
+  }
+
+  public abstract RestartVersion addEffectiveRestartVersion(RestartVersion restartVersion);
```
moreaut: EffectiveRestartVersion should not be on base configuration - it should be on ServerConfig

PR #568, opened 2018-12-05
WIP Feature/owls 69344

alai8:
In operator 1.x, the operator code use REST requests to the weblogic admin server to bring the dynamic cluster size up to the desired replica count, with a limit up to the max dynamic cluster size. The operator code does not adjust the dynamic cluster size down, nor would it adjust the max dynamic cluster size.

In operator 2.0, the operator could instead generate situational config during domain introspection to adjust the dynamic cluster size to the same as max dynamic cluster size.

However, this does not currently work with WebLogic 19.1. 

While dynamic-cluster-size attribute value can be modified through situational config, the newly added servers cannot be started. See bug 29017654.

Changes in this pull request includes removal of unused operator code related to issuing REST request to admin server for updating dynamic cluster size, and updating site/scaling.md to mention operator cannot create more manage servers than dynamic cluster size.


rjeberhard:
Deferring this from 2.0.

tbarnes-us:
Note that even if Bug 29017654 is fixed, this still isn't enough to expand the cluster size:   we'd also need some sort of Operator enhancement.  This is because the introspector job reads the existing cluster size without taking into account the sit-cfg, then passes this original value to the operator, and the operator in turn uses this original value as a maximum value for the cluster size.    (Applying sit-cfg would require booting the admin server.)

rjeberhard:
@alai8, this has been hanging out for a while...  Do we still need it?  If so, can you clean-up and prepare for committing?

PR #575, opened 2018-12-06
OWLS-70534 error handling

doxiao:
Check the results of the calls in the create domain job and exit early if any of the calls fails.

@vanajamukkara

### DIFF
--- a/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/common/create-domain-job.sh
+++ b/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/common/create-domain-job.sh
@@ -3,6 +3,13 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 #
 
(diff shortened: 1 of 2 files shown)


kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/common/create-domain-job.sh
```diff
@@ -3,6 +3,13 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 #
 
+function exitIfError {
+  if [ "$1" != "0" ]; then
+    echo "$2"
+    exit $1
```
tbarnes-us: Should echo/exit be replaced with "fail"?
Should this method be put in 'utilities'?

doxiao: This script is running in the job pod, and it does not use anything in utilities.

PR #579, opened 2018-12-06
OWLS-70527 : version.txt support in situational config

vkraemer:
Added code to generate an error if the user does not include a version.txt file in their situational config directory.

The version.txt must have the content "2.0" to be accepted in Operator 2.0.  Future versions could use this versioning to guide parsing or the custom config files. Could is the operative word...

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.py
+++ b/operator/src/main/resources/scripts/introspectDomain.py
@@ -878,6 +878,19 @@ class CustomSitConfigIntrospector(SecretManager):
     if not os.path.exists(self.env.CUSTOM_SITCFG_PATH):
       return
 
+    # We expect the user to include a 'version.txt' file in their situational
+    # config directory.
+    #
+    # That file is expected to contain '2.0'
+    #
+    versionPath=os.path.join(self.env.CUSTOM_SITCFG_PATH,"version.txt")
+    if not os.path.exists(versionPath):
+        self.env.addError("Error, Required file, "+versionPath+", does not exist")
+    else:
+        version=self.env.readFile(versionPath).strip()
+        if not version == "2.0":
+            self.env.addError("Error, "+versionPath+" does not have the value of '2.0'.")
+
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -878,6 +878,19 @@ def generateAndValidate(self):
...
+    #
+    # That file is expected to contain '2.0'
+    #
+    versionPath=os.path.join(self.env.CUSTOM_SITCFG_PATH,"version.txt")
+    if not os.path.exists(versionPath):
+        self.env.addError("Error, Required file, "+versionPath+", does not exist")
+    else:
+        version=self.env.readFile(versionPath).strip()
+        if not version == "2.0":
+            self.env.addError("Error, "+versionPath+" does not have the value of '2.0'.")
```
tbarnes-us: Can you add a quoted something to this error message that shows the first 100 characters of the invalid version?  (100 is arbitrary - just some value to prevent dumping too much if it's a huge file).     

Also, please add single quotes around 'versionPath' in the error messages.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -889,7 +902,7 @@ def generateAndValidate(self):
 
       # check if file name corresponds with config.xml or a module
 
-      if not self.moduleMap.has_key(the_file) and the_file != "config.xml":
+      if not self.moduleMap.has_key(the_file) and the_file != "config.xml" and the_file != "version.txt":
```
tbarnes-us: Add a if the_file == "version.txt": continue right at line 895...    See my next comment for the reasoning.

operator/src/main/resources/scripts/introspectDomain.py
vkraemer: not sure I understand your second comment... what is the point of quotes around versionPath?

PR #581, opened 2018-12-06
minor comment change

tbarnes-us:


tbarnes-us:
This change doesn't need review - it's only a comment change.

PR #587, opened 2018-12-07
OWLS-69882: Add support for an admin server administration node port …

jchermos:
Changes to support administration node port:
   

-  Updated AdminServer schema with AdminService object that contains:
-  Map of annotations,
-  Map of labels,
-  Map with channels to be exposed.
-  Updated Operator runtime to create such nodePort service.

When is merged #561 uncomment line 8656 `/*else if("adminSecure".equals(channel) && serverConfig.isAdminPortEnabled()){
            return serverConfig.getAdminPort();
          }*/` to include to admin service admin port.


rjeberhard:
Carlos, this looks good to me...  Based on your comment though, it looks like you need coordination with the work of @swapnanitin.  Correct?  @tbarnes-us Can you please help these two to coordinate based on the status of Swapna's work?

jchermos:
Yes, this needs @swapnanitin PR to add to the admin service the admin port that her changes are returning.

tbarnes-us:
@lennyphan is kindly helping to coordinate as needed.

PR #598, opened 2018-12-10
Remove schema elements for domainName, asName and asPort

rjeberhard:
Plus, resolve some other issues such as CRDHelper not updating the existing CRD when schema validation has changed.

### DIFF
--- a/model/src/main/java/oracle/kubernetes/weblogic/domain/v2/Domain.java
+++ b/model/src/main/java/oracle/kubernetes/weblogic/domain/v2/Domain.java
@@ -272,39 +272,12 @@ public class Domain {
   }
 
   /**
-   * Returns the name of the admin server.
-   *
-   * @return admin server name
-   */
-  public String getAsName() {
-    return spec.getAsName();
-  }
-
-  /**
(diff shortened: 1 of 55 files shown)


model/src/main/java/oracle/kubernetes/weblogic/domain/v2/Domain.java
```diff
@@ -271,44 +277,18 @@ public V1SecretReference getAdminSecret() {
...
-  }
-
-  /**
-   * Returns the domain name
-   *
-   * @return domain name
-   */
-  public String getDomainName() {
-    return spec.getDomainName();
+    return spec.getDomainUID() != null ? spec.getDomainUID() : getMetadata().getName();
```
russgold: In Java 8, use of Optional is preferred:

return Optional.ofNullable(spec.getDomainUID()).orElse(getMetadata().getName());

This removes the need to refer to spec.getDomainUID() twice, and ensures that the reader can tell it is the same value.

russgold:
With the exception of the one style issue, looks good.

PR #611, opened 2018-12-11
Add support for javaOptions field

russgold:
Adds the functionality that we had in 1.1 and which is documented in the wiki. It provides an alternative way to set the JAVA_OPTIONS environment variable.

### DIFF
--- a/model/src/test/resources/oracle/kubernetes/weblogic/domain/v2/domain-sample-3.yaml
+++ b/model/src/test/resources/oracle/kubernetes/weblogic/domain/v2/domain-sample-3.yaml
@@ -16,6 +16,8 @@ spec:
   # The Admin Server's ListenPort
   asPort: 8001
 
+  serverPod:
+    javaOptions: "-add-this"
   adminServer:
   # the T3 endpoints to export. Each entry is an endpoint name,
   # optionally with labels and annotations to be added to the generated external channel service.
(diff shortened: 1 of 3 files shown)


model/src/test/resources/oracle/kubernetes/weblogic/domain/v2/domain-sample-3.yaml
```diff
@@ -29,6 +31,9 @@ spec:
         annotations:
           time: midnight
     serverPod:
+      env:
```
doxiao: It seems that we have two ways of specifying Java Options.

  serverPod:
    javaOptions: "-add-this"

and

    serverPod:
      env:
      - name: JAVA_OPTIONS
         value: "-server"

Are both configurable at all (domain/server/cluster) levels? How do they work together if specified in both javaOptions and envs?

doxiao: It is confusing to have two ways of doing the same thing. We don't have to be backward compatible in 2.0.

rjeberhard: Russ, I thought we had agreed that we would leave the javaOptions field out of the domain schema and just use env/JAVA_OPTIONS?  It's simpler just to have one schema element.

russgold:
I’ve been receiving mixed signals on this - can you close the related bug as ‘will not fix’?

PR #625, opened 2018-12-13
OWLS-70659 Provide ability to retry introspection due to DeadlineExceeded status

alai8:
Problem description from OWLS-70659:

===========================
Default 'activeDeadlineSeconds' value for introspection pod execution is set to 60 seconds. Under machine load, introspection runs may exceed the default 60 seconds and fail with 'DealineExceeded' status:

sh-3.2$ kubectl get po
NAME READY STATUS RESTARTS AGE
domain1-introspect-domain-job-sxfqz 0/1 DeadlineExceeded 0 1m
DeadlineExceeded

This causes the introspection to hang since we're not detecting this failure.

We should provide a retry implementation to rerun the introspection up to a max retry count.

===========================
This pull request includes the following changes:

- The introspector job was not rerun due to the Complete callback not being called if the job status is failed. Fix is made in the receiveResponse method in JobWatcher class

- An issue is found in re-running the domain introspector job when the pod fails due to node manager not able to start because the nodemanager.lck file already exists, and nodemanager is writing log to nodemanager-1.log. This results in error in startNodeManager.sh as it fails to find nodemanager.log file. Added code to remove nodemanager.lck file at the start of startNodeManager.sh script.

- The retrying for domain introspector is part of the retry logic in DomainProcessorImpl. Currently it does not have a limit on number of retries. It is modified to retry makeRightDomainPresence() only up to a maximum retry count that is specified in TuningParameters$MainTuning, with a default value of 5.

- Set initial activeDeadlineSeconds for introspect domain job to 120 seconds, and each subsequent retry incremented by 60 seconds.

- update JobWatcher to ignore responses that are for a previous job. These responses interfere with the retry logic and causing domain to fail to startup even if the introspector job completed successfully in a retry.

Passed Jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/687


operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -813,7 +813,10 @@ void runDomainPlan(
         new CompletionCallback() {
           @Override
           public void onCompletion(Packet packet) {
-            // no-op
+            DomainPresenceInfo existing = getExistingDomainPresenceInfo(ns, domainUID);
```
rjeberhard: Move to TailStep so that there isn't a race condition that a new Fiber started.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -837,17 +840,29 @@ public void onThrowable(Packet packet, Throwable throwable) {
...
+                    MessageKeys.CANNOT_START_DOMAIN_AFTER_MAX_RETRIES,
+                    domainUID,
+                    ns,
+                    DomainPresence.getDomainPresenceFailureRetryMaxCount(),
+                    throwable);
+              } else {
+                gate.getExecutor()
+                    .schedule(
+                        () -> {
+                          existing.setPopulated(false);
```
rjeberhard: This changes the behavior incorrectly...  We need to get the existing domain presence at the time when the scheduled event happens.  You've moved the lookup to when the retry is scheduled.

PR #635, opened 2018-12-13
OWLS-68437: handle the introspection validation errors at operator runtime

swapnanitin:
Log the introspect validation errors to the operator log and set the introspect job to complete.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
@@ -14,6 +14,7 @@ import io.kubernetes.client.models.V1ObjectMeta;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -274,6 +275,12 @@ public NextAction apply(Packet packet) {
       if (topologyYaml != null) {
         LOGGER.fine("topology.yaml: " + topologyYaml);
         DomainTopology domainTopology = parseDomainTopologyYaml(topologyYaml);
+        if (!domainTopology.getValidationErrors().isEmpty()) {
+          for (String err : domainTopology.getValidationErrors()) {
+            LOGGER.severe(err);
```
rjeberhard: Throw an exception with the errors.  This will log as severe and trigger the domain processing retry logic.

swapnanitin: We don't want to trigger any retries when there's validation errors during introspection. validation errors are due to the domain being invalid, we need to stop processing and log this as severe, so that customer can fix the issue with the domain.

PR #640, opened 2018-12-14
Test for changes to numerous Pod settings

russgold:
OWLS-70504 Added unit tests for new PodHelper functionality, and updated the code to make the tests pass. Some minor refactoring to extract settings needed for both tests and production code into PodDefaults.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -7,24 +7,7 @@ package oracle.kubernetes.operator.helpers;
 import static oracle.kubernetes.operator.LabelConstants.forDomainUid;
 
 import io.kubernetes.client.custom.IntOrString;
-import io.kubernetes.client.models.V1ConfigMapVolumeSource;
-import io.kubernetes.client.models.V1Container;
-import io.kubernetes.client.models.V1ContainerPort;
-import io.kubernetes.client.models.V1DeleteOptions;
-import io.kubernetes.client.models.V1EnvVar;
-import io.kubernetes.client.models.V1ExecAction;
-import io.kubernetes.client.models.V1HTTPGetAction;
-import io.kubernetes.client.models.V1Handler;
-import io.kubernetes.client.models.V1Lifecycle;
-import io.kubernetes.client.models.V1ObjectMeta;
-import io.kubernetes.client.models.V1PersistentVolume;
-import io.kubernetes.client.models.V1PersistentVolumeList;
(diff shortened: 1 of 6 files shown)


moreaut:
These changes look good.

However, there are other fields we need to check too. 
Not sure if they should be handled in this PR or in another:

  - server services
    - metadata
       - check annotations (from domain resource's adminService's annotations)
       - check labels (from domain resource's adminService's labels)
  - server pods
    - pod spec
       - check securityContext (from domain resource's podSecurityContext)
       - check nodeSelector (from domain resource's nodeSelector)
    - container
        - check securityContext (from domain resource's containerSecurityContext)
        - check livenessProbe (from domain resource's livenessProbe)
        - check readinessProbe (from domain resource's readinessProbe)
        - check resources (from domain resource's resources)

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -300,6 +279,18 @@ private static boolean isCurrentPodValid(V1Pod build, V1Pod current) {
...
     }
 
+    if (areUnequal(
+        volumesWithout(current.getSpec().getVolumes(), ignoring), build.getSpec().getVolumes()))
+      return false;
+
+    if (areUnequal(current.getSpec().getImagePullSecrets(), build.getSpec().getImagePullSecrets()))
+      return false;
+
+    if (areUnequal(getCustomerLabels(current), getCustomerLabels(build))) return false;
```
chunshi-ora: Except resource version and restart versions, labels set by operator will not change, right?

russgold: Correct.

operator/src/test/java/oracle/kubernetes/operator/helpers/AdminPodHelperTest.java
```diff
@@ -152,6 +164,41 @@ public void whenExistingAdminPodHasBadVersion_replaceIt() {
         pod -> pod.getMetadata().putLabelsItem(RESOURCE_VERSION_LABEL, "??"));
   }
 
+  @Test
+  public void whenExistingManagedPodHasUnknownCustomerLabel_designateForRoll() {
```
chunshi-ora: Need to test the case a label is removed?

russgold: it should be the same, but I will see if there is an easy way to mutate the model rather than the existing pod in the test

moreaut:
Actually, the server service checks are more important since the customer would configure the labels to help set up load balancing.

PR #642, opened 2018-12-14
Config overrides site doc

tbarnes-us:
This is the initial draft of the config overrides documentation.


markxnelson:
@tbarnes-us I will go ahead and merge this, but please have a look at the user-guide.md and let me know where you think this belongs in the table of contents, so I can add a link. thanks

moreaut:
It should be 'configOverrides'

tbarnes-us:
But I guess this confusion isn't enough justification to make such a change.

PR #648, opened 2018-12-14
OWLS-70593: OPERATOR: CUSTOM OVERRIDE CONFIG MAP DOES NOT TAKE EFFECT WHEN DYNAMICALLY ADDED

lennyphan:
Fix for 'running servers' count for doing introspection job.  Also added some logic for looking at 'Server Start' policy for best effort in determining if we're  going from zero to one or more managed server pods.


lennyphan:
Jenkins run passed on http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/720/

PR #656, opened 2018-12-15
Add an option to not alway clone the docker images project plus bug-fixing

doxiao:
1) add a command line option to keep using the previously cloned project dir. This allows customers to re-generate the images after modifying the WDT model files.
2) Fix hard-coded image value in the domain yaml file
3) Correct credentials override for WDT case.

@anpanigr


doxiao:
Note: with this PR, the domain-home-in-image-wdt Docker image is really used in the Operator sample, which has reveled issues in the Docker image (docker-images/OracleWebLogic/samples/12213-domain-home-in-image-wdt). The issues are being investigated.

PR #657, opened 2018-12-17
Fixes for WebLogic Server and Node Manager .log and .out locations

tbarnes-us:
This check-in has two fixes:

- A: Adds missing support for putting logs in their default locations when logHomeEnabled is false (logHomeEnabled flag was being ignored)
- B: Generates correct sit-cfg overrides when original Log settings aren't set -- e.g. use "add" sit-cfg verb instead of "replace" (log files would end up in their default location instead of getting overridden)

Work-around, if not applying this check-in:

- Set 'logHome' to a location that has write-permission (helps with problem A above, not problem B)

Changes in this check-in:

- modify Operator to set LOG_HOME env var to '' when logHomeEnabled is false,
- modify startNodeManager.sh script to call mkdir ${DOMAIN_HOME}/servers/${SERVER_NAME}/logs instead of mkdir ${LOG_HOME} when LOG_HOME is '',
- modify startNodeManager.sh script to skip calling mkdir all-together when running inside an introspector pod,
- modify introspectDomain.py to skip creating sit-cfg overrides for log dirs if LOG_HOME is '' or null (it use to check for null, and did not check for ''),
- modify introspectDomain.py to use 'add' sit-config verbs instead of 'replace' if original log 'file-name' setting isn't explicitly set in the config.xml

Test status:

- Passes QUICKTEST integration testing when running locally.

- Passes Ad Hoc Testing
  - for both when setting log file-name on domain/server/server-template in config.xml, and when _not_ setting it:
    - For logHomeEnabled=true
      - confirm sit-cfg override works (log files for domain/server/server-template are in logHome)
      - also confirm NM log is in logHome too
      - also confirm .out from servers & node-manager end up in logHome too
    - For logHomeEnabled=false
      - verified sit-cfg isn't created
      - verified domain/server/server-tamplate/nodemanager logs end up at their respective default locations
      - verified server .out files end up in ${DOMAIN_HOME}/servers/${SERVER_NAME}/logs

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.py
+++ b/operator/src/main/resources/scripts/introspectDomain.py
@@ -797,7 +797,7 @@ class SitConfigGenerator(Generator):
     self.writeln("</d:security-configuration>")
 
   def customizeDomainLogPath(self):
-    self.customizeLog(self.env.getDomain().getName())
+    self.customizeLog(self.env.getDomain().getName(), self.env.getDomain(), true)
 
   def customizeServers(self):
     for server in self.env.getDomain().getServers():
@@ -824,7 +824,7 @@ class SitConfigGenerator(Generator):
       self.writeln("<d:listen-address f:combine-mode=\"replace\">" + listen_address + "</d:listen-address>")
       self.undent()
       self.writeln("</d:network-access-point>")
-    self.customizeLog(name)
+    self.customizeLog(name, server, false)
     self.undent()
     self.writeln("</d:server>")
 
@@ -840,18 +840,33 @@ class SitConfigGenerator(Generator):
     self.indent()
     self.writeln("<d:name>" + name + "</d:name>")
     self.writeln("<d:listen-address f:combine-mode=\"replace\">" + listen_address + "</d:listen-address>")
-    self.customizeLog(server_name_prefix + "${id}")
+    self.customizeLog(server_name_prefix + "${id}", template, false)
     self.undent()
     self.writeln("</d:server-template>")
 
-  def customizeLog(self, name):
+  def customizeLog(self, name, bean, isDomainBean):
     logs_dir = self.env.getDomainLogHome()
-    if logs_dir is not None:
-      self.writeln("<d:log f:combine-mode=\"replace\">")
-      self.indent()
-      self.writeln("<d:file-name>" + logs_dir + "/" + name + ".log</d:file-name>")
-      self.undent()
-      self.writeln("</d:log>")
+    if logs_dir is None or len(logs_dir) == 0:
+      return
+
+    logaction=''
+    fileaction=''
+    if bean.getLog() is None:
+      if not isDomainBean:
+        # don't know why, but don't need to "add" a missing domain log bean, and adding it causes trouble
+        logaction=' f:combine-mode="add"'
+      fileaction=' f:combine-mode="add"'
+    else:
+      if bean.getLog().getFileName() is None:
+        fileaction=' f:combine-mode="add"'
+      else:
+        fileaction=' f:combine-mode="replace"'
+
+    self.writeln("<d:log" + logaction + ">")
+    self.indent()
+    self.writeln("<d:file-name" + fileaction + ">" + logs_dir + "/" + name + ".log</d:file-name>")
+    self.undent()
+    self.writeln("</d:log>")
 
 class CustomSitConfigIntrospector(SecretManager):
 
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
@@ -327,7 +327,7 @@ public abstract class PodHelperTestBase {
             hasEnvVar("DOMAIN_UID", UID),
             hasEnvVar("NODEMGR_HOME", NODEMGR_HOME),
             hasEnvVar("SERVER_OUT_IN_POD_LOG", Boolean.toString(INCLUDE_SERVER_OUT_IN_POD_LOG)),
-            hasEnvVar("LOG_HOME", LOG_HOME + "/" + UID),
+            hasEnvVar("LOG_HOME", ""),
             hasEnvVar("SERVICE_NAME", LegalNames.toServerServiceName(UID, getServerName())),
             hasEnvVar("AS_SERVICE_NAME", LegalNames.toServerServiceName(UID, ADMIN_SERVER))));
   }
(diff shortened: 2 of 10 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
```diff
@@ -462,7 +464,8 @@ public void whenNoPod_retryOnFailure() {
     testSupport.verifyCompletionThrowable(ApiException.class);
   }
 
-  @Test
+  // TBD This test fails - don't know why.
```
russgold: The unit test fails because you changed the logHome as recorded in an environment variable, but didn't change the test to match.

The test includes (at about line 565):

       .addEnvItem(envItem("LOG_HOME", LOG_HOME + "/" + UID))

but that doesn't appear to be correct any more. Make that match what the log location should be, and the test ought to pass.

tbarnes-us: Unit test updated.

tbarnes-us: Thaks @russgold !

tbarnes-us: Ready for approval @rjeberhard ...

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -840,18 +840,33 @@ def customizeServerTemplate(self, template):
...
+
+    logaction=''
+    fileaction=''
+    if bean.getLog() is None:
+      if not isDomainBean:
+        # don't know why, but don't need to "add" a missing domain log bean, and adding it causes trouble
+        logaction=' f:combine-mode="add"'
+      fileaction=' f:combine-mode="add"'
+    else:
+      if bean.getLog().getFileName() is None:
```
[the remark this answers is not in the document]
tbarnes-us: I thought the same, and am already tracking this idea via the laundry list of wiki ideas in Introspector 2.0 wiki.   Here's what's in there:
{code}
JIRA/BUG TBD - A sit-cfg 'replace' of a non-existant (unset) mbean attribute sometimes works (it works for server ListenAddress, server log FileName), but sometimes doesn't and you need to use a sit-cfg 'add' instead to make it work ('replace' doesn't work for an unset T3 Channel Public Address, or a Server Max Message Size, but 'add' does).   This brings up two potential sit-cfg enhancement requests for WebLogic Server:
1 - automatically fail sit-cfg attempts if replacing an unset field when this won't override the field.
2 - perhaps add a convenient 'add-or-replace' verb that does an 'add' if a 'replace' won't work - this eliminates the need to carefully check the originally configuration before deciding which verb to use.
{code}

This is complicated by the fact that at least one bean _appears_ to be missing when checked via off-line WLST but is only 'partly missing' somehow (reminds of Princes Bride quote " It just so happens that your friend here is only MOSTLY dead. There's a big difference between mostly dead and all dead." ), so when you try to create sit-cfg to add it in order to add an internal attribute within it, this fails to take effect - and it turns out the 'right' way to do it is actually not to specify either 'add' or 'replace' at the bean level.    The is the "domain level log bean".  The behavior of the "server level log" bean is more intuitive.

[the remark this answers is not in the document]
tbarnes-us: I'll add that last paragraph as a note to the wiki.

tbarnes-us:
@CarolynRountree Please provide the steps to run the reproducer for your test case.

CarolynRountree:
1. Pull the changes in branch fixes-sample-domain-in-image-wdt. You will run the sample in kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image.
2. Add logHome and logHomeEnabled to the domain-template.yaml
     logHomeEnabled: true
     logHome: /u01/oracle/user_projects/domains/logs (or something different from default)
3. Create kubernetes secrets for admin credentials, as an example: /kubernetes/samples/scripts/create-weblogic-domain-credentials/create-weblogic-credentials.sh -u weblogic -p welcome1 -d domain1 -n domain1-image -s domain1-weblogic-credentials
4. Modify the sample domain custom resource inputs file create-domain-inputs.yaml to match up with your desired environment. Use either 12213-domain-home-in-image-wdt or 12213-domain-home-in-image for your domain image. 
5. Create the domain using:   ./create-domain.sh \
  -u <username> \
  -p <password> \
  -i create-domain-inputs.yaml \
  -o /path/to/output-directory
 -e
The user and password must be the same as the ones in your secrets file. This is passed to the image build and persisted in the domain home. The -e says to create and apply a kubernetes object. 
3. Once the admin or managed servers are up, attach to one of the pods and check both the logs directory you specified and the servers/<server>/logs

CarolynRountree:
Here is a more comprehensive list from Tom with steps for building and installing operator and cleaning up resources
Pull the changes from branch fixes-sample-domain-in-image-wdt into your branch.

e.g. `git merge origin/fixes-sample-domain-in-image-wdt`

Build the operator image:

  mvn clean install -DskipTest
  docker build -t weblogic-kubernetes-operator:2.0 --build-arg http_proxy=$http_proxy --build-arg https_proxy=$https_proxy --build-arg no_proxy=$no_proxy --build-arg VERSION=2.0-SNAPSHOT --no-cache=true

Deploy the operator image (by default it expects domains in 'default')

  Modify kubernetes/charts/weblogic-operator/values.yaml log-level to 'FINEST'.
  kubectl create ns weblogic-operator
  helm install kubernetes/charts/weblogic-operator --name weblogic-operator --values kubernetes/charts/weblogic-operator/values.yaml --wait --namespace weblogic-operator

You will run the sample in kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image:

  Add logHome and logHomeEnabled to the domain-template.yaml
    edit kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/domain-template.yaml
    logHomeEnabled: true
    logHome: /u01/oracle/user_projects/domains/logs (or something different from default)

  Create kubernetes secrets for admin credentials:
    ./kubernetes/samples/scripts/create-weblogic-domain-credentials/create-weblogic-credentials.sh -u weblogic -p welcome1 -d domain1 -n default -s domain1-weblogic-credentials

  Modify the sample domain custom resource inputs file create-domain-inputs.yaml to match up with your desired environment:
    Use either 12213-domain-home-in-image-wdt or 12213-domain-home-in-image for your domain image for image and imagePath.
    Make sure your secrets name matches.

  Create the domain using:
    cd ./kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image
    ./create-domain.sh -u weblogic -p welcome1 -i create-domain-inputs.yaml -o /path/to/your/output-directory -e
    The -o path must of course refer to an existing location
    The -e says to create/apply a kubernetes object once its yaml is created (e.g. the Domain yaml).
    The user and password must be the same as the ones in your secrets file, it is passed to the image build and persisted in the domain home.

  Once the admin or managed servers are up, attach to one of the pods and check both the logs directory you specified and the servers//logs directory

Cleanup

  kubernetes/samples/scripts/delete-domain/delete-weblogic-domain-resources.sh -d domain1
  helm list
  helm delete --purge weblogic-operator
  kubectl delete ns weblogic-operator

PR #673, opened 2018-12-17
first pass review

rosemarymarano:
Just a few notes:
*We are using sentence capitalization instead of title capitalization.
*Operator is not capitalized, as are most other terms, not capitalized unless they refer to a proper name.
*The doc reference will have to be to the 19c release, which of course, it not yet available.
In general, the doc looks very good.

### DIFF
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -79,7 +79,7 @@ The behavior when using an unsupported override is undefined.
 ---
 # Override template names and syntax
 
-Overrides leverage a built-in WebLogic feature called "Configuration Overriding" (See [References](#References)). Situational configuration consists of XML formated files that closely resemble the structure of WebLogic `config.xml` and system resource module XML files. In addition, the attribute fields in these files can embed `add`, `replace`, and `delete` verbs to specify the desired override action for the field.  
+Overrides leverage a built-in WebLogic feature called "Configuration Overriding" which is often informally called "Situational Configuration." (See [References](#References)). Situational configuration consists of XML formatted files that closely resemble the structure of WebLogic `config.xml` and system resource module XML files. In addition, the attribute fields in these files can embed `add`, `replace`, and `delete` verbs to specify the desired override action for the field.  
 
 ## Override template names
 
@@ -150,7 +150,7 @@ Two types of macros are supported `environment variable macros` and `secret macr
 
 The secret macro `SECRETNAME` field must reference the name of a Kubernetes secret, and the `SECRETKEY` field must reference a key within that secret. For example, if you have created a secret named `dbuser` with a key named `username` that contains the value `scott`, then the macro `${secret:dbuser.username}` will be replaced with the word `scott` before the template is copied into its WebLogic server pod.
 
-**SECURITY NOTE:** Use the `:encrypt` suffix in a secret macro to encrypt its replacement value with the WebLogic WLST `encrypt` command (instead of leaving it at its plain text value).  This is useful for overriding MBean attributes that expect encrypted values, such as the `password-encrypted` field of a data source, and is also useful for ensuring that a custom override situational configuration file the operator places in the domain home does not expose passwords in plain-text.
+**SECURITY NOTE: Use the `:encrypt` suffix in a secret macro to encrypt its replacement value with the WebLogic WLST `encrypt` command (instead of leaving it at its plain text value).  This is useful for overriding MBean attributes that expect encrypted values, such as the `password-encrypted` field of a data source, and is also useful for ensuring that a custom override situational configuration file the operator places in the domain home does not expose passwords in plain-text.**
 
 ## Override template samples
 
@@ -312,7 +312,7 @@ The WebLogic situational configuration feature provides advanced options and cap
 ---
 # References
 
-See the "Managing Configuration Changes" chapter in _Oracle Fusion Middleware Understanding Domain Configuration for Oracle WebLogic Sever_ version 19.1.0.0.0.
+See the "Managing Configuration Changes" chapter in _Oracle Fusion Middleware Understanding Domain Configuration for Oracle WebLogic Server_ version 19.1.0.0.0.
 
 ---
 # Release Notes
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -313,8 +313,3 @@ The WebLogic situational configuration feature provides advanced options and cap
 # References
 
 See the "Managing Configuration Changes" chapter in _Oracle Fusion Middleware Understanding Domain Configuration for Oracle WebLogic Server_ version 19.1.0.0.0.
-
----
-# Release Notes
-
-TBD These are to be moved to a central release notes section.
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -13,7 +13,26 @@ Use configuration overrides (also called _situational configuration_) to customi
   * A file named `version.txt` that contains the string `2.0`.
 * Set your domain resource `configOverrides` to the name of this configuration map.
 * Create Kubernetes secrets that contain template macro values.
-* Set your domain `configOverrideSecrets` to reference the aforementioned secrets.
+* Set your domain `configOverrideSecrets` to reference the aforementioned secrets. For example:
+
+```
+apiVersion: "weblogic.oracle/v2"
+kind: Domain
+metadata:
+  name: domain1
+  namespace: default
+  labels:
+    weblogic.resourceVersion: domain-v2
+    weblogic.domainUID: domain1
+spec:
+  [ ... ]
+  webLogicCredentialsSecret:
+    name: domain1-wl-credentials-secret
+  configOverrides: domain1-overrides-config-map
+  configOverrideSecrets: [domain1-overrides-db1-secret, domain1-config-jms1-secret]
+  [ ... ]
+```
+
 * Start or restart your domain.
 
 ## How do overrides work during runtime?
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -4,13 +4,13 @@
 
 * [Overview](#overview)
 * [Prerequisites](#prerequisites)
-* [Typical overrides](#typicaloverrides)
-* [Unsupported overrides](#unsupportedoverrides)
-* [Override template names and syntax](#overridetemplatenamesandsyntax)
-* [Step-by-step guide](#step-by-stepguide)
+* [Typical overrides](#typical-overrides)
+* [Unsupported overrides](#unsupported-overrides)
+* [Override template names and syntax](#override-template-names-and-syntax)
+* [Step-by-step guide](#step-by-step-guide)
 * [Debugging](#debugging)
-* [Internal design flow](#internaldesignflow)
-* [Advanced situational configuration](#advancedsituationalconfiguration)
+* [Internal design flow](#internal-design-flow)
+* [Advanced situational configuration](#advanced-situational-configuration)
 * [References](#references)
 ---
 # Overview
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -23,8 +23,10 @@ Use configuration overrides (also called _situational configuration_) to customi
   * A file named `version.txt` that contains the string `2.0`.
 * Set your domain resource `configOverrides` to the name of this configuration map.
 * Create Kubernetes secrets that contain template macro values.
-* Set your domain `configOverrideSecrets` to reference the aforementioned secrets. For example:
+* Set your domain `configOverrideSecrets` to reference the aforementioned secrets. See the example below.
+* Start or restart your domain.
 
+Example:
 ```
 apiVersion: "weblogic.oracle/v2"
 kind: Domain
@@ -42,8 +44,6 @@ spec:
   configOverrideSecrets: [domain1-overrides-db1-secret, domain1-config-jms1-secret]
   [ ... ]
 ```
-
-* Start or restart your domain.
 
 ## How do overrides work during runtime?
 * When a domain is first deployed, or is restarted, the operator will:
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -1,108 +1,108 @@
-# Configuration Overrides
+# Configuration overrides
 
-**PLEASE NOTE**:  This page is a work in progress. 
+**PLEASE NOTE**:  This page is a work in progress.
 
 ---
 # Overview
 
-Use configuration overrides to customize a WebLogic domain home configuration without modifying the domain's actual config.xml or system resource files. For example, you may want to override a JDBC Datasource xml module Username, Password, and URL so that it references a local database. 
+Use configuration overrides (also called _situational configuration_) to customize a WebLogic domain home configuration without modifying the domain's actual `config.xml` or system resource files. For example, you may want to override a JDBC datasource XML module user name, password, and URL so that it references a local database.
 
-How do you specify overrides? 
-* Create a Kubernetes config map that contains
-  * Override templates (also known as situational config templates).
-  * A file named 'version.txt' that contains the string '2.0'.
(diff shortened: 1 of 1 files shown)


site/config-overrides.md
```diff
@@ -1,108 +1,108 @@
...
+# Configuration overrides
 
-**PLEASE NOTE**:  This page is a work in progress. 
+**PLEASE NOTE**:  This page is a work in progress.
 
 ---
 # Overview
 
-Use configuration overrides to customize a WebLogic domain home configuration without modifying the domain's actual config.xml or system resource files. For example, you may want to override a JDBC Datasource xml module Username, Password, and URL so that it references a local database. 
+Use configuration overrides (also called _situational configuration_) to customize a WebLogic domain home configuration without modifying the domain's actual `config.xml` or system resource files. For example, you may want to override a JDBC datasource XML module user name, password, and URL so that it references a local database.
```
markxnelson: I am told that "situational config" is not used externally. @rjeberhard can you confirm?

site/config-overrides.md
```diff
@@ -138,34 +138,34 @@ _`wldf-MODULENAME.xml`_
...
 
-**SECURITY NOTE:** Use the `:encrypt` suffix in a secret macro to encrypt its replacement value with the WebLogic WLST encrypt command (instead of leaving it at its plain text value).  This is useful for overriding mbean attributes that expect encrypted values, such as the `password-encrypted` field of a data source, and is also useful for ensuring that a custom override situational config file the Operator places in the domain home does not expose passwords in plain-text.
+**SECURITY NOTE:** Use the `:encrypt` suffix in a secret macro to encrypt its replacement value with the WebLogic WLST `encrypt` command (instead of leaving it at its plain text value).  This is useful for overriding MBean attributes that expect encrypted values, such as the `password-encrypted` field of a data source, and is also useful for ensuring that a custom override situational configuration file the operator places in the domain home does not expose passwords in plain-text.
```
tbarnes-us: I wonder if we should just make this whole paragraph bold?   It's a pretty important fact.   Maybe I should also add an  entire security section that repeats this fact, as well as it's a best practice to use Secrets + macros in the first place rather than hard-coding passwords, and that only the introspector that populates the templates mounts/reads secrets where-as the wl-pods do not mount the secrets?

site/config-overrides.md
```diff
@@ -272,50 +272,49 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
...
+# Advanced situational configuration
 
-WebLogic Situational Config feature provides advanced options and capabilities that are supported, but aren't covered in this document. For example, you can use a wildcard character in place of an mbean name. See [References](#references).
+The WebLogic situational configuration feature provides advanced options and capabilities that are supported, but aren't covered in this document. For example, you can use a wildcard character in place of an MBean name. See [References](#references).
 
 ---
 # References
 
-See the 'Managing Configuration Changes' chapter in 'Oracle� Fusion Middleware Understanding Domain Configuration for Oracle WebLogic Sever' version TBD.
+See the "Managing Configuration Changes" chapter in _Oracle Fusion Middleware Understanding Domain Configuration for Oracle WebLogic Sever_ version 19.1.0.0.0.
```
tbarnes-us: Sever --> Server

site/config-overrides.md
rosemarymarano: Just so happens that I wrote the public doc for the feature. See https://docs-uat.us.oracle.com/en/middleware/weblogic-server/19.1/domcf/changes.html#GUID-3C5A2D0B-437C-41F1-B914-81F0184985CC  Look at the second para, third sentence.

markxnelson: ok. cool by me.  i don't know a thing about it :) i was just told that

rosemarymarano: The feature is called "Configuration Overriding" but it is referenced as situational configuration within the doc because many of the files names use the term.

tbarnes-us: Also pretty much all of the config override WL Server log messages use the term 'situational' - at least the ones I've seen far.    It's a nice thing to know - makes it easy to pick them out fo the clutter.

site/config-overrides.md
```diff
@@ -1,108 +1,108 @@
...
-  * Override templates (also known as situational config templates).
-  * A file named 'version.txt' that contains the string '2.0'.
-* Set your Domain resource `configOverrides` to the name of this config map.
+## How do you specify overrides?
+* Create a Kubernetes configuration map that contains:
+  * Override templates (also known as situational configuration templates).
+  * A file named `version.txt` that contains the string `2.0`.
+* Set your domain resource `configOverrides` to the name of this configuration map.
 * Create Kubernetes secrets that contain template macro values.
-* Set your Domain `configOverrideSecrets` to reference the aforementioned secrets.
```
lennyphan: Would it be good to have Domain CR snippets as examples for showing what it looks like when specifying 'configOverrides' and 'configOVerrideSecrets'?

tbarnes-us:
Sankar has some comments, but doesn't have github access.   I'll incorporate them into this review later today or tomorrow.

site/config-overrides.md
lennyphan: e.g.
==============================

# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.

#
# This is an example of how to define a Domain resource.
#
apiVersion: "weblogic.oracle/v2"
kind: Domain
metadata:
  name: domain1
  namespace: weblogic-domain
  labels:
    weblogic.resourceVersion: domain-v2
    weblogic.domainUID: domain1
spec:
    configOverrides: domain1-weblogic-domain-introspect-test-cm
    configOverridesSecrets:
        - domain1-weblogic-credentials
...

==============================

lennyphan: just a thought.

vanajamukkara:
@tbarnes-us I don't see any mention about automatic situational configuration overrides. Shouldn't that be included in this doc? along with the note for listen address value which should not be set.

tbarnes-us:
@vanajamukkara  Vanaja, automatic sit config overrides happen internally, and are unrelated to custom sit config overrides that this document discusses.   I don't think they need to be brought up here.   Also, the requirement to leave listen address unset or set exactly to the right value is discussed elsewhere (domains.md -- but it looks like a related creating-domains.md is out of date) -- this document simply states that a custom override must not override a network address.

site/config-overrides.md
tbarnes-us: Nice idea.   I'll leave this open and try get to this in the next day or two.

site/config-overrides.md
sankarpn: I think Weblogic domain configuration is more appropriate than Weblogic domain home configuration.

site/config-overrides.md
```diff
@@ -138,34 +138,34 @@ _`wldf-MODULENAME.xml`_
...
 
-## Override Template Samples
+## Override template samples
 
-Here are some sample template override files.  Use 'combine-mode="add"' to specify the values for mbean attributes that have not been set in the original configuration, and use the 'combine-mode="replace"' to specify the values for mbean fields that are already set in the original configuration.
+Here are some sample template override files.  Use `combine-mode="add"` to specify the values for MBean attributes that have not been set in the original configuration, and use `combine-mode="replace"` to specify the values for MBean fields that are already set in the original configuration.
 
-**IMPORTANT: Your overrides may not take effect if you use 'add' for an mbean attribute that already has a value specified in your original configuration, or use 'replace' for an attribute that does not already exist.**
+**IMPORTANT: Your overrides may not take effect if you use 'add' for an MBean attribute that already has a value specified in your original configuration, or use 'replace' for an attribute that does not already exist.**
```
sankarpn: of **if you** use 'replace' for an attribute....

site/config-overrides.md
```diff
@@ -179,15 +179,15 @@ The following `config.xml` override file demonstrates setting the 'max-message-s
 </domain>
 ```
 
-### Overriding a DataSource Module
+### Overriding a data source module
 
-The following `jdbc-testDS.xml` override template demonstrates setting the URL, username, and password-encrypted fields of a JDBC module named `testDS` via secret macros.  The generated situational config that replaces the macros with secret values will be located in the DOMAIN_HOME/optconfig/jdbc directory.   The `password-encrypted` field will be populated with an encrypted value since it uses a secret macro with a `:encrypted` suffix.  The secret is named `dbsecret` and contains three keys: `url`, `username`, and `password`.
+The following `jdbc-testDS.xml` override template demonstrates setting the URL, user name, and password-encrypted fields of a JDBC module named `testDS` via secret macros.  The generated situational configuration that replaces the macros with secret values will be located in the `DOMAIN_HOME/optconfig/jdbc` directory.   The `password-encrypted` field will be populated with an encrypted value because it uses a secret macro with an `:encrypted` suffix.  The secret is named `dbsecret` and contains three keys: `url`, `username`, and `password`.
```
sankarpn: with an encrypted value because it uses a secret macro with an :encrypted suffix. 
should be 
"with an encrypted value because it uses a secret macro with an :**encrypt** suffix."

site/config-overrides.md
```diff
@@ -205,61 +205,61 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
...
+  * Templates must not override the settings listed in [Unsupported overrides](#unsupported-overrides).
+  * Templates must be formatted and named as per [Override template names and syntax](#override-template-names-and-syntax) and [References](#references).
+  * Templates can embed macros that reference environement variables or Kubernetes secrets.  See [Override template macros](#override-template-macros).
+* Create a Kubernetes configuration map from the directory of templates.
+  * The configuration map must be in the same Kubernetes namespace as the domain.
+  * If the configuration map is going to be used by a single `DOMAIN_UID`, it is recommended to add the `weblogic.domainUID=<mydomainuid>` label to help track the resource.
+  * For example, assuming `./mydir` contains your `version.txt` and situation configuration template files:
     ```
     kubectl -n MYNAMESPACE create cm MYCMNAME --from-file ./mydir
     kubectl -n MYNAMESPACE label cm MYCMNAME weblogic.domainUID=DOMAIN_UID
```
sankarpn: kubectl -n MYNAMESPACE label cm MYCMNAME weblogic.domainUID=DOMAIN_UID
 should be 
"kubectl -n MYNAMESPACE label cm MYCMNAME weblogic.domainUID=<DOMAIN_UID>"

site/config-overrides.md
```diff
@@ -205,61 +205,61 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
...
     ```
-* Create any kubernetes secrets referenced by a template macro.
-  * Secrets can have multiple keys (files) that can hold either cleartext or base64 values
-  * Secrets must be in the same kubernetes namespace as the domain
-  * If a secret is going to be used by a single DOMAIN_UID, it is recommended to add the 'weblogic.domainUID=<mydomainuid>' label to help track the resource.
+* Create any Kubernetes secrets referenced by a template macro.
+  * Secrets can have multiple keys (files) that can hold either cleartext or base64 values.
```
sankarpn: Secrets can have multiple keys (files) that can hold either cleartext or base64 values. 

must be reworded to enforce password should be in base64 encoding.

site/config-overrides.md
```diff
@@ -205,61 +205,61 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
...
-  * For all other secrets, add them to Domain CR `configOverrideSecrets` field.
-* See [Debugging](#debugging) for ways to check if sit cfg is taking effect or if there are errors.
+* Configure the name of the configuration map in the domain CR `configOverrides` field.
+* Configure the names of each secret in domain CR.
+  * If the secret contains the WebLogic admin `username` and `password` keys, set the domain CR `webLogicCredentialsSecret` field.
+  * For all other secrets, add them to domain CR `configOverrideSecrets` field.
+* See [Debugging](#debugging) for ways to check if the situational configuration is taking effect or if there are errors.
 
 ---
 # Debugging
```
sankarpn: Debugging should include some info like

The following flags can be added in the domain resource file to enable Debug.

env:
    - name: JAVA_OPTIONS
      value: "-Dweblogic.debug.DebugSituationalConfig=true -Dweblogic.debug.DebugSituationalConfigDumpXml=true"

site/config-overrides.md
tbarnes-us: The idea is to keep using 'domain home' when referencing a domain home configuration. The word   'domain' alone is very ambiguous, since the Operator project uses domain do describe the Domain kubernetes resource.

site/config-overrides.md
tbarnes-us: I leave the change up to @rosemarymarano, she has far greater command of English grammar and Oracle's sentence structure guidelines than I.

site/config-overrides.md
tbarnes-us: Good catch.  `:encrypted` should be `:encrypt`.

site/config-overrides.md
tbarnes-us: The existing DOMAIN_UID without the <> is better.    The convention throughout this document is to purely use caps to represent replaceable values - as you can likely see from the other replaceable values in the same command line.

site/config-overrides.md
tbarnes-us: Rosemary, you can add something like '(It's recommended to use base64 values for passwords via 'Opaque' type secrets in their 'data' field, just so that they can't be easily read at a casual glance.  See https://kubernetes.io/docs/concepts/configuration/secret/ for more information.)'.

site/config-overrides.md
tbarnes-us: Rosemary, please add a final bullet to the debugging section, that's something like:  To cause the WebLogic Situational Config feature to produce additional debugging information in the WebLogic Server logs, configure the "JAVA_OPTIONS" env var in your domain resource with "-Dweblogic.debug.DebugSituationalConfig=true -Dweblogic.debug.DebugSituationalConfigDumpXml=true".

site/config-overrides.md
rosemarymarano: Got it.

site/config-overrides.md
rosemarymarano: Got it. Done.

site/config-overrides.md
rosemarymarano: Got it.

site/config-overrides.md
```diff
@@ -272,50 +272,49 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
...
-* WebLogic Servers read their overrides from their domain home's 'optconfig' directory.
+  * Deletes any situational configuration files in the `optconfig` directory that do not have corresponding template files in the configuration map.
+* WebLogic Servers read their overrides from their domain home's `optconfig` directory.
 
 ---
-# Advanced Situational Config
+# Advanced situational configuration
 
-WebLogic Situational Config feature provides advanced options and capabilities that are supported, but aren't covered in this document. For example, you can use a wildcard character in place of an mbean name. See [References](#references).
+The WebLogic situational configuration feature provides advanced options and capabilities that are supported, but aren't covered in this document. For example, you can use a wildcard character in place of an MBean name. See [References](#references).
```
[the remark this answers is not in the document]
tbarnes-us: We actually don't want folks to set the timeout most of the time.   The primary use case is overriding things like JDBC URLs, etc.    (Which reminds me, we should probably add a validation check for this -- I'll update the Introspector 2.0 wiki to add it to the list.)

site/config-overrides.md
```diff
@@ -138,34 +138,34 @@ _`wldf-MODULENAME.xml`_
 <wldf-resource>
 ```
```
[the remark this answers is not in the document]
tbarnes-us: It would help, but I chose not to because of time constraints.   Ideally, we need a working .xml that demonstrates overriding a "Foreign JMS" URL, username, and password.     We'll very likely add something at a later time in a point release.

PR #675, opened 2018-12-17
WIP - Feature server lifecycle site doc

moreaut:
Add server lifecycle site documentation.

Note: please look at the TBD in 'Properties that Cause Servers To Be Restarted'


moreaut:
Hi Carolyn,

Thanks for the feedback.

1) restartLabel was a typo - changed it to restartVersion

2) Ryan recently added support for the maxUnavailable property, which defaults to 1.  This changed the rolling restart behavior to roll one server at a time.  Added more info about maxUnavailable to the site doc.

3) Added info to the site doc stating that removing restartLabel from the domain resource also causes restarts.

PR #682, opened 2018-12-18
Fix the operator REST swagger endpoint

moreaut:
The operator's swagger REST endpoint was broken because the swagger file wasn't being copied into the operator's jar.  Changed pom.xml to fix this.

The operator's swagger's index.html changed because the build indented it differently.  Check in the newest version.


rjeberhard:
I'm making a change in the same area to generate swagger doc for the domain resource, so I've included your fixes in my PR.

PR #687, opened 2018-12-18
Test for changes to numerous Pod settings

russgold:
(replaces PR 640)

OWLS-70504 Added unit tests for new PodHelper functionality, and updated the code to make the tests pass. Some minor refactoring to extract settings needed for both tests and production code into PodDefaults.


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -332,12 +308,84 @@ private static boolean isCurrentPodValid(V1Pod build, V1Pod current) {
...
+  }
+
+  private static Map<String, String> getCustomerLabels(V1ObjectMeta metadata) {
+    Map<String, String> result = new HashMap<>();
+    for (Map.Entry<String, String> entry : metadata.getLabels().entrySet())
+      if (!isOperatorLabel(entry)) result.put(entry.getKey(), entry.getValue());
+    return result;
+  }
+
+  private static boolean isOperatorLabel(Map.Entry<String, String> label) {
```
moreaut: I don't think we tell customers they can't add labels that start with 'weblogic.'

Is there a way you could rework this to tell it to ignore the ones we know that k8s adds instead?

moreaut: How about adding a label to the server pod that lists the names of the labels the user added (since this code has access to the domain resource).

Then you could use that label to find the names of the labels to compare when deciding if a pod is valid.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodDefaults.java
```diff
@@ -0,0 +1,84 @@
+// Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Licensed under the Universal Permissive License v 1.0 as shown at
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.helpers;
+
+import static oracle.kubernetes.operator.KubernetesConstants.*;
```
markxnelson: do not use wildcard imports

russgold: Updated. What is the reason for that rule?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
moreaut: Russ said that Ryan is OK with telling customers that they're not allowed to add labels that start with 'weblogic.'.

Do/should we check for this anywhere?

And where is this restriction documented?

PR #710, opened 2018-12-20
discard client on watch failure; extend socket timeout

russgold:
When a watch fails, the client is sometimes rendered unusable, but we have been returning it to the pool, so that subsequent watch calls also fail. This is a short-term fix for that. When the exception is detected, we force the watch implementation to get a new client to be returned to the pool in place of the bad one.

A cleaner fix will move the responsibility for the client pooling to the Watcher itself.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
@@ -126,7 +126,13 @@ abstract class Watcher<T> {
                 .withResourceVersion(resourceVersion.toString())
                 .withTimeoutSeconds(tuning.watchLifetime))) {
       while (watch.hasNext()) {
-        Watch.Response<T> item = watch.next();
+        Watch.Response<T> item;
+        try {
+          item = watch.next();
+        } catch (Throwable e) {
+          watch.discardClient();
+          throw e;
+        }
 
         if (isStopping()) setIsDraining(true);
         if (isDraining()) continue;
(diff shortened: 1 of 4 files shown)


operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -126,7 +126,13 @@ private void watchForEvents() {
                 .withResourceVersion(resourceVersion.toString())
                 .withTimeoutSeconds(tuning.watchLifetime))) {
       while (watch.hasNext()) {
-        Watch.Response<T> item = watch.next();
+        Watch.Response<T> item;
+        try {
+          item = watch.next();
+        } catch (Throwable e) {
+          watch.discardClient();
+          throw e;
```
tbarnes-us: Is the socket exception still logged as a WARNING?   If so, can we (A) detect the exact exception is a socket exception, and (B) only log a FINE in taht case?

rjeberhard: Unfortunately, we can't.  These exceptions are getting printed by threads that the HTTP client creates.  I have some (not done) work where I override the ThreadFactory of the HTTP client library and attempt to suppress this and other exceptions.

russgold: Worth exploring. Not sure yet how to distinguish, as we just get an exception from the k8s api

rjeberhard: Yes, but the exception has already been printed by another thread.

PR #714, opened 2018-12-20
Develop server lifecle site doc pod obsolete fix

moreaut:
Make the site doc in server-lifecycle.md that talks about when the operator automatically rolls pods match the operator runtime.

Add domain resource schema doc to say that customer supplied pod and service label names can't start with 'weblogic.'.


moreaut:
I added the restriction to podLabels and serviceLabels.
I didn't add them to podAnnotations or serviceAnnotations.

PR #716, opened 2018-12-20
POST-2.0-rc1: Sit-cfg override fixes:

tbarnes-us:
Also tracked by https://github.com/oracle/weblogic-kubernetes-operator/issues/727

This is a must for the 2.0 operator.
- MAJOR
-- Move log location sit-cfg override stanzas to just after listen-address (work-around for XML parsing issue seen in domain-in-image wl server log).  This works around a problem which occurs with some configurations and is waiting on a WL fix, and allows the operator to override a domain home's listen-address and log-location instead of leaving them at their original values.  

- MINOR
-- Call 'add' instead of 'replace' if overriding a missing Listen address (this is a best practice, 'replace' works but that's not documented as supported).
-- Do not 'replace' or 'add' an unset NAP ListenAddress (corrects mock/unit test issue seen in a particular T3Channel, no override is actually needed as the default NAP LA comes from server default LA).
-- Do not add SSL listen-address overide stanzas since there's no listen address field in an SSL stanza (this wasn't causing an issue, discovered during code inspection).
-- Update introspect mock/unit tests to cover even more use cases.

Test status:

- Passes QUICKTEST, 'Jenkins' (full integration test run), ad-hoc mock/unit introspector testing, and ad-hoc domain-in-image sample testing.


tbarnes-us:
- The change passes QUICKTEST, ad-hoc mock/unit introspector testing, and ad-hoc domain-in-image sample testing.

- Even so, we probably should wait until after the holiday break to merge this into develop (and then eventually onto rc1), since the change likely won't get much testing until then...

PR #724, opened 2018-12-21
final doc review

markxnelson:


site/wlst.md
```diff
@@ -1,9 +1,5 @@
-**TODO** review and update
-
 # Using WLST
 
-Note that a video demonstration of using WLST to create a data source is available [here](https://youtu.be/eY-KXEk8rI4).
-
 You can use the WebLogic Scripting Tool (WLST) to manage a domain running in Kubernetes.  If the domain was configured to expose a T3 channel using the `exposeAdminT3Channel` setting when creating the domain, then the matching T3 service can be used to connect.  For example, if the `domainUID` is `domain1`, and the Administration Server name is `admin-server`, then the service would be called:
```
tbarnes-us: 1:  `exposeAdminT3Channel` is 1.0, 2.0 is `adminService`, 

2:  I think the service name has changed from `domain1-admin-server-extchannel-t3channel (but I don't know exactly what it changed too).

3:  An alternative to setting up a node-port (which can be a security risk) is to:
A:  Set up LB that directs traffic to a non-node-port service (ports still must match across the board).   

4: Here's an alternative to access  through a node-port that naive readers may not know about, and should use if that's all they need:  kubectl exec into a WL pod in your namespace, and then type 'wlst.sh'.

markxnelson: 1. thanks, will update
2. no i checked, its still called that
3. i dont think t3 will work that way, which is why we did it this way
2. yeah, but then they have to have kubernetes perms too, not just weblogic ones :)

tbarnes-us: 3. can work - I've seen customers do it - but haven't tried it myself

tbarnes-us: 4.  is definitely the most convenient if you already have the permissions.    plus note that it's fairly risky to open up an admin node-port -- this essentially exposes an RMI port that pokes through a firewall - especially risky if the node-port shows up on the public web

PR #730, opened 2018-12-22
Log when pods are replaced and use null for LOG_HOME when effective log home is not set

rjeberhard:


moreaut:
I manually tested this against a previously built image containing a domain home and it worked for me.
Initially, I configured my domain to start 2 managed servers.
Then I changed the replica count to 1, and one ms terminated.
Then I changed the replica count back to 2, and one ms started.
The other running servers did not restart (as expected).

PR #733, opened 2018-12-25
Update the domain home in image sample to keep server logs on PV

simon-meng-cn:
The change contains following:
1. Add the logHome and PVC configuration items for domain home in image sample.
2. Refactor the two samples domain home in image and domain home on PV  to share common code.

The function does not work as expected. Server logs are still written to the default location domainHome/servers/xxx-server/logs.
If setting logHomeEnabled to true in domain.yaml, server logs can be redirected to the configured logHome, but start admin-server failed, saw exception:

java.lang.RuntimeException: weblogic.descriptor.DescriptorException: Unmarshaller failed: com.bea.xml.XmlException: javax.xml.stream.XMLStreamException: Unexpected element '{http://xmlns.oracle.com/weblogic/domain}name' in text content.


simon-meng-cn:
The work-around works fine for me. Start server does not report error, server logs are redirected to the configured logHome.

I have a question for the logHomeEnabled property. It must be set to true in the domain-home-in-image sample. But in another sample domain-home-on-pv, the logHome override takes effect without setting this property explicitly.

doxiao:
I am wondering if we need to make the logs-on-pv part optional.  With what is in the current PR, PV and PVC are required.  It might be helpful if customers could experiment with the samples/Operator without having to create a PV/PVC when they just want to kick-the-tires, and they can add it once they get the basic things working. On the other hand, with logs persisted, it is easier for customers to debug issues. I am on the fence with this.

markxnelson:
i think it should be optional

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/README.md
```diff
@@ -1,14 +1,17 @@
 # WebLogic sample domain home in Docker image
 
-The sample scripts demonstrate the creation of a WebLogic domain home in a Docker image. The scripts also generate the domain YAML file, which can then be used to start the Kubernetes artifacts of the corresponding domain. Optionally, the scripts start up the domain, and WebLogic Server pods and services.
+The sample scripts demonstrate the creation of a WebLogic domain home in a Docker image. The log home is on an existing Kubernetes persistent volume (PV) and persistent volume claim (PVC). The scripts also generate the domain YAML file, which can then be used to start the Kubernetes artifacts of the corresponding domain. Optionally, the scripts start up the domain, and WebLogic Server pods and services.
```
rosemarymarano: Should this sentence be qualified regarding whether logHomeOnPV is enabled?
"The log home is on an existing Kubernetes persistent volume (PV) and persistent volume claim (PVC)."

For example, "If `logHomeOnPV` is enabled, then the log home resides on an existing Kubernetes persistent volume (PV) and persistent volume claim (PVC)."

simon-meng-cn: Fixed in the new change. Thanks

PR #734, opened 2018-12-29
Quickstart fix

lilyhe123:
The major change is to fix the step to pull wls operator image, otherwise installing operator will fail. 
Other changes are to provide the more specific cmds.


markxnelson:
looks reasonable @rjeberhard

PR #740, opened 2019-01-03
Clean up fix for failed clients being returned to the pool

russgold:
We had identified a problem, whereby a Kubernetes ApiClient would be returned to the client pool and reused, even if it had failed and could no longer succeed. I put in a quick fix for it.

This is a simpler change, with unit tests.


operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017, 2018 Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2017, 2018, 2019 Oracle Corporation and/or its affiliates.  All rights reserved.
```
rjeberhard: Per our discussion, this becomes "2017, 2019"

PR #745, opened 2019-01-07
Remove run.sh and its mentions

simon-meng-cn:
Jenkins passed. http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/801/


site/v1.1/README.md
```diff
@@ -167,7 +167,7 @@ For information about how to shut down a domain running in Kubernetes, see [Shut
 
 To permanently remove the Kubernetes resources for a domain from a Kubernetes cluster, run the [Delete WebLogic domain resources](/kubernetes/delete-weblogic-domain-resources.sh) script. This script will delete a specific domain, or all domains, and all the Kubernetes resources associated with a set of given domains. The script will also attempt a clean shutdown of a domain’s WebLogic pods before deleting its resources.  You can run the script in a test mode to show what would be shutdown and deleted without actually performing the shutdowns and deletions.   For script help, use its `-h` option.
 
-The script will remove only domain-related resources which are labeled with the `domainUID` label, such as resources created by the [Create WebLogic domain](/kubernetes/create-weblogic-domain.sh) script or the [integration tests](/src/integration-tests/bash/run.sh).  If you manually created resources and have not labelled them with a `domainUID`, the script will not remove them.   One way to label a resource that has already been deployed is:
```
doxiao: I don't think we should remove the references to run.sh from V1.1 doc because run.sh is only removed in V2.0.

src/integration-tests/bash/archive.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
-# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
+# Copyright 2017, 2019, Oracle Corporation and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
 #
 # archive.sh <source_dir> <target_dir>
-#   - internal helper method called by run.sh
```
doxiao: Is this helper script used anywhere else except for the run.sh that is removed in this PR? If not, shall we remove this helper script as well?

src/integration-tests/bash/cleanup.sh
```diff
@@ -321,7 +318,7 @@ if [ -x "$(command -v helm)" ]; then
      helm delete --purge  $helm_name
   done
 
-  # cleanup tiller artifacts that are created in run.sh
```
doxiao: If cleanup.sh only cleans up artifacts that are created in run.sh and run.sh is not removed, shall we remove this helper script as well? Loop in QA and see if cleanup.sh is still in use.

src/integration-tests/bash/job.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/bash
-# Copyright 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
+# Copyright 2017, 2019, Oracle Corporation and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
 #
 # job.sh <command> [<optarg1>] [<optarg2>] ...
 #
-#    This is a helper script called by run.sh & cleanup.sh.  It runs the given bash
```
doxiao: This helper script may need to be removed as well if we confirm that we need to remove cleanup.sh.

src/integration-tests/bash/cleanup.sh
simon-meng-cn: The cleanup.sh is called in the java test and other test scripts.

src/integration-tests/bash/job.sh
simon-meng-cn: The job.sh is also called in java test.

src/integration-tests/bash/archive.sh
simon-meng-cn: The archive.sh is not only called by run.sh, it is also called by other test scripts.

site/v1.1/README.md
simon-meng-cn: A question: why release V1.1 doc in 2.0 source code?

src/integration-tests/bash/cleanup.sh
doxiao: okay. thanks.

src/integration-tests/bash/job.sh
doxiao: okay.

src/integration-tests/bash/archive.sh
doxiao: Thanks for checking.

site/v1.1/README.md
doxiao: @markxnelson @rjeberhard I don't remember the details about this. can you also confirm if my understanding/comments are correct?

rjeberhard: The 1.1 docs were preserved so you could see the documentation for that release.  In general, I agree that we should not change the historical documents.

PR #746, opened 2019-01-07
(WIP) OWLS-70886: server services not recreated when labels/annotations changed	

chunshi-ora:
- Fixed so that a service is recreated in case that label/annotation/selector is changed.
- Added necessary tests (WIP).

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
@@ -624,20 +624,31 @@ public class ServiceHelper {
   }
 
   private static boolean validateCurrentService(V1Service build, V1Service current) {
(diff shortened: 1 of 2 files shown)


russgold:
I'm a bit confused, here. My understanding was that the intent was for the _Admin_ service to be recreated when labels or annotations were changed, and that we already had tests for replacing the other services (ServiceHelperTest).

This seems to duplicate the existing unit tests, but doesn't add the required functionality for the admin service.

Did you run the unit tests with coverage to see that your new tests were needed?

chunshi-ora:
From the JIRA, looks like the needed changes are applied across all services, cluster, server and external channel.

The ServiceHelpTest covers some of test cases. I was about to update it with additional tests. However, I realized that the tests are combination of a few instances to be tested against a set of similar test cases.

Therefore, I am trying to parameterize them to make less verbose and reduce duplication. Duplicate test cases in ServiceHelpTest would be removed.

PR #750, opened 2019-01-08
allow different startup policies for domains, servers and clusters

russgold:
Moved serverStartPolicy field down one level in order to allow different options for domain, servers and clusters.
Used qualifier to select enum values


rjeberhard:
Thanks, Russ!

PR #754, opened 2019-01-08
OWLS-70853 remove configured cluster and default domain-home-in-image sample to use WLST

doxiao:
Changes in this PR:
1) Remove support to Configured cluster from the create domain samples; updated Java integration test domain yaml files. 
2) Change the default of domain home in image sample to use WLST, instead of WDT.

http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/821/ (passed)


integration-tests/src/test/resources/domain5.yaml
```diff
@@ -4,7 +4,6 @@
 adminServerName: admin-server
 domainUID: domain5
 clusterName: cluster-1
-clusterType: CONFIGURED
```
simon-meng-cn: Why remove this property instead of change it to DYNAMIC?
It looks to me, clusterType should be changed to DYNAMIC in domain5,  don't be deleted from domain6 and domain9.

doxiao: The clusterType property does not existing in the inputs file any more.

PR #756, opened 2019-01-08
new quickstart

lilyhe123:
This is a new quickstart to setup from scratch in a reliable and automated way.
I try to make it simple, self-contained and straightforward. The major dependencies are the three helm charts: wls operator, traefik controller and ingress-per-domain. See README for detail.


moreaut:
Where did the requirement for this new sample come from?  I wasn't aware it was being developed.

rjeberhard:
Hi @lilyhe123, like Monica and Tom, I didn't know that you were working on this.  Is your intention to replace the existing quick start documentation or to augment and simplify it?  Let's hold back on proposing this for 2.0 and do a few more rounds of review and thinking about how this relates to the existing quick start.  Also, if I had known you were working on this then I would have kept you up-to-date on my work to correct some final Domain schema issues that, since it's now in, will impact your solution.

PR #759, opened 2019-01-09
OWLS-70886: server services not recreated when labels/annotations changed

chunshi-ora:
- Fixed so that a service is recreated in case that label/annotation/selector is changed.
- Added necessary tests.


chunshi-ora:
I am still working on tests for ExternalService.

chunshi-ora:
Please review if you have a chance.

PR #766, opened 2019-01-09
WIP OWLS-71193 - update affected integration tests

moreaut:
One of the integration tests was depending on there only being one config map with the domain uid label (weblogic-domain-cm).  Now there is another one (<domainuid>-weblogic-domain-introspector-cm).


moreaut:
Please wait to merge this until after http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/841/ passes.

tbarnes-us:
Can you remind me why there’s a domain-uid label on weblogic-domain-cm?   I assume it isn’t set to anything since it’s meant to work with multiple domains.   Has it always been there?   I don’t recall it being there before.

moreaut:
Sorry - error in my description.  The weblogic-domain-cm doesn't, and shouldn't, have a domain uid label.

However, the per-domain introspector cm was missing the label and should have it, so I added it.

The test was previously expecting to find one config map with the domain uid label (maybe from the create domain job), and started failing because now there are 2 (because of my change).

PR #769, opened 2019-01-10
fix selector of voyager-stats

lilyhe123:
otherwise the voyager stats can not show up in browser


moreaut:
Looks good do me.

markxnelson:
@rjeberhard i am ok to put this in 2.0-rc2

PR #771, opened 2019-01-10
doc update:  further emphasize need to restart domain after custom override changes

tbarnes-us:
custom override doc changes:
- further emphasize need to restart domain after custom override changes
- to aid clarity, move/copy some text, and update TOC


moreaut:
LGTM

Issue #772, opened 2019-01-11
Quickstart needs to be updated for rc1


namanshenoy:
Current quick start does not work with latest version.

moreaut:
Can you please try the 2.0-rc2 branch?

And, if you jave problems in the quick start, please provide more details.

Thanks, -Tom

moreaut:
Thanks for the clarification.

It's more than a fix to the readme - you modified the product too (i.e. the operator helm chart's values.yaml).

alai8:
Could you check if https://github.com/oracle/weblogic-kubernetes-operator/pull/776 has addressed the issue? Thanks

alai8:
The quick start guide for 2.0-rc2 has been updated with the correct image name for the operator rc2 image specified in the helm install command example:

```
$ helm install kubernetes/charts/weblogic-operator \
  --name sample-weblogic-operator \
  --namespace sample-weblogic-operator-ns \
  --set image=oracle/weblogic-kubernetes-operator:2.0-rc2 \
  --set serviceAccount=sample-weblogic-operator-sa \
  --set "domainNamespaces={}" \
  --wait
```

PR #773, opened 2019-01-11
Fix for #772

namanshenoy:
Fix for #772 
This will fix the quick start to make users pull in the cirrect weblogic-kubernetes-operator image from Dockerhub


moreaut:
Also, this is more than a doc fix - it modifies the operator helm chart's default values too.

namanshenoy:
You're right, I'm not sure if it is necessary to change the chart's values. I'll try that and get back to you. The doc change was a change to the command suggested by Mark in slack that I added to the doc

rjeberhard:
I've adjusted this slightly in #776.  Let me know what you think.

PR #776, opened 2019-01-11
Use oracle/weblogic-kubernetes-operator:2.0 and update quickstart instructions

rjeberhard:
Set the default for image in our helm chart values to oracle/weblogic-kubernetes-operator:2.0, so that it will be right when we release 2.0.

Update the quick start instructions to *NOT* say to tag 2.0-rc2 as 2.0 -- this would lead to confusion later.

The helm install step had already been updated to include "--set image=oracle/weblogic-kubernetes-operator:2.0-rc2".

When we release 2.0, we'll update the quick start to remove these references to 2.0-rc2.

### DIFF
--- a/kubernetes/charts/weblogic-operator/values.yaml
+++ b/kubernetes/charts/weblogic-operator/values.yaml
@@ -20,7 +20,7 @@ domainNamespaces:
   - "default"
 
 # image specifies the docker image containing the operator code.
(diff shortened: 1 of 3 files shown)


rjeberhard:
I've adjusted in this slightly in #776.  Let me know what you think.

PR #781, opened 2019-01-12
modify patching instructions

rosemarymarano:
Modified patching instructions according to text from Monica.

### DIFF
--- a/site/quickstart.md
+++ b/site/quickstart.md
@@ -45,9 +45,12 @@ e.	Pull the WebLogic 12.2.1.3 install image:
 ```
 $ docker pull store/oracle/weblogic:12.2.1.3
 ```
-f.	**TODO** remove this item when Monica has published the new image with the patch in it **TODO**
-    Then patch the WebLogic image according to these [instructions](https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12213-patch-wls-for-k8s),
-    and copy the image to all nodes in your cluster, or put it in a Docker registry that your cluster can access.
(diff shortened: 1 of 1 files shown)


markxnelson:
@mriccell don't we want to put the "docker pull xxxx" in the quickstart? we tell them how to pull the operator image, not how to build it, for example.   and save the patching details for the user guide?

mriccell:
We do tell them to pull the weblogic image from docker store in the quick start guide.
The modifications I have here is because the store image already has one of the patches so the patching sample needs to be modified.

rosemarymarano:
As per your suggestion, added detailed information regarding credential requirements.

site/quickstart.md
```diff
@@ -170,11 +174,19 @@ domain namespace (`sample-domain1-ns`) and the `domainHomeImageBase` (`oracle/we
 For example, assuming you named your copy `my-inputs.yaml`:
 ```
 $ cd kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image
-$ ./create-domain.sh -i my-inputs.yaml -o /some/output/directory -u username -p password -e
+$ ./create-domain.sh -i my-inputs.yaml -o /some/output/directory -u weblogic -p welcome1 -e
```
mriccell: Why did we change the username and password to specific ones?

rosemarymarano: It was a recommendation to me from Tom M to be more specific and emphasize that the user name and password needed to be the same as the one specified in the WebLogic credentials. (It also fits with the following sentence which refers to it: "You need to provide the WebLogic administration user name and password in the -u and -p options respectively, as shown in the example."

PR #782, opened 2019-01-12
Domain status schema in CRD

rjeberhard:
While working with Michael, we discovered that the operator was not updating status for domain resources.  This was bad for diagnosability, but also affected rolling scenarios, as the operator uses the current status to determine how many servers in the cluster are healthy.  He was seeing the operator restart servers in-mass because the operator thought they were already unhealthy.

I discovered that the API server was rejecting status updates.  There appear to be two problems: later versions of K8s (1.12+) want status updates to use the /status resource, and that the status didn't match the schema validation.

This proposed change corrects the schema generation for status.  Locally, with 1.13, I now see status updates.  However, I'm not seeing status updates on the OCI clusters and I'm not yet sure why.  That's why this PR is WIP.

Given that this touches the schema and because of the failure in status updating and rolling, we'll most likely want to get a completed fix into 2.0.


rjeberhard:
Here's what I found:

Background: Kubernetes has been slowing building out the CRD support so that CRD-defined types are feature compatible with built-in types (e.g. Pod).  In <= 1.9, there was limited support for the distinction between "spec" and "status" portions of the type.  There was no support for the "/status" path.  In 1.10, the CRD can now express a sub-resource that says to enable the "/status" path.  However, I've found that it's completely broken until 1.13.

For the time being, we'll go back to not using the status sub-resource.

Since all changes go in through the main REST path for a domain, the status has to also validate against the schema.  I found a few problems; the biggest that we were improperly generating the schema for DateTime.  

Finally, I cleaned-up a bunch of descriptions.

PR #787, opened 2019-01-14
incorporate edits from Bob D

rosemarymarano:
Including:
* Noting that the guide assumes a single node cluster.
* Moved up the instruction to grant the Helm service account the cluster-admin role.
* Changed the “kubectl log” command to “kubectl logs” (in both files).
* Added the path for the kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml file.

### DIFF
--- a/site/helm-charts.md
+++ b/site/helm-charts.md
@@ -495,7 +495,7 @@ The `helm install` eventually times out and creates a failed release.
 $ helm install kubernetes/charts/weblogic-operator --name op2 --namespace myuser-op2-ns --values o24.yaml --wait --no-hooks
 Error: release op2 failed: timed out waiting for the condition
 
-kubectl log -n kube-system tiller-deploy-f9b8476d-mht6v
+kubectl logs -n kube-system tiller-deploy-f9b8476d-mht6v
 ...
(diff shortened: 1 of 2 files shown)


tbarnes-us:
@rosemarymarano @moreaut Some general comments, unrelated to this specific change:
- The file name (helm-charts.md) and doc title are confusing (), since the focus is on installing and running the operator - not helm itself.

tbarnes-us:
Some general comments, unrelated to this specific change:

- The file name (helm-charts.md) and doc title ('using operator helm charts') are confusing to me since:
-- The focus is on installing and running the operator - not helm in particular.   Helm is just a tool we use to get the operator running.   
-- We also need a helm-charts instruction for the load balancers, but the focus of the document is the operator process.  
- So how about 
-- (A) move the bulk of the contents to a new file named something like 'operator.md' with title like 'Managing the Operator'
-- (B) leave only the helm/tiller install/setup guidance, retitle to 'setting up helm and tiller', and reference this file from from 'operator.md', 'install.md', 'ingress.md', and any other place that has helm install as a pre-req

Also:

- If removing the operator does not remove the CRD, then the directions for removing the operator in helm-charts.md should be updated to mention this step.

rosemarymarano:
All good suggestions, Tom. I will work on them.

PR #792, opened 2019-01-15
OWLS-71197

simon-meng-cn:
Changes in this PR:

1) replace imageRegistry, image, imageTag in values.yaml with just image that has the full image name
2) remove the 'release' label (for symmetry with the other samples) (service account, cluster role, cluster role binding, secret, service, deployment)
3) remove the 'app' label (for symmetry with the other samples (service account, cluster role, cluster role binding, secret, service - keep it on the deployment)
4) add a weblogic.resourceVersion: loadbalancer-v2 label (service account, cluster role, cluster role binding, secret, service, deployment)

### DIFF
--- a/kubernetes/samples/charts/apache-webtier/README.md
+++ b/kubernetes/samples/charts/apache-webtier/README.md
@@ -35,23 +35,21 @@ The following table lists the configurable parameters of the Apache webtier char
 
 | Parameter                          | Description                                                   | Default               |
 | -----------------------------------| ------------------------------------------------------------- | ----------------------|
-| `imageRegistry`                    | Docker registry used to pull image                            | `store/oracle`        |
-| `image`                            | Apache webtier Docker image                                   | `apache`              |
-| `imageTag`                         | Image tag for the Apache webtier Docker image                 | `12.2.1.3`            |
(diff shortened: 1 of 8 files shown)


rjeberhard:
Presently, I'm not planning on merging this to `2.0-rc2`.  Do you want to make the case that it should be?

rjeberhard:
The "weblogic.resourceVersion" labels are present to support future product versions' upgrade behavior.  The use case is a customer who has a domain, or in this case ingress, and then upgrades to a newer version of the operator and/or a newer version of these samples.  The product will need a way to detect outdated resources and then behave properly -- either working with the existing resources or replacing them with modern versions.  Note: we've chosen versions like "loadbalancer-v2" now, but we will only update these versions if the generated resources change.

doxiao:
Thanks for the explanation. 
In that case, should we change  ingress-per-domain samples to use loadbalancer-v2 as well?
In addition, the different load balancer samples may evolve in different paces. Shall we use different  labels in samples of different types, for example, apache-lb-v2, imgress-per-domain-lb-v2, and etc?

rjeberhard:
If you think that the samples could evolve independently then you're last example, e.g. "apache-lb-v2" is the correct value.

PR #793, opened 2019-01-15
incorporate suggesstions from Tom B

rosemarymarano:
Including:
* Moved text from install.md into the User Guide.
* Renamed helm-charts.md to install.md (changed title from "Using operator Helm charts" to "Create and manage the operator."
* Deleted helm-charts.md.
* Re-org'ed REST bulleted item in User Guide TOC.
* Added deleting the domain custom resource definition to the "Removing the operator" in install.md.


site/install.md
```diff
@@ -1,15 +1,556 @@
 # Create and manage the operator
 
-An operator is an application-specific controller that extends Kubernetes to create, configure, and manage instances
```
tbarnes-us: It looks like the information here that was moved to 'user-guide' could also be copied to the Overview/Intro section of the new 'install.md' (the former helm-charts.md), and would make its introduction nicer...   E.g. it'd be helpful if the introduction to 'create and manage the operator' described what the operator was in the first place.

site/install.md
```diff
@@ -1,15 +1,556 @@
 # Create and manage the operator
```
tbarnes-us: The file name 'operator-setup.md' or 'operator-install.md' would be more specific and more clear IMO.    In addition,  maybe 'rest.md' could become 'operator-rest.md'.    @rjeberhard what do you think?

site/install.md
rosemarymarano: Sure, I'll put some of the information into the Overview section.

rosemarymarano: Actually, it's going into the Intro section, prior to the Overview.

site/install.md
rjeberhard: Hmmm... I'm all for clarity, but I don't think this is necessary.  Since this is the operator project, my guess is that most people would assume that setup.md, install.md, and rest.md were for the operator.

site/install.md
```diff
@@ -1,8 +1,10 @@
...
 
+An operator is an application-specific controller that extends Kubernetes to create, configure, and manage instances of complex applications. The Oracle WebLogic Server Kubernetes Operator (the "operator") simplifies the management and operation of WebLogic domains and deployments.
+
 ## Overview
 
-The WebLogic Kubernetes Operator uses Helm to create and deploy any necessary resources and then run the operator in a Kubernetes cluster. Helm helps you manage Kubernetes applications. Helm charts help you define and install applications into the Kubernetes cluster. The operator's Helm chart is located in the `kubernetes/charts/weblogic-operator` directory.
+The operator uses Helm to create and deploy any necessary resources and then run the operator in a Kubernetes cluster. Helm helps you manage Kubernetes applications. Helm charts help you define and install applications into the Kubernetes cluster. The operator's Helm chart is located in the `kubernetes/charts/weblogic-operator` directory.
```
tbarnes-us: Since helm wraps the operator runtime, the phrase 'the operator uses Helm' can be a little confusing.  One needs to deduce that 'operator' in this context is the operator project as a whole, and then immediately switch to thinking the 'operator' in  this context is the operator runtime.      How about:

Helm is used to create and deploy and necessary operator resources and to run the operator in a Kubernetes cluster.   Helm is a framework that helps you manage Kubernetes applications, and `helm charts` help you define and install Helm applications into a Kubernetes cluster.

site/install.md
tbarnes-us: Hmm, I'm still not so sure.   We also have REST, setup, and install guidance that is not specific to the operator runtime.   One can call REST on the admin server, one can setup & install a domain, one can setup k8s itself, etc, etc.   But these three files 'setup.md' 'install.md' 'rest.md' are all very specific to the operator runtime -- they don't have anything directly to do with setting up a domain or setting up k8s itself.

PR #797, opened 2019-01-15
An override file sometimes needs a domain name

tbarnes-us:
_Description:_

- QA testing recently discovered that when a sit-cfg override file of config.xml doesn't include the domain-name, then some of the overrides within may be silently ignored.   This now confirmed by two different testers using two different tests @davisn9 and Huiling Zhao.
- The problem occurred in (1) a custom override file that had become moderately complex (it introduced debug overrides in addition to some other overrides), and (2) in a custom override file that only has a single override.
- Note that pretty much all internal testing of the sit-cfg feature prior to the operator project happened to specify the domain-name. 

_This update includes:_

- Add a domain-name stanza to the internal sit-cfg the introspector generates for overriding listen-addresses and log directories.
- Add support for an `${env:DOMAIN_NAME}` macro in custom override templates.
- Update public custom override doc to state the need for a domain name stanza and to mention the DOMAIN_NAME macro support.

_Test Status:_

- This change passes override unit/mock testing, Jenkins, Wercker, and 'p4' QA testing. 

_Side Notes:_

- This change also removes the 'expiration' field from the introspector generated sit-cfg (since we know it isn't needed).    Note that removing this can cause sit-cfg errors for customers that forget to use a supported image (e.g. they need to use a patched image or a version newer than 12.2.1.3, as is already documented).

- This problem is internally tracked via OWLS-71350.

### DIFF
--- a/site/config-overrides.md
+++ b/site/config-overrides.md
@@ -203,7 +203,7 @@ The following `config.xml` override file demonstrates setting the `max-message-s
 </domain>
 ```
 
-**IMPORTANT: To ensure all situational config takes effect, remember to reference the name of each bean in the hierarchy that is overridden, including the domain name as in the sample above.**
+**IMPORTANT: To ensure all situational configuration takes effect, remember to reference the name of each bean in the hierarchy that is overridden, including the domain name as in the sample above.**
 
 ### Overriding a data source module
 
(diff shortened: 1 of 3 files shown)


site/config-overrides.md
```diff
@@ -202,6 +203,8 @@ The following `config.xml` override file demonstrates setting the `max-message-s
 </domain>
 ```
 
+**IMPORTANT: To ensure all situational config takes effect, remember to reference the name of each bean in the hierarchy that is overridden, including the domain name as in the sample above.**
```
rosemarymarano: situational config -> situational configuration

site/config-overrides.md
```diff
@@ -267,6 +270,8 @@ The following `jdbc-testDS.xml` override template demonstrates setting the URL,
 
 **IMPORTANT: Incorrectly formatted override files are 'somewhat' silently ignored. WebLogic Servers log errors or warnings when they detect an incorrectly formatted configuration override template file, but will still boot, and will skip overriding. So it is important to make sure that the template files are correct in a QA environment by checking your WebLogic pod logs for situational configuration errors and warnings, before attempting to use them in production.**
 
+**IMPORTANT: To ensure all custom overrides takes effect, remember to reference the name of each bean in the hierarchy that is overridden, including the domain name when overriding config.xml fields. Also, remember to use 'replace' verb to modify existing fields in the domain home configuration, and the 'add' verb to add a field that doesn't yet exist in the domain home configuration.  Otherwise, situational config may be partially ignored without generating any errors or warnings. See the sample in [Override template samples](#override-template-samples).**
```
rosemarymarano: Also, remember to use 'replace' verb -> Also, remember to use the 'replace' verb
situational config -> the situational configuration

site/config-overrides.md
```diff
@@ -332,6 +337,8 @@ spec:
...
```
rosemarymarano: Also, remember to use 'replace' verb -> Also, remember to use the 'replace' verb
situational config -> the situational configuration

PR #802, opened 2019-01-16
move to okio 2.2.0

russgold:
We are seeing I/O errors in the okio library, but we are using a very old version, inherited from the k8s client. The simplest possible fix is to upgrade that library. If that doesn't solve it, we have at least ruled out one possible source of error.


rjeberhard:
I've messed this up... Cherry picking your commit to a new branch and creating new PR.

PR #804, opened 2019-01-16
move to okio 2.2.0

russgold:
Try this as a way to solve the I/O exceptions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -265,6 +265,11 @@
         <version>3.0.0</version>
       </dependency>
       <dependency>
+        <groupId>com.squareup.okio</groupId>
+        <artifactId>okio</artifactId>
+        <version>2.2.0</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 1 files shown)


markxnelson:
testing this change

markxnelson:
for comparison, here is a thread dump from an operator on the old version of okio, which had been running for several hours and stopped responding to cr changes
[thread-dump.txt](https://github.com/oracle/weblogic-kubernetes-operator/files/2771067/thread-dump.txt)

markxnelson:
i built a new operator with this change, just deployed it and now waiting for a few hours, then will get a thread dump to see if it exhibits the same behavior

markxnelson:
Here is a thread dump from an operator built with this change - does not appear to fix the issue. 
[new-thread-dump.txt](https://github.com/oracle/weblogic-kubernetes-operator/files/2773243/new-thread-dump.txt)

PR #805, opened 2019-01-16
READY TO MERGE: Change java tests to use 12.2.1.3 image 

vanajamukkara:
Jenkins and standalone(local) runs are modified to use 12.2.1.3 image from docker hub. This required  changes like using "kubectl exec" instead of "kubectl cp" and comment out set(Id) in all create domain scripts.

Java tests on Wercker will continue to use 19.1 image. Its difficult to move to 19.1 on wercker without interrupting the current runs, if needed I can work on this as a separate JIRA and PR.


integration-tests/src/test/resources/setupenv.sh
```diff
@@ -66,6 +66,9 @@ function pull_tag_images {
 
   docker pull wlsldi-v2.docker.oraclecorp.com/store-weblogic-12.2.1.3:latest
   docker tag wlsldi-v2.docker.oraclecorp.com/store-weblogic-12.2.1.3:latest store/oracle/weblogic:12.2.1.3
+  
+  docker pull wlsldi-v2.docker.oraclecorp.com/weblogic-for-rc2:12.2.1.3.0
```
rjeberhard: Let's update this to use the recently pushed store/oracle/weblogic:12.2.1.3, which is already patched.  Also, please don't use the "rc2" element in the name.

integration-tests/src/test/resources/setupenv.sh
```diff
@@ -169,7 +172,8 @@ elif [ "$JENKINS" = "true" ]; then
   export docker_pass=${docker_pass:?}
   export M2_HOME=${M2_HOME:?}
   export K8S_VERSION=${K8S_VERSION}
-
+  export IMAGE_NAME_WEBLOGIC="store/oracle/weblogic-for-rc2"
```
doxiao: Ryan's comment above applies here too.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -289,6 +335,7 @@ public void testWLDFScaling(Operator operator, Domain domain) throws Exception {
     domain.deployWebAppViaWLST(
         "opensessionapp",
         getProjectRoot() + "/src/integration-tests/apps/opensessionapp.war",
+        "/u01/oracle/apps/",
```
doxiao: shall this be appLocationInPod?

doxiao:
LGTM

integration-tests/src/test/resources/setupenv.sh
vanajamukkara: I tried running the tests against the latest  store/oracle/weblogic:12.2.1.3, which gets pulled using 
docker pull wlsldi-v2.docker.oraclecorp.com/store-weblogic-12.2.1.3:latest
  docker tag wlsldi-v2.docker.oraclecorp.com/store-weblogic-12.2.1.3:latest store/oracle/weblogic:12.2.1.3

Managed server logs contains these errors, the image doesn't seem to be a patched one.
<Jan 18, 2019 8:54:16,214 PM GMT> <Error> <Kernel> <BEA-000802> <ExecuteRequest failed
 java.lang.AssertionError: LocalGroup should atleast have the local server!.
java.lang.AssertionError: LocalGroup should atleast have the local server!
	at weblogic.cluster.messaging.internal.GroupImpl.send(GroupImpl.java:176)
	at weblogic.cluster.messaging.internal.server.UnicastFragmentSocket.send(UnicastFragmentSocket.java:97)
	at weblogic.cluster.FragmentSocketWrapper.send(FragmentSocketWrapper.java:84)
	at weblogic.cluster.UnicastSender.send(UnicastSender.java:53)
	at weblogic.cluster.UnicastSender.send(UnicastSender.java:21)
	Truncated. see log file for complete stacktrace

vanajamukkara:
@rjeberhard export IMAGE_NAME_WEBLOGIC and IMAGE_TAG_WEBLOGIC to use 12.2.1.3 for jenkins in  integration-tests/src/test/resources/setupenv.sh is not working as System.getenv("IMAGE_NAME_WEBLOGIC") is returning null even when they are set in setupenv.sh. I would have to export these variables in the job script at http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/configure, let me know if that's ok.

integration-tests/src/test/resources/setupenv.sh
doxiao: I don't think wlsldi-v2.docker.oraclecorp.com/store-weblogic-12.2.1.3:latest has been updated with the patched image. 
Can you pull store/oracel/weblogic:12.2.1.3 directly from the docker store, instead of pulling from the internal registry? Alternatively, you need to ask Brian to update the image in the internal registry.

vanajamukkara:
@rjeberhard using rc2 image from the local repository based on the response from Mark Nelson on slack https://proddev-paas-fmw.slack.com/archives/C998UPSCT/p1548259286656100 . As mentioned domain home in image tests are run only on Jenkins.

Jenkins job (still running) - http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/914/console

fyi @anpanigr

rjeberhard:
@vanajamukkara @doxiao @tbarnes-us This is a big change and an important set of tests.  Is this ready to go?  I'll admit I'm having a bit of trouble putting my head around what's being tested and not tested.

vanajamukkara:
> 
> 
> @vanajamukkara @doxiao @tbarnes-us This is a big change and an important set of tests. Is this ready to go? I'll admit I'm having a bit of trouble putting my head around what's being tested and not tested.

I am blocked with the issue mentioned on slack re the load balancing of the http requests after scaleup, other than that everything else is tested. - https://proddev-paas-fmw.slack.com/archives/C998UPSCT/p1549391017569900.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -168,12 +168,53 @@ public void testAdminT3Channel(Domain domain) throws Exception {
...
+
+          result =
+              ExecCommand.exec(
+                  "kubectl logs "
+                      + domain.getDomainUid()
+                      + ("-")
+                      + domainMap.get("managedServerNameBase")
+                      + "1 -n "
+                      + domainMap.get("namespace")
+                      + " | grep BEA-000802");
```
tbarnes-us: What is the contents of BEA-000802?  I'd like to understand how this ensures teh patch is applied...

vanajamukkara: When domain home on image test was using a non-patched 12.2.1.3 image, seeing unicast errors in managed server logs which makes them run as standalone servers, not part of the cluster. The servers would still be running, so had to check for the error message. To get the full error message, I need to look into the logs by running the test on a non-patched 12.2.1.3 image.

tbarnes-us: OK - that helps.  Since this is pretty obscure, but helpful to understand, I think it'd be nice to include your comment as a code comment.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -237,6 +283,11 @@ public void testClusterScaling(Operator operator, Domain domain) throws Exceptio
         "Scale domain " + domain.getDomainUid() + " Up to " + replicas + " managed servers");
     operator.scale(domainUid, domainMap.get("clusterName").toString(), replicas);
 
+    if (domainUid.equalsIgnoreCase("domain1")) {
+      logger.info("Adding sleep ");
```
tbarnes-us: Please add a code comment explaining why the sleep is needed when it wasn't before.  If you don't have an explaination, please have teh comment reference a Bug that tracks the issue...

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Sure. here is the error 
<Jan 18, 2019 8:54:16,214 PM GMT> <Error> <Kernel> <BEA-000802> <ExecuteRequest failed
 java.lang.AssertionError: LocalGroup should atleast have the local server!.
java.lang.AssertionError: LocalGroup should atleast have the local server!
	at weblogic.cluster.messaging.internal.GroupImpl.send(GroupImpl.java:176)
	at weblogic.cluster.messaging.internal.server.UnicastFragmentSocket.send(UnicastFragmentSocket.java:97)
	at weblogic.cluster.FragmentSocketWrapper.send(FragmentSocketWrapper.java:84)
	at weblogic.cluster.UnicastSender.send(UnicastSender.java:53)
	at weblogic.cluster.UnicastSender.send(UnicastSender.java:21)
	Truncated. see log file for complete stacktrace

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: this is the open issue we are still trying to track down. I don't think the sleep is needed when the actual issue is tracked.

tbarnes-us: OK - can you add a comment that discusses the problem, and references a tracking bug for the issue?

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
tbarnes-us: It'd be helpful to cut-and paste the log message and the stack trace into the code as a comment...

vanajamukkara: Ok

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Ok

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
tbarnes-us: Looks good! Thanks.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -48,26 +48,158 @@ def getEnvVar(var):
...
 set('Name', admin_server_name)
 
 create('T3Channel', 'NetworkAccessPoint')
 cd('/Servers/%s/NetworkAccessPoints/T3Channel' % admin_server_name)
 set('PublicPort', t3_channel_port)
-set('PublicAddress', 'junkvalue')
-# Dont set listenaddress, introspector overrides automatically with sit-config
-#set('ListenAddress', '%s-%s' % (domain_uid, admin_server_name_svc))
+set('PublicAddress', t3_public_address)
+set('ListenAddress', '%s-%s' % (domain_uid, admin_server_name_svc))
```
tbarnes-us: We should not be setting listen-addresses to their exact expected value, as automatic situational config is always generated and always expected to override the value, and we need to verify that the automatic situational config is working.   A public-address, on the other hand, requires 'manually' setting to the correct value if it's different than the listen-address - this can be done in the original configuration or via 'custom overrides'.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -48,26 +48,158 @@ def getEnvVar(var):
...
+    cd('/')
+
+    msIndex = index+1
+    name = '%s%s' % (managed_server_name_base, msIndex)
+    name_svc = '%s%s' % (managed_server_name_base_svc, msIndex)
+
+    create(name, 'Server')
+    cd('/Servers/%s/' % name )
+    print('managed server name is %s' % name);
+    set('ListenAddress', '%s-%s' % (domain_uid, name_svc))
```
tbarnes-us: We should not be setting listen-addresses to their exact expected value, as automatic situational config is always generated and always expected to override the value, and we need to verify that the automatic situational config is working. A public-address, on the other hand, requires 'manually' setting to the correct value if it's different than the listen-address - this can be done in the original configuration or via 'custom overrides'.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -48,26 +48,158 @@ def getEnvVar(var):
...
+else:
+  print('Configuring Dynamic Cluster %s' % cluster_name)
+
+  templateName = cluster_name + "-template"
+  print('Creating Server Template: %s' % templateName)
+  st1=create(templateName, 'ServerTemplate')
+  print('Done creating Server Template: %s' % templateName)
+  cd('/ServerTemplates/%s' % templateName)
+  cmo.setListenPort(server_port)
+  cmo.setListenAddress('%s-%s${id}' % (domain_uid, managed_server_name_base_svc))
```
tbarnes-us: We should not be setting listen-addresses to their exact expected value, as automatic situational config is always generated and always expected to override the value, and we need to verify that the automatic situational config is working. A public-address, on the other hand, requires 'manually' setting to the correct value if it's different than the listen-address - this can be done in the original configuration or via 'custom overrides'.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
tbarnes-us: The same comment applies all listen-addresses in this change.

integration-tests/README.md
```diff
@@ -194,6 +194,12 @@ Secret - to create secret
 
 Command to run the tests:
 ```
+export DOCKER_USERNAME=<docker_username>
+export DOCKER_PASSWORD=<docker_password>
+export DOCKER_EMAIL=<docker_email>
+or
+make sure the weblogic image i.e. store/oracle/weblogic:12.2.1.3 already exists locally
+
```
tbarnes-us: NODEPORT_HOST below should be K8S_NODEPORT_HOST

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
vanajamukkara: Update create-domain.py with the one from samples which doesn't set listen-address.

integration-tests/README.md
vanajamukkara: Yes, will fix it.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -293,27 +325,37 @@ public void deployWebAppViaREST(
...
+            + "/integration-tests/src/test/resources/deploywebapp.py");
+    TestUtils.kubectlexec(
+        adminPod,
+        domainNS,
+        " -- bash -c 'cat > "
+            + appLocationInPod
+            + "/callpyscript.sh' < "
+            + projectRoot
+            + "/integration-tests/src/test/resources/callpyscript.sh");
+
```
tbarnes-us: The same 'exec cat' pattern is used a few times in this change to cp files -- it'd be helpful to abstract this out via TestUtils.copyFileViaCat() or some-such function, just like was already done for kubectl cp via TestUtils.kubectlcp()...

tbarnes-us: In theory, you should be able to simply replace the implementation of kubectlcp()....

vanajamukkara: Ok. Will add a new method instead of replacing the implementation of kubectlcp() which could be confusing.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -233,11 +233,15 @@ public static void testWlsLivenessProbe(String domainUid, String serverName, Str
     new File(filePath).setExecutable(true, false);
 
     // copy file to pod
-    kubectlcp(filePath, "/shared/killserver.sh", podName, namespace);
+    kubectlexec(podName, namespace, " -- bash -c 'cat > /shared/killserver.sh' < " + filePath);
```
tbarnes-us: This could use copyFileViaCat

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -710,18 +724,20 @@ public static void createWLDFModule(String adminPodName, String domainNS, int t3
...
-    TestUtils.kubectlcp(
-        BaseTest.getProjectRoot() + "/integration-tests/src/test/resources/wldf/wldf.py",
-        "/shared/wldf.py",
+    TestUtils.kubectlexec(
         adminPodName,
-        domainNS);
+        domainNS,
+        " -- bash -c 'cat > /shared/wldf.py' < "
+            + BaseTest.getProjectRoot()
+            + "/integration-tests/src/test/resources/wldf/wldf.py");
```
tbarnes-us: Can use copyFileViaCat

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -710,18 +724,20 @@ public static void createWLDFModule(String adminPodName, String domainNS, int t3
...
-    TestUtils.kubectlcp(
-        BaseTest.getProjectRoot() + "/integration-tests/src/test/resources/callpyscript.sh",
-        "/shared/callpyscript.sh",
+    TestUtils.kubectlexec(
         adminPodName,
-        domainNS);
+        domainNS,
+        " -- bash -c 'cat > /shared/callpyscript.sh' < "
+            + BaseTest.getProjectRoot()
+            + "/integration-tests/src/test/resources/callpyscript.sh");
```
tbarnes-us: Can use copyFileViaCat

PR #819, opened 2019-01-22
Add pooled connection lifetime 

rjeberhard:


rjeberhard:
Let's punt this for now since we made progress by just clearing the networkInterceptors() causing the memory leak.

PR #824, opened 2019-01-23
OWLS-71147 Enhance markdown generation

russgold:
Adds to generation of the domain markdown:

1. Added a column for field types
2. The type for an object is now a link to the corresponding type
3. Generated markdown for the referenced Kubernetes artifacts
4. Added descriptions for model classes to the generated schema and markdown

To see this, go to https://github.com/oracle/weblogic-kubernetes-operator/blob/enhance_markdown_generation/docs/domains/Domain.md


rjeberhard:
Can we link this from the doc set, specifically from site/domain-resource.md?

russgold:
I think that does it. Try https://github.com/oracle/weblogic-kubernetes-operator/blob/enhance_markdown_generation/site/domain-resource.md

PR #825, opened 2019-01-23
Sit cfg sometimes needs domain name fix

tbarnes-us:
Fix for pulls #797 and #820:   the calculated DOMAIN_NAME value in introspectDomain.py should be discovered via wlst instead of from a domain env var.   @alai8 discovered that the domain env var isn't necessarily in sync with the 'actual' value.

### DIFF
--- a/site/weblogic-docker-images.md
+++ b/site/weblogic-docker-images.md
@@ -9,8 +9,8 @@ There are two main options available:
   and the domain directory.
 
 If you want to use the first option, you will need to obtain the standard
-WebLogic Server image from the Docker Store [see here](#obtaining-standard-images-from-the-docker-store)
-and then create a new image with the mandatory patches applied as described in [this section](#creating-a-custom-images-with-patches-applied).
+WebLogic Server image from the Docker Store, [see here](#obtaining-standard-images-from-the-docker-store),
+this image already contains the mandatory patches applied as described in [this section](#creating-a-custom-image-with-patches-applied).
 If you want to use additional patches, you can customize that process to include additional patches.
 
 If you want to use the second option, which includes the domain directory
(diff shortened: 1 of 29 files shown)


site/weblogic-docker-images.md
```diff
@@ -9,8 +9,8 @@ There are two main options available:
   and the domain directory.
 
 If you want to use the first option, you will need to obtain the standard
-WebLogic Server image from the Docker Store [see here](#obtaining-standard-images-from-the-docker-store)
-and then create a new image with the mandatory patches applied as described in [this section](#creating-a-custom-images-with-patches-applied).
+WebLogic Server image from the Docker Store, [see here](#obtaining-standard-images-from-the-docker-store),
+this image already contains the mandatory patches applied as described in [this section](#creating-a-custom-image-with-patches-applied).
```
tbarnes-us: Not sure about this change - the merge of the develop produced this conflict somehow...   I suspects 'master' might be more up to date, and the above matches master?

site/weblogic-docker-images.md
```diff
@@ -64,12 +68,16 @@ Docker Store.
 
 ## Creating a custom image with patches applied
 
+<<<<<<< HEAD
```
alai8: git merge conflict? There is another one above.

tbarnes-us: I just pushed a merge.  Not quite sure if I did it right.  Not sure if 'master' or 'develop' is supposed to be the lead for the change.   Checking.

tbarnes-us: I just pushed the version that was on develop.

tbarnes-us: So they should match now.

PR #831, opened 2019-01-24
OWLS-71300 sample updates and cleanup

doxiao:
Changes in this PR:
1) Allow an absolute directory for domainHomeImageBuildPath
2) Make store/oracle/weblogic:12.2.1.3 the default value for the image property in the domain-home-on-pv sample, and the domainHomeImageBase property in the domain-home-in-image sample.
3) Use the domainHomeImageBase's tag as the generated image's tag when the "image" property is not set.
4) Avoid generating an extra image tag by using the new TAG_NAME property in domain-home-in image Docker samples (WLST and WDT) to customize the image name.
5) When the -k option is specified, still clone the Docker image project if the domainHomeImageBuildPath is empty (which means there is no previously cloned project available)
6) Clean up generated temporary files after they are used.
7) Update README.md

The changes passed java integration test on Jenkins: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/920.


kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/README.md
```diff
@@ -102,13 +102,13 @@ The following parameters can be provided in the inputs file.
 | `adminServerName` | Name of the Administration Server. | `admin-server` |
 | `clusterName` | Name of the WebLogic cluster instance to generate for the domain. | `cluster-1` |
 | `configuredManagedServerCount` | Number of Managed Server instances to generate for the domain. | `5` |
-| `domainHomeImageBase` | Base WebLogic binary image used to build the WebLogic domain image. The operator requires WebLogic Server 12.2.1.3.0 with patch 29135930 applied. Refer to [WebLogic Docker images](../../../../../site/weblogic-docker-images.md) for details on how to create one. | |
+| `domainHomeImageBase` | Base WebLogic binary image used to build the WebLogic domain image. The operator requires WebLogic Server 12.2.1.3.0 with patch 29135930 applied. The existing WebLogic Docker image, `store/oracle/weblogic:12.2.1.3`, was updated on January 17, 2019,and has all the necessary patches applied; a `docker pull` is required if you already have this image. Refer to [WebLogic Docker images](../../../../../site/weblogic-docker-images.md) for details on how to obtain or create the image. | store/oracle/weblogic:12.2.1.3 |
```
rosemarymarano: store/oracle/weblogic:12.2.1.3 -> `store/oracle/weblogic:12.2.1.3`

kubernetes/samples/scripts/common/utility.sh
```diff
@@ -293,6 +303,7 @@ function createFiles {
     sed -i -e "s:%JAVA_OPTIONS%:${javaOptions}:g" ${domainPropertiesOutput}
     sed -i -e "s:%T3_CHANNEL_PORT%:${t3ChannelPort}:g" ${domainPropertiesOutput}
     sed -i -e "s:%T3_PUBLIC_ADDRESS%:${t3PublicAddress}:g" ${domainPropertiesOutput}
+    sed -i -e "s|%IMAGE_NAME%|${defaultImageName}|g" ${domainPropertiesOutput}
```
simon-meng-cn: Need check whether the image property is setting, then decide use which name. See the following snippet:
    if [ -z $image ]; then
      sed -i -e "s|%IMAGE_NAME%|${defaultImageName}|g" ${domainPropertiesOutput}
    else
      sed -i -e "s|% IMAGE_NAME%|${image}|g" ${domainPropertiesOutput}
    fi

doxiao: The proposed behaviors are
1) the Docker sample always tag the generated image with the default image name (defaultImageName)
2) Optionally if the "image" property is set, tag the generated image with the value of "image" property as well (see the comment in line 199 in create-domain.sh).

We could change it to tag with the defaultImageName only if the "image" property is set, which is probably cleaner.

simon-meng-cn: The confusing thing is that the image name in domain.properties may be different from the image name in domain.yaml

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/README.md
```diff
@@ -26,6 +26,12 @@ $ create-weblogic-credentials.sh
 
 ## Use the script to create a domain
 
+**Note**: The `create-domain.sh` script generates a new Docker image on each run with a new domain home and a different internal `domain secret` in it.  To prevent having disparate images with different domain secrets in the same domain, it is highly recommended that a new domain uses a `domainUID` that is different from any of the active domains, or that you delete the existing domain resource before you create a domain with the same `domainUID` using the following command (and wait until all the server pods are terminated):
```
rosemarymarano: it is highly recommended -> we strongly recommend

(and wait until all the server pods are terminated): -> (_after_ all the server pods are terminated):

kubernetes/samples/scripts/common/utility.sh
doxiao: I updated the script to tag with the default image name only if the "image" is not specified.

PR #842, opened 2019-01-28
OWLS-69366 Remove log and address setup from domain home on pv sample

simon-meng-cn:
The PR includes the following:
1. The setup logic is already in the Operator runtime,  so we remove the logic from the domain-home-on-pv sample scripts.
2. Remove the unused code.


doxiao:
If I understand the description of the task correctly, we also need to clean up the wlst script and wdt model and get rid of setting log location and server listen addresses.

tbarnes-us:
> If I understand the description of the task correctly, we also need to clean up the wlst script and wdt model and get rid of setting log location and server listen addresses.

+1

simon-meng-cn:
Jenkins report a test failure, so I need revert some code. I will close current PR, and delete branch, then create a new one.

PR #844, opened 2019-01-28
OWLS-69951 Add unit tests for Domain status processing and cleanup

russgold:
The rest of this task involves creating a wiki to recommend more status updates. The only thing that seems obvious, though, is semantic errors when processing the domain resource.


russgold:
I've just found more code that should probably be tested as part of this PR, so let's not check it in yet.

russgold:
In creating the new unit tests, I found a number of cases where the existing behavior appears to be wrong. I have marked them in the unit test with // todo tags

The replica status appears to make no sense. The tests `whenDomainUsesDefaultReplicaCount_statusLacksReplicaCount` and `whenDomainHasDefaultReplicaCount_statusReplicaCountIsSmallestNumberOfPodsInClustersWithoutReplica` describe what is currently happening. Perhaps it should set the replica count only when there is a single cluster? Or report for each cluster in the domain?

The logic to select servers to compute whether all servers are running seems to be backwards, selecting those in admin rather than running state.

When there is a pod failure, the code appears to want to delete any existing "available" conditions, but actually creates one if all servers are running.

My plan is to fix these once I have confirmation of the intended functionality and then to refactor the code in a subsequent commit, as cleaning it up would make it much easier to understand and modify.

russgold:
Went ahead and did the cleanup, deciding to implement DomainStatus.replicas as undefined if there is not exactly one cluster in the domain, and otherwise the size of that cluster.

docs/domains/Domain.json
```diff
@@ -297,6 +302,9 @@
           "type": "number",
           "minimum": 0
         },
+        "modified": {
```
rjeberhard: We shouldn't add this field.  The "modified" field in the processing logic was to just to track when the operator has a pending change to the status.

russgold: Now ignoring volatile fields in schema generation

PR #845, opened 2019-01-29
Owls 71453: Operator logs NPE after (correctly) logging introspector validation errors.

lennyphan:
If the introspector determines the domain is invalid  (due to validation errors) then log the errors and terminate the fiber by returning next action with and 'null' step.   Correctly terminating the fiber resolves the NPE.


tbarnes-us:
LGTM

PR #854, opened 2019-02-05
Test Network Access Point on WL Clusters

jgrundback:
This change is adding testing for cluster channels by adding a custom WLS channel to the cluster server template in the create domain WLST. This custom channel is picked up as part of the larger lifecycle tests as the domain is restarted. After the domain is restarted we check that the channel is present as a service in k8s.

WIP:  This change is gated on getting tests to pass on Wercker, which is in turn gated on internal JIRA owls-72310.   The suspected cause is that Wercker is using an old WebLogic image.   Have asked @rjeberhard to give it a new one.

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -320,22 +320,22 @@ public class BaseTest {
 
     // make sure cluster service endpoint is updated with the scaled up servers before verifying
     // load balancing
-    int i = 0;
-    while (i < BaseTest.getMaxIterationsPod()) {
-      int numberOfServersInEndpoint =
-          domain.getNumberOfServersInClusterServiceEndpoint((String) domainMap.get("clusterName"));
-      if (replicaCnt != numberOfServersInEndpoint) {
-        // check for last iteration
-        if (i == (BaseTest.getMaxIterationsPod() - 1)) {
-          throw new RuntimeException("FAILURE: Cluster Serice Endpoint is not updated");
-        }
-        Thread.sleep(BaseTest.getWaitTimePod() * 1000);
-        i++;
-      } else {
-        logger.info("Number of servers addresses in endpoint is same as replica count ");
-        break;
-      }
-    }
+//    int i = 0;
+//    while (i < BaseTest.getMaxIterationsPod()) {
+//      int numberOfServersInEndpoint =
+//          domain.getNumberOfServersInClusterServiceEndpoint((String) domainMap.get("clusterName"));
+//      if (replicaCnt != numberOfServersInEndpoint) {
+//        // check for last iteration
+//        if (i == (BaseTest.getMaxIterationsPod() - 1)) {
+//          throw new RuntimeException("FAILURE: Cluster Serice Endpoint is not updated");
+//        }
+//        Thread.sleep(BaseTest.getWaitTimePod() * 1000);
+//        i++;
+//      } else {
+//        logger.info("Number of servers addresses in endpoint is same as replica count ");
+//        break;
+//      }
+//    }
     // commenting the load balance check, bug 29325139
     // domain.verifyWebAppLoadBalancing(TESTWEBAPP);
 
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -320,22 +320,23 @@ public class BaseTest {
 
     // make sure cluster service endpoint is updated with the scaled up servers before verifying
     // load balancing
-//    int i = 0;
-//    while (i < BaseTest.getMaxIterationsPod()) {
-//      int numberOfServersInEndpoint =
-//          domain.getNumberOfServersInClusterServiceEndpoint((String) domainMap.get("clusterName"));
-//      if (replicaCnt != numberOfServersInEndpoint) {
-//        // check for last iteration
-//        if (i == (BaseTest.getMaxIterationsPod() - 1)) {
-//          throw new RuntimeException("FAILURE: Cluster Serice Endpoint is not updated");
-//        }
-//        Thread.sleep(BaseTest.getWaitTimePod() * 1000);
-//        i++;
-//      } else {
-//        logger.info("Number of servers addresses in endpoint is same as replica count ");
-//        break;
-//      }
-//    }
+    //    int i = 0;
+    //    while (i < BaseTest.getMaxIterationsPod()) {
+    //      int numberOfServersInEndpoint =
+    //          domain.getNumberOfServersInClusterServiceEndpoint((String)
+    // domainMap.get("clusterName"));
+    //      if (replicaCnt != numberOfServersInEndpoint) {
+    //        // check for last iteration
+    //        if (i == (BaseTest.getMaxIterationsPod() - 1)) {
+    //          throw new RuntimeException("FAILURE: Cluster Serice Endpoint is not updated");
+    //        }
+    //        Thread.sleep(BaseTest.getWaitTimePod() * 1000);
+    //        i++;
+    //      } else {
+    //        logger.info("Number of servers addresses in endpoint is same as replica count ");
+    //        break;
+    //      }
+    //    }
     // commenting the load balance check, bug 29325139
     // domain.verifyWebAppLoadBalancing(TESTWEBAPP);
 
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -254,6 +254,7 @@ public class BaseTest {
     c.close();
     logger.info("Done - testAdminT3ChannelWithJMS");
   }
+
   /**
    * Restarting the domain should not have any impact on Operator managing the domain, web app load
    * balancing and node port service
@@ -499,6 +500,13 @@ public class BaseTest {
     return branchName;
   }
 
+  /**
+   * Use this method to override the create_domain.py script with a custom one. Remember to call
+   * unoverrideCreateDomainScript below after you are done.
+   *
+   * @param scriptPath
+   * @throws IOException
+   */
   public void overrideCreateDomainScript(String scriptPath) throws IOException {
 
     if (scriptPath == null) {
@@ -521,6 +529,12 @@ public class BaseTest {
         StandardCopyOption.REPLACE_EXISTING);
   }
 
+  /**
+   * Reverts the custom create_domain.py script with the default one. Should be called after you are
+   * done with overrideCreateDomainScript above.
+   *
+   * @throws IOException
+   */
   public void unoverrideCreateDomainScript() throws IOException {
 
     String createDomainScriptDir =
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -4,11 +4,8 @@
 
 package oracle.kubernetes.operator;
 
-import java.io.File;
-import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.util.Map;
 import java.util.Properties;
 import java.util.logging.FileHandler;
@@ -500,52 +497,6 @@ public class BaseTest {
 
   public static String getBranchName() {
     return branchName;
-  }
-
-  /**
-   * Use this method to override the create_domain.py script with a custom one. Remember to call
-   * unoverrideCreateDomainScript below after you are done.
-   *
-   * @param scriptPath
-   * @throws IOException
-   */
-  public void overrideCreateDomainScript(String scriptPath) throws IOException {
-
-    if (scriptPath == null) {
-      return;
-    }
-
-    logger.info("Overriding create domain script with this: " + scriptPath);
-
-    String createDomainScriptDir =
-        BaseTest.getProjectRoot() + "/integration-tests/src/test/resources/domain-home-on-pv";
-
-    // cp py
-    Files.copy(
-        new File(createDomainScriptDir + "/create-domain.py").toPath(),
-        new File(createDomainScriptDir + "/create-domain.py.bak").toPath(),
-        StandardCopyOption.REPLACE_EXISTING);
-    Files.copy(
-        new File(BaseTest.getProjectRoot() + scriptPath).toPath(),
-        new File(createDomainScriptDir + "/create-domain.py").toPath(),
-        StandardCopyOption.REPLACE_EXISTING);
-  }
-
-  /**
-   * Reverts the custom create_domain.py script with the default one. Should be called after you are
-   * done with overrideCreateDomainScript above.
-   *
-   * @throws IOException
-   */
-  public void unoverrideCreateDomainScript() throws IOException {
-
-    String createDomainScriptDir =
-        BaseTest.getProjectRoot() + "/integration-tests/src/test/resources/domain-home-on-pv";
-
-    Files.copy(
-        new File(createDomainScriptDir + "/create-domain.py.bak").toPath(),
-        new File(createDomainScriptDir + "/create-domain.py").toPath(),
-        StandardCopyOption.REPLACE_EXISTING);
   }
 
   private void copyScalingScriptToPod(
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -4,8 +4,11 @@
 
 package oracle.kubernetes.operator;
 
+import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.Map;
 import java.util.Properties;
 import java.util.logging.FileHandler;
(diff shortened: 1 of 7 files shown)


vanajamukkara:
Describe the use cases this PR is testing.

vanajamukkara:
Can you explain how you are verifying all the custom overrides you have in the custom config.xml?

jgrundback:
@vanajamukkara I am not verifying that yet, at this point I have the test running, but I am having trouble checking the overrides and open ports

integration-tests/src/test/resources/domain13/custom/config.xml
```diff
@@ -0,0 +1,59 @@
...
+<domain xmlns="http://xmlns.oracle.com/weblogic/domain" xmlns:f="http://xmlns.oracle.com/weblogic/domain-fragment" xmlns:s="http://xmlns.oracle.com/weblogic/situational-config">
+  <log>
+    <file-name f:combine-mode="add">/shared/logs/base_domain.log</file-name>
+  </log>
+  <server>
+    <name>admin-server</name>
+    <log f:combine-mode="add">
+      <file-name f:combine-mode="add">/shared/logs/admin-server.log</file-name>
+    </log>
+    <listen-address f:combine-mode="replace">domain13-admin-server</listen-address>
```
tbarnes-us: It isn't acceptable for custom sit-cfg to override listen address or log files.  

See the custom sit config writeup in the site documentation.

tbarnes-us:
The scope of this test should only be about making sure that network-access-points are exposed as services on configured _and_ dynamic clusters.  

There's no need to create custom sit-cfg for this purpose.    Pretty much all custom sit-cfg is already tested, and teh current custom sit-cfg overrides in this test look suspect (for example, they override stuff that is not supported for overrides).    

It should be possible to find two existing test domains and re-use them (one configured and one dynamic) and then (A) add network-access-point(s) to its cluster's config, (B) add a test that confirms there's a 'cluster' service that has the expected port from the NAP, and (C) exercises this port in some fashion (call curl on the test web app on the cluster directly from the admin server pod to the cluster pod for example -- or set up a load balancer).   There are already tests that perform B & C, so you should be able re-use code from these tests - perhaps just by directly calling their helper functions.   Also, you can 'piggyback' the new checks on an existing test for the domain that assumes the domain is already up and running.

integration-tests/src/test/resources/domain13/custom/config.xml
jgrundback: @tbarnes-us This config comes from the output of the introspector, I pretty much took the output in the job files folder.

Sit-Cfg-CFG--introspector-situational-config.xml
  <d:server>
    <d:name>admin-server</d:name>
    <d:log f:combine-mode="add">
      <d:file-name f:combine-mode="add">/shared/logs/admin-server.log</d:file-name>
    </d:log>
    <d:listen-address f:combine-mode="replace">domain1-admin-server</d:listen-address>
    <d:network-access-point>
      <d:name>T3Channel1</d:name>
      <d:listen-address f:combine-mode="replace">domain1-admin-server</d:listen-address>
    </d:network-access-point>
    <d:network-access-point>
      <d:name>T3Channel2</d:name>
      <d:listen-address f:combine-mode="replace">domain1-admin-server</d:listen-address>
    </d:network-access-point>
  </d:server>

jgrundback:
@tbarnes-us Sure, I can take out this test and try to fold it into one of the existing tests. I have to go back to the test suite then and try to figure out how to re-approach this.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -414,6 +417,39 @@ public static String getBranchName() {
     return branchName;
   }
 
+  public void overrideCreateDomainScript(String scriptPath) throws IOException {
```
jchermos: Add javadoc to public method.

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
```diff
@@ -622,39 +616,55 @@ public void testCustomSitConfigOverrides() throws Exception {
...
+      this.unoverrideCreateDomainScript();
       if (domain12 != null && (JENKINS || testCompletedSuccessfully)) {
         domain12.destroy();
       }
     }
     logger.info("SUCCESS - testCustomSitConfigOverrides");
   }
 
+  @Test
+  public void testIntrospectorSitConfigOverrides() throws Exception {
```
jchermos: Add javadoc with description of the test.

integration-tests/src/test/resources/domain13/custom/config.xml
tbarnes-us: There are two kinds of sit-cfg overrides:
1) custom sit-cfg 
2) introspector generated
The former isn't needed in your testing.
The latter isn't needed, as the introspector is generating this automatically.

tbarnes-us: The latter needs to be _properly_ generated, and subsequently consumed by the operator, and finally the operator should expose the port(s) via services.   It's the last that needs to be tested.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
jgrundback: @jchermos Just did, thanks!

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
jgrundback: Test is now removed

integration-tests/src/test/resources/domain13/custom/config.xml
jgrundback: I removed all of this

jgrundback:
@tbarnes-us @vanajamukkara I completely rewrote this to integrate the channels test with the existing tests. I added a NAP port to the cluster server template. Then I check for this channel in the testDomainLifecyle test, which relies on a domain restart, at which point the introspector picks up the channel and creates it.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -243,6 +244,42 @@ public void verifyAdminServerExternalService(String username, String password) t
...
+  }
+
+  /**
+   * Verify that we have do not have the channel set for the cluster
+   *
+   * @param protocol
+   * @param port
+   * @throws Exception
+   */
+  public void verifyNotClusterServiceChannelPort(String protocol, int port) throws Exception {
```
tbarnes-us: This method doesn't look like it's getting called.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -106,6 +106,13 @@ def getEnvVar(var):
   cmo.setCluster(cl)
   print('Done setting attributes for Server Template: %s' % templateName);
 
+  templateChannelName = cluster_name + "-NAP"
+  create(templateChannelName, 'NetworkAccessPoint')
+  cd('NetworkAccessPoints/%s' % templateChannelName)
+  set('PublicPort', server_port + 10)
+  # set('PublicAddress', 'junkvalue')
+  set('ListenAddress', '%s-%s${id}' % (domain_uid, managed_server_name_base_svc))
```
tbarnes-us: It's not necessary to set the ListenAddress (same holds true for both domain11 and domain12).

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -1018,9 +1055,12 @@ private void initialize(String inputYaml) throws Exception {
     domainMap.values().removeIf(Objects::isNull);
 
     // create config map and secret for custom sit config
-    if (domainMap.get("configOverrides") != null) {
+    if ((domainMap.get("configOverrides") != null)
+        && (domainMap.get("configOverridesFile") != null)) {
       // write hostname in config file for public address
 
+      String configOverridesFile = domainMap.get("configOverridesFile").toString();
+
```
tbarnes-us: Why does this test change need to exercise custom config overrides?
The primary focus is testing whether the NAP has been exposed via a service, and the service port 'works' (actually redirects to the NAP).

jgrundback: @tbarnes-us It doesn't. That is a refactoring I wrote to make it dynamic. I was using a custom file earlier. But that has since been removed. Only the flexibility refactoring remains.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
tbarnes-us: Actually it might be best, for test purposes, to set the ListenAddress to a completely invalid value 'not-a-real-dns-address' or some-such.   This would help verify that the automatic situational config generated by the introspector properly overrides the NAP address.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
jgrundback: @tbarnes-us I thought it was a nice addition, but I still have to merge this with @vanajamukkara 's refactoring changes.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
jgrundback: Yep, I'll remove it.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
jgrundback: I'll fix that. Thanks!

tbarnes-us: Cool.  Once you do this, you can 'ad-hoc' check if the override takes effect at least at the mbean level using wlst.sh per the debugging section of 'config-overrides.md'.   Also, if you add a 'curl' test, as I'm recommending, that also implicitly tests if the override is working - as if it isn't, the curl couldn't succeed...

jgrundback:
@tbarnes-us Thank you for your review. Regarding the curl test of the port, yes I definitely remember us talking about that. Since I have been on JCS primary call all week this week, I haven't been able to get there yet.

jgrundback:
@vanajamukkara Sure. That is definitely on my list of todos, right now I am trying to debug why the integration tests suite is failing on wercker.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -116,6 +116,12 @@ def getEnvVar(var):
   cmo.setCluster(cl)
   print('Done setting attributes for Server Template: %s' % templateName);
```
tbarnes-us: This print statement belongs after the NAP creation, since NAP is part of the template too.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
```diff
@@ -116,6 +116,12 @@ def getEnvVar(var):
...
 
+  templateChannelName = cluster_name + "-NAP"
+  create(templateChannelName, 'NetworkAccessPoint')
+  cd('NetworkAccessPoints/%s' % templateChannelName)
+  set('PublicPort', server_port + 10)
+  set('ListenAddress', '%s-%s${id}' % (domain_uid, managed_server_name_base_svc))
+  set('ListenPort', server_port + 10)
```
tbarnes-us: The same set of changes also need to be made for configured clusters, otherwise, once configured cluster support is tested again, the test will fail...

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
```diff
@@ -622,39 +616,54 @@ public void testCustomSitConfigOverrides() throws Exception {
...
-          StandardCopyOption.REPLACE_EXISTING);
+      this.unoverrideCreateDomainScript();
       if (domain12 != null && (JENKINS || testCompletedSuccessfully)) {
         domain12.destroy();
       }
     }
     logger.info("SUCCESS - testCustomSitConfigOverrides");
   }
 
+  @Test
```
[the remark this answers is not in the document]
jgrundback: @vanajamukkara I have now a clean merge of develop in this branch, can you have a look again?

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -248,6 +249,57 @@ public void verifyAdminServerExternalService(String username, String password) t
...
+    curlCmdResCode.append(" -- --write-out %{http_code} -o /dev/null");
+
+    logger.info("Curl cmd with response code " + curlCmdResCode);
+    logger.info("Curl cmd " + curlCmd);
+
+    // call webapp iteratively till its deployed/ready
+    callWebAppAndWaitTillReady(curlCmdResCode.toString());
+
+    // execute curl and look for the managed server name in response
+    callWebAppAndCheckForServerNameInResponse(curlCmd.toString(), true);
```
vanajamukkara: do u need to verify load balancing of requests here? if not, you can set false.

integration-tests/src/test/resources/domain-home-on-pv/create-domain-auto-custom-sit-config.py
```diff
@@ -116,6 +116,13 @@ def getEnvVar(var):
   cmo.setCluster(cl)
   print('Done setting attributes for Server Template: %s' % templateName);
 
+  templateChannelName = cluster_name + "-NAP"
```
vanajamukkara: This file is used in testAutoAndCustomSitConfigOverrides, domain is not restarted. are you sure this config is needed here?

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -213,6 +217,7 @@ public void testDomainLifecyle(Operator operator, Domain domain) throws Exceptio
     domain.verifyDomainCreated();
     domain.verifyWebAppLoadBalancing(TESTWEBAPP);
     domain.verifyAdminServerExternalService(getUsername(), getPassword());
+    domain.verifyHasClusterServiceChannelPort("TCP", 8011, TESTWEBAPP + "/");
```
vanajamukkara: channel port config is there right from the beginning, does it really need domain restart to work?

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
jgrundback: Yes. I that case, should I add the NAP to each individual server?

integration-tests/src/test/resources/domain-home-on-pv/create-domain-auto-custom-sit-config.py
jgrundback: I took a look at the code, you are right. I thought we fun the full lifecycle test here too, but we don't. I'll take it out.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
jgrundback: The k8s service for the port/channel is not created until we restart the domain. Even though the port is created in the WLS config from the beginning, the introspector only opens is after the reboot.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
jgrundback: Yep

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
jgrundback: I.e. we pass along the URL to the cluster, which resolves in round robin to each managed server in turn.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain.py
```diff
@@ -102,6 +102,12 @@ def getEnvVar(var):
   cmo.setCluster(cl)
   print('Done setting attributes for Server Template: %s' % templateName);
 
+  templateChannelName = cluster_name + "-NAP"
```
vanajamukkara: You only need to change integration-tests/src/test/resources/domain-home-on-pv/create-domain.py. I think the change to kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain.py is not needed.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -1058,6 +1110,11 @@ private void initialize(Map<String, Object> inputDomainMap) throws Exception {
       domainMap.put("t3PublicAddress", TestUtils.getHostName());
     }
 
+    domainMap.put("domainHome", "/shared/domains/" + domainUid);
```
vanajamukkara: You don't need the next 5 lines. Its done in line 1129 - 1136.

jgrundback: Done

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain.py
jgrundback: @vanajamukkara Thank you! My bad, that was a remnant from when the tests was using the wrong domain creation scripts. I took it out.

integration-tests/src/test/resources/domain-home-on-pv/create-domain.py
jgrundback: @tbarnes-us Is it ok if I file a jira for this? Or should I simply check for the NAP in the lifecycle checks if this is a dynamic cluster?

vanajamukkara:
@jgrundback please run the tests on Jenkins once.

jgrundback:
@vanajamukkara Sure! I ran it on jenkins last week, but I can do so again.

vanajamukkara:
I checked jenkins run, its failing only for the wldf scaling issue for which there is a jira. Otherwise LG.

jgrundback:
@vanajamukkara Thank you!

PR #855, opened 2019-02-05
WIP: Display Jenkins build status

rjeberhard:


README.md
```diff
@@ -1,8 +1,8 @@
 # Oracle WebLogic Server Kubernetes Operator
 
-Built with [Wercker](http://www.wercker.com)
+Built with [Jenkins](http://build.weblogick8s.org:8080/job/WebLogic%20Kubernetes%20Operator%20(any%20branch)/)
 
-[![wercker status](https://app.wercker.com/status/68ce42623fce7fb2e52d304de8ea7530/m/develop "wercker status")](https://app.wercker.com/project/byKey/68ce42623fce7fb2e52d304de8ea7530)
+[![Build Status](http://build.weblogick8s.org:8080/job/WebLogic%20Kubernetes%20Operator%20(any%20branch)/badge/icon)](http://build.weblogick8s.org:8080/job/WebLogic%20Kubernetes%20Operator%20(any%20branch)/)
```
markxnelson: will this work for a user who is not logged in?  we might need to create an anonymous user or something?

rjeberhard: There are different options depending on the perms you expect callers to have.  I was going to update Jenkins to do role based access control and give anonymous just privs to read status.

However, there is still an oddity that since the job is for any branch, it looks confusing to say that the build is passed or failed based on the last run (which is likely for some other branch).

PR #858, opened 2019-02-06
OWLS-71901-OPERATOR FAILS TO BRING UP WEBLOGIC DOMAIN WITH SITCONFIG OVERRIDE CONFIGURATION

alai8:
This is a regression caused by owls-71580 when we start passing en vars to introspector job. The WLS container is preconfigured with USER_MEM_ARGS="-Djava.security.egd=file:/dev/./urandom". When this env var is overridden, the introspector script would take a longer time to run. In Pani's environment, it took longer than the timeout of 60s specified in job.template.spec.activeDeadlineSeconds. This results in the introspector job failing with DeadlineExceeded

Changes made in this PR:

- The operator should always explicitly set USER_MEM_ARGS to `-Djava.security.egd=file:/dev/./urandom` if the domain.yaml does not specify this env var.  (This will override any value that the docker image itself might be setting.)

- If the domain.yaml specifies USER_MEM_ARGS of “” or any value, then honor the custom value instead.

- Use the same approach for both the introspector job and for wl pods- Append USER_MEM_ARGS with "-Djava.security.egd=file:/dev/./urandom" in introspector.sh instead of in operator code so we are not relying on it being set in the weblogic docker image

- there are 2 activeDeadlineSeconds, one in job.spec and another in job.template.spec. The former is for the job and the second is for the pod started by the job. We have been setting the former to 120s, with an increment of 60s in subsequent retries. Now we are also setting the latter with the same values.

- tested by Pani
- passed jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1008/

### DIFF
--- a/site/domains.md
+++ b/site/domains.md
@@ -85,7 +85,7 @@ running in Kubernetes.
 * _Host Path Persistent Volumes:_ If using a `hostPath` persistent volume, then it must be available on all worker nodes in the cluster and have read/write/many permissions for all container/pods in the WebLogic Server deployment.  Be aware
   that many cloud provider's volume providers may not support volumes across availability zones.  You may want to use NFS or a clustered file system to work around this limitation.
 
-* _Security Note:_ The `USER_MEM_ARGS` environment variable defaults to `-Djava.security.egd=file:/dev/./urandom` in all WebLogic server pods and the WebLogic introspection job. It can be explicitly set to another value in your domain resource yaml using the `env` attribute under the `serverPod` configuration.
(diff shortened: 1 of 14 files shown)


site/domains.md
```diff
@@ -85,6 +85,8 @@ running in Kubernetes.
 * _Host Path Persistent Volumes:_ If using a `hostPath` persistent volume, then it must be available on all worker nodes in the cluster and have read/write/many permissions for all container/pods in the WebLogic Server deployment.  Be aware
   that many cloud provider's volume providers may not support volumes across availability zones.  You may want to use NFS or a clustered file system to work around this limitation.
 
+* _Security Note:_ The `USER_MEM_ARGS` environment variable defaults to `-Djava.security.egd=file:/dev/./urandom` in all WebLogic server pods and the WebLogic introspection job. It can be explicitly set to another value in your domain resource yaml using the `env` attribute under the `serverPod` configuration.
```
rosemarymarano: in all WebLogic server -> in all WebLogic Server
domain resource yaml -> domain resource YAML file

PR #860, opened 2019-02-07
WIP: cache the JobWatcher by namespace

vkraemer:
Resolve https://jira.oraclecorp.com/jira/browse/OWLS-71734

This has passed quicktest and  the full integration test suite locally

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import javax.validation.constraints.NotNull;
 import oracle.kubernetes.operator.JobWatcher;
 import oracle.kubernetes.operator.LabelConstants;
 import oracle.kubernetes.operator.ProcessingConstants;
@@ -117,27 +118,15 @@ public class JobHelper {
    *
    * @param tuning Watch tuning parameters
    * @param next Next processing step
-   * @return Step for creating job
-   */
-  @Deprecated
-  public static Step createDomainIntrospectorJobStep(WatchTuning tuning, Step next) {
-
-    // return new DomainIntrospectorJobStep(
-    //    readDomainIntrospectorPodLogStep(ConfigMapHelper.createSitConfigMapStep(next)));
-    return new DomainIntrospectorJobStep(tuning, next, null, new AtomicBoolean(false));
-  }
-
-  /**
-   * Factory for {@link Step} that creates WebLogic domain introspector job
-   *
-   * @param tuning Watch tuning parameters
-   * @param next Next processing step
    * @param jws Map of JobWatcher objects, keyed by the string value of the name of a namespace
    * @param isStopping
    * @return Step for creating job
    */
   public static Step createDomainIntrospectorJobStep(
-      WatchTuning tuning, Step next, Map<String, JobWatcher> jws, AtomicBoolean isStopping) {
+      WatchTuning tuning,
+      Step next,
+      @NotNull Map<String, JobWatcher> jws,
+      @NotNull AtomicBoolean isStopping) {
 
     return new DomainIntrospectorJobStep(tuning, next, jws, isStopping);
   }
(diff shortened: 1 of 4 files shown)


operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
```diff
@@ -33,15 +37,22 @@ public NextAction apply(Packet packet) {
...
-          JobWatcher.create(
-              ThreadFactorySingleton.getInstance(),
-              namespace,
-              initialResourceVersion,
-              tuning,
-              new AtomicBoolean(false));
-
-      return doNext(jw.waitForReady(domainIntrospectorJob, getNext()), packet);
+      JobWatcher jw = null;
+      if (jws == null || null == jws.get(namespace)) {
```
rjeberhard: Can you use !containsKey here?

operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
```diff
@@ -33,15 +37,22 @@ public NextAction apply(Packet packet) {
...
+        jw =
+            JobWatcher.create(
+                ThreadFactorySingleton.getInstance(),
+                namespace,
+                initialResourceVersion,
+                tuning,
+                new AtomicBoolean(false));
```
rjeberhard: I see that you didn't write this line, but it was wrong before.  I purpose of passing in a boolean is so that the watcher can eventually be stopped if and when the operator stops watching this namespace.  The AtomicBoolean should be passed down just like the other use cases for Watchers.

vkraemer: not sure i follow what you are saying here...

operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
vkraemer: done

operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
vkraemer: is this better?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -117,19 +119,40 @@ Domain getDomain() {
    * @param next Next processing step
    * @return Step for creating job
    */
+  @Deprecated
```
rjeberhard: Does anyone call this method?  Let's just delete it.

operator/src/main/java/oracle/kubernetes/operator/steps/WatchDomainIntrospectorJobReadyStep.java
```diff
@@ -33,15 +39,22 @@ public NextAction apply(Packet packet) {
...
-      return doNext(jw.waitForReady(domainIntrospectorJob, getNext()), packet);
+      JobWatcher jw = null;
+      if (jws == null || !jws.containsKey(namespace)) {
+        jw =
+            JobWatcher.create(
+                ThreadFactorySingleton.getInstance(),
+                namespace,
+                initialResourceVersion,
+                tuning,
+                isStopping != null ? isStopping : new AtomicBoolean(false));
```
rjeberhard: isStopping must not be null, so I don't want to pass in a new AtomicBoolean(false)

vkraemer: ummm. ok. so if the caller passes in a null at JobHelper.createDomainIntrospectorJobStep(), how do we want to handle it there? should we put a default in place there or throw an exception?

vkraemer: I think these will help. Let me know.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
vkraemer: will do.

PR #862, opened 2019-02-07
Doc update: we no longer need listen addresses to be blank or exactly…

tbarnes-us:
… correct, so remove that admonition from the doc.


rjeberhard:
@tbarnes-us Should this go to `master` as well?  When did we fix the issue that means we no longer need this workaround?

tbarnes-us:
@tbarnes-us If you have WL with the latest sit-cfg patches, is should be fine.  So yes, it can go into master.

PR #865, opened 2019-02-08
second pass edits

rosemarymarano:
Plus, reverted the LB port number (to the correct setting) in the Quick Start curl example.

### DIFF
--- a/site/quickstart.md
+++ b/site/quickstart.md
@@ -18,7 +18,7 @@ refer to the [User guide](user-guide.md).
 ## Prerequisites
 For this exercise, you’ll need a Kubernetes cluster. If you need help setting one up, check out our [cheat sheet](k8s_setup.md). This guide assumes a single node cluster.
 
-The operator uses Helm to create and deploy necessary resources and then run the operator in a Kubernetes cluster. For Helm installation and usage information, see [Install Helm and Tiller](install.md#install-helm-and-tiller).
+The operator uses Helm to create and deploy the necessary resources and then run the operator in a Kubernetes cluster. For Helm installation and usage information, see [Install Helm and Tiller](install.md#install-helm-and-tiller).
 
 You should clone this repository to your local machine so that you have access to the
 various sample files mentioned throughout this guide:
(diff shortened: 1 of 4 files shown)


site/user-guide.md
```diff
@@ -66,7 +68,7 @@ You can have one or more operators in your Kubernetes cluster that manage one or
 We provide a Helm chart to manage the installation and configuration of the operator.
 Detailed instructions are available [here](install.md).
 
-## Exposing applications outside the Kubernetes cluster
+### Exposing applications outside the Kubernetes cluster
```
doxiao: I am not sure if the changes to the level of heading here and below are correct. We need to make sure they match what is in the table of contents section at least.

rosemarymarano: You're right. I will move both Exposing applications outside the Kubernetes cluster and Operator Docker images down a level in the TOC. Thanks.

site/quickstart.md
```diff
@@ -237,10 +237,10 @@ e.	To confirm that the load balancer noticed the new Ingress and is successfully
 **NOTE**: Be sure to include the trailing forward slash on the URL, otherwise the command won't work.
 
 ```
-$ curl -v -H 'host: sample-domain1.org' http://your.server.com:30701/weblogic/
-* About to connect() to your.server.com port 30701 (#0)
+$ curl -v -H 'host: sample-domain1.org' http://your.server.com:30305/weblogic/
```
doxiao: I have not tested the port. Invited Lily to take a look at this.

rosemarymarano: Just a bit of history. The port number, 30305, came from Lily (I believe) and has always been the port number in the curl command in the Quick Start guide. Recently, Monica gave me the port number to use for WLS Administration Console access, which is port 30701. I changed all the port numbers to 30701, then was informed by Monica that 30305 is the correct port number for the LB whereas 30701 is the correct port number for the admin console, hence the edit. I expect that Lily will confirm this change.

doxiao: Thanks, Rosemary, for the explanation. The changes look good to me then.

PR #866, opened 2019-02-09
Develop rest cert chain tests

AdrianPadilla:


integration-tests/src/test/java/oracle/kubernetes/operator/utils/Operator.java
```diff
@@ -17,6 +17,16 @@
 /** Operator class with all the utility methods for Operator. */
 public class Operator {
 
+  public static enum RESTCertType {
+    /*self-signed certificate and public key stored in a kubernetes tls secret*/
+    SELF_SIGNED,
+    /*Certificate signed by an auto-created CA signed by an auto-created root certificate,
+     * both and stored in a kubernetes tls secret*/
+    CHAIN,
+    /*Certificate and public key, and stored in a kubernetes tls secret*/
```
TheFrogPad: When I read the comment for LEGACY RESTCertType this does not seem to be correct in use of TLS secret type?

I believe that LEGACY indicates the test would be using the Helm chart to pass the actual values and then only the private key is kept a secret, is that correct?

vanajamukkara:
@AdrianPadilla can you use one of the existing operators to test this rather than adding a new operator? Existing operators manages domains, so it covers that flow as well. 

fyi @tbarnes-us @anpanigr

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Operator.java
AdrianPadilla: Yeah, the comment is wrong, it seems like a typical copy/paste issue from the above. Will change to   /*Certificate and public key are stored in a helm values file*/

AdrianPadilla:
Hi @vanajamukkara, yes that was my first approach, but then I realized there is no current test that uninstall the operator i.e. nowhere in ITOperator the Operator#destroy() method is invoked, and I would have to use some logic to wait for the k8s elements to be fully removed before continue with other test.

vanajamukkara:
@AdrianPadilla You are right, operator is not removed in the integration tests. You can add that logic if its needed for your test.

PR #868, opened 2019-02-11
Fix ES_JAVA_OPTS to prevent Initial heap size failure

bobdonat:
The following change in elasticsearch_and_kibana.yaml from the samples has broken the deployment of elasticsearch.

      - name: ES_JAVA_OPTS
        value: -Xms2014m -Xmx1024m

This results in "Error occurred during initialization of VM Initial heap size set to a larger value than the maximum heap size".

The fix is to make the values the same which is the recommendation from elasticsearch.


denismaggior8:
Really sorry @rjeberhard and @bobdonat about that, it was my first PR and I should have confused the two params. Locally I tested with 512Mb each so I did't spot the problem.

markxnelson:
No problem @denismaggior8 keep them coming!

PR #871, opened 2019-02-12
Patch pods to add labels or annotations, rather than restarting

russgold:
When a pod is missing a label or annotation specified in the domain, we no longer recycle it; instead, we patch it. 

We should probably do the same for services, but there is some cleanup that will be needed to prepare for that.


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -472,6 +529,8 @@ public NextAction apply(Packet packet) {
       V1Pod currentPod = getSko().getPod().get();
       if (currentPod == null) {
         return doNext(createNewPod(getNext()), packet);
+      } else if (mustPatchPod(currentPod)) {
```
rjeberhard: Question on the intention here (and we might need another test)...  What if the domain resource has changed such that there would both be an additional label and an additional volume?  For that use case, I believe that the pod needs to be replaced, but I think the else clause ordering would just patch the label.

site/server-lifecycle.md
```diff
@@ -132,7 +132,6 @@ The operator runtime does rolling restarts of clustered servers so that service
 
 ### Properties that cause servers to be restarted
 The operator will restart servers when any of the follow properties on the domain resource that affect the server are changed:
-* `annotations`
```
rjeberhard: Any thoughts on how we should document that servers may have labels or annotations that were previously specified on the domain resource given that the server won't be restarted?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
russgold: I've now changed the order of conditionals to handle that.

site/server-lifecycle.md
russgold: I've added a note.

PR #872, opened 2019-02-12
HOLD: Initial JRF support, including initContainers

markxnelson:
This PR adds initial support for running a JRF domain.  It includes the following changes: 

- the node manager scriptStartMode is changed to true, this is required for JRF-enabled domains
- the readiness probe is fixed (was already merged into develop in a separate PR)
- domain template is updated to use container aware memory settings (-XX:+UseContainerSupport)
- livenessProbe is updated to detect that a server has been shutdown using a standard WebLogic mechanism (admin console, WLST, etc.) and restart the pod

The main change in this PR is the addition of initContainer support. 
This PR allows you to specify initContainers in the serverSpec, as shown in the example below.  These follow the standard Kubernetes initContainers object schema, i.e. V1Container.  When initContainers are included, they will be run before the main containers are started, as per the standard Kubernetes behavior,

```
  serverPod:
    # an (optional) list of environment variable to be set on the servers
    env:
    - name: JAVA_OPTIONS
      value: "-Dweblogic.StdoutDebugEnabled=false"
    - name: USER_MEM_ARGS
      value: "-Xms64m -Xmx256m "
    initContainers:
    - name: test
      image: busybox
      command: ['sh', '-c', 'echo The app is running! && sleep 120']
 ```

Unit tests are added for initContainers and a successful run of the full integration test suite shows no regressions in any of those tests.  (see comment below for caveats)

This PR does not include documentation updates yet!

### CODE REVIEW

**Comment by markxnelson** [comment]
Need to HOLD this PR until discussion with @rjeberhard about the need to rev the schema version for these additions

**Comment by markxnelson** [comment]
I have pulled in @vanajamukkara 's `feature/java-integration-tests

File: operator/src/main/resources/scripts/livenessProbe.sh [hunk 1]
```diff
@@ -34,4 +34,8 @@ if [ -f ${STATEFILE} ] && [ `grep -c "FAILED_NOT_RESTARTABLE" ${STATEFILE}` -eq
   trace "Error: WebLogic Server state is FAILED_NOT_RESTARTABLE."
   exit $RETVAL
 fi
+if [ -f ${STATEFILE} ] && [ `grep -c "SHUTDOWN" ${STATEFILE}` -eq 1 ]; then
```
tbarnes-us: Is this required for JRF?   Checking for shutdown would force a liveness probe failure after a normal shutdown, which leaves no approach for bringing down a server within a pod without causing the pod to immediately restart.
@mriccell @rjeberhard Please comment.

markxnelson: no it is not required for jrf. i just did it in this same branch since the people wanting this are already using this branch. and making it restart the pod after you shut down the server is the desired behavior here.  it is only restarted if the domain CR says it should be running

tbarnes-us: How can one shutdown WL and leave the pod running?

tbarnes-us: @moreaut FYI

markxnelson: why would you want to do that? what's the use case?

tbarnes-us: I thought there was a use case to bring down WL, perform some local update using the container, bring up WL...

moreaut: I'm a bit confused - are you saying that if you go to the console and shutdown a server, then the operator will kill the pod and restart it?  If so, that doesn't sound intuitive.

russgold:
A bunch of changes have happened with regard to how we detect pod changes. That needs to be merged in here, and the appropriate changes made.

kubernetes/samples/scripts/common/utility.sh
```diff
@@ -363,6 +363,10 @@ function createFiles {
     sed -i -e "s:%DOMAIN_ROOT_DIR%:${domainPVMountPath}:g" ${createJobOutput}
     sed -i -e "s:%CREATE_DOMAIN_SCRIPT_DIR%:${createDomainScriptsMountPath}:g" ${createJobOutput}
     sed -i -e "s:%CREATE_DOMAIN_SCRIPT%:${createDomainScriptName}:g" ${createJobOutput}
+    # extra entries for FMW Infra domains
+    sed -i -e "s:%RCU_CREDENTIALS_SECRET_NAME%:${rcuCredentialsSecret}:g" ${createJobOutput}
+    sed -i -e "s:%CUSTOM_RCUPREFIX%:${rcuSchemaPrefix}:g" ${createJobOutput}
+    sed -i -e "s|%CUSTOM_CONNECTION_STRING%|${rcuDatabaseURL}|g" ${createJobOutput}
```
xiancao: is this a typo?
should it be
sed -i -e "s:%CUSTOM_CONNECTION_STRING%:${rcuDatabaseURL}:g" ${createJobOutput}

markxnelson: No that is correct

PR #873, opened 2019-02-12
Optimization and refactoring of java integration tests

vanajamukkara:
Combined multiple use cases into a single test where its appropriate, so that a new domain is not started for every use case. Some refactoring like test method names, domain input file names are changed. We still cover all the use cases, takes less time than before. On Jenkins, full test suite on develop branch takes 1 hr 35 mins. With the optimized tests, it takes 1 hr.

New test methods - testDomainOnPVUsingWLST, testDomainOnPVUsingWDT, testTwoDomainsManagedByTwoOperators, testCreateDomainWithStartPolicyAdminOnly, testCreateDomainPVReclaimPolicyRecycle, testCreateDomainWithDefaultValuesInSampleInputs, testAutoAndCustomSitConfigOverrides, testOperatorRESTIdentityBackwardCompatibility, testOperatorRESTUsingCertificateChain

Jenkins run - http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1031/consoleFull


integration-tests/src/test/resources/DomainOnPVUsingWLST.yaml
```diff
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
 adminServerName: admin-server
-domainUID: domain1
+domainUID: domainonpvwlst
```
tbarnes-us: Can we have the file name match the domainUID as a convention?  That would simplify correlating them, and keep the filenames short.   

Also, can we have a comment at the beginning of each domain setup file that explains its topology?   Ideally, we should be able to easily determine what makes the domain unique by looking at the first few lines of each domain file...

Similarly, it'd be helpful if every operator setup file matched its namespace name.

integration-tests/src/test/resources/domain-home-on-pv/create-domain-auto-custom-sit-config.py
```diff
@@ -48,17 +48,17 @@ def getEnvVar(var):
 # Configure the Administration Server
 # ===================================
 cd('/Servers/AdminServer')
-# Give incorrect listenaddress, introspector overrides with sit-config
-set('ListenAddress', '%s-%s' % (domain_uid, admin_server_name_svc))
+# Dont set listenaddress, introspector overrides automatically with sit-config
+#set('ListenAddress', '%s-%s' % (domain_uid, admin_server_name_svc))
```
tbarnes-us: Do we have formal 'p4 tests' or similar that check (a) setting admin server & managed server listen address/nap address to a completely invalid DNS address (one that doesn't resolve)? (b) or to a DNS address that resolves to an incorrect location ('oracle.com' would probably work internally and externally as an example)?

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
```diff
@@ -30,38 +31,27 @@
...
-  private static String domain11YamlFile = "domain11.yaml";
-  private static String domain12YamlFile = "domain12.yaml";
+  // file used to customize domain properties for domain, PV and LB inputs yaml
+  private static String domainOnPVUsingWLSTYamlFile = "DomainOnPVUsingWLST.yaml";
+  private static String domainOnPVUsingWDTYamlFile = "DomainOnPVUsingWDT.yaml";
+  private static String domainWithServerStartPolicyAsAdminOnlyYamlFile =
+      "DomainWithServerStartPolicyAsAdminOnly.yaml";
+  private static String domainWithStorageReclaimPolicyRecycleYamlFile =
+      "DomainWithStorageReclaimPolicyRecycle.yaml";
+  private static String domainWithDefaultValuesForSamplesYamlFile =
```
tbarnes-us: Similar to my naming suggestion for domain files, maybe have var names like this be something like 'DOMAIN_UIDFile'.

integration-tests/src/test/resources/DomainOnPVUsingWLST.yaml
vanajamukkara: Ok. Are you saying operator setup file name should have the namespace where operator is deployed?

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
```diff
@@ -144,202 +134,192 @@ public static void staticUnPrepare() throws Exception {
...
    * @throws Exception
    */
   @Test
-  public void test5CreateConfiguredDomainInTest2NS() throws Exception {
+  public void testTwoDomainsManagedByTwoOperators() throws Exception {
     Assume.assumeFalse(QUICKTEST);
-
-    logTestBegin("test5CreateConfiguredDomainInTest2NS");
+    String testMethodName = new Object() {}.getClass().getEnclosingMethod().getName();
+    logTestBegin(testMethodName);
```
tbarnes-us: To help avoid cut-and-paste code for logging the begin/end of each test, it looks like you can use an alternate approach to get the current method:
```
  private static final int CALLER_INDEX;
  static {
    int i = 0;
    for (StackTraceElement ste : Thread.currentThread().getStackTrace()) {
      i++;
      if (ste.getClassName().equals(ITOperator.class.getName())) {
        break;
      }
    }
    CALLER_INDEX = i;
  }

  private static String currentTest = "";

  public static void logTestBegin() {
    currentTest = Thread.currentThread().getStackTrace()[CALLER_INDEX].getMethodName();
    System.out.println("Test '" + currentTest + "' - BEGIN");
  }

  public static void logTestSuccess() {
    System.out.println("Test '" + currentTest + "' - SUCCESS");
  }

  public static void logTestInfo(String s) {
    System.out.println("Test '" + currentTest + "' - INFO: " + s);
  }

  /*
  // Sample usage
  public void someTest() {
    logTestBegin();
    logTestInfo("Something something.");
    logTestSuccess();
  }
  */
```
This won't always work on all JVMs according to 'https://stackoverflow.com/questions/442747/getting-the-name-of-the-currently-executing-method/8592871#8592871', but since the operator project always uses the Sun JVM it should be fine.

tbarnes-us:
LGTM overall - I only had some relatively minor naming and code simplification suggestions (provided in other comments).

integration-tests/src/test/resources/domain-home-on-pv/create-domain-auto-custom-sit-config.py
vanajamukkara: I don't think so. If you want to cover these use cases in java tests, I can create a JIRA for this.

integration-tests/src/test/resources/DomainOnPVUsingWLST.yaml
vanajamukkara: @tbarnes-us what should be the operator file name for OperatorManagingdefaultAndtest1NS.yaml based on your suggestion?

tbarnes-us: It can be the same as the operator's namespace, which I assume is unique...

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
vanajamukkara: I tried the above code, its not working Tom, its printing invoke0 for the method name.

integration-tests/src/test/resources/DomainOnPVUsingWLST.yaml
vanajamukkara: ok, changes made.

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
vanajamukkara: ok, changes made.

vanajamukkara:
@rjeberhard This PR is ready to merge.

PR #898, opened 2019-02-22
PR: voyager-java-integ-test

maggiehe00:
In this PR 1) An env var "LB_TYPE" is added. Set it to "VOYAGER" to switch  loadBlancer from the default TRAEFIK to VOYAGER; 2) Correspondingly VOYAGER operator and VOYAGER ingress-per-domain resource will be installed/created; 3) Add a new property voyagerWebPort in domain.yaml to set unique loadBalacerWebPort for the voyager; 4) In the code both "LB_TYPE" and loadBalancerWebPort are handed in the class Domain; 5) Move "INGRESSPERDOMAIN" env var and logic to the class Domain from ITOperator class
The latest Voyager Jenkins job passed at http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1150/
The latest Traefik Jenkins job passed at http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1151/


maggiehe00:
@vanajamukkara On Jenkins I can see Voyager logic is kicked in on http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1102/
but didn't get luck. It looks an intermittent issue.
I just started another run

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
```diff
@@ -146,7 +151,15 @@ public void testDomainOnPVUsingWLST() throws Exception {
     Domain domain = null;
     boolean testCompletedSuccessfully = false;
     try {
-      domain = TestUtils.createDomain(domainonpvwlstFile);
+      // domain = TestUtils.createDomain(domainonpvwlstFile);TODO
+      if (VOYAGER) {
```
vanajamukkara: The logic to check the system env var LB_TYPE and adding into Map can go into Domain.java, may be in createLoadBalancer() method rather than doing the same thing in all test methods.

maggiehe00:
@vanajamukkara...there are some "TODO" commented out code I'm going to delete.

integration-tests/src/test/java/oracle/kubernetes/operator/ITOperator.java
maggiehe00: Here is to add corresponding fields into the domain Map. Later In Domain.java I adde those fields into LB map.

vanajamukkara: are they needed in domain Map?

maggiehe00: Yes. Here we configure the domain to use a certain LB.

vanajamukkara: Are they used in domain creation?

maggiehe00: In our sample domain creation is completely separated from LB creation. My understanding is that domain map here includes both fields used to create the domain and the fields we want to override in the domain.yaml file, which includes both domain creation ones and LB creation ones.

vanajamukkara: Yes, domain and LB creation are separate. domain input yaml in our test framework contains all the properties just so that there is only one input file, but the LB properties are not needed in domain map if you are adding at runtime.

maggiehe00: Under current integ test framework we have only one domain.yaml serving as the input file for the domain (we don't have separate PV.yaml or LB.yaml) . Last time for TRAEFIK ingress-per-domain I added "loadBalancer" and "ingressPerDomain" into the domain.yaml to define LB's properties. This time I need to switch LB to VOYAGER.

vanajamukkara: Yes, you had to add them in domain.yaml as there is no LB properties. But if you are adding at runtime, you don't need to add in the domain map. Even if you want to do so, you can do that in Domain.java instead of repeating in all the test methods.

maggiehe00: Creating LB is just like creating PV and creating domain, not a runtime thing. (I may misunderstand you though).  I only added the VOYAGER logic into the test methods that have "BasicUseCases" --> LB testing, which are the majority of cases of ITOperator.java. If a method doesn't need to test LB ideally we don't even need to create any LB for that method.

maggiehe00: Correction: In 2 methods: testCreateDomainPVReclaimPolicyRecycle and testCreateDomainWithStartPolicyAdminOnly there is no BasicUseCases" --> LB testing but I still added VOYAGER logic because If without VOYAGE, the default TRAEFIk LB will be created.

vanajamukkara: Sorry, it is still not clear to me, why we have to have this logic in every test rather than in the framework classes. Also, not sure why loadBalancerWebPort has to be unique for every domain when VOYAGER is used, but not for TRAEFIK. If cleaning/deleting LB or ingress after every test helps, then may be you can try that.

PR #899, opened 2019-02-26
Container port not added to admin server pod

bobdonat:
This pull request fixes a bug found while testing istio with the WLS operator where the admin service port was not being added as a pod container port.  For managed servers, the port was already being added as a pod container port.

istio has the following requirement for pods:

**Pod ports: Pods must include an explicit list of ports each container will listen on. Use a containerPort in the container spec for each port. Any unlisted ports will bypass the Istio Proxy.**

Jira: https://jira.oraclecorp.com/jira/browse/OWLS-72015


operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -1012,8 +1012,8 @@ private static Step bringAdminServerUp(DomainPresenceInfo info, Step next) {
 
   private static Step[] bringAdminServerUpSteps(DomainPresenceInfo info, Step next) {
     List<Step> resources = new ArrayList<>();
-    resources.add(PodHelper.createAdminPodStep(null));
     resources.add(new BeforeAdminServiceStep(null));
```
mgianatagh: Should we rename this to be BeforeAdminPodStep instead of BeforeAdminServiceStep?

russgold:
Can you explain why changing the order of the steps fixes the issue?

bobdonat:
@russgold The order was changed because the ports needed to create the admin server pod are setup in the method BeforeAdminServiceStep.  This change was suggested by @rjeberhard.

russgold:
Can you please add a unit test to show the problem? That is, without your change, the test should show that the ports are not set up, and with your change, that they are.

bobdonat:
I am not familiar with the unit tests.  Is there an existing test that this should be added to?  Can you point me to an example test to follow.

russgold:
The tests would be added to the test class you already changed, but it is not clear to me how easy it would be to add right now. Where is the code that is actually setting up the ports?

russgold:
After examination, it looks as though a unit test for this would not be trivial, so don't worry about it. I will do some study.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
bobdonat: @rjeberhard is this something you think needs renaming

rjeberhard: I kind of remember discussing this and suggesting BeforeAdminStep.

PR #902, opened 2019-02-27
Handle watch status that is missing or for 410 doesn't include good resource version

rjeberhard:
We found two issues with watches while certifying on OpenShift.  Both issues involved use cases where the API server needs to return an HTTP 401 "Gone" status because the resource version used when the watch was started has aged out.

The expected behavior is that the V1Status contains a message that encodes a "good" resource version as part of the error message.  The error message from OpenShift does not include this good resource.

Additionally, we found that the Kubernetes Java client can incorrectly parse the status if the "target" type of the watch contains similar enough fields, e.g. V1ConfigMap.  In these cases, the watch response is of type ERROR; however, the status is not available.

In each case, the correct behavior is to use a resource version of "0".

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
@@ -193,19 +193,25 @@ abstract class Watcher<T> {
       // not available to our layer, so respond defensively by resetting resource version.
       resourceVersion = 0l;
     } else if (status.getCode() == HTTP_GONE) {
-      resourceVersion = 0l;
-      String message = status.getMessage();
+      resourceVersion = computeNextResourceVersionFromMessage(status);
+    }
+  }
+
+  private long computeNextResourceVersionFromMessage(V1Status status) {
+    String message = status.getMessage();
+    if (message != null) {
       int index1 = message.indexOf('(');
       if (index1 > 0) {
         int index2 = message.indexOf(')', index1 + 1);
         if (index2 > 0) {
           String val = message.substring(index1 + 1, index2);
           if (!isNullOrEmptyString(val)) {
-            resourceVersion = Long.parseLong(val);
+            return Long.parseLong(val);
           }
         }
       }
     }
+    return 0l;
   }
 
   /**
--- a/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
@@ -187,14 +187,22 @@ abstract class Watcher<T> {
 
   private void handleErrorResponse(Watch.Response<T> item) {
     V1Status status = item.status;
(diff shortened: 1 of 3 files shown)


operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -187,14 +187,22 @@ private void handleRegularUpdate(Watch.Response<T> item) {
...
       int index1 = message.indexOf('(');
       if (index1 > 0) {
         int index2 = message.indexOf(')', index1 + 1);
         if (index2 > 0) {
           String val = message.substring(index1 + 1, index2);
-          resourceVersion = !isNullOrEmptyString(val) ? Long.parseLong(val) : 0;
+          if (!isNullOrEmptyString(val)) {
+            resourceVersion = Long.parseLong(val);
+          }
         }
```
russgold: The longer the method, the harder it is for the next person in to understand it without reverse engineering. I think this method has gotten too long at this point. In-line comments can actually aggravate this. We're better off extracting a method. Let me suggest the following:

static final Pattern RESOURCE_PATTERN = Pattern.compile("(.?\\()(\\\\d+)");

private void handleErrorResponse(Watch.Response<T> item) {
  resourceVersion = computeNextResourceVersion(item.status);
    : 

(rest of method)
}

// The kubernetes client parsing logic can mistakenly parse a status as a type
// with similar fields, such as V1ConfigMap. In this case, the actual returned status is
// null, so respond defensively by resetting resource version.
// In addition, the message returned by OpenShift does not include the next valid resource
// in parenthesis, so once again we reset the resource version.
private long computeNextResourceVersion(V1Status status) {
  if (status == null) return 0L;

 Matcher matcher = RESOURCE_PATTERN.matcher(text);
 if (!matcher.find()) return 0L;
else return Long.parseLong(matcher.group(0));
}

operator/src/test/java/oracle/kubernetes/operator/WatcherTestBase.java
```diff
@@ -152,6 +160,36 @@ public void afterHttpGoneError_nextRequestSendsIncludedResourceVersion() {
     }
   }
 
+  @SuppressWarnings({"unchecked", "rawtypes"})
```
russgold: Usually, it is better to let any thrown exceptions propagate from the method, so that they indicate any problems. Catching them can just hide those problems. So remove the catch class, and just declare the needed exceptions on the methods' throw clauses.

russgold: Beyond that, everything looks good.

rjeberhard: Funny...  I just copied these tests, but I'll fix.

PR #907, opened 2019-03-01
READY TO MERGE: Adding domain home in image tests

vanajamukkara:
Adding tests for domain home in image using WLST and WDT.

For domain home in image, logs are written on PV when logHomeOnPv=true, the logs are created by root user. The current cleanup.sh and statedump.sh(which archives logs) uses job.sh which doesn't run as root and can't access these logs. So, modified these scripts to use krun.sh provided by Tom Barnes which gives the option to select the image you want to use, using jre image solves the issue as it uses root user. 

Jenkins run passed - http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1163/console


integration-tests/README.md
```diff
@@ -181,11 +183,15 @@ test methods -  testDomainOnPVUsingWLST, testDomainOnPVUsingWDT, testTwoDomainsM
...
+Domain - constructor takes Map with domain, LB, PV properties and creates domain crd, LB operator/ingress and PV artifacts using the sample scripts provided in the project. Also contains helper methods to destroy domain by deleting domain crd, verify domain created and servers are ready, deploy webapp, verify load balancing of http requests, etc <p>
+PersistentVolume - runs k8s job to create PV directory and creates PV and PVC using sample scripts  <p>
+LoadBalancer - creates load balancer, currently TREFIK and VOYAGER are supported <p>
+Secret - creates a k8s secret <p>
+TestUtils - mostly runs kubectl commands. Contains utility methods to check if a pod is created, ready, deleted, service created, get pod restart cnt, get cluster replica, delete PVC, check PV released, create rbac policy, create wldf module, etc. <p>
+ExecCommand - Class for executing shell commands from java <p>
+ExecResult - Class that holds the results of using java to exec a command (i.e. exit value, stdout and stderr) <p>
+K8sTestUtils - uses k8s java client api, this is used only for delete domain use cases for now. <p>
 
 # How to run the Java integration tests
```
tbarnes-us: Please update these instructions to look something like:
* Maven and latest Git should be in PATH
* export JAVA_HOME
* export WEBLOGIC_IMAGE_NAME and WEBLOGIC_IMAGE_TAG if different from `store/oracle/weblogic` and `12.2.1.3`.
* Setup docker access to WebLogic 12c Images
  * Method 1
    * Setup a personal account on hub.docker.com
    * Then sign in to hub.docker.com and signup for access to WebLogic 12c Images via https://hub.docker.com/_/oracle-weblogic-server-12c
    * Then export the following before running the tests:
      * export DOCKER_USERNAME=<docker user>
      * export DOCKER_PASSWORD=<docker pass>
      * export DOCKER_EMAIL=<docker email>
  * Method 2
      * Make sure the weblogic image i.e. store/oracle/weblogic:12.2.1.3 already exists locally
      * Make sure the weblogic image has patch p29135930 (required for the WebLogic Kubernetes Operator).
        * If not, see [https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12213-patch-wls-for-k8s].

* Command to run the tests: ```
 mvn clean verify -P java-integration-tests 2>&1 | tee log.txt
 ```

integration-tests/src/test/resources/deploywebapp.py
```diff
@@ -1,5 +1,22 @@
 # Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
+import time as systime
+
 connect(sys.argv[1],sys.argv[2],sys.argv[3])
-deploy(sys.argv[4],sys.argv[5],sys.argv[6],upload='false',remote='false')
\ No newline at end of file
+deploy(sys.argv[4],sys.argv[5],sys.argv[6],upload='false',remote='false')
+systime.sleep(30)
```
tbarnes-us: This seems like a long sleep.   It might be too long on some systems - making the tests run too long, or too short on others - cuasing the 'state print' you added below to be invalid.   Is there a way you can have WLST loop every  second to wait until '/AppDeployments/'+appName.getName()+'/Targets' and its associated runtime exist?

integration-tests/src/test/resources/statedump.sh
```diff
@@ -60,16 +63,46 @@ function state_dump {
...
+	 	if [ "$?" = "0" ]; then
+   			base64 -di $RESULT_DIR/pvarchive.b64 > $ARCHIVE
+   			if [ "$?" = "0" ]; then
+   				echo Run complete. Archived to $ARCHIVE
+   			else 
+   				echo Run failed. 
+   			fi
+	 	else
+     		# command failed
+  			cat $RESULT_DIR/pvarchive.b64 | head -30
```
tbarnes-us: maybe make this '-100' instead of '-30' - on an error the output may include a kubectl describe, and sometimes a kubetcl describe can be pretty big...

integration-tests/src/test/resources/deploywebapp.py
vanajamukkara: Actually I have added the app status code for my debugging. I think I should remove it. Is it needed?

src/integration-tests/bash/krun.sh
```diff
@@ -0,0 +1,276 @@
...
+  PWDFILTER="{{DNE=DNE}}"
+fi
+
+cat $TEMPFILE | sed "s/{{EXITCODE=0=EXITCODE}}//" \
+              | sed "s/$IDFILTER//" \
+              | sed "s/$PWDFILTER//" \
+              | grep -v "Unable to use a TTY - input is not a terminal or the right kind of file" \
+              | grep -v "If you don't see a command prompt, try pressing enter." \
+              | grep -v "Error attaching, falling back to logs:"
+
```
tbarnes-us: FYI:   When I wrote/tested this script, I only tried it out on 1.10.11 (client and server).   I don't know if the above 'grep -v' commands are sufficient to filter out the stderr/stdout extraneous cruft on all version of k8s.   This means a krun.sh output might unexpectedly  include some unexpected extra output sometimes on some versions -- which in turn would mean the 'base64' copy trick won't work.

tbarnes-us: If you do see such cruft, I have an idea about a potentially more generic way to filter it out (print sentinel values at start and end of a run, and filter out everything before and after these sentinel values).

integration-tests/src/test/resources/deploywebapp.py
tbarnes-us: Since it adds 30 seconds, and its only debug, maybe just make it optional (default off), or leave it commented out...   Unless you can figure out how to do a loop/test of some sort...

integration-tests/README.md
vanajamukkara: Done

src/integration-tests/bash/krun.sh
vanajamukkara: Ok.

integration-tests/src/test/resources/statedump.sh
vanajamukkara: Done

src/integration-tests/bash/krun.sh
vanajamukkara: Ok

integration-tests/src/test/resources/deploywebapp.py
vanajamukkara: commented.

vanajamukkara:
@tbarnes-us please review

integration-tests/README.md
```diff
@@ -179,27 +193,43 @@ staticUnPrepare() - releases the cluster lease on wercker env.
...
+  - Then sign in to hub.docker.com and signup for access to WebLogic 12c Images via https://hub.docker.com/_/oracle-weblogic-server-12c
+  - Then export the following before running the tests: 
+	```
+	export DOCKER_USERNAME=<docker_username>
+	export DOCKER_PASSWORD=<docker_password>
+	export DOCKER_EMAIL=<docker_email>
+	```
+ 
+ Method 2 
+ - Make sure the weblogic image i.e. store/oracle/weblogic:12.2.1.3 already exists locally
```
tbarnes-us: locally --> in a docker repository the k8s cluster can access

integration-tests/README.md
```diff
@@ -179,27 +193,43 @@ staticUnPrepare() - releases the cluster lease on wercker env.
...
+ExecResult - Class that holds the results of using java to exec a command (i.e. exit value, stdout and stderr) <p>
+K8sTestUtils - uses k8s java client api, this is used only for delete domain use cases for now. <p>
 
 # How to run the Java integration tests
 
 * Maven and latest Git should be in PATH
 * export JAVA_HOME
-
-Command to run the tests:
+* export WEBLOGIC_IMAGE_NAME and WEBLOGIC_IMAGE_TAG if different from store/oracle/weblogic and 12.2.1.3
```
tbarnes-us: + export IMAGE_PULL_SECRET_WEBLOGIC if a k8s secret is needed to access the image (plus deploy the k8s secret before running the test)

vanajamukkara: @tbarnes-us secret IMAGE_PULL_SECRET_WEBLOGIC is created using the docker credentials provided by DOCKER_USERNAME, DOCKER_PASSWORD, DOCKER_EMAIL. Its not required to export IMAGE_PULL_SECRET_WEBLOGIC as the test takes care of that.

vanajamukkara: @tbarnes-us secret IMAGE_PULL_SECRET_WEBLOGIC is created using the docker credentials provided by DOCKER_USERNAME, DOCKER_PASSWORD, DOCKER_EMAIL. Its not required to export IMAGE_PULL_SECRET_WEBLOGIC as the test takes care of that.

tbarnes-us: Oh - I see.  Isn't it possible that it is required when USER/PASS/EMAIL are not specified?   I should have made that clear.

vanajamukkara: Currently tests expect USER/PASS/EMAIL if the image doesn't exist locally, otherwise it fails with the error message asking to provide the docker login details.

PR #910, opened 2019-03-04
OWLS-72421 envvar macro substitution got wrong value for managed server pods

alai8:
We need to create copies of the V1EnvVars from ServerSpec before performing value translation to prevent the managed server pods from getting the same translated V1EnvVar values as in the admin server pods.

Fix verified by Huiling

Passes jenkins: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1157/


russgold:
LGTM

PR #911, opened 2019-03-05
fix format

rosemarymarano:
Because of an error in the number of code marks (tics), the format was incorrect in the"Use Helm Chart from Github chart repository" section code examples.


rosemarymarano:
Also, fixed section title word capitalization, "GitHub."

PR #912, opened 2019-03-05
docker pull is needed in setup.env to get the wlthinclient jar

vanajamukkara:
docker login and docker pull was commented in my previous PR which makes the tests to fail if the image doesn't image locally. reverting the change.


vanajamukkara:
@rjeberhard - this PR is ready, it should go ASAP.

PR #917, opened 2019-03-06
READY TO MERGE: Add new property createDomainPyScript to domain inputs in java tests

vanajamukkara:
https://jira.oraclecorp.com/jira/browse/OWLS-72639

domain input files like domainonpvwlst.yaml can have a new property to override the create-domain.py.

createDomainPyScript: integration-tests/src/test/resources/domain-home-on-pv/create-domain-custom-nap.py

or it can be added into the domain Map from the test method.

Also copy the samples scripts to RESULTS_DIR location at the beginning for every domain.
The file given in createDomainPyScript is copied to the samples dir or downloaded docker-images dir before calling the create-domain.sh script. This gives the option to use the  custom domain configuration than the one provided in the samples for the domains created with WLST. This property can be used for both domain on pv and domain home in image domains using WLST.

Jenkins passed - http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1213/


anpanigr:
I verified the change by adding a custom py script to add JMS Configurations. It works

tbarnes-us:
The description should be updated to describe the change (the JIRA link isn't available publicly).

vanajamukkara:
> 
> 
> The description should be updated to describe the change (the JIRA link isn't available publicly).

Updated the description.

PR #918, opened 2019-03-06
OWLS 72452: WLDF scaling failing on Jenkins with Operator Cert File not found

lennyphan:
Internal operator certificate intermittently not available from weblogic-operator-cm config map as it is dependent on when the config map is updated.  We now read the internal operator certificate from /operator/internal-identity/internalOperatorCert file.


operator/src/main/java/oracle/kubernetes/operator/TuningParametersImpl.java
```diff
@@ -134,4 +137,28 @@ public PodTuning getPodTuning() {
...
+
+  // path - a file containing a base64 encoded string containing the operator's cert in pem format
+  public String getFileContents(String path) {
+    // in pem format
+    String result = null;
+    if (checkFileExists(path)) {
+      try {
+        result = new String(Files.readAllBytes(Paths.get(path)));
+      } catch (Throwable t) {
+      }
```
rjeberhard: LOGGER.warning(MessageKeys.EXCEPTION, t);

lennyphan: Thanks.  Fixed.  I didn't know there was a generic exception message for logging.

operator/src/test/java/oracle/kubernetes/operator/helpers/AdminPodHelperTest.java
```diff
@@ -41,7 +41,8 @@
 
 @SuppressWarnings("SameParameterValue")
 public class AdminPodHelperTest extends PodHelperTestBase {
-  private static final String INTERNAL_OPERATOR_CERT_FILE_PARAM = "internalOperatorCert";
+  private static final String INTERNAL_OPERATOR_CERT_FILE_PARAM =
```
moreaut: Do we really want another copy of this pathname? (i.e. RestConfigImpl.java has it too) - should we have a shared constant instead?

lennyphan: I agree. Any opinions on an existing interface to put the shared constants or create a new interface?

lennyphan: I've refactored out the path names and the reading of the certificate data to TuningParameters to remove duplicate code.  Also easiest to add it to TuningParameters as convenience methods as well as easier for unit test changes.  Please take a look and comment.

operator/src/main/java/oracle/kubernetes/operator/rest/RestConfigImpl.java
```diff
@@ -63,12 +58,12 @@ public int getInternalHttpsPort() {
 
   @Override
   public String getOperatorExternalCertificateData() {
-    return getCertificate(EXTERNAL_CERTIFICATE);
+    return TuningParameters.getInstance().getFileContents(EXTERNAL_CERTIFICATE);
```
alai8: Should we move EXTERNAL_CERTIFICATE to OperatorConstants too? Seems odd that it is not defined in the same place as INTERNAL_CERTIFICATE

operator/src/main/java/oracle/kubernetes/operator/TuningParametersImpl.java
```diff
@@ -134,4 +137,38 @@ public PodTuning getPodTuning() {
...
+
+  public boolean checkFileExists(String path) {
+    LOGGER.entering(path);
+    File f = new File(path);
+    boolean result = false;
+    if (f.exists()) {
+      if (f.isFile()) {
+        result = true;
+      } else {
+        LOGGER.warning(path + " is not a file");
```
alai8: should we create MessageKeys for these messages?

operator/src/main/java/oracle/kubernetes/operator/rest/RestConfigImpl.java
moreaut: I think EXTERNAL_CERTIFICATE, EXTERNAL_CERTIFICATE_KEY and INTERNAL_CERTIFICATE_KEY should all move to OperatorConstants.

moreaut: Actually, maybe instead of moving the constants to OperatorConstants, and writing yet more code that reads the certificate from the file, how about if AdminPodStepContext is somehow given a RestConfig instance - since RestConfig has methods that read the contents of the file - e.g.:

  /**
   * Gets the internal https port's certificate.
   *
   * @return base64 encoded PEM containing the certificate, or null if
   *     getOperatorInternalCertificateFile should be used instead to get the certificate.
   */
  String getOperatorInternalCertificateData();

lennyphan: I refactored the code that reads the certificate from RestConfigImpl to TuningParameters as my first pass didn't look like it was going to be easy to pass the RestConfig to the AdminPodContext.  I can change it back and try and pass RestConfig if you prefer?

lennyphan: Ok, I'll try and make change to give AdminPodContext a RestConfig instance.

operator/src/main/java/oracle/kubernetes/operator/rest/RestConfigImpl.java
```diff
@@ -4,26 +4,21 @@
...
 import java.util.Collection;
+import oracle.kubernetes.operator.OperatorConstants;
+import oracle.kubernetes.operator.TuningParameters;
 import oracle.kubernetes.operator.logging.LoggingFacade;
 import oracle.kubernetes.operator.logging.LoggingFactory;
 import oracle.kubernetes.operator.rest.backend.RestBackend;
 
 /** RestConfigImpl provides the WebLogic Operator REST api configuration. */
-public class RestConfigImpl implements RestConfig {
+public class RestConfigImpl implements RestConfig, OperatorConstants {
```
russgold: Are you implementing an interface in order to import constants? That's a very old practice, and should be avoided. Instead, you can statically import them into a class if you want to access them without specifying the interface on every reference.

lennyphan: I'm working on a different change to reuse the RestConfig.  Might want to wait on the review.  Thank you!

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -63,14 +65,29 @@
     SSL_PROTOCOL
   }; // ONLY support TLSv1.2 (by default, we would get TLSv1 and TLSv1.1 too)
 
+  public static synchronized void create(RestConfig restConfig) {
+    if (INSTANCE == null) {
+      INSTANCE = new RestServer(restConfig);
+    }
+    // throw new IllegalStateException();
```
moreaut: uncomment the throw

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -63,14 +65,29 @@
...
+  }
+
+  public static synchronized RestServer getInstance() {
+    return INSTANCE;
+  }
+
+  public static void destroy() {
+    INSTANCE = null;
```
moreaut: throw if INSTANCE != null ?

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -168,6 +185,18 @@ public void stop() {
     LOGGER.exiting();
   }
 
+  public String getInternalCertificate() {
```
moreaut: This is supposed to return base64 encoded pem - right?
Might want to rename it to match.
Add javadoc and LOGGER.entering ... since this is a public method (i.e. the RESTServer traces all public methods).
Please add javadoc & LOGGING for the new static create/destroy methods too.

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -321,6 +350,13 @@ private SSLContext createSSLContext(KeyManager[] kms) throws Exception {
     return result;
   }
 
+  private static String readCertFromDataOrFile(String data, String file) throws IOException {
```
moreaut: This seems weird - readFromDataOrFile always returns PEM.  either data is pem, or the file is base64 pem.

What's this method supposed to return?  base64 pem? it will receive the same data/file as readFromDataOrFile.
If it's supposed to return base64 pem, then if it uses data, it needs to base64 encode it.

operator/src/test/java/oracle/kubernetes/operator/rest/RestTest.java
```diff
@@ -44,7 +44,6 @@
 /** Tests the Weblogic Operator REST api */
 public class RestTest {
```
moreaut: add a new test method for the new getInternalCertificate method

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -168,6 +185,18 @@ public void stop() {
     LOGGER.exiting();
   }
 
+  public String getInternalCertificate() {
+    try {
+      return readCertFromDataOrFile(
+          this.config.getOperatorInternalCertificateData(),
+          this.config.getOperatorInternalCertificateFile());
+    } catch (IOException e) {
+      LOGGER.warning("Unable to read internal certificate data", e);
```
rjeberhard: Add log value info() and above requires i18n.

PR #919, opened 2019-03-07
update prereqs

rosemarymarano:
Added OpenShift version info and how to check the WLS version/patches.

### DIFF
--- a/site/user-guide.md
+++ b/site/user-guide.md
@@ -98,7 +98,7 @@ Oracle recommends careful consideration before deciding to expose any administra
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
 * Docker 18.03.1.ce (check with `docker version`).
 * Helm 2.8.2+ (check with `helm version`).
-* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930.
+* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930 (check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`).
    * The existing WebLogic Docker image, `store/oracle/weblogic:12.2.1.3`,
 was updated on January 17, 2019, and has all the necessary patches applied.
    * A `docker pull` is required if you pulled the image prior to that date.
@@ -106,4 +106,6 @@ was updated on January 17, 2019, and has all the necessary patches applied.
 
 ### OpenShift
 
-When using the operator in OpenShift, the `anyuid` security context constraint is required to ensure that WebLogic containers run with a Unix UID that has the correct permissions on the domain filesystem.
+The operator is certified for use on OpenShift 3.11.43+
+
+When using the operator in OpenShift, the `anyuid` security context constraint is required to ensure that WebLogic containers run with a UNIX UID that has the correct permissions on the domain filesystem.
--- a/site/user-guide.md
+++ b/site/user-guide.md
@@ -98,7 +98,9 @@ Oracle recommends careful consideration before deciding to expose any administra
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
 * Docker 18.03.1.ce (check with `docker version`).
 * Helm 2.8.2+ (check with `helm version`).
-* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930 (check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`).
+* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930.
+   * Check WLS version with `docker run store/oracle/weblogic:12.2.1.3 sh -c` `'source $ORACLE_HOME/wlserver/server/bin/setWLSEnv.sh > /dev/null 2>&1 && java weblogic.version'`.
+   * Check WLS patches with `docker run store/oracle/weblogic:12.2.1.3 sh -c` `'$ORACLE_HOME/OPatch/opatch lspatches'`.
    * The existing WebLogic Docker image, `store/oracle/weblogic:12.2.1.3`,
 was updated on January 17, 2019, and has all the necessary patches applied.
    * A `docker pull` is required if you pulled the image prior to that date.
--- a/site/user-guide.md
+++ b/site/user-guide.md
@@ -106,6 +106,6 @@ was updated on January 17, 2019, and has all the necessary patches applied.
 
 ### OpenShift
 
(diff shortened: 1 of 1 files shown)


site/user-guide.md
```diff
@@ -98,12 +98,14 @@ Oracle recommends careful consideration before deciding to expose any administra
...
+* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930 (check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`).
    * The existing WebLogic Docker image, `store/oracle/weblogic:12.2.1.3`,
 was updated on January 17, 2019, and has all the necessary patches applied.
    * A `docker pull` is required if you pulled the image prior to that date.
 * You must have the `cluster-admin` role to install the operator.
 
 ### OpenShift
 
-When using the operator in OpenShift, the `anyuid` security context constraint is required to ensure that WebLogic containers run with a Unix UID that has the correct permissions on the domain filesystem.
+The operator is certified for use on OpenShift 3.11.43+
```
tbarnes-us: According to @mriccell Monica's slack, OpenShift also requires v1.11 k8s (without OpenShift just need 1.10.11) - actually this should be 1.11.5 to match the 1.11 pre-req above, and 2.0.1+ operator (not 2.0)

site/user-guide.md
```diff
@@ -98,12 +98,14 @@ Oracle recommends careful consideration before deciding to expose any administra
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
 * Docker 18.03.1.ce (check with `docker version`).
 * Helm 2.8.2+ (check with `helm version`).
-* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930.
+* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930 (check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`).
```
tbarnes-us: Monica:

As far as I can tell, the opatch lsinventory command won't actually report the WL version in a clear way even if adding `-details` or `-all`.  Go figure.  It doesn't seem to be designed to tell you what products are installed so much as which patches are installed.  Is there a way to invoke it to get the WL version out of it in a clear way?

Alternatively, the following will report the version and the patches:    `docker run store/oracle/weblogic:12.2.1.3 sh -c 'source /u01/oracle/wlserver/server/bin/setWLSEnv.sh && java weblogic.version -verbose'`, so maybe that's better?

mriccell: It is lspatches that tell you WLS version and the patches that have been applied.  
  $ cd OPatch
    $ ./opatch version
    $ ./opatch lspatches

mriccell: ./opatch version is specific to the version of opatch which sometimes needs to be updated.  For this documentation lspatches should be enough.
Monica

rjeberhard:
What's the current state of this PR?  Looks like both Tom and Monica approved, but then have withdrawn approval?

rosemarymarano:
@mriccell Yes, there was a bit of hashing back and forth. Last I read, Monica gave a final ok, but she can confirm to be sure. Thanks.

tbarnes-us:
OK, let's change "(check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`)" to (check wl version with `docker run store/oracle/weblogic:12.2.1.3 sh -c 'source $ORACLE_HOME/wlserver/server/bin/setWLSEnv.sh > /dev/null 2>&1 && java weblogic.version'`, check wl patches with `docker run store/oracle/weblogic:12.2.1.3 sh -c '$ORACLE_HOME/OPatch/opatch lspatches'`)

site/user-guide.md
tbarnes-us: OK, let's change "(check with docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory)" to (check wl version with docker run store/oracle/weblogic:12.2.1.3 sh -c 'source $ORACLE_HOME/wlserver/server/bin/setWLSEnv.sh > /dev/null 2>&1 && java weblogic.version', check wl patches with docker run store/oracle/weblogic:12.2.1.3 sh -c '$ORACLE_HOME/OPatch/opatch lspatches')

tbarnes-us:
@mriccell Suggested a change to use 'lspatches' instead of 'lsinventory'.   And I suggest we add a call to 'java weblogic.version' (opatch doesn't seem to have a clear way to reveal the WL version).    I've incorporatged both suggestions in the comment chain above.

site/user-guide.md
tbarnes-us: (The reason I think we need 'weblogic.version' too is that none of the opatch options - lspatches/version/lsinventory - clearly reveal the WL version.)

rosemarymarano:
Monica approved Tom's suggestions. I incorporated them. 
Tom, please double-check the syntax.

site/user-guide.md
```diff
@@ -98,7 +98,9 @@ Oracle recommends careful consideration before deciding to expose any administra
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
 * Docker 18.03.1.ce (check with `docker version`).
 * Helm 2.8.2+ (check with `helm version`).
-* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930 (check with `docker run store/oracle/weblogic:12.2.1.3 /u01/oracle/OPatch/opatch lsinventory`).
+* Oracle WebLogic Server 12.2.1.3.0 with patch 29135930.
+   * Check WLS version with `docker run store/oracle/weblogic:12.2.1.3 sh -c` `'source $ORACLE_HOME/wlserver/server/bin/setWLSEnv.sh > /dev/null 2>&1 && java weblogic.version'`.
+   * Check WLS patches with `docker run store/oracle/weblogic:12.2.1.3 sh -c` `'$ORACLE_HOME/OPatch/opatch lspatches'`.
    * The existing WebLogic Docker image, `store/oracle/weblogic:12.2.1.3`,
 was updated on January 17, 2019, and has all the necessary patches applied.
    * A `docker pull` is required if you pulled the image prior to that date.
```
tbarnes-us: Maybe make the last two bullets the first two bullets - so the references to 'store/oracle/weblogic:12.2.1.3' in the 'Check' bullets have more context.  Otherwise, LGTM.

tbarnes-us: I've verified the syntax - cut/pasting from the `rich diff`, then running the command, worked fine...

rosemarymarano: @rjeberhard I think it's ready now to be merged.

PR #921, opened 2019-03-08
Domain CRD version upgrade and move to v3alpha1

rjeberhard:
Current Domain CRD version is "v2".  Version values must be well-formed, which is `"v<version>[alpha|beta[<pre-release version>]]"`.

Sadly, this means that "v2.1" is not a valid schema version.  Therefore, we are working toward version "v3".

Since I'm assuming we will at least internally have multiple schema changes before we are satisfied, I've started the schema version at "v3alpha1" and included the support for schema version ordering and updating the CRD when appropriate.

Kubernetes currently only supports a CRD having one schema definition!!!  Therefore, while we can change schema, the changes have to be backward compatible (i.e. only add optional fields).

The CRD can then specify which schema versions can be served and also which can be stored.

If the operator is listening for "v3alpha1" it will get domains that are of type "v2" should "v2" still be included in the list of stored versions.

This change:
- Makes the schema version "v3alpha1"
- Preserves "v2" as a served version, which means 2.0 operators will continue to work

Finally, there is forward looking code so that the operator will not overwrite the CRD if a future operator version has already updated the CRD for later versions (e.g. "v4", "v5"), but that this operator will ensure that it's current version is included in the served versions.


russgold:
The inclusion of java doc (and generated files in general) in git makes it really hard to review changes like this; There are 141 files listed as changed. That would not be an issue if we officially generated the site, as with the maven site plugin.

operator/src/main/java/oracle/kubernetes/operator/helpers/CRDHelper.java
```diff
@@ -112,6 +119,20 @@ static V1beta1CustomResourceDefinitionSpec createSpec(KubernetesVersion version)
       return spec;
     }
 
+    static List<V1beta1CustomResourceDefinitionVersion> getCRDVersions() {
+      List<V1beta1CustomResourceDefinitionVersion> versions =
+          Arrays.stream(KubernetesConstants.DOMAIN_ALTERNATE_VERSIONS)
+              .map(e -> new V1beta1CustomResourceDefinitionVersion().name(e).served(true))
+              .collect(Collectors.toList());
+      versions.add(
```
markxnelson: i assume this results in the right order overall...

operator/src/main/java/oracle/kubernetes/operator/helpers/CRDHelper.java
```diff
@@ -208,8 +231,54 @@ private boolean isOutdatedCRD(V1beta1CustomResourceDefinition existingCRD) {
...
+      if (versions != null) {
+        for (V1beta1CustomResourceDefinitionVersion v : versions) {
+          if (KubernetesConstants.DOMAIN_VERSION.equals(v.getName())) {
+            found = true;
+            break;
+          }
+        }
+      }
+
+      return !found;
```
markxnelson: i find the naming and negation confusing and suspect it will create future problems, would prefer to fix now

russgold:
Given that we are allowed alpha characters in a version, could we not call it something like "v2revision1"?

operator/src/main/java/oracle/kubernetes/operator/helpers/CRDHelper.java
rjeberhard: Yes, correct.  For k8s 1.10 - 1.12, I found that the version field had to be set to the first entry in versions.  K8s 1.13 removed that restriction.

operator/src/main/java/oracle/kubernetes/operator/helpers/HealthCheckHelper.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017, 2018 Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2017, 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: if you change this file, you have to update the year to 2019 :)

operator/src/main/java/oracle/kubernetes/operator/helpers/CRDHelper.java
```diff
@@ -248,18 +333,45 @@ boolean isOutdatedCRD(
...
+      }
+
+      return getSchemaValidation(actual) == null
+          || !getSchemaValidation(expected).equals(getSchemaValidation(actual))
+          || !getSchemaSubresources(expected).equals(getSchemaSubresources(actual));
       // Similarly, we will later want to check:
       // VersionHelper.matchesResourceVersion(existingCRD.getMetadata(), DEFAULT_OPERATOR_VERSION)
     }
 
+    // true, if right is later than left
```
markxnelson: don't like the names left and right.....

rjeberhard:
> Given that we are allowed alpha characters in a version, could we not call it something like "v2revision1"?

No, this isn't legal.

The only values after `v` and a positive integer are "alpha" or "beta".

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesVersion.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: 2018, 2019,

operator/src/main/java/oracle/kubernetes/operator/helpers/ResourceVersion.java
```diff
@@ -0,0 +1,174 @@
...
+
+    version = Integer.valueOf(value.substring(n, i));
+
+    if (i < value.length()) {
+      String remainder = value.substring(i);
+      if (remainder.startsWith("alpha")) {
+        prerelease = "alpha";
+        i += 5;
+        n = i;
+        for (; i < value.length(); i++) {
```
markxnelson: this is fine, but a regex might be nicer

operator/src/test/java/oracle/kubernetes/operator/builders/EventMatcher.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: 2018, 2019,

operator/src/test/java/oracle/kubernetes/operator/builders/StubWatchFactory.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Oracle Corporation and/or its affiliates.  All rights reserved.
+// Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: 2018, 2019,

operator/src/main/java/oracle/kubernetes/operator/helpers/ResourceVersion.java
rjeberhard: I have 97% code coverage, so I'm going to leave it alone. :)

PR #927, opened 2019-03-11
Create Operator and weblogic domain with managed coherence servers. Verify the domain is started successfully and coherence is enabled on each managed servers.

chpatel3:
Signed-off-by: chinmay patel <chinmay.patel@oracle.com>

### DIFF
--- a/integration-tests/src/test/resources/coh/domaincreation.yaml
+++ b/integration-tests/src/test/resources/coh/domaincreation.yaml
@@ -1,0 +1,13 @@
+# Copyright 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+
+domainUID: cohdomain
+namespace: coh
+configuredManagedServerCount: 4
+initialManagedServerReplicas: 2
+exposeAdminT3Channel: true
+exposeAdminNodePort: true
+domainHomeImageBuildPath: ./docker-images/OracleWebLogic/samples/12213-domain-home-in-image-wdt
+domainHomeImageBase: "store/oracle/weblogic:12.2.1.3"
+logHomeOnPV: true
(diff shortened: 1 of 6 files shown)


vanajamukkara:
@chpatel3 Please update description in the PR and what is the use case for the new test.

rjeberhard:
I just realized that we don't yet have https://jira.oraclecorp.com/jira/browse/OWLS-71966...  This is the support for correctly doing a rolling restart of a Coherence cluster.  Should we wait to add tests until we can correctly support a cluster?

chpatel3:
Currently test only verifies that coherence is enabled on each managed servers in a domain started by weblogic-operator and does not do rolling restart of the cluster. So I'm fine with either way. As this test currently does not do rolling restart this can be merged as it is or wait until OWLS-71966 is implemented with the test that do rolling restart of a cluster.

integration-tests/src/test/resources/coh/domaincreation.yaml
```diff
@@ -0,0 +1,13 @@
...
+domainUID: cohdomain
+namespace: coh
+configuredManagedServerCount: 4
+initialManagedServerReplicas: 2
+exposeAdminT3Channel: true
+exposeAdminNodePort: true
+domainHomeImageBuildPath: ./docker-images/OracleWebLogic/samples/12213-domain-home-in-image-wdt
+domainHomeImageBase: "store/oracle/weblogic:12.2.1.3"
+logHomeOnPV: true
+replaceDomainModelYaml: coh/coherence.yaml
```
vanajamukkara: can you change the property name to domainModelYaml, give path as integration-tests/src/test/resources/coh/coherence.yaml and change the code accordingly? Otherwise looks good.

chpatel3: Thanks @vanajamukkara! I will make that change and resubmit the PR with that change.

PR #934, opened 2019-03-14
READY TO MERGE: Fix wercker lease failures and some code refactoring

vanajamukkara:
Changed new test ITUsabilityOperatorHelmChart not to be called on wercker, this will fix the lease failures.
archive and cleanup is done after every IT class, so made changes to create PV_ROOT and RESULT_ROOT dirs in initialize() method in BaseTest and also moved some common code to BaseTest.java

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -27,6 +27,25 @@ public class BaseTest {
   public static final Logger logger = Logger.getLogger("OperatorIT", "OperatorIT");
   public static final String TESTWEBAPP = "testwebapp";
 
+  // property file used to customize operator properties for operator inputs yaml
+
+  public static final String OPERATOR1_YAML = "operator1.yaml";
+  public static final String OPERATOR2_YAML = "operator2.yaml";
+  public static final String OPERATORBC_YAML = "operator_bc.yaml";
+  public static final String OPERATOR_CHAIN_YAML = "operator_chain.yaml";
+
+  // file used to customize domain properties for domain, PV and LB inputs yaml
+  public static final String DOMAINONPV_WLST_YAML = "domainonpvwlst.yaml";
+  public static final String DOMAINONPV_WDT_YAML = "domainonpvwdt.yaml";
+  public static final String DOMAIN_ADMINONLY_YAML = "domainadminonly.yaml";
+  public static final String DOMAIN_RECYCLEPOLICY_YAML = "domainrecyclepolicy.yaml";
+  public static final String DOMAIN_SAMPLE_DEFAULTS_YAML = "domainsampledefaults.yaml";
+  public static final String DOMAININIMAGE_WLST_YAML = "domaininimagewlst.yaml";
+  public static final String DOMAININIMAGE_WDT_YAML = "domaininimagewdt.yaml";
+
+  // property file used to configure constants for integration tests
+  public static final String APP_PROPS_FILE = "OperatorIT.properties";
+
   private static String resultRoot = "";
   private static String pvRoot = "";
   private static String resultDir = "";
@@ -40,6 +59,28 @@ public class BaseTest {
   private static String branchName = "";
   private static String appLocationInPod = "/u01/oracle/apps";
   private static Properties appProps;
+
+  public static boolean QUICKTEST;
+  public static boolean SMOKETEST;
+  public static boolean JENKINS;
+  public static boolean INGRESSPERDOMAIN = true;
+
+  // Set QUICKTEST env var to true to run a small subset of tests.
+  // Set SMOKETEST env var to true to run an even smaller subset of tests
+  // set INGRESSPERDOMAIN to false to create LB's ingress by kubectl yaml file
+  static {
+    QUICKTEST =
+        System.getenv("QUICKTEST") != null && System.getenv("QUICKTEST").equalsIgnoreCase("true");
+    SMOKETEST =
+        System.getenv("SMOKETEST") != null && System.getenv("SMOKETEST").equalsIgnoreCase("true");
+    if (SMOKETEST) QUICKTEST = true;
+    if (System.getenv("JENKINS") != null) {
+      JENKINS = new Boolean(System.getenv("JENKINS")).booleanValue();
+    }
+    if (System.getenv("INGRESSPERDOMAIN") != null) {
+      INGRESSPERDOMAIN = new Boolean(System.getenv("INGRESSPERDOMAIN")).booleanValue();
+    }
+  }
 
   public static void initialize(String appPropsFile) throws Exception {
 
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
@@ -609,6 +609,8 @@ public class BaseTest {
       logger.info("Execution of statedump.sh failed, " + result.stderr() + "\n" + result.stdout());
     }
 
+    TestUtils.renewK8sClusterLease(getProjectRoot(), getLeaseId());
+
     if (JENKINS) {
       result = cleanup();
       if (result.exitValue() != 0) {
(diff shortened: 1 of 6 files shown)


integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -95,6 +136,28 @@ public static void initialize(String appPropsFile) throws Exception {
...
+      TestUtils.exec(
+          "/usr/local/packages/aime/ias/run_as_root \"mkdir -p "
+              + pvRoot
+              + "/acceptance_test_pv\"");
+      TestUtils.exec(
+          "/usr/local/packages/aime/ias/run_as_root \"chmod 777 "
+              + pvRoot
+              + "/acceptance_test_pv\"");
+    }
+
```
tbarnes-us: Just an FYI - We'll need a different approach for setting up directories on the new public Jenkins boxes.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -520,4 +583,29 @@ private void callWebAppAndVerifyScaling(Domain domain, int replicas) throws Exce
       TestUtils.checkServiceCreated(podName, domainNS);
     }
   }
+
+  public static void tearDown() throws Exception {
+    StringBuffer cmd =
```
tbarnes-us: Not a new issue.  

It's sometimes confusing that the test cleanup can occur *before* the system reports errors from a failed a test.  So the logging shows a test abruptly ending without explaining why, then a ton of cleanup logging, and only after than an error is reported...   This can get even more confusing if the tear-down itself reports problems tearing down.   I assume this is a side-effect of how jUnit works, but it turns the output of a failed test run into a bit of a jumble.    

To address this, maybe do two things:

(1) add a log message here something like:   "TEARDOWN:  Starting Test Run TearDown (cleanup and state-dump).   Note that if the test failed previous to tearDown, the error that caused the test failure may be reported after the tearDown completes.   Note that tearDown itself may report errors, but this won't affect the outcome of the test results."

(2) Minimize logging during the tear-down (unless errors occur).

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -520,4 +583,29 @@ private void callWebAppAndVerifyScaling(Domain domain, int replicas) throws Exce
...
+    ExecResult result = ExecCommand.exec(cmd.toString());
+    if (result.exitValue() == 0) {
+      logger.info("Executed statedump.sh " + result.stdout());
+    } else {
+      logger.info("Execution of statedump.sh failed, " + result.stderr() + "\n" + result.stdout());
+    }
+
+    // if (JENKINS) {
+    result = cleanup();
+    logger.info("cleanup result =" + result.stdout() + "\n " + result.stderr());
```
tbarnes-us: Make sure statedump and cleanup both try to renew the lease before they start (and exit with a failure if the lease doesn't renew).

tbarnes-us: Or renew the lease right before calling each...

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -520,4 +583,29 @@ private void callWebAppAndVerifyScaling(Domain domain, int replicas) throws Exce
...
+
+    ExecResult result = ExecCommand.exec(cmd.toString());
+    if (result.exitValue() == 0) {
+      logger.info("Executed statedump.sh " + result.stdout());
+    } else {
+      logger.info("Execution of statedump.sh failed, " + result.stderr() + "\n" + result.stdout());
+    }
+
+    // if (JENKINS) {
+    result = cleanup();
```
tbarnes-us: Always calling cleanup on test completing is a relatively big change in some ways.  If nothing else, it adds more time to the end of the run, plus makes it harder to diagnose a failed run.    Is this needed on Wercker and locally?

vanajamukkara: On Wercker, currently only ITOperator class is executed so I don't think its needed. Locally also, I doubt if its needed, I can try by skipping cleanup locally and see.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
rjeberhard: @bsmoyers has already created the new cluster on OCI and created the OCI file system.  This will be available through NFS PV's, but won't be auto-mounted on any of the Wercker nodes.  So, as Tom points out, we'll have to copy files with exec or cp onto a pod that has the proper volume.

vanajamukkara: Sure. we will work out a new approach at that time.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Changes are done

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Done

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Done

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -520,4 +584,41 @@ private void callWebAppAndVerifyScaling(Domain domain, int replicas) throws Exce
...
+            + "after the tearDown completes. Note that tearDown itself may report errors,"
+            + " but this won't affect the outcome of the test results.");
+    StringBuffer cmd =
+        new StringBuffer("export RESULT_ROOT=$RESULT_ROOT && export PV_ROOT=$PV_ROOT && ");
+    cmd.append(BaseTest.getProjectRoot())
+        .append("/integration-tests/src/test/resources/statedump.sh");
+    logger.info("Running " + cmd);
+
+    // renew lease before callin statedump.sh
+    TestUtils.renewK8sClusterLease(getProjectRoot(), getLeaseId());
```
tbarnes-us: If this renew (or _any_ renew) fails, we must exit immediately -- all future results , state dumps, and cleanup are invalid.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
tbarnes-us: Looks good!

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
```diff
@@ -520,4 +584,41 @@ private void callWebAppAndVerifyScaling(Domain domain, int replicas) throws Exce
...
+
+    ExecResult result = ExecCommand.exec(cmd.toString());
+    if (result.exitValue() == 0) {
+      // logger.info("Executed statedump.sh " + result.stdout());
+    } else {
+      logger.info("Execution of statedump.sh failed, " + result.stderr() + "\n" + result.stdout());
+    }
+
+    if (JENKINS) {
+      result = cleanup();
```
tbarnes-us: Unless cleanup() itself calls renew the lease should be renewed just before calling cleanup() - if you're not sure, it wouldn't hurt to call it here regardless...

vanajamukkara: cleanup is called only for Jenkins, currently we don't have lease concept on Jenkins.

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
vanajamukkara: Yes, it exits with Exception.

tbarnes-us: Cool - resolved!

integration-tests/src/test/java/oracle/kubernetes/operator/BaseTest.java
tbarnes-us: ok - the 'new Jenkins' may pick up the leasing concept, so might as well...

vanajamukkara:
@rjeberhard This PR is ready to merge, this will fix the lease failures on wercker.

PR #948, opened 2019-03-18
Cluster scaling

russgold:
This resolves issue https://jira.oraclecorp.com/jira/browse/OWLS-72885

I have refactored access to the service fields in DomainPresenceInfo and ServerKubernetesObjects, so that now all accesses should go through ServiceHelper.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -16,13 +16,11 @@ import io.kubernetes.client.util.Watch;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import oracle.kubernetes.operator.TuningParameters.MainTuning;
 import oracle.kubernetes.operator.calls.CallResponse;
@@ -30,6 +28,7 @@ import oracle.kubernetes.operator.helpers.CallBuilder;
 import oracle.kubernetes.operator.helpers.ConfigMapHelper;
 import oracle.kubernetes.operator.helpers.DomainPresenceInfo;
 import oracle.kubernetes.operator.helpers.JobHelper;
+import oracle.kubernetes.operator.helpers.KubernetesUtils;
 import oracle.kubernetes.operator.helpers.PodHelper;
 import oracle.kubernetes.operator.helpers.ResponseStep;
 import oracle.kubernetes.operator.helpers.ServerKubernetesObjects;
@@ -56,17 +55,13 @@ import oracle.kubernetes.weblogic.domain.model.AdminService;
 import oracle.kubernetes.weblogic.domain.model.Channel;
 import oracle.kubernetes.weblogic.domain.model.Domain;
 import oracle.kubernetes.weblogic.domain.model.DomainSpec;
-import org.joda.time.DateTime;
 
 public class DomainProcessorImpl implements DomainProcessor {
   static final DomainProcessor INSTANCE = new DomainProcessorImpl();
 
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
 
-  private static final FiberGateFactory FACTORY =
-      () -> {
-        return new FiberGate(Main.engine);
-      };
+  private static final FiberGateFactory FACTORY = () -> new FiberGate(Main.engine);
   private static final ConcurrentMap<String, FiberGate> makeRightFiberGates =
       new ConcurrentHashMap<>();
   private static final ConcurrentMap<String, FiberGate> statusFiberGates =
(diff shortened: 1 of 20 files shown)


rjeberhard:
@russgold, is there a test that captures how to handle existing services that don't have the serviceType label?  What will be the behavior?

russgold:
That turns out to be a problem in the current system. What happened in the past is that a service with a cluster name was considered a cluster service, and one with a server name but no cluster name was be treated as a server service. The problem is, some server services _do_ have cluster names. That means that events regarding them have been putting them in the wrong place: in the clusters array rather than in the sko service field.

The new behavior reverses that: an old service with a server name is treated as a server service, while one with a cluster name but no server name is treated as a cluster service. I will add tests to validate that.

But since this version also includes validation by recipe checksum, all services in an existing install will be replaced, and the new ones will have the service type annotation.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
```diff
@@ -71,28 +71,95 @@ private ServerKubernetesObjects getSko(String serverName) {
...
+        clusters,
+        clusterName,
+        s -> !KubernetesUtils.isFirstNewer(getMetadata(s), getMetadata(event)));
+  }
+
+  private static <K, V> boolean removeIfPresentAnd(
+      ConcurrentMap<K, V> map, K key, Predicate<? super V> predicateFunction) {
+    Objects.requireNonNull(predicateFunction);
+    for (V oldValue; (oldValue = map.get(key)) != null; ) {
+      if (!predicateFunction.test(oldValue)) return false;
```
markxnelson: brackets

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
```diff
@@ -71,28 +71,95 @@ private ServerKubernetesObjects getSko(String serverName) {
...
-    getSko(serverName).setExternalService(service);
+    getSko(serverName).getExternalService().set(service);
+  }
+
+  void setExternalServiceFromEvent(String serverName, V1Service event) {
+    getSko(serverName).getExternalService().accumulateAndGet(event, this::getNewerService);
+  }
+
+  boolean deleteExternalServiceFromEvent(String serverName, V1Service event) {
+    if (serverName == null) return false;
```
markxnelson: {}

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -141,4 +142,26 @@ static V1ObjectMeta getResourceMetadata(Object resource) {
...
+
+  /**
+   * Returns true if the first metadata indicates a newer resource than does the second.
+   *
+   * @param first the first item to compare
+   * @param second the second item to compare
+   * @return true if the first object is newer than the second object
+   */
+  public static boolean isFirstNewer(V1ObjectMeta first, V1ObjectMeta second) {
+    if (second == null) return true;
```
markxnelson: {}

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -141,4 +142,26 @@ static V1ObjectMeta getResourceMetadata(Object resource) {
...
+    if (first == null) return false;
+
+    DateTime time1 = first.getCreationTimestamp();
+    DateTime time2 = second.getCreationTimestamp();
+
+    if (time2.isAfter(time1)) {
+      return false;
+    }
+    return Integer.parseInt(second.getResourceVersion())
```
markxnelson: i would prefer this to be written so the intent is clear and i dont have to rtfm to work out what it does

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -220,192 +216,26 @@ public void dispatchPodWatch(Watch.Response<V1Pod> item) {
...
-    for (V oldValue; (oldValue = map.get(key)) != null; ) {
-      if (predicateFunction.test(oldValue)) {
-        if (map.remove(key, oldValue)) {
-          return true;
-        }
-      } else {
-        return false;
-      }
+    switch (item.type) {
+      case "ADDED":
```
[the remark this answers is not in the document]
russgold: Huh. I find this to be a really common case, and clear when there is nothing between the cases.

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
russgold: Agreed, and will rewrite.

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -141,4 +102,33 @@ static V1ObjectMeta getResourceMetadata(Object resource) {
...
+    DateTime time1 = first.getCreationTimestamp();
+    DateTime time2 = second.getCreationTimestamp();
+
+    if (time1.isAfter(time2)) {
+      return true;
+    } else if (time2.isAfter(time1)) {
+      return false;
+    } else {
+      return getResourceVersion(first) > getResourceVersion(second);
+    }
```
tbarnes-us: The original version you had for isNewer compares version numbers even if the timestamp of time2 'isAfter' time1 or time2 equals time1, and returns true if if the second version number is higher.   The new code only compares version numbers if time1 equals time2.   Is this intentional?

In other words, here's the logic of the new code (simplified a little vs. what you have above):

if (time1.isEquals(time2)) {
  return getResourceVersion(first) > getResourceVersion(second);
} else {
  return time1.isAfter(time2);
}

And here's the logic of the old code you had before:

if (time2.isAfter(time1)) {
  return false;
} else if (getResourceVersion(second) < getResourceVersion(first)) {
  return true;
} else {
  return time2.isBefore(time1);
}

russgold: Yes, thank you. That is an improvement on what I had, and does exactly the same thing. Resource versions are just a tie-breaker for when the times are equal.

operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesUtilsTest.java
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Licensed under the Universal Permissive License v 1.0 as shown at
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.helpers;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.*;
```
markxnelson: generally we do not allow wildcard imports, though this particular one is probably ok

PR #951, opened 2019-03-19
configure Voyager to only watch Voyager ingress class

lilyhe123:
By default, Voyager controller monitors both k8s standard Ingress class and Voyager Ingress class. With this change Voyager controller will only monitor Voyager Ingress class.  This can help to run the two supported Ingress Controllers: Voyager and Traefik, on the same k8s cluster at the same time without interfering with each other. Voyager is monitoring its own Voyager Ingress class and Traefik is monitoring k8s-provided Ingress class.

Another change is to fix annotation problem in chart ingress-per-domain.


lilyhe123:
No, we need not. I have tested it. With this change Voyager controller will only watch Voyager Ingress. 
I think what matters is the `apiVersion` in the Ingress yamls. For Voyager Ingress it's `apiVersion: voyager.appscode.com/xxx` while for k8s-provided ingress it's `apiVersion: extensions/xxx`

PR #960, opened 2019-03-20
minor edits

rosemarymarano:
mostly punctuation, one fix (duplicated sentence)


docs-source/content/quickstart/cleanup.md
```diff
@@ -13,7 +13,7 @@ weight: 7
     ```bash
     $ helm delete --purge sample-domain1-ingress
     ```
-1.	Remove the domain resources by using the sample [`delete-weblogic-domain-resources`](http://github.com/oracle/weblogic-kubernetes-operator/blob/master/kubernetes/samples/scripts/delete-domain/delete-weblogic-domain-resources.sh) script.
+1.	Remove the domain resources by using the sample [`delete-weblogic-domain-resources`](http://github.com/oracle/weblogic-kubernetes-operator/blob/master/kubernetes/samples/scripts/delete-domain/delete-weblogic-domain-resources.sh) script:
```
rjeberhard: @markxnelson @rosemarymarano, In the scrum yesterday we discussed that the doc. would use relative paths so that we could copy the doc. set for a particular release...  However, I'm seeing more and more absolute references.  Since we will have a release soon (next couple weeks?), we need to figure out a solution.

markxnelson: these absolute links are to things that are outside the doc @rjeberhard

rjeberhard: Hmm... that's true, but do we need to update these links then to appropriate release branch when it's time: "blob/master" -> "blob/release/v2.1" or whatever?

markxnelson: probably yes - we always did that manually in the past for each release, so i don't think this is any different...

PR #963, opened 2019-03-20
Update documentation on Kubernetes secrets used by the operator

TheFrogPad:
Updates the information about the secrets being used that was noted by Tom B.

### DIFF
--- a/docs-source/content/security/secrets.md
+++ b/docs-source/content/security/secrets.md
@@ -8,15 +8,24 @@ description: "Kubernetes secrets for the WebLogic operator"
 #### Contents
 * [WebLogic domain credentials secret](#weblogic-domain-credentials-secret)
 * [WebLogic domain image pull secret](#weblogic-domain-image-pull-secret)
+* [WebLogic operator configuration override secrets](#weblogic-operator-configuration-override-secrets)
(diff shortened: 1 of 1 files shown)


docs-source/content/security/secrets.md
```diff
@@ -84,7 +106,7 @@ under **Securty**.
 #### WebLogic operator internal REST interface secret
 
 The operator exposes an internal REST HTTPS interface with a self-signed certificate.
-The certificate is kept in a Kubernetes `ConfigMap` with the name `weblogic-operator-cm ` using the key `internalOperatorCert`.
+The certificate is kept in a Kubernetes `ConfigMap` with the name `weblogic-operator-cm` using the key `internalOperatorCert`.
```
tbarnes-us: Do customers need to directly reference this cert?  If not, I wonder if this is an internal implementation detail that could change over time - and maybe it (a) shouldn't be documented, or (b) he wording should include cautions like `the location of this secret may change without notice`.

TheFrogPad: There are examples for making a REST call to the operator using the internal port which get this cert for the trust setup, the main example is the WLDF action scaling up the cluster.

I doubt we'd ever change the handling for internal cert and if we did then we should make the change compatible and/or document clearly.

PR #973, opened 2019-03-24
Promote new docs

markxnelson:
This PR will promote the new documentation to be the main documentation for the operator. 
In particular:
- the main readme will now point to the new docs
- the 2.0.1 "site" directory is archived as per previous releases
- links are fixed in the old 2.0.1 docs so that they still work
In addition to this PR, it should be noted that there is already a Jenkins job in place that detects changes to the `docs-source` directory and automatically builds and deploys the updated doc site as needed.  The main build job is configured to ignore those commits to prevent unnecessary resource usage running integration tests.


README.md
```diff
@@ -35,7 +35,7 @@ This release was published on March 4, 2019.
 
 Documentation for the operation is available [here](https://oracle.github.io/weblogic-kubernetes-operator) and includes
```
rosemarymarano: Documentation for the operation -> Documentation for the operator

PR #974, opened 2019-03-25
fix capitalization of 'Docker' in example commands

markxnelson:
This PR fixes the capitalization of "Docker" in example commands where it should be `docker` not `Docker`


markxnelson:
my fault for trusting find & replace !!

PR #981, opened 2019-03-26
Rise max_wait_secs in startNodeManager.sh

flistell:
15 seconds may be ta  too tight timeout. In some cases we saw the pod restarting several times until it can starts the nodemanager in the expected interval. We propose an arbitrary value of 60 seconds.

### DIFF
--- a/operator/src/main/resources/scripts/startNodeManager.sh
+++ b/operator/src/main/resources/scripts/startNodeManager.sh
@@ -290,7 +290,7 @@ ${stm_script} > ${nodemgr_out_file} 2>&1 &
 
 wait_count=0
 start_secs=$SECONDS
-max_wait_secs=15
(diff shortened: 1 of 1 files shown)


rjeberhard:
@tbarnes-us, what do you think?  Is it reasonable to just bump the arbitrary value from 15 to 60 seconds or should we find a way to make this value configurable?

tbarnes-us:
@rjeberhard 60 seconds - or even a bit higher - is fine by me.   It'd be nice to make it configurable - maybe simply as an undocumented env var (eg just use ${NODE_MANAGER_TIMEOUT_SECS:-60} in the script).

rjeberhard:
We put in a version similar to what Tom suggested with the updated default and the ability to override with an environment variable.

PR #989, opened 2019-03-28
WIP - OWLS-73066 - - WLDF SCALING IS NOT WORKING ON OPERATOR JENKINS RUN

alai8:
Get WLDF scaling test to pass in jenkins by updating domain crd version in k8s rest api call in scalingAction.sh

Passed jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1382/

### DIFF
--- a/src/scripts/scaling/scalingAction.sh
+++ b/src/scripts/scaling/scalingAction.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright 2017, 2018, Oracle Corporation and/or its affiliates.  All rights reserved.
+# Copyright 2017, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
 echo "called scalingAction.sh" >> scalingAction.log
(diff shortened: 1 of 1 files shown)


src/scripts/scaling/scalingAction.sh
```diff
@@ -118,7 +118,7 @@ port=`echo ${STATUS} | python cmds.py`
 echo "port: $port" >> scalingAction.log
 
 # Reteive Custom Resource Domain 
-DOMAIN=`curl -v --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" $kubernetes_master/apis/weblogic.oracle/v2/namespaces/$wls_domain_namespace/domains/$domain_uid`
+DOMAIN=`curl -v --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" $kubernetes_master/apis/weblogic.oracle/v3alpha1/namespaces/$wls_domain_namespace/domains/$domain_uid`
```
lennyphan: wonder if we should parameterize the version so we don't have to keep updating the script?  can we retrieve the version?

rjeberhard: We'll need to parameterize and pass in this value as the version will change depending on the domain.  For instance, if the customer had a domain resource with version "v2" and upgraded to 2.1, then the domain version would still be "v2".  The 2.1 operator will support both "v2" and "v3" (currently, "v3alpha1").

rjeberhard: The value we need here is metadata.selfLink from the domain resource

src/scripts/scaling/scalingAction.sh
```diff
@@ -118,7 +118,7 @@ port=`echo ${STATUS} | python cmds.py`
 echo "port: $port" >> scalingAction.log
 
 # Reteive Custom Resource Domain 
-DOMAIN=`curl -v --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" $kubernetes_master/apis/weblogic.oracle/v2/namespaces/$wls_domain_namespace/domains/$domain_uid`
```
markxnelson: you also need to update the copyright notice please

alai8: Thanks. Missed that. Updated now.

src/scripts/scaling/scalingAction.sh
alai8: Updated script to query the domain CRD version to be used in the "retrieve custom resource domain" API call.

rjeberhard: Based on my testing, the implementation you picked, @alai8, is the correct one.  Nice work!

PR #1021, opened 2019-04-10
Search for free ports for REST unit test

rjeberhard:
We've had some test failures where the unit test ports for the REST unit tests conflict.  We've tried random ports, but even these sometimes conflict.  Now, I propose that we explicitly search for and choose free ports.


russgold:
What exactly are we testing, here? The REST client? The services? Maybe we can avoid any socket I/O at all. That would not only make the tests more reliable; it would greatly speed them up.

rjeberhard:
This worked locally but didn't on Wercker so, I'd prefer your help to have no socket IO at all.  I didn't write these REST server unit tests.  They were using hardcoded ports, then we tried random ports, but even so we still see common failures on Jenkins about ports in use.

PR #1027, opened 2019-04-11
WIP: NEXT version of ci/cd documentation

markxnelson:
This PR adds an initial set of CI/CD documentation that forms a basis which we can then iterate on and improve.


docs-source/content/userguide/cicd/_index.md
```diff
@@ -0,0 +1,64 @@
...
+[Server JRE image](https://hub.docker.com/_/oracle-serverjre-8), for example.  On 
+top of this, you need the WebLogic Server binaries (the "Oracle Home").  On top 
+of that, you may wish to have some patches or updates installed.  And then 
+you need your domain, that is the configuration.
+
+There is also other information associated with a domain that needs to live 
+somewhere, for example leasing tables, message and transaction stores, and so 
+on.  We recommend that all of these be kept in a database so that they are 
+separate from the Docker image and so that we get all of the management 
+capabilities offered by a database server.
```
tbarnes-us: Also to take advantage of built-in database server HA, and the fact that disaster recovery of sites across all but the shortest distances almost always requires using a single database server to consolidate and replicated data (DataGuard).

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
...
+  domain layer.  You will need to determine if you need to keep the same domain encryption keys.
+
+The diagram below summarizes these concerns in a decision tree for the “domain in image” case:
+
+![Decision model for the "domain in image" approach](/weblogic-kubernetes-operator/images/flowchart.png)
+
+If you are using the "domain on persistent storage" approach, many of these concerns become 
+moot, because you have an effective separation between your domain and the Docker image.
+There is still the possibility that an update in the Docker image could affect your domain, 
+for example, if you updated the JDK, you may need to update some of your domain scripts
```
mriccell: In a Docker  image the JDK path is always the same location.

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+draft: false
+weight: 5
+description: "How to copy domains."
+---
+
+The recommended approach to save a copy of a domain is to simply zip (or tar) 
+the domain directory.  However there is a very important caveat with this 
+recommendation - when you unzip the domain, it must go back into exactly 
+the same location (Domain Home) in the (new) file system.  Using this 
+approach will maintain the same domain encryption key.
```
mriccell: In a Docker image the domain home is always the same as long as the domain name is the same.

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+
+Every time you run your CI/CD pipeline to create a new mutation of the domain, 
+it should retrieve and unzip the primordial domain first, and then apply changes 
+to that domain using tools like WDT or WLST (see below). 
+
+
+> Always use external state.
+
+You should always keep state outside the Docker image.  This means that you should 
+use JDBC stores for leasing tables, JMS and Transaction stores, HTTP sessions,
```
mriccell: Not HTTP sessions those are kept in memory

docs-source/content/userguide/cicd/why-layering-matters.md
```diff
@@ -0,0 +1,101 @@
...
+
+If we just want to update the domain configuration itself, that is the top layer, 
+then it is pretty easy.  We can make the necessary changes and save a new version 
+of that layer, and then roll the domain.  We could also choose to just build another 
+layer on top of the existing top layer that contains our delta.  If the change is 
+small, then we will just end up with another small layer, and as we have seen, 
+the small layers are no problem. 
+
+But consider a more complicated scenario - let's take updating the JDK as an example 
+to understand the impact of layers.  Say we want to update from JDK 11.0.2 to 11.0.4
```
mriccell: Instead of using JDK 11 in the example use JDK 8

mriccell: 8u201 and 8u202 can be an example

docs-source/content/userguide/cicd/choose-an-approach.md
markxnelson: no it's not, i checked

tbarnes-us:
About docs-source/static/images/flowchart.png - the flowchart should ask the question about rolling upgrades _before_ directing folks down the config overrides path, since config overrides don't currently support rolling upgrades...

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+Let's review what we have discussed and talk about when we might want to use 
+various approaches.  We can start by asking ourselves questions like these:
+
+
+- *Can you make the desired change with a configuration override?*  
+  The WebLogic Kubernetes Operator allows you to inject a number of [configuration 
+  overrides]({{< relref "/userguide/managing-domains/configoverrides/_index.md" >}}) 
+  into your pods at startup time.  This allows you to use the same image for multiple
```
tbarnes-us: at startup time. --> before starting any servers in the domain.   (E.g. try make it clear taht it's not something useful for rolling upgrades -- or maybe just flat out state that its not suited for rolling upgrades.)

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
...
+  instead of building a new image.  
+- *Are you only changing the WebLogic configuration, i.e. deploying or updating an 
+  application, changing a resource configuration in a way that is not supported by 
+  configuration overrides, etc.?*  
+  If your changes fit into this category, and you have used the "domain-in-image" 
+  approach and the Docker layering model, then you only need to update the top layer 
+  of your image.  This is relatively easy compared to making changes in lower layers.
+  You could create a new layer with the changes, or your could rebuild/replace the 
+  existing top layer with a new one.  Which approach you choose depends most on 
+  whether you need to maintain the same domain encryption keys or not.
```
tbarnes-us: Maybe add something like:    There's also trade-off in coming up with edits to the current configuration via a custom additive WDT/off-line script, rather than rerunning an changed version of the original WDT/off-line script.   The latter tends to be easier, but see the next bullet item before you consider it.

docs-source/content/userguide/cicd/why-layering-matters.md
```diff
@@ -0,0 +1,101 @@
...
+
+If we just want to update the domain configuration itself, that is the top layer,
+then it is pretty easy.  We can make the necessary changes and save a new version
+of that layer, and then roll the domain.  We could also choose to just build another
+layer on top of the existing top layer that contains our delta.  If the change is
+small, then we will just end up with another small layer, and as we have seen,
+the small layers are no problem.
+
+But consider a more complicated scenario - let's take updating the JDK as an example
+to understand the impact of layers.  Say we want to update from JDK 11.0.2 to 11.0.4
```
ddsharpe: Why wouldn't we use an example that DOES work with 12.2.1.3?

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
...
+  they do not require rebuilding the Docker image.  If all of your changes fit into
+  this category, it is probably much better to just use configuration overrides
+  instead of building a new image.  
+- *Are you only changing the WebLogic configuration, for example, deploying or updating an
+  application, changing a resource configuration in a way that is not supported by
+  configuration overrides, and such?*  
+  If your changes fit into this category, and you have used the "domain-in-image"
+  approach and the Docker layering model, then you only need to update the top layer
+  of your image.  This is relatively easy compared to making changes in lower layers.
+  You could create a new layer with the changes, or your could rebuild/replace the
```
lfeigen: Your could -> you could

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
...
+  existing top layer with a new one.  Which approach you choose depends mainly on
+  whether you need to maintain the same domain encryption keys or not.
+- *Do you need to be able to do a rolling restart?*
+  If you need to do a rolling restart, for example to maintain the availability of
+  your applications, then you need to make sure the new domain layer has the same
+  domain encryption keys.  You cannot perform a rolling restart of a domain if the
+  new members have a different encryption key.
```
lfeigen: I'm not sure that it's just rolling restart that's an issue.  We don't know where other layers may use encryption keys in, say, MDS.  I think I'd rather say something more conservative about encryption keys.

docs-source/content/userguide/cicd/_index.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "CI/CD"
```
TheFrogPad: Reader view comment: When you look at this in context of the website, the title name seem like it should be more along the line of  "CI/CD Processing" to parallel topics at the same level of managing domains, managing operators, etc.

docs-source/content/userguide/cicd/why-layering-matters.md
```diff
@@ -0,0 +1,101 @@
+---
+title: "Why layering matters"
```
TheFrogPad: Reader view comment: Use "Why image layering matters"?

docs-source/content/userguide/cicd/choose-an-approach.md
```diff
@@ -0,0 +1,59 @@
+---
+title: "Choose an approach"
```
TheFrogPad: Reader view comment: Use "Choosing an approach"?

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
+---
+title: "How to copy domains"
```
TheFrogPad: Would this be more appropriate to call "How to create primordial domain layer" along with a description to match?

docs-source/content/userguide/cicd/_index.md
```diff
@@ -0,0 +1,65 @@
...
+in the Docker image, but the domain is kept on a separate persistent storage
+outside of the image.  The second approach puts the JDK, WebLogic binaries
+and the domain all in the Docker image.  Both of these approaches are perfectly
+valid (and fully supported) and they have various advantages and disadvantages.
+
+We have listed the [relative advantages of these two approaches here]({{< relref "/userguide/managing-domains/choosing-a-model/_index.md" >}}).
+
+One of the key differences between these approaches is how many Docker images
+you have, and therefore, how you build and maintain them - your image CI/CD
+process.  Let's take a short detour and talk about Docker image layering.
```
TheFrogPad: Reader view comment: Given that you have the macro to expand the topics below the main title, this wording looks out of place:

" ... how you build and maintain them - your image CI/CD process.  Let's take a short detour and talk about Docker image layering."

Simple suggestion to would seem to flow better is:

" ... how you build and maintain them - your image CI/CD process.  Let's take a walk through the various topics below and start with a short detour about Docker image layering."

docs-source/content/userguide/cicd/_index.md
```diff
@@ -0,0 +1,65 @@
...
+server to consolidate and replicated data (DataGuard).
+
+There are two common approaches on how to structure these components.  The first,
+which we call "domain on persistent volume," places the JDK and WebLogic binaries
+in the Docker image, but the domain is kept on a separate persistent storage
+outside of the image.  The second approach puts the JDK, WebLogic binaries
+and the domain all in the Docker image.  Both of these approaches are perfectly
+valid (and fully supported) and they have various advantages and disadvantages.
+
+We have listed the [relative advantages of these two approaches here]({{< relref "/userguide/managing-domains/choosing-a-model/_index.md" >}}).
```
TheFrogPad: Suggest remove "here" from the link and have the text read: 

"We have listed the relative advantages of these two approaches under managing domains."

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+> The domain ZIP must be created before starting servers.  
+
+When servers are started the first time, they will encrypt various other data.
+Make sure you create the ZIP file before starting servers for the first time.
+The primordial domain ZIP file should be stored in a safe place where the CI/CD
+can get it when needed, for example in a secured Artifactory repository (or
+something similar).  Remember, anyone who gets access to this ZIP file can get access
+to the domain encryption key, so it needs to be protected appropriately.
```
TheFrogPad: Suggest a note or warning block with existing text about protected appropriately.

docs-source/content/userguide/cicd/why-layering-matters.md
mriccell: Because JDK 11 does not work with 12.2.1.3 only JDK 8.  It is 19.2 (or whatever it will be called) that works with JDK 11.  This is confusing to customers.

docs-source/content/userguide/cicd/choose-an-approach.md
rosemarymarano: We've been using imperative verbs versus gerunds, therefore, "Choose an approach" is correct.

docs-source/content/userguide/cicd/choose-an-approach.md
mriccell: The WebLogic image always extends the server jre image which sets the Java Home.  This path will always be the same in the images. JAVA_HOME=/usr/java/default

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+---
+
+The recommended approach to save a copy of a domain is to simply ZIP (or tar)
+the domain directory.  However, there is a very important caveat with this
+recommendation - when you unzip the domain, it must go back into exactly
+the same location (Domain Home) in the (new) file system.  Using this
+approach will maintain the same domain encryption key.  
+
+The best practice/recommended approach is to create a "primordial domain"
+(as described previously) which does not contain any applications or resources,
```
lilyhe123: "as described previously"? Where? I didn't see any in previous document.

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+When servers are started the first time, they will encrypt various other data.
+Make sure you create the ZIP file before starting servers for the first time.
+The primordial domain ZIP file should be stored in a safe place where the CI/CD
+can get it when needed, for example in a secured Artifactory repository (or
+something similar).  Remember, anyone who gets access to this ZIP file can get access
+to the domain encryption key, so it needs to be protected appropriately.
+
+Every time you run your CI/CD pipeline to create a new mutation of the domain,
+it should retrieve and unzip the primordial domain first, and then apply changes
+to that domain using tools like WDT or WLST (see below).
```
lilyhe123: You may want to add link to tools.md instead of "see below".

docs-source/content/userguide/cicd/mutate-the-domain-layer.md
```diff
@@ -0,0 +1,43 @@
...
+draft: false
+weight: 4
+description: "How to mutate the domain layer."
+---
+
+If you need to mutate the domain layer, and keep the same domain encryption keys,
+then there are some choices about how to implement that, as alluded to previously.
+Let's explore those in some more detail now.
+
+The first option is to implement each mutation as a delta to the previous state.
```
lilyhe123: It's not clear to me how to implement each mutation as a delta to the previous state. It will be helpful if you provide concrete simple samples.

docs-source/content/userguide/cicd/_index.md
```diff
@@ -0,0 +1,65 @@
...
+you need your domain, that is the configuration.
+
+There is also other information associated with a domain that needs to live
+somewhere, for example leasing tables, message and transaction stores, and so
+on.  We recommend that these be kept in a database to take advantage of built-in 
+database server HA, and the fact that disaster recovery of sites across all 
+but the shortest distances almost always requires using a single database 
+server to consolidate and replicated data (DataGuard).
+
+There are two common approaches on how to structure these components.  The first,
```
lilyhe123: It just occur to me that is it might be helpful to introduce a third approach: a hybrid of domaim-on-a-pv and domain-in-image. 
1. Create and update a domain to a pv, probably via `docker run`.
1. When come to build the domain image, just copy the domainHome from a PV to the image.

In this way we can keep all the benefits of domain-in-image while get rid of regenerating the domain when rebuild the domain image.
With this approach we can mutate the domain layer while keep the same domain encryption keys.

markxnelson: good idea @lilyhe123 - @rjeberhard  and I were actually discussing the same thing last week, and I think @rjeberhard  is working on testing it.  maybe you could help to test it?

mriccell: I am not sure customers will be happy with this option since they will need to maintain both domain image and PV.   The second problem with this approach is that it goes against Docker best practices.   I would prefer to spend the effort in enhancing WebLogic to export the key so that then a new domain image can be create using the same key.

TheFrogPad: Not clear what implications this would have since you are taking the domain and trying to copy that domain at various snapshots, are the server's booting while on PV?

lilyhe123: @markxnelson I'm sure basic thing should work with these hybrid approach but we need to brainstorm the downsides of it. Monica already have one.

@mriccell I agree that this is only a workaround. If we can enhance Weblogic to regenerate a domain without a key. That will be great!  But what Docker best practice you mean here?

@TheFrogPad the copy is done when creating the image, not in the running container.

mriccell: @lilyhe123 maybe I misunderstood "copy the domainHome from a PV to the image" do you mean build the image by doing a copy of the domain from the PV in the Dockerfile?

mriccell: I still not thrilled with this 3rd option just the fact of having to create both a domain inside of the image and then have to create the PV and persist the domain to a PV.  I think just archiving the domain, applications and then creating a new image from the archived domain is less steps and simpler for the user until we have the WebLogic enhancement.

TheFrogPad: @lilyhe123 @mriccell What is meant "regenerate domain without a key"?

Two WLS domains are not the same from an administrative and security perspective, I do no think anyone would expect that the administrator of another domain could simply come to your domain and make changes without valid credentials in your domain and one would certainly not expect that a server from another domain would join your domain unexpectedly as well?

Perhaps like Monica I'm unclear of the benefit of a PV vs creating an archive with the primordial domain that is re-used as application updates are made?

lilyhe123: Correction: "regenerate a domain without changing the encryption key".

mriccell: wait/defer

docs-source/content/userguide/cicd/mutate-the-domain-layer.md
mriccell: wait/defer

docs-source/content/userguide/cicd/how-to-copy-domains.md
mriccell: wait/defer

docs-source/content/userguide/cicd/_index.md
lilyhe123: If you talk about some manual steps to avoid regenerating the domain and keep the same keys, then any approach that can achieve it is fine, e.g. archive and recreate. 

Just one thing I want to add. Because of the nature of docker image layers, if the lower image is updated all the upper images that depend on it need to be rebuilt. Ideally the CI/CD system needs to find a way to do this **automatically**. That means the same Dockerfile need to be run when the image is first created and any time need to do rebuilding.

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+description: "How to copy domains."
+---
+
+The recommended approach to save a copy of a domain is to simply ZIP (or tar)
+the domain directory.  However, there is a very important caveat with this
+recommendation - when you unzip the domain, it must go back into exactly
+the same location (Domain Home) in the (new) file system.  Using this
+approach will maintain the same domain encryption key.  
+
+The best practice/recommended approach is to create a "primordial domain"
```
lilyhe123: Do you mean creating a "primordial domain" can bypass the limitation that domainHome directory need to be the same?

docs-source/content/userguide/cicd/how-to-copy-domains.md
```diff
@@ -0,0 +1,39 @@
...
+(as described previously) which does not contain any applications or resources,
+and to create a ZIP file of this domain before starting any servers.  
+
+
+> The domain ZIP must be created before starting servers.  
+
+When servers are started the first time, they will encrypt various other data.
```
lilyhe123: Why does this matter since the same encryption keys are used?

docs-source/content/userguide/cicd/how-to-copy-domains.md
mriccell: No what Mark is describing is that primordial domain must go in the same domain home.

lilyhe123: Then what's the benefits of using "primordial domain"? I didn't get it.

docs-source/content/userguide/cicd/_index.md
markxnelson: done

docs-source/content/userguide/cicd/why-layering-matters.md
markxnelson: done

docs-source/content/userguide/cicd/choose-an-approach.md
markxnelson: done

docs-source/content/userguide/cicd/how-to-copy-domains.md
markxnelson: fixed

docs-source/content/userguide/cicd/why-layering-matters.md
markxnelson: fixed

docs-source/content/userguide/cicd/choose-an-approach.md
markxnelson: fixed

docs-source/content/userguide/cicd/_index.md
markxnelson: done

docs-source/content/userguide/cicd/choose-an-approach.md
markxnelson: aha!!

docs-source/content/userguide/cicd/how-to-copy-domains.md
markxnelson: I am trying to keep the titles as brief as possible so the table of contents is not full of long ugly things

docs-source/content/userguide/cicd/how-to-copy-domains.md
markxnelson: done

docs-source/content/userguide/cicd/how-to-copy-domains.md
markxnelson: fixed

docs-source/content/userguide/cicd/how-to-copy-domains.md
markxnelson: fixed

tbarnes-us:
There are a couple more caveats that you might want to mention somewhere - forgive me if you already mentioned them - too lazy to check, but I'm somewhat sure you haven't.  

(1) If you're making a change for which there's already an override then there's no way to support rolling.  The override will of course continue to have precedence.  

(2) We don't support rolling upgrade and/or dynamic changes to network/cluster membership/cluster-size/log-dirs  - the mnain reason is that the operator gathers this information once, before the domain starts, and needs it to setup services and the like - the operator doesn't detect any subsequent dchanges.

tbarnes-us:
It could be nice to explain what CI/CD is an abbreviation for somewhere in the introductory pieces/parts ('Overview' or 'Review of the problem space').

tbarnes-us:
The flowchart is hard to read in the deployed documentation:   https://oracle.github.io/weblogic-kubernetes-operator/userguide/cicd/choose-an-approach/   .  At least on my browser it looks like it was shrunk slightly - enough that the words become a bit of a guessing game.

bsmoyers:

rjeberhard:
Closing these older pull requests.

PR #1030, opened 2019-04-12
Graceful shutdown

rjeberhard:
Graceful shutdown options (timeout, ignore sessions) can be controlled either by domain resource content or by environment variables.  New documentation added to server lifecycle to describe the shutdown of servers.

Full test suite is clean: http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1509/


docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -126,18 +139,81 @@ However, sometimes some of the Managed Servers are different (for example, suppo
...
+    managedServers:
+    - serverName: "cluster1_server1"
+      serverPod:
+        shutdown:
+          timeoutSeconds: 60
+          ignoreSessions: false
+    ...
+```
+
+Graceful shutdown is used for all servers in the domain since this is specified at the domain level and is not overridden at
```
rosemarymarano: since  -> because

docs/domains/Domain.md
```diff
@@ -157,6 +158,16 @@ ServerPod describes the configuration for a Kubernetes pod for a server.
 | periodSeconds | number | The number of seconds between checks |
 | timeoutSeconds | number | The number of seconds with no response that indicates a failure |
 
+### Shutdown
+
+Shutdown describes the configuration for shutting down a server instance.
+
+| Name | Type | Description |
+| --- | --- | --- |
+| ignoreSessions | boolean | For graceful shutdown only, indicates to ignore pending HTTP sessions during in-flight work handling. Not required. Defaults to false. |
```
rosemarymarano: ignoreSessions -> `ignoreSessions`
boolean -> Boolean
shutdownType -> `shutdownType`
timeoutSeconds -> `timeoutSeconds`

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -126,18 +139,81 @@ However, sometimes some of the Managed Servers are different (for example, suppo
...
+
+#### Shutdown environment variables
+
+The operator runtime configures shutdown behavior with the use of the following environment variables. Customers may
+instead simply configure these environment variables directly.  When a customer configured environment variable is present,
+the operator will not override the environment variable based on the shutdown configuration.
+
+| Environment Variables | Default Value | Supported Values |
+| --- | --- | --- |
+| `SHUTDOWN_FORCED` | true | true or false |
```
tbarnes-us: 1) Modify the doc here to note here that forced == false implies gracefull.
2) Consider using an enumerated type for SHUTDOWN type instead of true/false - this (a) would align with the domain resource configuration, and (b) would easily allow more options to be added in the future if more are ever needed...

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -126,18 +139,81 @@ However, sometimes some of the Managed Servers are different (for example, suppo
...
+| `SHUTDOWN_TIMEOUT` | 30 | Whole number in seconds where 0 means no timeout |
+| `SHUTDOWN_IGNORE_SESSIONS` | false | true or false |
```
tbarnes-us: Indicate here that 'ignore sessions' is for graceful shutdown only and also its purpose...

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -126,18 +139,81 @@ However, sometimes some of the Managed Servers are different (for example, suppo
...
+### Shutdown options
+
+The domain resource includes the element `serverPod` that is available under `spec`, `adminServer` and each entry of 
+`clusters` and `managedServers`. The `serverPod` element controls many details of how pods are created for server instances.
+
+The `shutdown` element of `serverPod` controls how servers will be shutdown.  This element has three properties:
+`shutdownType`, `timeoutSeconds`, and `ignoreSessions`.  The `shutdownType` property can be set to either `Graceful` 
+or `Forced` specifying the type of shutdown.  The `timeoutSeconds` property configures how long the server is given to 
+complete shutdown before the server is killed.  The `ignoreSessions` property when "false" allows the shutdown
+process to take longer to give time for any active sessions to complete up to the configured timeout.
```
tbarnes-us: note here that 'ignoreSessions' is applicable to graceful shutdown only...

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -126,18 +139,81 @@ However, sometimes some of the Managed Servers are different (for example, suppo
...
 {{%/ notice %}}
 
+### Shutdown options
+
+The domain resource includes the element `serverPod` that is available under `spec`, `adminServer` and each entry of 
+`clusters` and `managedServers`. The `serverPod` element controls many details of how pods are created for server instances.
+
+The `shutdown` element of `serverPod` controls how servers will be shutdown.  This element has three properties:
+`shutdownType`, `timeoutSeconds`, and `ignoreSessions`.  The `shutdownType` property can be set to either `Graceful` 
+or `Forced` specifying the type of shutdown.  The `timeoutSeconds` property configures how long the server is given to
```
tbarnes-us: Note here that 'Graceful' is the default.

docs/domains/Domain.md
```diff
@@ -157,6 +158,16 @@ ServerPod describes the configuration for a Kubernetes pod for a server.
...
+| shutdownType | string | Tells the operator how to shutdown server instances. Not required. Defaults to graceful shutdown. |
+| timeoutSeconds | number | For graceful shutdown only, number of seconds to wait before aborting in-flight work and shutting down the server. Not required. Defaults to 30 seconds. |
```
tbarnes-us: This documentation contradicts documentation elsewhere in this check-in.  Here it's written that timeoutSeconds controls the time until in-flight work is quiesced during graceful shutdown, elsewhere it's documented that it controls when a graceful/forced shutdown is aborted via a kill-9.  

Is an additional configurable needed?

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
tbarnes-us: NOte here the default for ignoreSessions

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -32,10 +32,13 @@
 import oracle.kubernetes.operator.work.Packet;
 import oracle.kubernetes.operator.work.Step;
 import oracle.kubernetes.weblogic.domain.model.ServerSpec;
+import oracle.kubernetes.weblogic.domain.model.Shutdown;
 
 public class PodHelper {
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
 
+  private static final int DEFAULT_ADDITIONAL_DELETE_TIME = 10;
```
tbarnes-us: Why the 10 second fudge factor?

tbarnes-us: Would this cause a shutdown timeout of 1 second to actually become 11 seconds?  This seems significant for folks like me that are often just trying to get rid of a domain as fast as possible

operator/src/main/java/oracle/kubernetes/operator/http/HttpClient.java
```diff
@@ -249,44 +248,54 @@ public static HttpClient createAuthenticatedClient(final byte[] username, final
...
+            }
+          }
+          if (port == -1) {
+            return null;
+          }
+        } else {
+          port = spec.getPorts().iterator().next().getPort();
+        }
         portalIP += ":" + port;
         String serviceURL = HTTP_PROTOCOL + portalIP;
```
tbarnes-us: Just an FYI:  it looks like HTTP_PROTOCOL is still hard coded, I think there's a JIRA out there somewhere to fix this in case it needs to be SSL instead of plain-text.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
```diff
@@ -156,6 +157,52 @@ public boolean isAdminPortEnabled() {
...
+    if (adminProtocolChannel == null) {
+      if (adminPort != null) {
+        adminProtocolChannel = "default-admin";
+      } else if (sslListenPort != null) {
+        adminProtocolChannel = "default-secure";
+      } else if (listenPort != null) {
+        adminProtocolChannel = "default";
+      }
+    }
+
```
tbarnes-us: Cool: you're figuring out the 'true' admin port here.   I think one addition is needed - I suspect the admin protocol needs to be deduced too.   Default-admin and default-secure requires https instead of http.   I assume admin nap does too - but not sure.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsDomainConfig.java
```diff
@@ -510,22 +513,48 @@ public WlsDomainConfig withCluster(WlsClusterConfig clusterConfig) {
...
-        + ", wlsMachineConfigs="
-        + wlsMachineConfigs
-        + '}';
+    return new ToStringBuilder(this)
+        .append("name", name)
+        .append("adminServerName", adminServerName)
+        .append("configuredClusters", configuredClusters)
+        .append("servers", servers)
+        .append("serverTemplates", serverTemplates)
+        .append("wlsMachineConfigs", wlsMachineConfigs)
```
tbarnes-us: FYI - I suspect wlsMachineConfigs is a dead code path - the operator now gets config from the introspector, and I don't think the introspector returns machine configs...

operator/src/main/java/oracle/kubernetes/operator/http/HttpClient.java
tbarnes-us: Looking at your other changes, I suspect this needs to be fixed.  The current update detects admin port/channel/default-secure channel should be used, and I suspect all of these can require https.

operator/src/main/resources/scripts/stop-server.py
```diff
@@ -24,6 +24,13 @@ def getEnvVar(var):
 domain_name = getEnvVar('DOMAIN_NAME')
 domain_path = getEnvVar('DOMAIN_HOME')
 service_name = getEnvVar('SERVICE_NAME')
+local_admin_port = sys.argv[1]
+local_admin_protocol = sys.argv[2]
+timeout = sys.argv[3]
+ignore_sessions = sys.argv[4]
+force = sys.argv[5]
+
```
tbarnes-us: Could these be passed as env vars?  IMO, it's confusing to pass values in both as env vars and as command line args - and now this script is doing both, plus env vars are easier to maintain over time...

operator/src/main/resources/scripts/stop-server.py
```diff
@@ -36,29 +43,42 @@ def getEnvVar(var):
...
 
 try:
-  nmConnect(userConfigFile='/weblogic-operator/introspector/userConfigNodeManager.secure',
+  connect(userConfigFile='/weblogic-operator/introspector/userConfigNodeManager.secure',
             userKeyFile='/tmp/userKeyNodeManager.secure.bin',
-            host='127.0.0.1',port='5556',
+            url=connect_url,
             domainName=domain_name,
             domainDir=domain_path,
             nmType='plain')
```
tbarnes-us: Ideally, the on-line call could be done via REST instead of a shutdown script.  This would eliminate the need to wait for WLST to spin up during shutdown.

operator/src/main/resources/scripts/stopServer.sh
```diff
@@ -22,18 +22,66 @@ if [ "${MOCK_WLS}" == 'true' ]; then
...
 fi
 
+# stop node manager process
+#
+trace "Stopping NodeManager" &>> /u01/oracle/stopserver.out
+pid=$(jps | grep "NodeManager" | awk '{print $1}')
+trace "PID=[${pid}]" &>> /u01/oracle/stopserver.out
+if [ ! -z $pid ]; then
+  echo "Killing NodeManager process $pid" &>> /u01/oracle/stopserver.out
+  kill -15 $pid
```
tbarnes-us: Maybe this can be just a -9 - it's just the NM.

operator/src/main/resources/scripts/stopServer.sh
```diff
@@ -22,18 +22,66 @@ if [ "${MOCK_WLS}" == 'true' ]; then
...
+trace "After stop-server.py" &>> /u01/oracle/stopserver.out
+
+# at this point node manager should have terminated the server
+# but let's try looking for the server process and
+# kill the server if the process still exists,
+# just in case we failed to stop it via wlst
+pid=$(jps -v | grep "[D]weblogic.Name=${SERVER_NAME}" | awk '{print $1}')
+if [ ! -z $pid ]; then
+  echo "Killing the server process $pid" &>> /u01/oracle/stopserver.out
+  kill -15 $pid
```
tbarnes-us: might be nice to note here that '-15' initiates forced shutdown (e.g. it's not as abrupt as a -9).

docs/domains/Domain.md
rjeberhard: I'll see if I can fix this... This is generated content :)

rosemarymarano: Not a big deal.

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -159,25 +159,25 @@ The domain resource includes the element `serverPod` that is available under `sp
...
+the default, allows the shutdown process to take longer to give time for any active sessions to complete up to the configured timeout.
 The operator runtime monitors this property but will not restart any server pods solely to adjust the shutdown options.
 Instead, server pods created or restarted because of another property change will be configured to shutdown, at the appropriate
 time, using the shutdown options set when the server pod is created.
 
 #### Shutdown environment variables
 
-The operator runtime configures shutdown behavior with the use of the following environment variables. Customers may
+The operator runtime configures shutdown behavior with the use of the following environment variables. Users may
 instead simply configure these environment variables directly.  When a customer configured environment variable is present,
```
rosemarymarano: When a customer configured environment variable -> When a user-configured environment variable (change customer to user, globally)

docs/domains/Domain.md
rjeberhard: https://github.com/oracle/weblogic-kubernetes-operator/pull/1032 😄

tbarnes-us:
@rjeberhard I see many comments marked as 'resolved' without an accompanying update or a reason for skipping the update. Do you have changes pending that haven't been pushed?

tbarnes-us:
Just recording the issue we discussed in slack (which can wait for another pull):   according to Brian Dueck, pods may delay shutdown even after WL itself exits since the main container thread waits on a 'tail' or a 'sleep' command.  If this is true, then something will need to interrupt the tail and/or sleep to speedup pod exit; to implement, perhaps the start script could write it's pid to a file somewhere so that the the stop script can kill the pid once it thinks the server pid has gone away...

operator/src/main/resources/scripts/stop-server.py
rjeberhard: @tbarnes-us, I'd like to punt on this now if you will agree.  I have some code that makes the REST call and it mostly works, but I'm seeing some hard to diagnose HTTP 500 errors and then I just don't have the code (right) yet to watch for the server to finish shutting down before deleting the pod.  So, I'd like to merge this change and then work on that as a separate PR.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
rjeberhard: I'd like to punt on resolving this too...  (See the comment below).  I'd like to, if you'll agree, work in a separate PR on getting the operator to connect via REST to shutdown the server.  Then, when that has worked, we won't need to fudge factor for the WLST initialization time.

operator/src/main/resources/scripts/stop-server.py
tbarnes-us: Sounds good to me.  Would be nice to track the separate PR via a JIRA if you aren't going to get to it soon.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
tbarnes-us: OK.

PR #1034, opened 2019-04-17
OWLS-72193 - Operator logs repeated NPE when a domain resource credential secret is missing

alai8:
Prevents repeated logging of same error messages in periodic invocation of ReadHealthStep by using a new logging filter mechanism that only logs each log message once until ReadHealthStep is successful again. The logging filter remembers what messages were logged and is passed to the Steps via packet.

Also added a few null checks in HttpClient and ReadHealthStep.

Passed jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1523


operator/src/main/java/oracle/kubernetes/operator/http/HttpClient.java
```diff
@@ -196,11 +196,15 @@ public NextAction apply(Packet packet) {
       if (secretData != null) {
         username = secretData.get(SecretHelper.ADMIN_SERVER_CREDENTIALS_USERNAME);
         password = secretData.get(SecretHelper.ADMIN_SERVER_CREDENTIALS_PASSWORD);
-      }
-      packet.put(KEY, createAuthenticatedClient(username, password));
+        packet.put(KEY, createAuthenticatedClient(username, password));
 
-      Arrays.fill(username, (byte) 0);
-      Arrays.fill(password, (byte) 0);
+        if (username != null) {
```
russgold: I see that you are addressing a potential bug here; it would be better, though, if you took the opportunity to clarify what is going on. For example. You could create a method:
```
// Erase authentication credential so that it is not sitting in memory where a rogue program can find it.
private void clearCredential(byte[] credential) {
  if (credential != null) Arrays.fill(credential, (byte) 0);
}
```
and then use that method for both variables. That would not only eliminate duplication, but would also make it clear why we're doing it.

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFilter.java
```diff
@@ -0,0 +1,35 @@
...
+   * Checks if the message should be logged according to an optional LoggingFilter
+   *
+   * @param loggingFilter LoggingFilter that decides whether the log message should be logged, can
+   *     be null
+   * @param msg The log message to be loggd
+   * @return the canLog() return value from the provided loggingFilter, message, and parameters, or
+   *     true if loggingFilter is null
+   */
+  static boolean canLog(LoggingFilter loggingFilter, String msg) {
+    if (loggingFilter == null) {
```
russgold: You can simplify this as:
```
return loggingFilter == null || loggingFilter.canLog();
```
or, if you don't like those kinds of conditionals:
```
return Optional.ofNullable(loggingFilter).map(loggingFilter::canLog).orElse(true);
```
although I prefer the first as simpler. In either case, it should be easier for most programmers to understand with less reading.

operator/src/main/java/oracle/kubernetes/operator/logging/OncePerMessageLoggingFilter.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Turn off filtering of log messages and allow all log messages to be logged
+   *
+   * @param resetLogHistory Whether to also clear the history of messages logged so those messages
+   *     will be logged again next time filtering is turned on
+   */
+  public void setFilteringOff(boolean resetLogHistory) {
+    filtering = false;
+    if (resetLogHistory) {
```
russgold: Using the same name for both the flag and the method is a bit confusing. The long descriptive comment in the Javadoc is a clue that the parameter isn't as obvious as it could be.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -92,6 +93,14 @@ public NextAction apply(Packet packet) {
       try {
         HttpClient httpClient = (HttpClient) packet.get(HttpClient.KEY);
```
russgold: The apply method is already confusingly long, and now you are adding an exit in the middle. This would be a good time to do some refactoring to simplify it, probably via "extract method" in the right place.

operator/src/test/java/oracle/kubernetes/operator/logging/LoggingFacadeTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.*;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.junit.Test;
+
+public class LoggingFacadeTest {
+
```
russgold: Every test seems to invoke:
```
MockLogger mockLogger = new MockLogger();
LoggingFacade loggingFacade = new LoggingFacade(mockLogger);
```
That suggests that those should be fields on the test, which get initialized in their definitions. That reduces code duplication.

Also, MockLogger isn't actually a mock. See https://blog.pragmatists.com/test-doubles-fakes-mocks-and-stubs-1a7491dfa3da

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFilter.java
alai8: I prefer the first too as it is easier to read. will make the change. Thanks

operator/src/main/java/oracle/kubernetes/operator/logging/OncePerMessageLoggingFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** A LoggingFilter that logs each log message, which are typically message keys, at most once */
+public class OncePerMessageLoggingFilter implements LoggingFilter {
+
+  // allow all messages to be logged when filtering is off
+  volatile boolean filtering = false;
+
+  List messagesLogged = new ArrayList();
```
tbarnes-us: Just a minor performance note:  this could be a HashSet instead of an ArrayList.   Avoids needing to internally iterate through the whole list to implement 'contains' check...

operator/src/main/java/oracle/kubernetes/operator/logging/OncePerMessageLoggingFilter.java
alai8: Should setFilteringOff(boolean resetLogHistory) be separated into 2 method - setFilteringOff() and resetLogHistory()?

russgold: That would certainly be easier to understand. If doing both is something that happens a lot, you could either have a third method which does both, or have both methods return *this* and then call them both something like this:

```
filter.setFiltering(false).resetLogHistory();
```

operator/src/main/java/oracle/kubernetes/operator/logging/OncePerMessageLoggingFilter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/** A LoggingFilter that logs each log message, which are typically message keys, at most once */
+public class OncePerMessageLoggingFilter implements LoggingFilter {
+
+  // allow all messages to be logged when filtering is off
+  volatile boolean filtering = false;
+
```
tbarnes-us: filtering and the hashmap are protected by two different locks -- it's probably fine for this change since I beleive that instances of this class are only exercised single threaded - but still, this means that multi-threading class won't work in all situations.   Would be safer if filtering setter was protected by the same synchronize.

PR #1039, opened 2019-04-19
OWLS-73633 Exclude istio sidecar auto injection in Operator k8s jobs.

davidzl:
Add annotation in job's template to exclude istio auto injection.

Jenkins test passed, http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1527/?auto_refresh=false

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
@@ -214,7 +214,10 @@ public abstract class JobStepContext extends StepContextBase {
   }
 
   private V1PodTemplateSpec createPodTemplateSpec(TuningParameters tuningParameters) {
-    V1ObjectMeta metadata = new V1ObjectMeta().name(getJobName());
+    V1ObjectMeta metadata =
+        new V1ObjectMeta()
+            .name(getJobName())
+            .putAnnotationsItem("sidecar.istio.io/inject", "false");
     V1PodTemplateSpec podTemplateSpec =
         new V1PodTemplateSpec().metadata(metadata).spec(createPodSpec(tuningParameters));
     return podTemplateSpec;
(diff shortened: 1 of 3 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -214,7 +214,10 @@ protected V1JobSpec createJobSpec(TuningParameters tuningParameters) {
   }
 
   private V1PodTemplateSpec createPodTemplateSpec(TuningParameters tuningParameters) {
-    V1ObjectMeta metadata = new V1ObjectMeta().name(getJobName());
+    V1ObjectMeta metadata =
+        new V1ObjectMeta()
+            .name(getJobName())
+            .putAnnotationsItem("sidecar.istio.io/inject", "false");
```
rjeberhard: Why is it that we don't want Istio injection for the domain job?  Is there ever a use case where a customer would want injection?

Separately, what do you think about creating a domain resource field for conforming to Istio requirements?  I know there are going to be other Istio-related changes such as to port names and listen addresses.  I think it would make sense to guard all such changes behind a common flag.

davidzl: 1) Why we don't want istio auto injection for the jobs:
By auto injection, istio will automatically add an container named istio-proxy into every pod. Thus when the Operator job is completed, its status is still RUNNING by "kubectl get pod xxx" because the istio-proxy container is still in running state. So the operator script will not return and trap in the infinite loop until timeout.
Adding this annotation is helpful in istio environment and is harmless in non-istio environment.
So I don't think there is a customer want injection for the job.

2) Creating a domain resource field is a good idea, but now we don't need it in the current stage. I suggest we can add it as a separate task in the future.

davidzl:
> -

Yes, I will add it into delete-domain-job-template.yaml too. should I continue adding it in this branch, or I need to create a new branch for a new pull request?

davidzl:
Move this PR to https://github.com/oracle/weblogic-kubernetes-operator/pull/1043

PR #1042, opened 2019-04-22
OWLS-73633 Exclude istio sidecar auto injection in Operator k8s jobs.

davidzl:
Add annotation in job's template to exclude istio auto injection.


davidzl:
will create new branch for merge

PR #1048, opened 2019-04-23
WIP: Introduce a delay between Terminated and Running status checks.

sankarpn:
Added a new util method to check for restart of the pods. This is to address the NPE when the pod is gone and test tries to get the non existing pod. This fix adds a little delay between the checks of Terminated and Running pod.

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/ITPodsRestart.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/ITPodsRestart.java
@@ -355,18 +355,12 @@ public class ITPodsRestart extends BaseTest {
       logger.log(Level.INFO, "kubectl apply -f {0}", path.toString());
       ExecResult exec = TestUtils.exec("kubectl apply -f " + path.toString());
       logger.info(exec.stdout());
-
-      logger.info("Verifying if the admin server is terminating");
-      verifyPodStatus(podName, "Terminating");
-      verifyPodStatus(podName, "Running");
-
+      checkPodRestartStatus(podName);
     } finally {
       logger.log(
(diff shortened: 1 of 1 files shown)


vanajamukkara:
As the NPE is coming from K8sTestUtils.isPodTerminating() line 414, can you check for null for getPod() and return false if null?
```
 public boolean isPodTerminating(String namespace, String labelSelectors, String podName) {
    V1ObjectMeta metadata;
    metadata = getPod(namespace, labelSelectors, podName).getMetadata();
    if (metadata == null) {
      return false;
    }
    if (metadata.getDeletionTimestamp() != null) {
      logger.info(metadata.getDeletionTimestamp().toString());
    } else {
      logger.info("DeletionTimestamp is null, which means pod is Running");
    }
    return metadata.getDeletionTimestamp() != null;
  }
```

sankarpn:
> 
> 
> As the NPE is coming from K8sTestUtils.isPodTerminating() line 414, can you check for null for getPod() and return false if null?
> 
> ```
>  public boolean isPodTerminating(String namespace, String labelSelectors, String podName) {
>     V1ObjectMeta metadata;
>     metadata = getPod(namespace, labelSelectors, podName).getMetadata();
>     if (metadata == null) {
>       return false;
>     }
>     if (metadata.getDeletionTimestamp() != null) {
>       logger.info(metadata.getDeletionTimestamp().toString());
>     } else {
>       logger.info("DeletionTimestamp is null, which means pod is Running");
>     }
>     return metadata.getDeletionTimestamp() != null;
>   }
> ```

Looks like @rjeberhard already added the check for null.

commit ab3936bc33af601b4c929cad6295d0f7a280dfff
Author: Ryan Eberhard <ryan.eberhard@oracle.com>
Date:   Tue Apr 23 13:01:38 2019 -0400

    Check for null

PR #1050, opened 2019-04-24
WIP: Fix jenkins archives

vanajamukkara:
cleanup.sh which runs at the end of each IT test is unable to delete the directories acceptance_test_tmp and acceptance_test_pv, failing with permission denied errors which makes the log files carried to next IT Test and its log file size will grow.

At the beginning of each test, these directories are created using run_as_root, I modified to delete the directories using run_as_root for jenkins run.


tbarnes-us:
I don't have an exact solution off the top of my head, but can we have a solution that doesn't depend on run_as_root?  Run_as_root is only available internally. The public Jenkins tests don't need this, and can't use this, so why do the internal Jenkins need it?

@rjeberhard

bsmoyers:

PR #1051, opened 2019-04-25
OWLS-73109 - Overloaded server handling

alai8:
** Draft review for comments - not ready for merging **

The scenario in bug owls-73109 is that the server template for a dynamic cluster is configured with low memory overload protection. When the low memeory threshold is reached,  the server would be in an OVERLOADED state, as reported by ServerRuntimeMBean.getHealthState(). Upon entering the OVERLOADED state, server would start rejecting requests from the Work Manager queue (if a Work Manager is configured), and HTTP requests return a 503 Error (Service Unavailable).
 
Filer of bug expects that the server failure action would take place when the server is in OVERLOADED state and shuts down server. But that is not the expected behavior. Server failure action is not taken when server is in OVERLOADED state, only when the server is in FAILED state. An OVERLOADED server will go back to a OK state once the high free memory threshold is reached. So it is working as designed.

But in kubernetes environment, It is not ideal if a load balancer such as traefik continues to direct traffic to the overloaded server, especially if there are other healthy servers available in the same cluster. So the idea is to update the ingress to exclude the overloaded server, so traffic would be directed to other healthy servers. This can be done through the readiness probe.

Currently in operator the readiness probe hard-coded the httpGet path to "/weblogic/ready". However, this does not return 503 Error when the server is overloaded. Thus this change is making the httpGet path of the readiness probe configurable, with a default value of "/weblogic/ready", so customers are allowed to use the URL to their app for the readiness probe.

Pros:
- when server is overloaded, the HTTP request used in the readiness probe would return 503 Error, and ingress would be updated to exclude the overloaded server. When the server returns back to a healthy state, the readiness probe would be successful and ingress will be updated to include the server again.

- when the customer application is not ready to server requests for any other reasons, the ingress should also be updated to not direct traffic to that server, even though the server is in RUNNING state

Cons:
- server pod would not get to ready state if the httpGet path is misconfigured

Please let me know if there are any concerns with this approach, or suggestions on a better way to deal with overloaded servers in the operator.

### DIFF
--- a/docs/domains/Domain.json
+++ b/docs/domains/Domain.json
@@ -389,6 +389,10 @@
     "ProbeTuning": {
       "type": "object",
       "properties": {
+        "httpGetPath": {
+          "description": "Path for httpGet",
+          "type": "string"
+        },
         "periodSeconds": {
           "description": "The number of seconds between checks",
           "type": "number"
--- a/docs/domains/Domain.md
+++ b/docs/domains/Domain.md
@@ -154,6 +154,7 @@ ServerPod describes the configuration for a Kubernetes pod for a server.
 
 | Name | Type | Description |
 | --- | --- | --- |
+| `httpGetPath` | string | Path for httpGet |
 | `initialDelaySeconds` | number | The number of seconds before the first check is performed |
 | `periodSeconds` | number | The number of seconds between checks |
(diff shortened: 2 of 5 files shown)


alai8:
Decided in standup today that we should not be making a pod not ready when server is in overloaded state

PR #1053, opened 2019-04-25
Improvement to non-repeated logging of ReadHealthStep messages

alai8:
PR #1034 prevents repeated logging of same error messages in periodic invocation of ReadHealthStep by using a new logging filter mechanism that only logs each log message once until ReadHealthStep is successful again. The logging filter remembers what messages were logged and is passed to the Steps via packet.

However, it only enables the logging filter if ReadHealthStep for all WLS servers fail, which was the case if the weblogic credentials secret was missing.

This change modifies the change in PR #1034 by enabling the logging filter and prevent repeated logging of messages if ReadHealthStep for any of the WLS servers fails, such as if one of the managed servers fails to response to the REST request due to it being in overloaded health state.


alai8:
Closing this for now as changes for code that read overloaded health status would likely affect ReadHealthStep.java too.

PR #1059, opened 2019-04-30
add more jrf tests in Operator

xiancao:


xiancao:
The JRF tests and integration tests passed on the local run. The integration test also passed on the Jenkins run. http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1582/

PR #1073, opened 2019-05-10
OWLS-73691 - readiness probe needs to work with domain with administration port enabled

alai8:
Add support for weblogic domain with administration port enabled by fixing readiness probe httpGetAction, and also Admin URL for managed servers, so that they will use localAdminProtocolChannelPort instead of defaultPort, and HTTPS instead of HTTP if the admin channel is secure.

Also added -Dweblogic.ssl.AcceptKSSDemoCertsEnabled=true in starting servers in the test testJRFDomainAdminPortEnabled() in JrfInOperatorAdvancedTest.java

Notes:
- Noticed NullPointerExceptions when HTTPS requests were received before server reached running state. Filed OWLS-74406.

- introspectDomain.py was already writing the admin port to topology.yaml if server.isAdministrationPortEnabled() is true. However, server.isAdministrationPortEnabled() does not return true if the admin port is enabled only on DomainMBean. Thus added checking for getDomain().isAdministrationPortEnabled() to introspectDomain.py

- WlsDynamicServerConfig was not created with admin port even if it is set in topology.yaml. This is also fixed.


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -789,6 +803,7 @@ void overrideContainerWeblogicEnvVars(List<V1EnvVar> vars) {
     addEnvVar(vars, "DOMAIN_HOME", getDomainHome());
     addEnvVar(vars, "ADMIN_NAME", getAsName());
     addEnvVar(vars, "ADMIN_PORT", getAsPort().toString());
+    addEnvVar(vars, "ADMIN_PORT_SECURE", Boolean.toString(isLocalAdminProtocolChannelSecure()));
```
rjeberhard: @alai8 This will cause existing pods to roll during an operator upgrade (correct, @russgold?) Assuming that's correct, I would suggest that you only add this environment variable if the value is "true" and then account for the variable not being set inside the pod.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -825,14 +840,21 @@ private V1Probe createReadinessProbe(TuningParameters.PodTuning tuning) {
...
     return readinessProbe;
   }
 
   @SuppressWarnings("SameParameterValue")
-  private V1HTTPGetAction httpGetAction(String path, int port) {
+  private V1HTTPGetAction httpGetAction(String path, int port, boolean useHTTPS) {
     V1HTTPGetAction getAction = new V1HTTPGetAction();
     getAction.path(path).port(new IntOrString(port));
+    if (useHTTPS) {
+      getAction.scheme("HTTPS");
```
rjeberhard: Shouldn't this be lower case, "https"?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -172,6 +172,7 @@ checkEnv \
   SERVICE_NAME \
   ADMIN_NAME \
   ADMIN_PORT \
+  ADMIN_PORT_SECURE \
```
rjeberhard: Handle case where this isn't set, as per earlier comment

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
alai8: The docs I found suggest valid values for scheme are "HTTP" and "HTTPS"
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#examples
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

PR #1077, opened 2019-05-15
Use the kubectl command instead to get pod status

sankarpn:
Use the kubectl command instead of java api to get the status of pods. Apparently the java api's are not reliable to get the pod status.


sankarpn:
Jenkins results.
http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/1664/console

PR #1081, opened 2019-05-22
change image pull, jenkins-ignore

rosemarymarano:
I edited this one file according to what I think needs to be changed for the relocation of WebLogic images to OCR. Please let me know if I am on the right track with the changed text, so I can fix the other mentions in the repo.


rosemarymarano:
Please re-review these steps to make sure that I have the procedure correct for obtaining and using the OCR WLS image. Thank you.

rosemarymarano:
Revised OCR image pull directions, for first time users, to point to Monica's revised Base Image doc (which contains detailed info for logging into OCR, accepting the Oracle Standard Terms and Restrictions, and creating a Kubernetes secret containing the registry credentials.

PR #1086, opened 2019-05-28
Add global option for istio support

davidzl:
1. This adds a global option for istio.
2. If it is under istio environment by the global option, then new channels will be added by WLST when creating the domain.
3. Now this change applies to domain on pv only. If this solution is approved, then it will be applied to domian in image in the next step.


kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+managed_server_name_base     = getEnvVar("MANAGED_SERVER_NAME_BASE")
+managed_server_name_base_svc = getEnvVar("MANAGED_SERVER_NAME_BASE_SVC")
+domain_logs                  = getEnvVar("DOMAIN_LOGS_DIR")
+script_dir                   = getEnvVar("CREATE_DOMAIN_SCRIPT_DIR")
+
+# Update Domain for Istio
+readDomain(domain_path)
+
+cd('/Servers/' + admin_server_name)
+create('istio-probe', 'NetworkAccessPoint')
```
bobdonat: Should this be http-probe.  The istio convention is to use a prefix of http- when using the http protocol.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+set('TwoWaySslEnabled', false)
+set('ClientCertificateEnforced', false)
+
+cd('/Servers/%s/NetworkAccessPoints/T3Channel' % admin_server_name)
+set('ListenAddress', '127.0.0.1')
+set('PublicAddress', domain_uid + '-' + admin_server_name)
+
+templateName = cluster_name + "-template"
+cd('/ServerTemplates/%s' % templateName)
+create('istio-probe', 'NetworkAccessPoint')
```
bobdonat: The istio convention is to use a prefix of http- when using the http protocol.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+set('TwoWaySslEnabled', false)
+set('ClientCertificateEnforced', false)
+
+cd('/ServerTemplates/%s' % templateName)
+create('istio-http', 'NetworkAccessPoint')
+cd('/ServerTemplates/%s/NetworkAccessPoints/istio-http' % templateName)
+set('Protocol', 'http')
+set('ListenAddress', '127.0.0.1')
+set('PublicAddress', domain_uid + '-' + managed_server_name_base + '${id}')
+set('ListenPort', 31111)
```
bobdonat: I am thinking a different port should be used here since 31111 is in the nodePort range and this is not a nodePort.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -837,7 +837,7 @@ def customizeNetworkAccessPoint(self, nap, listen_address):
     # FWIW there's theoretically no need to 'add' or 'replace' when empty
     #   since the runtime default is the server listen-address.
     nap_name=nap.getName()
-    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0:
+    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0 and not (nap_name.startswith('istio-')) and not (nap_name == 'T3Channel'):
```
tbarnes-us: Why does the change suppress overriding the 'T3Channel'?  That seems arbitrary.

operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -250,12 +250,14 @@ EOF
     if [ "${ADMIN_PORT_SECURE}" = "true" ]; then
       admin_protocol="https"
     fi  
-    echo "AdminURL=$admin_protocol\\://${AS_SERVICE_NAME}\\:${ADMIN_PORT}" >> ${wl_props_file}
+    if [ "${ISTIO_ENABLED}" == "true" ]; then
+      echo "AdminURL=t3\\://${AS_SERVICE_NAME}\\:${ADMIN_PORT}" >> ${wl_props_file}
```
tbarnes-us: Wouldn't non-istio case also have same 'admin_protocol' need?   @rjeberhard

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -841,12 +841,22 @@ private V1Probe createReadinessProbe(TuningParameters.PodTuning tuning) {
...
-        .httpGet(
-            httpGetAction(
-                READINESS_PATH,
-                getLocalAdminProtocolChannelPort(),
-                isLocalAdminProtocolChannelSecure()));
+        .failureThreshold(FAILURE_THRESHOLD);
+    try {
+      boolean istioEnabled = Boolean.valueOf(System.getenv("ISTIO_ENABLED"));
+      if (istioEnabled) {
+        readinessProbe = readinessProbe.httpGet(httpGetAction(READINESS_PATH, 8888, false));
```
tbarnes-us: Hard coded port?

tbarnes-us:
General questions:  
1) Where is the detailed description of this change's intent?  (Can you update the description with JIRAs & wiki links?)
2) At one point I recall someone suggested using config overrides to add the ISTIO channels.  This change has the customer 'manually' adding them in the original configuration instead.   Why?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-job-template.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
 spec:
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
```
alai8: Are we adding this annotation regardless of whether istio_enabled is true or false? Is it better not to add it if istio_enabled is false?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain-script.sh
```diff
@@ -10,4 +10,8 @@ export DOMAIN_HOME=${DOMAIN_HOME_DIR}
 # Create the domain
 wlst.sh -skipWLSModuleScanning ${CREATE_DOMAIN_SCRIPT_DIR}/create-domain.py
 
+echo "###### ISTIO_ENABLED=${ISTIO_ENABLED}"
 
+if [ "${ISTIO_ENABLED}" == "true" ]; then
+  wlst.sh -skipWLSModuleScanning ${CREATE_DOMAIN_SCRIPT_DIR}/istio-fix-domain.py
```
alai8: should we combine istio-fix-domain.py logic into create-domain.py and save a wlst.sh call?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+set('HttpEnabledForThisProtocol', true)
+set('TunnelingEnabled', false)
+set('OutboundEnabled', false)
+set('Enabled', true)
+set('TwoWaySslEnabled', false)
+set('ClientCertificateEnforced', false)
+
+cd('/Servers/' + admin_server_name)
+create('istio-t3', 'NetworkAccessPoint')
+cd('/Servers/%s/NetworkAccessPoints/istio-t3' % admin_server_name)
```
alai8: could you point me to some wiki etc to help me understand why we are adding these NAPs and what are they for? Thanks

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: The "istio-" prefix is used as condition in introspectDomain.py to skip overriding listen address in sitconfig. I had ever tried istio-http-probe, but it doesn't work as a wls channel name. Odd!

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: This is the wls channel name, not the http protocol name, so it should be fine.

operator/src/main/resources/scripts/introspectDomain.py
davidzl: The listen address has already been set to the right value: T3_PUBLIC_ADDRESS in create-domain.py and has been updated to 127.0.0.1 in istio-fix-domain.py. So I think it doesn't need to be overridden to service name in sitconfig, right?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: http://aseng-wiki.us.oracle.com/asengwiki/pages/viewpage.action?pageId=5251717305

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain-script.sh
davidzl: As sample codes, the clear structure is prior to the runtime efficiency. And this wlst only run once when creating the domain, so one more wlst call should not affect performance a lot.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -841,12 +841,22 @@ private V1Probe createReadinessProbe(TuningParameters.PodTuning tuning) {
...
+      } else {
+        readinessProbe =
+            readinessProbe.httpGet(
+                httpGetAction(
+                    READINESS_PATH,
+                    getLocalAdminProtocolChannelPort(),
+                    isLocalAdminProtocolChannelSecure()));
+      }
+    } catch (Exception e) {
+      // do nothing
```
tbarnes-us: Why swallow the exception?

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -371,6 +394,11 @@ void addNapServicePort(NetworkAccessPoint nap) {
     abstract void addServicePortIfNeeded(String portName, Integer port);
 
     V1ServicePort createServicePort(String portName, Integer port) {
+      StringBuffer sb = new StringBuffer();
+      StackTraceElement[] stes = Thread.currentThread().getStackTrace();
+      for (StackTraceElement ste : stes) {
+        sb.append(ste.toString()).append("\r\n");
+      }
```
tbarnes-us: This looks like dead (unused) code.

rjeberhard:
I've created a new branch from latest develop, merged the changes from this PR, resolved some conflicts, and then changed the top-level domain resource element to be "istio" instead of just a top-level boolean.  That PR is here: https://github.com/oracle/weblogic-kubernetes-operator/pull/1143.  Let's take the concerns here over to that PR.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+set('HttpEnabledForThisProtocol', true)
+set('TunnelingEnabled', false)
+set('OutboundEnabled', false)
+set('Enabled', true)
+set('TwoWaySslEnabled', false)
+set('ClientCertificateEnforced', false)
+
+cd('/Servers/%s/NetworkAccessPoints/T3Channel' % admin_server_name)
+set('ListenAddress', '127.0.0.1')
+set('PublicAddress', domain_uid + '-' + admin_server_name)
```
tbarnes-us: (1) For all addresses, the name needs to be converted to be DNS legal - see introspectDomain.py for example usage:
```  
def toDNS1123Legal(self, address):
    return address.lower().replace('_','-')
``` 
(2) Ideally we should not be setting addresses in the original configuration.  Instead, we should be setting the ListenAddress and Public Address of istio enabled channels for the user via situational configuration generated in 'introspectDomain.py' instead of requiring customers to set them exactly correctly.
(3) An issue is that the public-address of a T3Channel needs to be appropriate for the client's location, so "domain_uid + '-' + admin_server_name" can be wrong if the client is external.   This is also a problem for (2) - as we should not generate an automatic override for PublicAddress if the user is going to put something else there.

rjeberhard:
I've merge the proposed configuration change from the `istio-enablement` branch to here and also merged latest `develop`.

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -96,3 +96,6 @@ elasticSearchHost: "elasticsearch.default.svc.cluster.local"
 # elasticSearchPort specifies the port number of where elasticsearch is running.
 # This parameter is ignored if 'elkIntegrationEnabled' is false.
 elasticSearchPort: 9200
+
+# istioEnabled specifies whether or not it is deployed under Istio.
```
tbarnes-us: Consider adding '(experimental)' to the comment.

kubernetes/samples/scripts/common/domain-template.yaml
```diff
@@ -47,6 +47,8 @@ spec:
       value: "%JAVA_OPTIONS%"
     - name: USER_MEM_ARGS
       value: "-XX:+UseContainerSupport -Djava.security.egd=file:/dev/./urandom "
+    - name: ISTIO_ENABLED
+      value: "%ISTIO_ENABLED%"
```
tbarnes-us: We shouldn't be requiring users to set this env value - the Operator should set it for both the introspector job and the wl pods. If we want to keep it here, maybe rename it ISTIO_ENABLED_EXPERIMENTAL or at least put a clear comment above the env var stating that it's experimental.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -133,3 +134,6 @@ createDomainScriptName: create-domain-job.sh
 # Kubernetes config map, which in turn is mounted to the `createDomainScriptsMountPath`,
 # so that the Kubernetes pod can use the scripts and supporting files to create a domain home.
 createDomainFilesDir: wlst
+
+# istioEnabled specifies whether or not it is deployed under Istio.
+istioEnabled: false
```
tbarnes-us: Consider adding '(experimental)' to the comment.

operator/src/main/resources/scripts/introspectDomain.py
tbarnes-us: We need a reliable heuristic to detect if a nap needs overriding - 'T3Channel' is just an arbitrary name that happens to be in our main sample, where-as introspectDomain.py is intended to handle all NAP use cases.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -853,7 +853,7 @@ def customizeNetworkAccessPoint(self, nap, listen_address):
...
       self.writeln("<d:network-access-point>")
       self.indent()
```
tbarnes-us: In general, checking for the specific nap name 'T3Channel' looks iffy - I think this is the name we use in our sample and so introspectDomain.py should only check for it in the 'experimental' path.  Minimally, any test for it should be keyed keyed off of the ISTIO_ENABLED flag - something like:

if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0 and not (nap_name.startswith('istio-')) and (not (self.env.ISTIO_ENABLED == 'true') or not (nap_name == 'T3Channel')):

assuming that self.ISTIO_ENABLED = self.getEnvOrDef('ISTIO_ENABLED', 'false') has been added in the global var init section at the top of this script

docs/domains/Domain.md
```diff
@@ -168,6 +175,12 @@ ServerPod describes the configuration for a Kubernetes pod for a server.
 | `annotations` | Map | The annotations to be attached to generated resources. |
 | `labels` | Map | The labels to be attached to generated resources. The label names must not start with 'weblogic.'. |
 
+### Istio
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `istioEnabled` | Boolean | True, if this domain is deployed under an Istio service mesh. Defaults to true. Not required. |
+
```
tbarnes-us: Move this to the new Experimental section?

rjeberhard: It's already there.  This is the definition of the Istio object, which is only referenced in Experimental.

rjeberhard: David is going to be adding other fields to this object to address other comments, e.g. ports.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,131 @@
...
+set('Protocol', 'http')
+set('ListenAddress', '127.0.0.1')
+set('PublicAddress', domain_uid + '-' + admin_server_name)
+set('ListenPort', 8888)
+set('HttpEnabledForThisProtocol', true)
+set('TunnelingEnabled', false)
+set('OutboundEnabled', false)
+set('Enabled', true)
+set('TwoWaySslEnabled', false)
+set('ClientCertificateEnforced', false)
```
tbarnes-us: If istio always needs to add a special port for the ready probe, the introspector should add it via custom sit-config, and the port-number should be configurable so that end-users don't need to make any changes to their config to handle the situation.  

Also, even though the listen address is 127.0.0.1, which I suppose 'hides' the port from resources external to the pod?, I wonder if it should be locked down further and made SSL instead of plain-text.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -600,6 +600,9 @@ void overrideContainerWeblogicEnvVars(List<V1EnvVar> vars) {
     if (mockWls()) {
       addEnvVar(vars, "MOCK_WLS", "true");
     }
+    if (getDomain().istioEnabled()) {
+      addEnvVar(vars, "ISTIO_ENABLED", "true");
+    }
```
tbarnes-us: Need to add the same env var for Jobs.
@rjeberhard

PR #1096, opened 2019-06-04
pmackin-OWLS-72195

pfmackin:
OWLS-72195 - Log the domainUID that is currently being used by a fiber.   Handle the case where the DomainPresenceInfo is null.


pfmackin:
> Looks good to me.
> 
> Have you also looked into if it is possible to leverage the "key" as Ryan suggested in the JIRA, "The FiberGate knows the domainUID when it starts a Fiber (it's the "key") and the LoggingFormatter already acquires the Fiber."
> 
> Were you able to find any other "scopes fields" that might be useful to add to the log entries? Thanks

I didn't seem anything obvious and Derek suggested that i just get this in.  I can investigate this for future work.

PR #1106, opened 2019-06-11
OWLS-69370: Introspect: Domain-in-Image: Add data home override support

lennyphan:
Provide ability to centrally override default file store, and custom file stores for both domain-home-in-image and domain-on-PV installations.

Justification:

- Aligns data storage with log storage
- Makes it easy to centralize data storage

Implementation:

-     Add dataHome to Domain CR
-     dataHome is an optional configuration property.
-     If dataHome is not specified or  its value is either not set or empty (e.g. "") , means 'don't override'. The data storage directories are determined from the WebLogic domain home configuration. 
-     If setting is anything else, use this specific value.
-     Pass through as a well known env var (DATA_HOME) to introspector and to 'custom sit-config'
-     Have introspector override location via sit-config
-     Overrides both default and custom file store 'Directory' attribute
-     Support for creating a symbolic link from server's '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' directory to the centralized in the dataHome location ('${DATA_HOME}/${SERVER_NAME}/data') has been disabled by default and is an experimental feature that isn't supported in production.  The purpose of the symbolic link is to handle the specific case of web-applications with file-based session state, as there's no WebLogic mbean configurable to set the location of this state.  What this means is that the 'dataHome' location is solely designed to support the centralization of the default and custom file stores.  Support for the symbolic link feature for centralization of a server's data directory can be enabled by setting the environment variable 'EXPERIMENTAL_LINK_SERVER_DEFAULT_DATA_DIR' (e.g. EXPERIMENTAL_LINK_SERVER_DEFAULT_DATA_DIR=true). When enabled, during startup of the WebLogic server, the startServer.sh script will:

  - If possible, create a symbolic link from server's directory located in DOMAIN_HOME at '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' and the directory located in the domain resource dataHome directory at '${DATA_HOME}/${SERVER_NAME}/data' in order to centralize each server's data files.

  - Will fail server startup if both the server's directory located in DOMAIN_HOME at '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' and the directory located in the domain resource dataHome directory at '${DATA_HOME}/${SERVER_NAME}/data' contain persistent files and the Operator cannot resolve which directory to use.  Error message is logged directing the user to manually resolve the conflict.

  - The environment variable 'KEEP_DEFAULT_DATA_HOME' can be defined (e.g.  'KEEP_DEFAULT_DATA_HOME=true' ) to bypass creating the symbolic link and allow both the server's data files to exist in the default location ('${DOMAIN_HOME}/servers/${SERVER_NAME}/data' ) and the WebLogic custom and default stores to be centralized in the dataHome location ('${DATA_HOME}/${SERVER_NAME}/data').


kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml
```diff
@@ -85,6 +85,12 @@ logHomeOnPV: false
 # Otherwise, it is ignored.
 logHome: /shared/logs/domain1
 
+# The in-pod location for data storage of default and custom file stores.
+# If not specified, the value is derived from the domainUID as /shared/data/<domainUID>
+# This parameter is required if 'domainHomeOnPV' is true.
+# Otherwise, it is ignored.
```
tbarnes-us: If not specified, the value should be ignored.   This is necessary IMO for backward compatibility.
Also, the value should not be required.  
This is necessary to (a) ensure backward compatibility of old operator domain configuration, (b) provide an intuitive way for users to tell us to use the configured value in the config.xml instead of the override value, and most of all (c) prevent dataloss on upgrade as otherwise the restarted stores will not be able to find their original data.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -203,6 +207,15 @@ private String getEffectiveLogHome() {
     return logHome;
   }
 
+  private String getEffectiveDataHome() {
+    String dataHome = getDataHome();
+    if (dataHome == null) {
+      // dataHome not specified, use default value
+      return DEFAULT_DATA_HOME + File.separator + getDomainUID();
+    }
+    return dataHome;
```
tbarnes-us: If not specified, the value should be ignored. IMO this is necessary to (a) ensure backward compatibility of old operator domain configuration, (b) provide an intuitive way for users to tell us to use the configured value in the config.xml, or the default mbean value, instead of the override value, and most of all (c) prevent dataloss on upgrade as otherwise the restarted stores will not be able to find their original data.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -869,6 +879,54 @@ def customizeLog(self, name, bean, isDomainBean):
...
+    if data_dir is None or len(data_dir) == 0:
+      # do not override if dataHome not specified or empty ("")
+      return
+
+    dfsaction=''
+    fileaction=''
+    if bean.getDefaultFileStore() is None:
+      # don't know why, but don't need to "add" a missing domain log bean, and adding it causes trouble
+      dfsaction=' f:combine-mode="add"'
+      fileaction=' f:combine-mode="add"'
```
tbarnes-us: 1 - Comment is incorrect - mentions log bean.
2 - This combination needs to be tested (overriding data dir when default store bean is missing).

tbarnes-us:
Why is there a 'docs/charts/weblogic-operator-2.3.0.tgz' in the diff? Is this expected now?

tbarnes-us:
It isn't strictly needed, but I think it'd be helpful that if the the dataHome is set, then the introspector and wl-start bash scripts should try and create this directory via a 'mkdir -p', and the script should fail if the command fails.  This is user-friendly as it will cause the job/pod to 'fail fast' before WebLogic itself attempts to create the directory and fails at runtime.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml
lennyphan: Thanks Tom.  Just to clarify, so now we want to change requirement to:

If dataHome is not set or  is "" means means 'don't override'.

tbarnes-us: Yes.

tbarnes-us:
I'm not familiar with web-apps, but it looks like they can persist session state to files. I'm not sure, but it looks like this is configured on weblogic.management.configuration.WebAppComponentMBean /  SessionPersistentStoreDir and on WebServiceComponentBean which inherits from WebAppComponentMBean

lennyphan:
Thanks. Let me take a  look at the WebAppComponentMBean and WebServiceComponentBean and get an understanding if we need to address this for them too.

lennyphan:
> Why is there a 'docs/charts/weblogic-operator-2.3.0.tgz' in the diff? Is this expected now?

When looking at the develop and master branches, these files looked to be committed.

tbarnes-us:
> > Why is there a 'docs/charts/weblogic-operator-2.3.0.tgz' in the diff? Is this expected now?
> 
> When looking at the develop and master branches, these files looked to be committed.

Still looks strange to me - I don't remember seeing that before.

lennyphan:
2nd commit incorporates changes based on the the code review comments:

1.  Added tests for defining a dataHome with an empty ("") path.
2. When dataHome is not set or is "" means means 'don't override'.
3. bash scripts will try and create the dataHome directory (if override path provided) via a 'mkdir -p', and the script will fail (exit) if the command fails.

Still investigating the WebAppComponentMBean and WebServiceComponentBean to determine if we need to override the persistent store directory path.

docs/domains/Domain.md
```diff
@@ -20,6 +20,7 @@ DomainSpec is a description of a domain.
 | `clusters` | array of [Cluster](#cluster) | Configuration for the clusters. |
 | `configOverrides` | string | The name of the config map for optional WebLogic configuration overrides. |
 | `configOverrideSecrets` | array of string | A list of names of the secrets for optional WebLogic configuration overrides. |
+| `dataHome` | string | The in-pod name of the directory for data storage of default and custom file stores. |
```
tbarnes-us: Maybe we can provide a more detail here?  For example, change "The" to "Optional", and add a sentence something like:  "If `dataHome` is not set or set to an empty string, data storage directories are determined from WebLogic domain home configuration.".

docs/domains/Domain.json
```diff
@@ -238,6 +238,10 @@
           "description": "If present, every time this value is updated the operator will restart the required servers.",
           "type": "string"
         },
+        "dataHome": {
+          "description": "The in-pod name of the directory for data storage of default and custom file stores.",
```
tbarnes-us: Maybe we can provide a more detail here? For example, change "The" to "Optional", and add a sentence something like: "If dataHome is not set or is set to an empty string, data storage directories are determined from WebLogic domain home configuration.".

docs/domains/index.html
```diff
@@ -1158,6 +1158,10 @@
           "description": "If present, every time this value is updated the operator will restart the required servers.",
           "type": "string"
         },
+        "dataHome": {
+          "description": "The in-pod name of the directory for data storage of default and custom file stores.",
```
tbarnes-us: Maybe we can provide a more detail here? For example, change "The" to "Optional", and add a sentence something like: "If dataHome is not set or is set to an empty string, data storage directories are determined from WebLogic domain home configuration.".

kubernetes/samples/scripts/common/domain-template.yaml
```diff
@@ -34,6 +34,8 @@ spec:
   %LOG_HOME_ON_PV_PREFIX%logHomeEnabled: %LOG_HOME_ENABLED%
   # The in-pod location for domain log, server logs, server out, and Node Manager log files
   %LOG_HOME_ON_PV_PREFIX%logHome: %LOG_HOME%
+  # The in-pod location for data storage of default and custom file stores files
```
tbarnes-us: Maybe we can provide a more detail here? For example, change "The" to "Optional", and add a sentence something like: "If dataHome is not set or is set to an empty string, data storage directories are determined from WebLogic domain home configuration.".

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml
```diff
@@ -85,6 +85,11 @@ logHomeOnPV: false
 # Otherwise, it is ignored.
 logHome: /shared/logs/domain1
 
+# The in-pod location for data storage of default and custom file stores.
+# If not specified or the value is either not set or empty (e.g. dataHome: "") then the
+# configured 'Directory' value in the config.xml, or the default mbean value, will be used.
+#dataHome: /shared/data/domain1
```
tbarnes-us: I like you're "If not specified" wording here better than what I suggested above for the other locations that dataHome has documentation.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -30,6 +30,9 @@
   /** The pattern for computing the default shared logs directory. */
   private static final String LOG_HOME_DEFAULT_PATTERN = "/shared/logs/%s";
 
+  /** The pattern for computing the default data storage of default and custom file stores. */
+  private static final String DATA_HOME_DEFAULT_PATTERN = "/shared/data/%s";
```
tbarnes-us: This is dead code - no longer used.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -95,6 +95,11 @@
           + "Defaults to false if domainHomeInImage is true. ")
   private Boolean logHomeEnabled; // Boolean object, null if unspecified
 
+  /** The in-pod name of the directory for data storage of default and custom file stores. */
+  @Description(
+      "The in-pod name of the directory for data storage of default and custom file stores.")
```
tbarnes-us: Please enhance this description.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -396,6 +401,19 @@ public void setLogHomeEnabled(boolean logHomeEnabled) {
     this.logHomeEnabled = logHomeEnabled;
   }
 
+  /**
+   * Data Home.
+   *
+   * @return The in-pod name of the directory for data storage of default and custom file stores.
```
tbarnes-us: Please enhance this description as per my other comments.

operator/src/main/java/oracle/kubernetes/operator/helpers/StepContextConstants.java
```diff
@@ -19,6 +19,7 @@
   static final String STORAGE_MOUNT_PATH = "/shared";
   static final String NODEMGR_HOME = "/u01/nodemanager";
   static final String DEFAULT_LOG_HOME = "/shared/logs";
+  static final String DEFAULT_DATA_HOME = STORAGE_MOUNT_PATH + "/data";
```
tbarnes-us: No longer used - dead code.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -41,6 +42,9 @@
   private static final String RAW_VALUE_1 = "find uid1 at $(DOMAIN_HOME)";
   private static final String END_VALUE_1 = "find uid1 at /u01/oracle/user_projects/domains";
 
+  private static final String DATA_HOME = "DATA_HOME";
+  private static final String DEFAULT_DATA_HOME = "/shared/data/JobHelperTestDomain";
```
tbarnes-us: No longer used - dead code.

tbarnes-us:
Update ```If not set, default to overiding as /shared/data/[domain-uid]
If setting is "", means 'don't override'.``` in the description so that it reflects the current behavior.

tbarnes-us:
I should have noticed this earlier (sorry!), but it'd be good to update the documentation so that it covers dataHome.   Look for any mention of 'logHome' and twin up a similar discussion for dataHome - for example:
```
$ find ./docs-source -name "*.md" | xargs grep -l logHome
./docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
./docs-source/content/userguide/managing-domains/configoverrides/_index.md
./docs-source/content/userguide/managing-domains/_index.md
./docs-source/content/samples/simple/domains/domain-home-in-image/_index.md
./docs-source/content/samples/simple/domains/fmw-domain/_index.md
./docs-source/content/samples/simple/domains/domain-home-on-pv/_index.md
```

tbarnes-us:
Apart from the missing doc, the change LGTM.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
tbarnes-us: Confirmed done/resolved.

docs/domains/Domain.md
tbarnes-us: Confirmed done/resolved.

docs/domains/index.html
tbarnes-us: Confirmed resolved.

kubernetes/samples/scripts/common/domain-template.yaml
tbarnes-us: Confirmed resolved.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml
tbarnes-us: confirmed resolvved

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: resolved

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
tbarnes-us: resolbved

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
tbarnes-us: resolved

operator/src/main/java/oracle/kubernetes/operator/helpers/StepContextConstants.java
tbarnes-us: resolved

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
tbarnes-us: resolved

lennyphan:
> LGTM so far apart from missing WebAppComponentMBean/WebServiceComponentBean file based session state - which I assume should be included but still needs research. If this is added, the doc for dataHome throughout will need updated too...

Queried Steven Liu for verification of File based session state persistent store configuration and there isn't any way to override the persistent store directory through any mbean configuration.  Here is Steven's reply:

Hi Lenny,

There’s no configuration on MBean to specify the persistent store directory for File persistent session data.

There are different cases for how web container determins the directory:

    If there’s no <persistent-store-dir> configured in session-descriptor or the <persistent-store-dir> is configured to be relative directory, the store directory will be created under ${domain_home}/servers/${server_name}/data/ directory.
    If the <persistent-store-dir> is configured to be absolute directory, the configured directory will be used for persistent store.


Hope this helps!
-Steven

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml
tbarnes-us: Resolved.

lennyphan:
From the last review from Tom Barnes, added the ability to:

1. Create a symbolic link from the server's domain home 'data' directory to the centralized 'data home' directory specified by the 'dataHome' attribute in the CRD.  This allows each server's data directory to be centralized.

2. We provide a warning description/instructions and will fail server startup if there are files in both the  server's domain home 'data' directory and the centralized data directory (specified by 'dataHome' attribute) as we have no idea which files are important to the user's or are in conflict between the two directory locations.  As per the usual WebLogic approach to these kind of issues, we ask the user's to resolve the differences before restarting the server, if they intend to centralize their server's data storage.

3. Provided a container/server environment variable  ('KEEP_DEFAULT_DATA_HOME') that will  bypass the server startup failure due to data files being located in both the  server's domain home 'data' directory and the centralized data directory (specified by 'dataHome' attribute).  This allows the server to continue to startup and operate and keep its domain home 'data' directory in the default location while still allowing the override of the default file and custom file stores.

4. NOTE: This change does not override the WLDF diagnostics store as they are not uniquely named and so cannot be centralized, whereas the default and custom file stores are uniquely named.

operator/src/main/resources/scripts/introspectDomain.py
lennyphan: Tom, can you point me to any test framework/code we have for creating a WLS Domain and running multiple scenarios in order to test the different configurations above by the instrospector?  I've looked at src/integration-tests/introspector but it's unclear if it can be run multiple individual tests or it just runs  a full suite of tests based on one configuration.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -37,7 +37,7 @@ function exitOrLoop {
 #
 
 function createFolder {
-  mkdir -m 750 -p $1
+  mkdir -m 777 -p $1
```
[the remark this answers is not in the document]
lennyphan: Sorry, this was a carryover for when I was figuring out the symbolic link.  I reverted it back to 750.

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -74,6 +87,17 @@ for dir_var in DOMAIN_HOME JAVA_HOME WL_HOME MW_HOME ORACLE_HOME; do
   [ ! -d "${!dir_var}" ] && trace SEVERE "Missing ${dir_var} directory '${!dir_var}'." && exit 1
 done
 
+trace "DATA_HOME=${DATA_HOME}"
+
+#
+# DATA_HOME env variable exists implies override directory specified.  Attempt to create directory
+#
+if [ -n ${DATA_HOME} ] && [ ! -d ${DATA_HOME} ]; then
```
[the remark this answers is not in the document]
lennyphan: I found the following description from some online docs for bash conditional exresssions but I changed it to use '! -z' as you suggested since it's probably more portable and compatible with different bash versions.

-z string
    True if the length of string is zero. 
string
-n string
    True if the length of string is non-zero.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +211,100 @@ checkEnv \
...
+        fi
+
+        # forcefully delete the server's data directory so we can create symbolic link
+        deleteDir ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
+
+        # Create the symbolic link from server's data directory to $DATA_HOME
+        createSymbolicLink ${DATA_HOME}/${SERVER_NAME}/data ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
+      elif [ ${fileCountServerDataDir} -ne 0 ]; then
+        if [ ${fileCountServerDomainHomeDir} -ne 0 ]; then
+           trace SEVERE "Both ${DOMAIN_HOME}/servers/${SERVER_NAME}/data and ${DATA_HOME}/${SERVER_NAME}/data directories contain persistent files which WebLogic cannot resolve.  You must manually move any persistent files from ${DOMAIN_HOME}/servers/${SERVER_NAME}/data directory to ${DATA_HOME}/${SERVER_NAME}/data directory and then delete the ${DOMAIN_HOME}/servers/${SERVER_NAME}/data directory.  Once this is done you can then restart the Domain. Alternatively, you can avoid this validation by setting the 'KEEP_DEFAULT_DATA_HOME' environment variable."
```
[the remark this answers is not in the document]
lennyphan: I like your description a lot so I used it. Thanks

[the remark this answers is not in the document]
tbarnes-us: Sorry! thinking about this some more.    

"You must manually move any persistent files from the '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' directory to '${DATA_HOME}/${SERVER_NAME}/data'"

   -->  You must manually move any persistent files from the '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' directory to '${DATA_HOME}/${SERVER_NAME}/data'  or remove them, and then delete

operator/src/main/resources/scripts/introspectDomain.py
tbarnes-us: I'd be tempted to just try this out 'ad hoc', and the write-up a test plan JIRA for the QA team to create formal tests.You can overload the existing test in src/integration-tests/introspector  - just add more config as needed to the domain create (wl-create-domain-pod.pyt), and then, at the end, verify (a) that overrides are reflected in the bean tree, and (b)

tbarnes-us: To 'ad hoc' test one way might be temporarily modify introspector integration tests:
Modify domain config (wl-create-domain-pod.pyt) 'admin server' to have one style of default store config, 'cluster' to have a second style, plus add two custom file stores on admin server, and something that puts data in the custom file stores  to force them to create .DAT files (JMS Server + persistent messages?).   Then (a) verify overrides are taking effects via WLST inspect of the bean tree, and (b) verify overrides taking effect by noting the .DAT files are in the expected location.  Maybe just do (b) as (b) implies (a) must be working.

lennyphan: Thank you for the guidance, I will look into how to implement your 'ad hoc' tests.
Also, QA has already filed a JIRA: https://jira.oraclecorp.com/jira/browse/OWLS-75000 which we can update with your test coverage scenarios.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +204,105 @@ checkEnv \
...
+
+        # count number of files found under directory ${DATA_HOME}/${SERVER_NAME}/data
+        countFilesInDir ${DATA_HOME}/${SERVER_NAME}/data
+        fileCountServerDataDir=$?
+
+        # Use file counts to determine whether or not we can create a symbolic link to centralize
+        # data directory in specified ${DATA_HOME}/${SERVER_NAME}/data directory.
+        if [ ${fileCountServerDataDir} -eq 0 ]; then
+          if [ ${fileCountServerDomainHomeDir} -ne 0 ]; then
+            cp -rf ${DOMAIN_HOME}/servers/${SERVER_NAME}/data ${DATA_HOME}/${SERVER_NAME}
```
tbarnes-us: If this command fails, we should fail and exit.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +204,105 @@ checkEnv \
...
+            trace "Recursively copied directory/files from '${DOMAIN_HOME}/servers/${SERVER_NAME}/data' to '${DATA_HOME}/${SERVER_NAME}' directory"
+          else
+            trace "'${DOMAIN_HOME}/servers/${SERVER_NAME}/data' directory is empty"
+          fi
+
+          # forcefully delete the server's data directory so we can create symbolic link
+          frm -rf ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
```
tbarnes-us: I'm not familiar with the 'frm' command.  Is this a typo?  If so, then we need more testing before merging this pull as this would mean that this path wasn't tested...

tbarnes-us: Also, if this command fails, then should the script exit?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +204,105 @@ checkEnv \
...
+          trace "Deleted directory '${DOMAIN_HOME}/servers/${SERVER_NAME}/data'"
+
+          # Create the symbolic link from server's data directory to $DATA_HOME
+          createSymbolicLink ${DATA_HOME}/${SERVER_NAME}/data ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
```
tbarnes-us: If this command fails, should the script report an error and exit?   The createSymbolicLink function itself succeeds even if the ln command it has fails.

tbarnes-us:
Overall comments:
- Unfortunately, the symbolic link support still seems fragile. Maybe turn it off via an experimental flag which can be enabled via a new env var, or, if you're ambitious, via a new experimental domain resource attribute?

lennyphan:
> Overall comments:
> 
>     * dataHome should be documented in docs-source - I've posted potential locations in a previous comment
> 
>     * if a line of a script can fail, this needs to cause the script to log an error and fail - I've just posted comments about two such locations
> 
>     * otherwise all LGTM assuming 'ad hoc' testing is all good.
> 
>     * unfortunately, the symbolic link support still seems fragile. Maybe turn it off pending future hardening JIRA (via an experimental flag which can be enabled via a new env var, or, if you're ambitious, via a new experimental domain resource attribute)?  maybe it should be enabled via a separate setting?

I agree that the symbolic link support is still a little fragile.  It's mainly contained in the startServer.sh script. We could just punt on it and go back to the original intent of just supporting file stores.  That is, separate the symbolic link feature as you previously suggested.

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -74,6 +87,17 @@ for dir_var in DOMAIN_HOME JAVA_HOME WL_HOME MW_HOME ORACLE_HOME; do
...
 done
 
+trace "DATA_HOME=${DATA_HOME}"
+
+#
+# DATA_HOME env variable exists implies override directory specified.  Attempt to create directory
+#
+if [ ! -z "${DATA_HOME}" ] && [ ! -d "${DATA_HOME}" ]; then
+  createFolder ${DATA_HOME}
+  trace "Created data home directory: '${DATA_HOME}'"
```
tbarnes-us: Move the trace statement before 'createFolder' so that if createFolder fails the tracing will show why it was being called.   Change 'Created' to 'Creating'.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -162,6 +162,24 @@ function copySitCfg() {
   fi
 }
 
+# Returns the count of the number of files in the specified directory
+function countFilesInDir() {
+  dir=${1}
+  cnt=`find ${dir} -type f | wc -l`
```
tbarnes-us: I think the $? test below will always return success since 'wc -l' succeeds even though the find failed.   To counteract, set 'pipefail'.   I think you can set it for the entire script at the beginning of the script and it will cover functions as well (but I haven't tested to confirm), but if you want to localize it to this function then do this:
```
function countFilesInDir() {
   (
   set -o pipefail
   dir=${1}
   cnt=`find ${dir} -type f | wc -l`
   [ $? -ne 0 ] && trace SEVERE "failed determining number of files in '${dir}'" && exitOrLoop
   trace "file count in directory '${dir}': ${cnt}"
   return ${cnt}
   )
 }
```

src/integration-tests/introspector/util_test_adminfilestores.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
```
tbarnes-us: Way cool test!

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +206,109 @@ checkEnv \
...
+    trace "Creating directory '${DATA_HOME}/${SERVER_NAME}/data'"
+    createFolder ${DATA_HOME}/${SERVER_NAME}/data
+  else
+    trace "Directory '${DATA_HOME}/${SERVER_NAME}/data' exists"
+  fi
+
+  # If 'LINK_SERVER_DATA_DIR' env is defined and 'KEEP_DEFAULT_DATA_HOME' environment variable is not defined then
+  # try to link server's default 'data' directory (${DOMAIN_HOME}/servers/${SERVER_NAME}/data) to $DATA_HOME/${SERVER_NAME}/data.
+  if [ ! -z ${LINK_SERVER_DATA_DIR} ] && [ -z ${KEEP_DEFAULT_DATA_HOME} ]; then
+    # if server's default 'data' directory (${DOMAIN_HOME}/servers/${SERVER_NAME}/data) does not exist than create
```
tbarnes-us: This is 80+ of complex code that we only plan to enable experimentally, and it significantly lengthens this script, so it's a bit distracting to include it in-line.  If you agree, maybe we can hide it a bit?  Maybe move the guts of the 'if' to a function that's included at the end utils.sh, and only leave the 'if' in this script?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +206,109 @@ checkEnv \
   AS_SERVICE_NAME || exitOrLoop
 
 trace "LOG_HOME=${LOG_HOME}"
+trace "DATA_HOME=${DATA_HOME}"
 trace "SERVER_OUT_IN_POD_LOG=${SERVER_OUT_IN_POD_LOG}"
 trace "USER_MEM_ARGS=${USER_MEM_ARGS}"
 trace "JAVA_OPTIONS=${JAVA_OPTIONS}"
+trace "KEEP_DEFAULT_DATA_HOME=${KEEP_DEFAULT_DATA_HOME}"
+trace "LINK_SERVER_DATA_DIR=${LINK_SERVER_DATA_DIR}"
```
tbarnes-us: Since this env var isn't supported in production, can we rename it?
EXPERIMENTAL_LINK_SERVER_DEFAULT_DATA_DIR

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +206,109 @@ checkEnv \
...
+
+#
+# If DATA_HOME env variable exists than this implies override directory (dataHome attribute of CRD) specified
+# so we need to try and link the server's 'data' directory to the centralized DATA_HOME directory
+# If 'KEEP_DEFAULT_DATA_HOME' env variable is defined then we will NOT link the server's 'data' directory to the
+# centralized DATA_HOME directory and instead keep the server's 'data' directory in its default location of
+# ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
```
tbarnes-us: Please add an explanation why we have this experimental code:   something like "This experimental code handles the specific case of services that don't provide a configurable way to control the location of their file data, specifically it's to cover stateful web-apps that use file state instead of database state.  The strongly recommended work-around is to have the web-app use database storage."  (I don't know web-app terminology so that might not be quite right.)

tbarnes-us: On a related note, I updated your push description with a similar statement.  Please review the change and fix it up.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +147,41 @@ checkEnv \
...
 trace "SERVER_OUT_IN_POD_LOG=${SERVER_OUT_IN_POD_LOG}"
 trace "USER_MEM_ARGS=${USER_MEM_ARGS}"
 trace "JAVA_OPTIONS=${JAVA_OPTIONS}"
+trace "KEEP_DEFAULT_DATA_HOME=${KEEP_DEFAULT_DATA_HOME}"
+trace "EXPERIMENTAL_LINK_SERVER_DEFAULT_DATA_DIR=${EXPERIMENTAL_LINK_SERVER_DEFAULT_DATA_DIR}"
+
+# The following is experimental code that handles the specific case of services that don't provide a configurable way to
+# control the location of their persistent file data.  For example, web applications can configure file-based
+# session persistence where the default persistent file store location is automatically created in the
+# <server-name>\data\store\default directory.
```
tbarnes-us: Minor comment:  The above comment is out-of-place it should be moved/merged into into line 176 comment.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -188,12 +147,41 @@ checkEnv \
...
+#
+# If DATA_HOME env variable exists than this implies override directory (dataHome attribute of CRD) specified
+# so we need to try and link the server's 'data' directory to the centralized DATA_HOME directory
+# If 'KEEP_DEFAULT_DATA_HOME' env variable is defined then we will NOT link the server's 'data' directory to the
+# centralized DATA_HOME directory and instead keep the server's 'data' directory in its default location of
+# ${DOMAIN_HOME}/servers/${SERVER_NAME}/data
```
tbarnes-us: The "KEEP" comments from line 164 - 166 are out of place.  Perhaps move/merge it into comments at 176.

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -48,6 +48,19 @@ trace "Introspecting the domain"
...
 env | tracePipe "Current environment:"
 
+#
+# Define helper fn to create a folder
+#
+
+function createFolder {
+  mkdir -m 750 -p $1
+  if [ ! -d $1 ]; then
+    trace "Unable to create folder $1"
```
tbarnes-us: Add SEVERE into this statement.   `trace SEVERE "Unable to create folder $1"`

lennyphan: createFolder() function exists in utils.sh so removed this duplicate in introspectDomain.sh

lennyphan:
> Looks good! I added a couple minor suggestions in case you'd like to address them.

Implemented your latest suggestions. Thank you.

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -74,6 +75,17 @@ for dir_var in DOMAIN_HOME JAVA_HOME WL_HOME MW_HOME ORACLE_HOME; do
...
 done
 
+trace "DATA_HOME=${DATA_HOME}"
+
+#
+# DATA_HOME env variable exists implies override directory specified.  Attempt to create directory
+#
+if [ ! -z "${DATA_HOME}" ] && [ ! -d "${DATA_HOME}" ]; then
+  trace "Creating data home directory: '${DATA_HOME}'"
+  createFolder ${DATA_HOME}
```
tbarnes-us: I wonder if createFolder should be locally defined.   This depends on utils.sh createFolder, which was moved out of startServer.sh to utils.sh and is specific to startServer.sh. It's specific to startServer.sh as it calls 'exitOrLoop' instead of 'exit', plus if it does 'loop' it ultimately exits 0 instead of 1.

operator/src/main/resources/scripts/utils.sh
```diff
@@ -419,6 +419,147 @@ function getAdminServerUrl() {
...
+  mkdir -m 750 -p $1
+  if [ ! -d $1 ]; then
+    trace SEVERE "Unable to create folder $1"
+    exitOrLoop
+  fi
+}
+
+# Returns the count of the number of files in the specified directory
+function countFilesInDir() {
+  set -o pipefail
```
tbarnes-us: If you intend to scope pipefail just to this method, then I think you may need to put a new line with an open paren at the beginning of the function and a closing paren at the end.

tbarnes-us: If you intend to scope pipefail to all scripts (which I think would be just find), then move the set -o pipefail to the top of utils.sh.

lennyphan: moved 'set -o pipefail' to the top of utils.sh.

operator/src/main/resources/scripts/introspectDomain.sh
lennyphan: added back createFolder to be locally defined in introspectDomain.sh with 'exit 1'.

lennyphan:
> Well, that was long strange trip. My apologies that the soft linking idea turned into a bit of a garden path. Here's my approval good sir.

No apologies needed.  Always learn something new, especially scripting as I hardly write scripts.  Thank you for your great reviews.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -257,6 +260,45 @@ public void whenDomainHasEnvironmentItemsWithVariables_introspectorPodStartupWit
...
+        domainIntrospectorJobStepContext.createJobSpec(TuningParameters.getInstance());
+
+    MatcherAssert.assertThat(
+        getContainerFromJobSpec(jobSpec, domainPresenceInfo.getDomainUID()).getEnv(),
+        allOf(hasEnvVar(DATA_HOME, DEFAULT_DATA_HOME)));
+  }
+
+  @Test
+  public void whenDomainHasEnvironmentVars_introspectorPodStartupVerifyDataHomeDefined() {
+    final String EMPTY_DATA_HOME = "/u01/data/JobHelperTestDomain";
```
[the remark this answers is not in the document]
lennyphan: Done

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -589,6 +594,10 @@ void addStartupEnvVars(List<V1EnvVar> vars) {
     addEnvVar(vars, ServerEnvVars.SERVER_OUT_IN_POD_LOG, isIncludeServerOutInPodLog());
     addEnvVar(vars, ServerEnvVars.SERVICE_NAME, LegalNames.toServerServiceName(getDomainUid(), getServerName()));
     addEnvVar(vars, ServerEnvVars.AS_SERVICE_NAME, LegalNames.toServerServiceName(getDomainUid(), getAsName()));
+    String dataHome = getDataHome();
```
russgold: Where is the unit test for this added behavior?

lennyphan: Thanks Russ.  Added it for JobStepContext but not for PodStepContext.  Will add unit tests for PodStepContext.

rjeberhard: @russgold, are you satisfied with the new tests?

PR #1117, opened 2019-06-18
Start WLS with -Dweblogic.SituationalConfig.failBootOnError=true

alai8:
This feature would require a WebLogic that supports the weblogic.SituationalConfig.failBootOnError system property in the docker image.

By setting the `FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR` environment variable in the Kubernetes containers for the WebLogic Servers to `false`, customers can start up the WebLogic Servers even with incorrectly formatted override files.

Prints configuration overrides on server pod log

startServer.sh looks for BEA-141335 which suggests WebLogic fails to start due to situational config errors, and logs message "WebLogic server failed to start due to missing or invalid situational configuration files. Please check ${SERVER_OUT_FILE} for details" to the pod log.

Updated configoverrides/_index.md doc


tbarnes-us:
The change LGTM except that:

(1) it'd be good to make the new behavior configurable via an environment variable (default to true

(2) it'd be good to update the config-override .md doc to explain the new behavior on a failure for 12.2.1.4 and later (the current doc explicitly details the current behavior - which is to 'keep on truckin' even though there are errors), and to perhaps even mention the env var in (1).

In addition we've been discussing additional 'bonus work' in slack which could be done by a different JIRA/change, or maybe in this change:

- When an error occurs, it'll be a generic 'situational configuration' error that might make it hard for many Kubernetes Operator administrators to figure out the root cause.  So it'd be helpful if we could detect and report it with additional information.  For example, we could hint that the customer's 'domain CR configuration overrides' may have an error.

- One approach might be to grep the servers stdout capture files during shutdown for the new error and log a more detailed exception to stdout.  (Search for 'doShutdown' in the scripts.)   

- A more advanced approach would be to add an Event to the Domain CR.

@TheFrogPad FYI

docs/domains/Domain.md
```diff
@@ -23,6 +23,7 @@ DomainSpec is a description of a domain.
 | `domainHome` | string | The folder for the WebLogic Domain. Not required. Defaults to /shared/domains/domains/domainUID if domainHomeInImage is false Defaults to /u01/oracle/user_projects/domains/ if domainHomeInImage is true |
 | `domainHomeInImage` | Boolean | True if this domain's home is defined in the docker image for the domain. Defaults to true. |
 | `domainUID` | string | Domain unique identifier. Must be unique across the Kubernetes cluster. Not required. Defaults to the value of metadata.name |
+| `failBootOnSituationalError` | Boolean | If true (the default), on WebLogic server that supports this feature, the WebLogic server  boot would fail if any errors occur when applying situational configuration  during server startup. If false, WebLogic server would start if there are errors in the situational configuration files, but some configuration overrides may be skipped. |
```
rosemarymarano: (suggested re-write) In WebLogic Server versions that support this feature (greater than 12.2.1.3.0):
<ul><li>If true (the default), then WebLogic Server  would fail to boot if any errors occurred when applying the situational configuration  during server startup. </li><li>If false, and if there were errors in the situational configuration files, then WebLogic Server would start but some configuration overrides may be skipped.</li></ul>

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
       * This directory, or a subdirectory within this directory, should contain each of your custom situational configuration files.
       * If it doesn't, then this likely indicates your domain resource `configOverrides` was not set to match your custom override configuration map name, or that your custom override configuration map does not contain your override files.
 
+* If the Administration Server pod do start but fails to reach ready state:
+  * Check if there is a message ` WebLogic server failed to start due to missing or invalid situational configuration files` in the Administration Server pod's `kubectl log`
```
rosemarymarano: Check if there is a message -> Check for this message

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -338,6 +338,11 @@ spec:
 
 Incorrectly formatted override files may be accepted without warnings or errors and will not prevent WebLogic pods from booting. So, it is important to make sure that the template files are correct in a QA environment, otherwise your WebLogic Servers may start even though critically required overrides are failing to take effect.
 
+On WebLogic servers that support the weblogic.SituationalConfig.failBootOnError system property ( Note: it is not supported in WebLogic 12.2.1.3 ), 
+by default the WebLogic server will fail to boot if any situational configuration files are invalid.
```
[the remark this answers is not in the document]
tbarnes-us: are we allowed to say "It is not supported in WebLogic Server 12.2.1.3 and is scheduled to be supported in future releases, and may also be supported by a WebLogic Patch generated via Bug XXXXXX."?

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
       * This directory, or a subdirectory within this directory, should contain each of your custom situational configuration files.
       * If it doesn't, then this likely indicates your domain resource `configOverrides` was not set to match your custom override configuration map name, or that your custom override configuration map does not contain your override files.
 
+* If the Administration Server pod do start but fails to reach ready state:
+  * Check if there is a message ` WebLogic server failed to start due to missing or invalid situational configuration files` in the Administration Server pod's `kubectl log`
+    * This suggests that the Administration Server failed to start may be caused by errors found in the a configuration override file.
+      * Lines containing the String `<Warning>` may be found in the Administration Server pod log to provide more hints.
```
tbarnes-us: I think it's better to search for `situational` instead of `<Warning>`, or at least in addition to Warning, as this will catch all types of log messages (Error, Critical, etc.)    (Also because I've noticed that Warnings are extremely common in some customers and layered products, even though the system is running acceptably as far as the admin is concern, so reading through all Warnings can be a bewildering exercise trying to figure out which ones are important.)

docs/domains/Domain.json
```diff
@@ -270,6 +270,10 @@
           "description": "The in-pod name of the directory in which to store the domain, node manager, server logs, and server  *.out files",
           "type": "string"
         },
+        "failBootOnSituationalError": {
+          "description": "If true (the default), on WebLogic server that supports this feature, the WebLogic server  boot would fail if any errors occur when applying situational configuration  during server startup. If false, WebLogic server would start if there are errors in the situational configuration files, but some configuration overrides may be skipped.",
+          "type": "boolean"
+        },
```
tbarnes-us: Would 'failBootOnSituationalConfigurationError' be too long?  I think the official name of teh WL sit-cfg feature is 'situational configuration'.

tbarnes-us: I had originally suggested we just document an environment variable customers could specify instead of adding another Domain attribute.   Now that you've already done the work that's fine IMO, but, OTOH I still have my doubts if we should be cluttering the main domain spec with this parameter.   Up to you.  Or ask Ryan to weigh in...

docs-source/content/userguide/managing-domains/configoverrides/_index.md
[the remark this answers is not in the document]
rosemarymarano: No, no, and no. All you can accurately state here (and perhaps, we should revise my advice given below to say the same thing) is, "It is not supported in WebLogic Server 12.2.1.3.0." We can change this sentence when we officially support the property.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -99,6 +99,21 @@
...
+   */
+  @Description(
+      "If true (the default), on WebLogic server that supports this feature, the WebLogic server "
+          + " boot would fail if any errors occur when applying situational configuration "
+          + " during server startup."
+          + " If false, WebLogic server would start if there are errors in the"
+          + " situational configuration files, but some configuration overrides may"
+          + " be skipped.")
+  private Boolean failBootOnSituationalError;
+
```
tbarnes-us: 'any errors' is promising too much - the setting doesn't capture all errors.   There are multiple errors that aren't captured -- you mispell an attribute name (IIRC), or reference an attribute that isn't there, or use 'add' when 'replace' was needed, etc, etc.    

How about:
 "If true (the default), on a WebLogic server that supports this feature, the WebLogic server "
	          + " boot will fail if it detects errors when applying situational configuration "
	          + " during server startup, as may happen if incorrectly configuration overrides templates"
                  + " are specified (see configOverrides attributes)."
	          + " If false, WebLogic server will still start even if there are errors detected in the"
	          + " situational configuration files but some or all configuration overrides may"
	          + " be skipped.")

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -110,7 +110,12 @@ function waitForShutdownMarker() {
   # Wait forever.   Kubernetes will monitor this pod via liveness and readyness probes.
   #
   trace "Wait indefinitely so that the Kubernetes pod does not exit and try to restart"
+  logged=false
   while true; do
+    if [ "$logged" != 'true' ] && grep -q "BEA-141335" ${SERVER_OUT_FILE} ; then
+      trace "WebLogic server failed to start due to missing or invalid situational configuration files. Please check ${SERVER_OUT_FILE} for details"
+      logged=true
+    fi
```
tbarnes-us: trace "WebLogic server failed to start due to missing or invalid situational configuration files. Please check ${SERVER_OUT_FILE} for details"
	     
-->

trace "WebLogic server failed to start due to missing or invalid situational configuration files, which may be due to invalid configOverride templates (these are specified via the configOverride attribute in the Domain custom resource).  For details, please search your pod log or ${SERVER_OUT_FILE} for the keyword 'situational'."

docs/domains/Domain.json
alai8: @rjeberhard Would like your input on whether we should add a Domain attribute here or not. Thanks.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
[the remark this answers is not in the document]
tbarnes-us: ok.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
tbarnes-us: Similarly, the phrasing 'any situational configuration files are invalid' and similar phrasing that's used in other places within the change also implies to me an overly strong guarantee that WL will detect all situational configuration errors.  Not sure how to word it though.   Maybe something along the lines of 'common situational configuration file problems' 'most situational configuration file problems', etc...

alai8: How about using something along the line of the actual WebLogic messages that were logged while loading sit config files:
BEA-141322 - The situational config file has an invalid format
BEA-141329 - Error loading the situational config files
So something along the line of "..the WebLogic server boot will fail if it encounters an error while loading the situational configuration files, or if a situational configuration file has an invalid format"
I don't know if that is better, but at least what we document here would be consistent with the error messages customers will find in the WebLogic log.

tbarnes-us: Sure.  Anything that implies 'detects most errors' as opposed to 'all errors' or 'any errors'.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -338,6 +338,11 @@ spec:
 
 Incorrectly formatted override files may be accepted without warnings or errors and will not prevent WebLogic pods from booting. So, it is important to make sure that the template files are correct in a QA environment, otherwise your WebLogic Servers may start even though critically required overrides are failing to take effect.
 
+On WebLogic servers that support the weblogic.SituationalConfig.failBootOnError system property ( Note: it is not supported in WebLogic 12.2.1.3 ),
```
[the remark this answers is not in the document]
alai8: Fixed. WLS Version number also modified to 12.2.1.3.0

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -338,6 +338,11 @@ spec:
 
 Incorrectly formatted override files may be accepted without warnings or errors and will not prevent WebLogic pods from booting. So, it is important to make sure that the template files are correct in a QA environment, otherwise your WebLogic Servers may start even though critically required overrides are failing to take effect.
 
+On WebLogic servers that support the weblogic.SituationalConfig.failBootOnError system property ( Note: it is not supported in WebLogic 12.2.1.3 ), 
+by default the WebLogic server will fail to boot if any situational configuration files are invalid.
+This can be configured by using the `failBootOnSituationalError` attribute in the Domain spec to `false` to start up the WebLogic servers even with
```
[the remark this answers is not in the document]
alai8: The latest update removed the `failBootOnSituationalError` attribute from the Domain spec. The sentence now read:
By setting the `FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR` environment variable in the Kubernetes containers for the WebLogic Servers to `false`, you can start up the WebLogic Servers even with incorrectly formatted override files.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
       * This directory, or a subdirectory within this directory, should contain each of your custom situational configuration files.
       * If it doesn't, then this likely indicates your domain resource `configOverrides` was not set to match your custom override configuration map name, or that your custom override configuration map does not contain your override files.
 
+* If the Administration Server pod do start but fails to reach ready state:
```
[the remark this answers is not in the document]
alai8: Fixed. Thanks

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
...
+  * Check if there is a message ` WebLogic server failed to start due to missing or invalid situational configuration files` in the Administration Server pod's `kubectl log`
+    * This suggests that the Administration Server failed to start may be caused by errors found in the a configuration override file.
```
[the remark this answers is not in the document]
alai8: Sentence now read:
This suggests that the Administration Server failure to start may have been caused by errors found in a configuration override file.

docs/domains/Domain.md
alai8: The `failBootOnSituationalError` attribute has been removed in the latest update

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: Fixed

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: Agree

docs/domains/Domain.json
alai8: Agree with your concern. I decided to remove the Domain spec attribute and just go with the env variable.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
alai8: This is outdated since the Domain spec attribute has been removed. I have applied the suggestion to the changes in configoverrides doc.

operator/src/main/resources/scripts/startServer.sh
alai8: Fixed as suggested. Thanks.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -338,6 +338,11 @@ spec:
 
 Incorrectly formatted override files may be accepted without warnings or errors and will not prevent WebLogic pods from booting. So, it is important to make sure that the template files are correct in a QA environment, otherwise your WebLogic Servers may start even though critically required overrides are failing to take effect.
```
tbarnes-us: Change 'will not prevent' to 'may not prevent'.   (Due to the new fail-boot-on-error behavior)

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
       * This directory, or a subdirectory within this directory, should contain each of your custom situational configuration files.
       * If it doesn't, then this likely indicates your domain resource `configOverrides` was not set to match your custom override configuration map name, or that your custom override configuration map does not contain your override files.
 
+* If the Administration Server pod does start but fails to reach ready state:
```
tbarnes-us: Maybe 'ready state' --> 'ready state or tries to restart:'

tbarnes-us: Also, please move this bullet up above the previous bullet, so it appears just after the other 'does start' bullet.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -366,6 +371,14 @@ Incorrectly formatted override files may be accepted without warnings or errors
...
+  * Check for this message ` WebLogic server failed to start due to missing or invalid situational configuration files` in the Administration Server pod's `kubectl log`
```
tbarnes-us: Since message text can change over time, can you also mention the BEA-XXXXX value that corresponds with this message?

operator/src/main/resources/scripts/startServer.sh
tbarnes-us: Text LGTM.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -137,6 +149,9 @@ function copySitCfg() {
   if [ $? = 0 ]; then
     for local_fname in ${src_dir}/${fil_prefix}*.xml ; do
       copyIfChanged $local_fname $tgt_dir/`basename ${local_fname/${fil_prefix}//}`
+      trace "Printing contents of situational configuration file $local_fname:"
+      file_content=`cat $local_fname`
+      echo "$file_content"
```
tbarnes-us: Can 
```
file_content=`cat $local_fname`
echo "$file_content"``` 
be replaced with 
```
cat $local_fname```?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -110,7 +110,19 @@ function waitForShutdownMarker() {
...
   while true; do
+    if [ "$logged" != 'true' ] && grep -q "BEA-141335" ${SERVER_OUT_FILE} ; then
+      msg=("WebLogic server failed to start due to missing or invalid"
+           "situational configuration files, which may be due to invalid"
+           "configOverride templates (these are specified via the"
+           "configOverride attribute in the Domain custom resource)."
+           "For details, please search your pod log or"
+           "${SERVER_OUT_FILE} for the keyword 'situational'."
+          )
+      trace "${msg[*]}"
```
tbarnes-us: Two comments:
1) Looks like spaces are missing?  Do you need a space just before each trailing double-quote?
2) I keep going back and forth on this one - not sure if it's a concern - but I assume calling grep on the .out log every three seconds might actually incur significant overhead - maybe even enough to cause a big enough blip so as to affect latency sensitive apps. Maybe call this less often then 'doShutdown' is checked?

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: That is the new trace message we added to startServer.sh

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: Do you mean move this bullet to before line 370?

tbarnes-us: Yes.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
tbarnes-us: Oops.   I withdraw the comment.
(BTW There's actually a JIRA to number each of our script trace logs...)

operator/src/main/resources/scripts/startServer.sh
TheFrogPad: For Tom's #2, what about running another script in the background like tailLog.sh that does the grep at periodic interval and that interval could even have an environment variable to change the sleep window?

I'm not suggesting this to make things more complicated but from trying to decipher the whole server lifecycle handling the process looks rather intricate so having a script to check the log means that same script can also be run or even exec'd independent of the lifecycle or other k8s probes.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: Done

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: Fixed

alai8: Done

operator/src/main/resources/scripts/startServer.sh
alai8: 1. No need to add spaces in each line
2. Created monitorLog.sh script. Since we only monitor the fail to start server due to sit config error message, the script also monitor BEA-000360 (Server started) and exit when server has started. Sleep window by default is 30 seconds, and is configurable via environment variable.

operator/src/main/resources/scripts/startServer.sh
alai8: Reduced to one line:
echo `cat $local_fname`

operator/src/main/resources/scripts/startServer.sh
TheFrogPad: Nice! This looks much cleaner too...

docs-source/content/userguide/managing-domains/configoverrides/_index.md
```diff
@@ -362,9 +367,18 @@ Incorrectly formatted override files may be accepted without warnings or errors
...
+      * Lines containing the String `situational` may be found in the Administration Server pod log to provide more hints. 
+      * For example:
+        * `<Jun 20, 2019 3:48:45 AM GMT> <Warning> <Management> <BEA-141323> <The situational config file has an invalid format, it is being ignored: XMLSituationalConfigFile[/shared/domains/domain1/optconfig/jdbc/testDS-0527-jdbc-situational-config.xml] because org.xml.sax.SAXParseException; lineNumber: 8; columnNumber: 3; The element type "jdbc:jdbc-driver-params" must be terminated by the matching end-tag "</jdbc:jdbc-driver-params>".`
+      * The warning message suggests a syntax error is found in the provided configuration override file for the testDS JDBC datasource.
+  
+* Look in your `DOMAIN_HOME/optconfig` directory.
+  * This directory, or a subdirectory within this directory, should contain each of your custom situational configuration files.
+  * If it doesn't, then this likely indicates your domain resource `configOverrides` was not set to match your custom override configuration map name, or that your custom override configuration map does not contain your override files.
```
tbarnes-us: I think this bullet still belongs in its original location.

tbarnes-us: E.g. current lines 379-381 are more clear if they stay in their original location...

alai8: move lines 379-381 back to before line 371?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -137,6 +142,8 @@ function copySitCfg() {
   if [ $? = 0 ]; then
     for local_fname in ${src_dir}/${fil_prefix}*.xml ; do
       copyIfChanged $local_fname $tgt_dir/`basename ${local_fname/${fil_prefix}//}`
+      trace "Printing contents of situational configuration file $local_fname:"
+      echo `cat $local_fname`
```
tbarnes-us: Why ```echo `cat $local_fname` ``` instead of just ``` cat $local_fname ```?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -102,6 +102,11 @@ function waitUntilShutdown() {
     trace "Showing the server out file from ${SERVER_OUT_FILE}"
     ${SCRIPTPATH}/tailLog.sh ${SERVER_OUT_FILE} &
   fi
+  FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR=${FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR:-true} 
+  SERVER_OUT_MONITOR_INTERVAL=${SERVER_OUT_MONITOR_INTERVAL:-30}
```
tbarnes-us: Since the monitor script only runs until the server reports that it started successfully - (I don't recall the original change did that - maybe I missed that) - it seems like it'd be OK to run the monitor more frequently than every 30 seconds.

alai8: It was added in the latest change. any suggestion on interval? 10 seconds?

tbarnes-us: Maybe even back to 3 seconds?  Boots can occur pretty quick these days.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
tbarnes-us: back to being a sub-bullet of the 'If WebLogic pods do start, then:' bullet.

operator/src/main/resources/scripts/startServer.sh
alai8: Updated interval back to 3 seconds.

operator/src/main/resources/scripts/startServer.sh
alai8: agree. not sure why didn't do that earlier.

docs-source/content/userguide/managing-domains/configoverrides/_index.md
alai8: moved

PR #1118, opened 2019-06-18
Resolve failure in introspection following a console edit

rjeberhard:
We discovered that when a domain is restarted following an edit of configuration using the console, that the introspector script would sometimes fail reading beans that were no longer present.

This change wraps those bean lookups in a function that handles the exceptions better.

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.py
+++ b/operator/src/main/resources/scripts/introspectDomain.py
@@ -1,4 +1,4 @@
-# Copyright 2018, Oracle Corporation and/or its affiliates. All rights reserved.
+# Copyright 2018, 2019, Oracle Corporation and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 #
 # ------------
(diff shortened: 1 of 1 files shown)


rjeberhard:
Ha, ha... It wouldn't show in the diff @markxnelson , but there was already a function there with that naming pattern.  I'm more than happy to change it.  It's "WA" for "work around" ;)

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -325,7 +332,7 @@ def validateAdminServer(self):
       addError("The admin server '" + adminServerName + "' does not exist.")
       return
     cluster = adminServer.getCluster()
```
tbarnes-us: Do we need a 'WA(' method for 'getCluster()' too?

rjeberhard: I wondered about that...  That method is all over and we've not seen this use case fail on getCluster().

markxnelson: ah, i was trying to work out who "wrapper" became "WA"

markxnelson: what about something like getClusterOrNone
or a more generic getOrElse( thingYouWant, None )

markxnelson: not a biggie

tbarnes-us: Maybe ignore this comment.  It'd be painful to replace every mbean get in this script with a 'WA(' variant, so maybe we can assume the 'WA' issue only applies to certain mbeans (log, ssl, and dynamic)?   

Any thoughts Johnny? @jshum2479

rjeberhard: I like the "OrNone" option

PR #1127, opened 2019-06-26
PR (ready to be merged): test for sharing PV across domains

maggiehe00:
This test is to create 2 domains that share the same PV/PVC. Both domains are  in the same namespace (default) and  managed by a single operator.  Verify
- scaling for domain2 cluster from 2 to 3 servers and back to 2, plus verify no impact on domain1
- cycle domain1 down and back up, plus verify no impact on domain2

After merging to the latest develop branch
Internal Jenkins passed at: http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/1883/
external Jenkins passed at: http://build.weblogick8s.org:8080/job/weblogic-kubernetes-operator-quicktest/420/


vanajamukkara:
Does shared PV work if two domains managed by two operators? If so, can the existing test testTwoDomainsManagedByTwoOperators be modified for this usecase instead of bringing up two new domains?

maggiehe00:
In order to share one PV 2 domains need to in the same namespace.

markxnelson:
is this even a supported use case?

mriccell:
Do we not support two domain homes in the same  PV but in two different directories?  I would think yes.

mriccell:
> is this even a supported use case?

I think the answer is yes.  You could two domain homes persisted to the same PV but in separate Domain Homes because the names are unique.

vanajamukkara:
@maggiehe00 how about the weblogicDomainStoragePath, isn't that same for both domains when they share PV?

rjeberhard:
PV's are globally scoped, but PVC's are per-namespace.  We would have to experiment to see if it's possible to have a shared PV with different PVC's in the different namespaces.  It's likely possible and it makes sense if the volume is accessed over NFS and the domains are kept in separate directory structures.

maggiehe00:
@rjeberhard, right now the PVCs of both domains are in the same namespace. I'm going to modify the tests to put them in the different PVC and give it a try.

maggiehe00:
@rjeberhard, i tried a couple of things. These are my findings.
1) under our current sample PV/PVC creation template

kind: PersistentVolumeClaim
apiVersion: v1
metadata:
 name: %DOMAIN_UID%%SEPARATOR%%BASE_NAME%-pvc
apiVersion: v1
kind: PersistentVolume
metadata:
 name: %DOMAIN_UID%%SEPARATOR%%BASE_NAME%-pv

So if PV name is same, PVC name must be same also, given for sharing PV, we have to leave domain uid empty.
2) I tried same PV/PVC name for 2 domains(namespace: default) but 2nd PVC in the different namespace(d2pvc). Both domains are created and tests passed but in fact the PVC that in the namespace d2pvc didn't bound to PV. 2 domains use the same PV/PVC.
ash-4.2$ kubectl get pv
NAME                 CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                         STORAGECLASS                    REASON   AGE
weblogic-shaing-pv   10Gi       RWX            Retain           Bound    default/weblogic-shaing-pvc   weblogic-shaing-storage-class            57m

I think this is because in the domain creation input file there is no option to specify the namespace of PVC.

Please let me know how you think about it. Thanks!

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
```diff
@@ -873,12 +874,16 @@ protected void createPV() throws Exception {
                     + "/samples/scripts/create-weblogic-domain-pv-pvc/create-pv-pvc-inputs.yaml"));
     pvMap = yaml.load(pv_is);
     pv_is.close();
-    pvMap.put("domainUID", domainUid);
 
-    // each domain uses its own pv for now
-    if (domainUid != null)
+    logger.info("pvSharing for this domain is: " + pvSharing);
+    if (!pvSharing) pvMap.put("domainUID", domainUid);
+    else pvMap.put("baseName", "weblogic-shaing");
```
vanajamukkara: typo weblogic-sharing

integration-tests/src/test/resources/domainonsharingpvwlst.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
```
vanajamukkara: you can remove 2018 as its a new file

integration-tests/src/test/java/oracle/kubernetes/operator/utils/Domain.java
maggiehe00: Done. Thanks!

integration-tests/src/test/resources/domainonsharingpvwlst.yaml
maggiehe00: Done. Thanks!

maggiehe00:
@vanajamukkara, @npanigr, @rjeberhard I hope this PR can be merged soon so that I can use it to test the fix for Bug29967859:PV/PVC CREATION LOGIC CAN'T PRECISELY DETECT IF PV/PVC EXISTS OR NOT

PR #1130, opened 2019-06-28
WIP: Fix SSL Managed Server start issue due to incorrect ADMIN_URL

pfmackin:
Fix SSL by generating a correct ADMIN_URL, which is used by the managed server to connect to the admin server.  The existing code uses http with the SSL port resulting in a connection failure.

This is WIP, do not merge

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -177,7 +177,25 @@ public abstract class PodStepContext extends StepContextBase {
         .getLocalAdminProtocolChannelPort();
   }
 
+  /**
+   * Check if the server is listening on a secure port. NOTE: If the targetted server is a managed
+   * server, this methd is overriden to check if the managed server has a secure listen port rather
+   * than the admin server. See PodHelper.ManagedPodStepContext
+   *
+   * @return true if server is listening on a secure port
+   */
   boolean isLocalAdminProtocolChannelSecure() {
+    return domainTopology
+        .getServerConfig(domainTopology.getAdminServerName())
+        .isLocalAdminProtocolChannelSecure();
+  }
+
+  /**
+   * Check if the admin server is listening on a secure port.
+   *
+   * @return true if admin server is listening on a secure port
+   */
+  private boolean isAdminServerProtocolChannelSecure() {
     return domainTopology
         .getServerConfig(domainTopology.getAdminServerName())
         .isLocalAdminProtocolChannelSecure();
@@ -805,6 +823,9 @@ public abstract class PodStepContext extends StepContextBase {
     addEnvVar(vars, "ADMIN_PORT", getAsPort().toString());
     if (isLocalAdminProtocolChannelSecure()) {
       addEnvVar(vars, "ADMIN_PORT_SECURE", "true");
+    }
+    if (isAdminServerProtocolChannelSecure()) {
+      addEnvVar(vars, "ADMIN_SERVER_PORT_SECURE", "true");
     }
     addEnvVar(vars, "SERVER_NAME", getServerName());
     addEnvVar(vars, "DOMAIN_UID", getDomainUID());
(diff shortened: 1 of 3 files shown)


operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -246,16 +246,13 @@ EOF
...
 
   if [ ! "${ADMIN_NAME}" = "${SERVER_NAME}" ]; then
-    admin_protocol="http"
-    if [ "${ADMIN_PORT_SECURE}" = "true" ]; then
-      admin_protocol="https"
-    fi  
-    echo "AdminURL=$admin_protocol\\://${AS_SERVICE_NAME}\\:${ADMIN_PORT}" >> ${wl_props_file}
+    ADMIN_URL=$(getAdminUrl)
+    export ADMIN_URL
+    trace "ADMIN_URL"=$ADMIN_URL >> ${wl_props_file}
```
tbarnes-us: Is any other script using the ADMIN_URL env var?  If not, it shouldn't be exported.

The props file entry needs to be in a special format that's acceptable to the NM, so line 251 should be:
  echo "AdminURL=$ADMIN_URL" >> ${wl_props_file}
or even
  echo "AdminURL=$(getAdminUrl)" >>  ${wl_props_file}
not
  trace "ADMIN_URL"=$ADMIN_URL >> ${wl_props_file}

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -806,6 +824,9 @@ void overrideContainerWeblogicEnvVars(List<V1EnvVar> vars) {
     if (isLocalAdminProtocolChannelSecure()) {
       addEnvVar(vars, "ADMIN_PORT_SECURE", "true");
     }
+    if (isAdminServerProtocolChannelSecure()) {
+      addEnvVar(vars, "ADMIN_SERVER_PORT_SECURE", "true");
```
TheFrogPad: Per discussion, looks like we are creating a new env variable when we already have one.

Also, we should see about being consistent in the naming, for example:
- SERVER_NAME
- SERVER_PORT
- SERVER_PORT_SECURE
- SERVER_SERVICE_NAME
- ADMIN_NAME
- ADMIN_PORT
- ADMIN_PORT_SECURE
- ADMIN_SERVICE_NAME
- ...

and then the scripts can always know the local + admin server content

operator/src/main/resources/scripts/traceUtils.sh
```diff
@@ -104,6 +104,19 @@ function checkEnv() {
   return 0
 }
 
+#
+# getAdminUrl
+#   purpose: Get the ADMIN_URL used to connect to the admin server internally.
+#   sample:
+#     ADMIN_URL=$(getAdminUrl)
+#
+function getAdminUrl() {
```
TheFrogPad: Would be good to document all the env vars used and even consider one for the local server as well?

TheFrogPad:
On the HNV property, we should consider how to pass this property given that the DemoIdentity,jks would not have a valid hostname for each POD. Also, if there was customer supplied certificates that trust property may be required as well.

We should discuss these implications and check with @alai8 as I think he added a property for FMW/JRF when server boots with KSS/JRF demo certificate.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
TheFrogPad: I also realized that changing names may cause upgrade compatibility issue when going operator.current to operator.next so even some doc/readme on the various env vars will help?

TheFrogPad: After revisiting this PR today I think we do need the new ADMIN_SERVER_PORT_SECURE variable (along with the new method) because on the managed server there needs to a property used by the scripts to know if the local managed server port is secure in addition to knowning if the admin server needs to be connected to over a secure port with https.

TheFrogPad:
For the ignore HNV property that is needed because POD names and generated DemoIdentity.jks will not align, I think the best course for now would be document the need to add the system property to the javaOptions on the Domain CRD.

We should get the fix in that allows for booting the domain with SSL/TLS enabled and then work out best approach for customers to configure SSL/TLS for the domain (i.e. non-Demo setup).

bsmoyers:

PR #1137, opened 2019-07-10
Add WebLogic version and WebLogic patch checks

tbarnes-us:
This change adds WebLogic version and WebLogic patch checks.

- It fails the introspector job and wl pod startup for unsupported versions.

   - Requires 12.2.1.3 or higher.

   - If 12.2.1.3, also requires patch 29135930.  

- The failure messages are logged to job/pod's log and include the name of an env var users can specify in the Domain resource to bypass checks: 

  - `"Error: WebLogic version='$cur_wl_ver' and the Operator requires WebLogic version '${exp_wl_ver}' or higher. To bypass this check, set env var SKIP_WL_VERSION_CHECK to 'true'."` 

  - `Error: The Operator requires that WebLogic version '12.2.1.3' have patch '$exp_wl_12213_patches'. To bypass this check, set env var SKIP_WL_VERSION_CHECK to 'true'.`

- Instead of using `weblogic.version` or `opatch`, this check determines the WebLogic version by parsing the installer's inventory file at `$ORACLE_HOME/inventory/registry.xml`.  This is because `weblogic.version` or `opatch` can take a few seconds to run, which is too long.  _If the parse itself fails, it ignores the problem and simply assumes the version is fine and reports the following Info message._

  - `@[2019-07-12T17:45:03.584 UTC][bash:65] Info: Could not determine WebLogic version. Assuming version is fine. (The Operator requires WebLogic version '12.2.1.3' or higher, and also requires patches '29135930' for version '12.2.1.3'.).`

### DIFF
--- a/operator/src/main/resources/scripts/traceUtils.sh
+++ b/operator/src/main/resources/scripts/traceUtils.sh
@@ -231,7 +231,8 @@ getWebLogicVersion()
 
   [ ! -f $reg_file ] && echo "9999.9999.9999.9999" && return
 
-  local wlver="`grep 'name="WebLogic Server".*version=' $reg_file \
+  # The following grep captures both "WebLogic Server" and "WebLogic Server for FMW"
+  local wlver="`grep 'name="WebLogic Server.*version=' $reg_file \
                | sed 's/.*version="\([0-9.]*\)".*/\1/g'`"
 
   echo ${wlver:-"9999.9999.9999.9999"}
--- a/operator/src/main/resources/scripts/traceUtils.sh
+++ b/operator/src/main/resources/scripts/traceUtils.sh
@@ -164,3 +164,107 @@ function exportEffectiveDomainHome() {
   trace "Error: More than one config.xml found at DOMAIN_HOME/config/config.xml and DOMAIN_HOME/*/config/config.xml, DOMAIN_HOME='$DOMAIN_HOME': ${found_configs}. Configure your 'domainHome' setting in your WebLogic Operator Domain resource to reference a single WebLogic domain."
   return 1
 }
+
+
+# versionCmp
+#   Compares two wl versions $1 $2 up to the length of $2
+#     Expects form N.N.N.N
+#     Uses '0' for an N in $1 if $1 is shorter than $2
+#   echo "1"  if v1 >  v2
+#   echo "-1" if v1 <  v2
+#   echo "0"  if v1 == v2
+versionCmp()
+{
+  IFS='.' read -r -a v1_arr <<< "`echo $1`"
+  IFS='.' read -r -a v2_arr <<< "`echo $2`"
+
+  for i in "${!v2_arr[@]}"
+  do
+    [ ${v1_arr[i]:-0} -gt ${v2_arr[i]:-0} ] && echo "1" && return
+    [ ${v1_arr[i]:-0} -lt ${v2_arr[i]:-0} ] && echo "-1" && return
+  done
+  echo "0"
+}
+
+# versionGE
+#   return success if WL v1 >= v2
+versionGE()
+{
+  [ `versionCmp "$1" "$2"` -ge 0 ] && return 0
+  return 1
+}
+
+# versionEQ
+#   return success if v1 == v2
+versionEQ()
+{
+  [ `versionCmp "$1" "$2"` -eq 0 ] && return 0
+  return 1
+}
+
+# hasWebLogicPatches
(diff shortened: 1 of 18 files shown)


operator/src/main/resources/scripts/traceUtils.sh
```diff
@@ -164,3 +164,107 @@ function exportEffectiveDomainHome() {
...
+  for pnum in "$@"; do
+    grep --silent "patch-id=\"$1\"" $reg_file || return 1
+  done
+}
+
+# getWebLogicVersion
+#   parse wl version from install inventory
+#   - if we can't get a version number then we return
+#     a high dummy version number that's sufficient
+#     to pass version checks "9999.9999.9999.9999"
```
alai8: do we log error if unable to find WLS version number?

operator/src/main/resources/scripts/traceUtils.sh
```diff
@@ -164,3 +164,107 @@ function exportEffectiveDomainHome() {
...
+#   - we parse the install inventory as this is far faster than
+#     using opatch or weblogic.version
+getWebLogicVersion()
+{
+  local reg_file=$ORACLE_HOME/inventory/registry.xml
+
+  [ ! -f $reg_file ] && echo "9999.9999.9999.9999" && return
+
+  local wlver="`grep 'name="WebLogic Server".*version=' $reg_file \
```
alai8: The registry.xml entry in FMW image ( container-registry.oracle.com/middleware/fmw-inf:12.2.1.3) and it is different: 
      <distribution status="installed" name="WebLogic Server for FMW" version="12.2.1.3.0">

operator/src/main/resources/scripts/traceUtils.sh
```diff
@@ -164,3 +164,107 @@ function exportEffectiveDomainHome() {
   trace "Error: More than one config.xml found at DOMAIN_HOME/config/config.xml and DOMAIN_HOME/*/config/config.xml, DOMAIN_HOME='$DOMAIN_HOME': ${found_configs}. Configure your 'domainHome' setting in your WebLogic Operator Domain resource to reference a single WebLogic domain."
   return 1
 }
+
```
alai8: Should we rename traceUtils.sh now that it is not just for trace utilities? Or perhaps we should update the "Purpose" section in the beginning of the file.

operator/src/main/resources/scripts/traceUtils.sh
tbarnes-us: No.  I prefer not to since parsing the inventory file is not a 'standard' way to get the version number and I'm not 100% sure it's reliable.  I think this is OK since checking the version number isn't critical, and we don't want to fail the pod/job or unnecessarily concern end-users if the check itself cannot determine the version number...

alai8: Yes, agree that failing pod/job for this is not appropriate. What about just logging a message? 
Would users be surprised to see version number "9999.9999.9999.9999" in the trace message "trace "Info: WebLogic version='$cur_wl_ver'. Version check passed. (The Operator requires WebLogic version '${exp_wl_ver}' or higher)."?

tbarnes-us: OK. I'll update to report a different message - something like  "Could not determine WebLogic Version, so assuming that it's correct."

operator/src/main/resources/scripts/traceUtils.sh
tbarnes-us: Ah - you're right.  It adds ' for FMW'!  Good catch!  I've modified the grep expression from `name="WebLogic Server".*` to `name="WebLogic Server.*`.

tbarnes-us: I also ran a little test with the FMW image to see if the version was obtained - it was...

operator/src/main/resources/scripts/traceUtils.sh
tbarnes-us: Good point.  I'll go ahead and do both.

operator/src/main/resources/scripts/traceUtils.sh
tbarnes-us: I've made the change.  Please resolve if you're fine with it.

operator/src/main/resources/scripts/traceUtils.sh
tbarnes-us: I've made the changes, please resolve if you're fine with it.

PR #1143, opened 2019-07-17
Modify domain resource to have "istio" as top-level element

rjeberhard:


tbarnes-us:
There are comments in https://github.com/oracle/weblogic-kubernetes-operator/pull/1086 that may also apply to this pull.

docs/domains/Domain.md
```diff
@@ -168,6 +175,12 @@ ServerPod describes the configuration for a Kubernetes pod for a server.
 | `annotations` | Map | The annotations to be attached to generated resources. |
 | `labels` | Map | The labels to be attached to generated resources. The label names must not start with 'weblogic.'. |
 
+### Istio
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `istioEnabled` | Boolean | True, if this domain is deployed under an Istio service mesh. Defaults to true. Not required. |
```
tbarnes-us: Move this to the new Experimental section?

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -96,3 +96,6 @@ elasticSearchHost: "elasticsearch.default.svc.cluster.local"
 # elasticSearchPort specifies the port number of where elasticsearch is running.
 # This parameter is ignored if 'elkIntegrationEnabled' is false.
 elasticSearchPort: 9200
+
+# istioEnabled specifies whether or not it is deployed under Istio.
```
tbarnes-us: add '(experimental)' at the end of the comment on line 100

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -133,3 +134,6 @@ createDomainScriptName: create-domain-job.sh
 # Kubernetes config map, which in turn is mounted to the `createDomainScriptsMountPath`,
 # so that the Kubernetes pod can use the scripts and supporting files to create a domain home.
 createDomainFilesDir: wlst
+
+# istioEnabled specifies whether or not it is deployed under Istio.
```
tbarnes-us: add '(experimental)' to comment

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -853,7 +853,7 @@ def customizeNetworkAccessPoint(self, nap, listen_address):
     # FWIW there's theoretically no need to 'add' or 'replace' when empty
     #   since the runtime default is the server listen-address.
     nap_name=nap.getName()
-    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0:
+    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0 and not (nap_name.startswith('istio-')) and not (nap_name == 'T3Channel'):
```
tbarnes-us: Checking for the specific nap name 'T3Channel' looks iffy - I think this is the name we use in our sample and suspect it should be keyed keyed off of the ISTIO_ENABLED flag - something like:

`if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0 and not (nap_name.startswith('istio-')) and ((self.env.ISTIO_ENABLED == 'true') or not (nap_name == 'T3Channel')):`

assuming that `self.ISTIO_ENABLED = self.getEnvOrDef('ISTIO_ENABLED', 'false')` has been added in the global var init section at the top of this script

kubernetes/samples/scripts/common/domain-template.yaml
```diff
@@ -47,6 +47,8 @@ spec:
       value: "%JAVA_OPTIONS%"
     - name: USER_MEM_ARGS
       value: "-XX:+UseContainerSupport -Djava.security.egd=file:/dev/./urandom "
+    - name: ISTIO_ENABLED
+      value: "%ISTIO_ENABLED%"
```
tbarnes-us: We shouldn't be requiring users to set this value - the Operator should set it.  If we want to keep it here, maybe rename it ISTIO_ENABLED_EXPERIMENTAL or at least put a clear comment above the env var stating that it's experimental.

PR #1149, opened 2019-07-19
OWLS-74263 - NPE in operator logs during 50 domains test case

alai8:
Issue 1: NPE and ConcurrentModificationExceptions reported in OWLS-74263 and OWLS-74220
NullPointerException and ConcurrentModificationExceptions are seen in opertor logs with DomainStatusUpdater in the call stack.

Fix: synchronize various domain status updates from DomainStatusUpdater

Issue 2: SEVERE level log messages are seen in the operator log
"*kubernetes.client.ApiException: Bad Request* at
io.kubernetes.client.ApiClient.handleResponse(ApiClient.java:886)
...
"status\":\"Failure\",\"message\":\"container
\\\"domain42-introspect-domain-job\\\" in pod
\\\"domain42-introspect-domain-job-6lp9s\\\" is not
available\",\"reason\":\"BadRequest\",\"code\":400}

This is caused by the introspector pod being terminated due to DeadlineExceeded, thus the container is no
longer available for the read pod log operation.

Fix: by not performing rest of the introspector steps starting with ReadDomainInstrospectorPodLogStep if
the job has status of Failed and reason of DeadlineExceeded


operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -231,14 +243,21 @@ public NextAction apply(Packet packet) {
...
                   if (!shouldProcessJob(job)) {
                     return;
                   }
                   completeCallbackRegistrations.remove(job.getMetadata().getName());
                   if (didResume.compareAndSet(false, true)) {
                     LOGGER.fine("Job status: " + job.getStatus());
                     packet.put(ProcessingConstants.DOMAIN_INTROSPECTOR_JOB, job);
+                    // Do not proceed to next step such as ReadDomainIntrospectorPodLog if job
+                    // failed due to DeadlineExceeded, as the pod container would likely not
+                    // be available for reading
```
tbarnes-us: Hmm.  It seems like there's a good chance that the timeout could be because the job's scripts themselves were running slowly.  I assume the pod log would still be available in this case?  If so, it'd still be useful to at least echo the contents of the pod log as a 'LOGGER.fine'.

tbarnes-us:
The change looks fine to me apart from my one relatively minor comment.   I'm not an Operator java expert though, so please take that with a grain of salt.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
alai8: unfortunately not as k8s has started terminating the pod. And the readPodLog call that follows would result in "ApiException: Bad Request" error quoted in the PR comment.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -283,4 +302,17 @@ public NextAction onSuccess(
...
+  static class DeadlineExceededException extends Exception {
+    final String job;
+
+    public DeadlineExceededException(String job) {
+      super();
+      this.job = job;
+    }
+
+    public String toString() {
+      return "Job " + job + " failed. Reason: DeadlineExceeded";
```
tbarnes-us: Could we include more information to help the user figure out the next steps?  For example:

- the deadline seconds and a hint about how to increase it  

- the number of seconds that passed since the job was submitted

- maybe something like `Ensure all domain dependencies have been deployed (any secrets, config-maps, PVs, and PVCs that the domain resource references).`  (I assume a missing dependency would be a common cause of exceeding the dead-line?)

- And finally even `kubectl describe the job and its pod for more job failure information`.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
tbarnes-us: ah, ok, resolving

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
rjeberhard: This is good feedback; yet, I'd prefer to do this work as part of a different PR specifically around diagnosing introspector job failures.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -283,4 +302,39 @@ public NextAction onSuccess(
...
+    final V1Job job;
+
+    public DeadlineExceededException(V1Job job) {
+      super();
+      this.job = job;
+    }
+
+    public String toString() {
+      StringBuilder sb = new StringBuilder("Job ")
+          .append(job.getMetadata().getName()).append(" failed due to reason: DeadlineExceeded.")
```
rjeberhard: Sigh... Now we need a change as this needs to be internationalized.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -283,4 +302,39 @@ public NextAction onSuccess(
...
+          .append(" ActiveDeadlineSeconds of the job is configured with "
+              + job.getSpec().getActiveDeadlineSeconds()
+              + " seconds.")
+          .append(getJobStartedSecondsMessage())
+          .append(" Ensure all domain dependencies have been deployed")
+          .append(" (any secrets, config-maps, PVs, and PVCs that the domain resource references).")
+          .append(" Use kubectl describe the job and its pod for more job failure information.")
```
tbarnes-us: describe the job --? for the job

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
rjeberhard: You can use code like here to format using the a message in our resource bundle: https://docs.oracle.com/javase/tutorial/i18n/format/messageFormat.html.  Our bundle is Operator.properties and the keys are in MessageKeys.java.

PR #1154, opened 2019-07-23
Owls 75227 - simplify debugging the operator

russgold:
This change:
* simplifies putting the operator into debug by disabling the probes when debug is enabled, and
* allows the operator to be set not to suspend, meaning that a developer can attach after it is running. This is controlled by the `haltOnDebugStartup` setting.

Note that the default behavior is, as before, that the operator halts on startup when debug is enabled. I think that the default should be changed to _not_ halt, but did not make that change, here. @rjeberhard what do you think?

### DIFF
--- a/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
+++ b/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
@@ -40,7 +40,7 @@ spec:
         - name: "REMOTE_DEBUG_PORT"
           value: {{ .internalDebugHttpPort | quote }}
         - name: "DEBUG_SUSPEND"
-          {{- if .haltOnDebugStartup }}
+          {{- if .suspendOnDebugStartup }}
           value: "y"
           {{- else }}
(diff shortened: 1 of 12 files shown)


pfmackin:
From a functional viewpoint it looks like there are 2 changes:

1) Automatically disable liveness / readiness probe if remote debugging enabled
2) Allow operator to run without halting in debug mode.

Is that correct?

russgold:
Yes; that's correct.

rjeberhard:
> I think that the default should be changed to not halt, but did not make that change, here. @rjeberhard what do you think?

I agree.  When we were first working with the operator, we had to frequently debug operator start-up.  Now, that is much less common and debugging use cases are almost always of the operator's response to some later change (namespace deleted, domain resource changed, etc.)  So, I'd be fine with the default being that the operator doesn't suspect.

I would prefer saying "suspend on startup" or not as "halt" sounds permanent.

russgold:
Replaced with PR #1156

PR #1155, opened 2019-07-24
Pod/job script logging improvements

tbarnes-us:
- Reflect each introspector job log message in the operator log using the same log-level
- Ensure job/pod log messages that are generated by scripts (a) use same log-levels as the operator, and (b) can be parsed by the operator if needed
- If a bash script 'trace' call doesn't specify a log-level, try deduce it from the first word in the log text, and if that doesn't help, assume the log-level is FINE
- **Note:** Since the Operator does not have an 'error' log level, we convert any 'error' to a 'severe'.
- **Note:** If a script log message has no log level, we assume it's a 'fine'.

Sample Operator output when job fails due to a missing DOMAIN_HOME:
```
[slc16okc weblogic-kubernetes-operator]$ ../notes/oplog.sh -ll SEVERE
@@ Reading log from 'kubectl logs weblogic-operator-5c96bb9489-794vs -n weblogic-operator1 ':
 1:"07-24-2019T15:03:57.782" 4:"" 5:"SEVERE" 9:"Job domainonpvwlst-introspect-domain-job has failed" 10:""
 1:"07-24-2019T15:03:57.854" 4:"domainonpvwlst" 5:"SEVERE" 9:"Introspector Job Log: @[2019-07-24T15:03:57.523 UTC][introspectDomain.sh:74][SEVERE] Missing DOMAIN_HOME directory '/shared/domains/domainonpvwlstDNE'." 10:""
```

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -476,7 +476,7 @@ public class JobHelper {
         }
       }
     } catch (IOException ioe) {
-      LOGGER.fine("Unexpected exception " + ioe + " while parsing string:\n" + jobLogs);
+      LOGGER.fine(MessageKeys.JOB_LOG_PARSE_FAILURE, ioe.toString(), jobLogs);
     }
   }
 }
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -4,6 +4,9 @@
 
 package oracle.kubernetes.operator.helpers;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
 import java.util.List;
 
 import io.kubernetes.client.models.V1DeleteOptions;
@@ -33,6 +36,7 @@ import oracle.kubernetes.weblogic.domain.model.ConfigurationConstants;
 import oracle.kubernetes.weblogic.domain.model.Domain;
 import oracle.kubernetes.weblogic.domain.model.DomainSpec;
 import oracle.kubernetes.weblogic.domain.model.ManagedServer;
+
 
 public class JobHelper {
 
@@ -329,6 +333,11 @@ public class JobHelper {
       // Log output to Operator log
       LOGGER.fine("+++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
 
+      // Parse out each job log message and log to Operator as SEVERE, FINE, etc...
+      if (result != null) {
+        convertJobLogsToOperatorLogs(result);
+      }
+
       V1Job domainIntrospectorJob = (V1Job) packet.get(ProcessingConstants.DOMAIN_INTROSPECTOR_JOB);
       if (domainIntrospectorJob != null && JobWatcher.isComplete(domainIntrospectorJob)) {
         if (result != null) {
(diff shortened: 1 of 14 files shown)


operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -309,11 +312,11 @@ while [ 1 -eq 1 ]; do
     break
   fi
   if [ $((SECONDS - $start_secs)) -ge $max_wait_secs ]; then
-    trace "Info: Contents of node manager log '$nodemgr_log_file':"
+    trace INFO "Contents of node manager log '$nodemgr_log_file':"
```
tbarnes-us: @rjeberhard @alai8 @lennyphan Should I make lines 315 & 317 SEVERE instead of INFO?  

My thinking was that the SEVERE at line 319 was sufficient to get folks to dig deeper into the pod log and fine the INFO messages...

tbarnes-us: (The INFO messages will contain the output from the following 'cat' when they are parsed and added to the operator log.)

alai8: Agree. I think the SEVERE at line 319 should be sufficient.

operator/src/main/resources/scripts/stopServer.sh
```diff
@@ -11,7 +11,7 @@
 
 SCRIPTPATH="$( cd "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
 source ${SCRIPTPATH}/utils.sh
-[ $? -ne 0 ] && echo "Error: missing file ${SCRIPTPATH}/utils.sh" && exit 1
+[ $? -ne 0 ] && echo "[SEVERE] Missing file ${SCRIPTPATH}/utils.sh" && exit 1
```
alai8: I assume we can't use trace here for some reason?

tbarnes-us: Correct.  This echo is reporting that the we couldn't load the utils.sh file which contains the 'trace' function.

operator/src/main/resources/scripts/readState.sh
```diff
@@ -26,7 +26,7 @@ if [ `jps -v | grep -c " -Dweblogic.Name=${SERVER_NAME} "` -eq 0 ]; then
 fi
 
 if [ ! -f ${STATEFILE} ]; then
-  trace "Error: WebLogic Server state file not found."
+  trace "WebLogic Server state file not found."
```
alai8: Should this be SEVERE?

operator/src/main/resources/scripts/utils.py
```diff
@@ -2,10 +2,15 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 
 # Usage: trace('string')
+#        trace(logLevel,'string')
 #
-# This matches format of bash utils.sh trace, and the operator's log format.
+# Valid values for logLevel are SEVERE|WARNING|ERROR|INFO|CONFIG|FINE|FINER|FINEST
+#    'ERROR' is converted to 'SEVERE'
```
alai8: Do we need to keep the ERROR log level?

tbarnes-us: I lean towards keeping it.  I kept it around as an extra precaution in case someone used it by accident.  It's just too easy to assume 'error' is a valid, and since bash isn't compiled such a mistake wouldn't be caught until runtime - which is too late for a rare error path...

operator/src/main/resources/scripts/readState.sh
tbarnes-us: I don't think so.  I removed it since this file is called by stopServer.sh, which assumes things are OK when readState.sh itself returns non-zero.

operator/src/main/resources/scripts/utils.py
alai8: ok

operator/src/main/resources/scripts/readState.sh
alai8: I think readState.sh is also called from ServerStatusReader.java

tbarnes-us: I just checked ServerStatusReader.java, and it assumes that it's normal to hit this path:
```
               proc = kubernetesExec.exec("/weblogic-operator/scripts/readState.sh");
               ...
               int exitValue = proc.exitValue();
               ...
                } else if (exitValue == 1 || exitValue == 2) {
                  state =
                      PodHelper.isDeleting(pod)
                          ? WebLogicConstants.SHUTDOWN_STATE
                          : WebLogicConstants.STARTING_STATE;
                } else {
```

tbarnes-us: So it looks like this path also doesn't merit an 'error' or 'severe' at line 29.

rjeberhard: Yes, if we read the status very early in the life of the pod.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -406,4 +415,68 @@ public NextAction onSuccess(Packet packet, CallResponse<V1PodList> callResponse)
       return doNext(packet);
     }
   }
+
+  // Convert a job log message to an operator log message
+  //   - assumes the messages contins a string like '[SEVERE]', etc, if
+  //     not, then assumes the log level is 'FINE'.
+  private static void logToOperator(StringBuffer logStr, StringBuffer extraStr) {
+    String logMsg = "Introspector Job Log: "
```
rjeberhard: i18n.  Can this please come from the resource bundle?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -406,4 +415,68 @@ public NextAction onSuccess(Packet packet, CallResponse<V1PodList> callResponse)
...
+            }
+            logExtra.append(line);
+          }
+        }
+        if (logString.length() > 0) { 
+          logToOperator(logString, logExtra);
+        }
+      }
+    } catch (IOException ioe) {
+      LOGGER.fine("Unexpected exception " + ioe + " while parsing string:\n" + jobLogs);
```
rjeberhard: i18n

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -930,7 +930,7 @@ def __init__(self, env):
         self.macroStr+=', '
       self.macroStr+='${' + key + '}'
 
-    trace("available macros: '" + self.macroStr + "'")
+    trace("Available macros: '" + self.macroStr + "'")
```
rjeberhard: Hmmm... okay, I guess we don't have any of the introspector internationalized. :(

rjeberhard:
I'm okay punting on the i18n concerns since the whole introspector isn't coming from the resource bundle.  It "would be good" if the Java code was at least internationalized so that we can easily find places that need to be fixed later.

operator/src/main/resources/scripts/introspectDomain.py
tbarnes-us: Yah. You and I lowered the priority of doing that type of work a week or two ago, only relatively 'P1' priority stuff is in this check-in (see OWLS-74901).

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: I'm thinking this should be a string that never changes and is easy to grep out of the log - so it's not a straightforward i18n candidate IMO.   Can you suggest something that can be hard-coded?   How about "[INTROSPECTORJOB] "?

tbarnes-us: @rjeberhard

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: Updated.

PR #1157, opened 2019-07-24
smarter defaulting for ORACLE_HOME, MW_HOME, and WL_HOME

tbarnes-us:
smarter defaulting for ORACLE_HOME, MW_HOME, and WL_HOME

also includes a minor fix to utils.sh tracing (affects only introspector mock/unit test)


alai8:
LGTM

PR #1160, opened 2019-07-29
Make istio object in Domain yaml work

davidzl:


davidzl:
Ryan,

Please merge my latest commit (https://github.com/oracle/weblogic-kubernetes-operator/pull/1160/commits/d2ca72b2c0cbbb2dc0b5535ea3e13016e20f3f1d) to make it work.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -136,4 +136,7 @@ createDomainScriptName: create-domain-job.sh
 createDomainFilesDir: wlst
 
 # istioEnabled specifies whether or not it is deployed under Istio.
-istioEnabled: false
\ No newline at end of file
+istioEnabled: true
```
markxnelson: istio should not be enabled by default in the sample

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Istio.java
```diff
@@ -11,25 +11,47 @@
 
 public class Istio {
 
-  @Description("True, if this domain is deployed under an Istio service mesh. Defaults to true. Not required.")
-  private Boolean istioEnabled = true;
+  @Description(
+      "True, if this domain is deployed under an Istio service mesh. Defaults to true. Not required.")
```
markxnelson: istio should default to false

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Istio.java
```diff
@@ -11,25 +11,47 @@
...
    *
-   * @param istioEnabled True, if this domain is deployed under an Istio service mesh.
+   * @param enabled True, if this domain is deployed under an Istio service mesh.
+   */
+  public void setEnabled(Boolean enabled) {
+    this.enabled = enabled;
+  }
+
+  /**
+   * Get the readyness port.
```
markxnelson: spelling error

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain-script.sh
```diff
@@ -10,4 +10,6 @@ export DOMAIN_HOME=${DOMAIN_HOME_DIR}
 # Create the domain
 wlst.sh -skipWLSModuleScanning ${CREATE_DOMAIN_SCRIPT_DIR}/create-domain.py
 
-
+if [ "${ISTIO_ENABLED}" == "true" ]; then
+  wlst.sh -skipWLSModuleScanning ${CREATE_DOMAIN_SCRIPT_DIR}/istio-fix-domain.py
```
markxnelson: seems like there should be some error checking in here?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2018, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: "2018" should not be in the copyright notice - this is a new file right? so just "2019"

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,132 @@
...
+t3_public_address            = getEnvVar("T3_PUBLIC_ADDRESS")
+number_of_ms                 = int(getEnvVar("CONFIGURED_MANAGED_SERVER_COUNT"))
+cluster_type                 = getEnvVar("CLUSTER_TYPE")
+managed_server_name_base     = getEnvVar("MANAGED_SERVER_NAME_BASE")
+managed_server_name_base_svc = getEnvVar("MANAGED_SERVER_NAME_BASE_SVC")
+domain_logs                  = getEnvVar("DOMAIN_LOGS_DIR")
+script_dir                   = getEnvVar("CREATE_DOMAIN_SCRIPT_DIR")
+istio_readiness_port         = int(getEnvVar("ISTIO_READINESS_PORT"))
+
+# Update Domain for Istio
```
markxnelson: i would like to see some logging in this file, print out what is going on, so it is easier for us to debug it of we need to

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -486,6 +505,20 @@ public void setConfigOverrideSecrets(@Nullable List<String> overridesSecretNames
     this.configOverrideSecrets = overridesSecretNames;
   }
 
+  boolean istioEnabled() {
```
markxnelson: javadoc missing

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -114,8 +111,7 @@ static V1beta1CustomResourceDefinition createModel(KubernetesVersion version) {
 
     static V1ObjectMeta createMetadata() {
       return new V1ObjectMeta()
-          .name(KubernetesConstants.CRD_NAME)
-          .putLabelsItem(LabelConstants.RESOURCE_VERSION_LABEL, DEFAULT_OPERATOR_VERSION);
```
markxnelson: why are these labels removed?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -196,6 +200,7 @@ protected V1JobSpec createJobSpec(TuningParameters tuningParameters) {
 
   private V1PodTemplateSpec createPodTemplateSpec(TuningParameters tuningParameters) {
     V1ObjectMeta metadata = new V1ObjectMeta().name(getJobName());
+    if (istioEnabled()) metadata.putAnnotationsItem("sidecar.istio.io/inject", "false");
```
markxnelson: if statements need braces

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -853,7 +853,7 @@ def customizeNetworkAccessPoint(self, nap, listen_address):
     # FWIW there's theoretically no need to 'add' or 'replace' when empty
     #   since the runtime default is the server listen-address.
     nap_name=nap.getName()
-    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0:
+    if not (nap.getListenAddress() is None) and len(nap.getListenAddress()) > 0 and not (nap_name.startswith('istio-')) and not (nap_name == 'T3Channel'):
```
markxnelson: check the logic here - does this break the non-istio case?

davidzl:
Added 2 more commits to address Mark's comments.

operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -255,12 +255,14 @@ EOF
     if [ "${ADMIN_PORT_SECURE}" = "true" ]; then
       admin_protocol="https"
     fi  
-    echo "AdminURL=$admin_protocol\\://${AS_SERVICE_NAME}\\:${ADMIN_PORT}" >> ${wl_props_file}
+    if [ "${ISTIO_ENABLED}" == "true" ]; then
+      echo "AdminURL=t3\\://${AS_SERVICE_NAME}\\:${ADMIN_PORT}" >> ${wl_props_file}
```
tbarnes-us: Why is AdminURL forced to t3 here?

tbarnes-us: If it needs to be t3 for some reason, then t3s should be used if the admin port is secured.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-job-template.yaml
```diff
@@ -67,6 +71,10 @@ spec:
               value: "%DOMAIN_HOME%"
             - name: DOMAIN_LOGS_DIR
               value: "%DOMAIN_ROOT_DIR%/logs/%DOMAIN_UID%"
+            - name: ISTIO_ENABLED
+              value: "%ISTIO_ENABLED%"
+            - name: ISTIO_READINESS_PORT
+              value: "%ISTIO_READINESS_PORT%"
```
tbarnes-us: We shouldn't be requiring users to set these values - the Operator itself should set them for both the introspector job and the wl pods when istio is enabled via the domain yaml. If we want to keep it here, maybe rename it ISTIO_ENABLED_EXPERIMENTAL or at least put a clear comment above the env var stating that it's experimental.

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
```diff
@@ -0,0 +1,143 @@
...
+readDomain(domain_path)
+
+cd('/Servers/' + admin_server_name)
+create('istio-probe', 'NetworkAccessPoint')
+cd('/Servers/%s/NetworkAccessPoints/istio-probe' % admin_server_name)
+set('Protocol', 'http')
+set('ListenAddress', '127.0.0.1')
+set('PublicAddress', domain_uid + '-' + admin_server_name)
```
tbarnes-us: (1) For all addresses, the name needs to be converted to be DNS legal - see introspectDomain.py for example usage:
```
def toDNS1123Legal(self, address):
    return address.lower().replace('_','-')
```
(2) Ideally we should not be requiring that users set addresses in the original configuration. Instead, we should be setting the ListenAddress and Public Address of istio enabled channels for the user via situational configuration generated in 'introspectDomain.py' instead of requiring customers to set them exactly correctly.
(3) Note that the public-address of a T3Channel needs to be appropriate for the client's location, so "domain_uid + '-' + admin_server_name" can be wrong if the client is external. The sample should somehow note this (I assume most customers will use config-overrides to set the value).

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -36,6 +36,8 @@ spec:
           value: "false"
         - name: "JAVA_LOGGING_LEVEL"
           value: {{ .javaLoggingLevel | quote }}
+        - name: ISTIO_ENABLED
+          value: {{ .istioEnabled | quote }}
```
tbarnes-us: Why is this needed?  Can the value instead be derived from the domain resource as needed?

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -96,3 +96,6 @@ elasticSearchHost: "elasticsearch.default.svc.cluster.local"
 # elasticSearchPort specifies the port number of where elasticsearch is running.
 # This parameter is ignored if 'elkIntegrationEnabled' is false.
 elasticSearchPort: 9200
+
+# istioEnabled specifies whether or not it is deployed under Istio.
+istioEnabled: false
```
tbarnes-us: Maybe make clear that this is an experimental setting (mark it with '(experimental)')

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -133,3 +134,9 @@ createDomainScriptName: create-domain-job.sh
 # Kubernetes config map, which in turn is mounted to the `createDomainScriptsMountPath`,
 # so that the Kubernetes pod can use the scripts and supporting files to create a domain home.
 createDomainFilesDir: wlst
+
+# istioEnabled specifies whether or not it is deployed under Istio.
+istioEnabled: false
```
tbarnes-us: Mark this with '(experimental)'?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
```diff
@@ -133,3 +134,9 @@ createDomainScriptName: create-domain-job.sh
...
+
+# istioReadinessPort specifies the WebLogic port for readiness probe which is used under Istio.
+istioReadinessPort: 8888
```
tbarnes-us: Mark this with '(experimental)'?

operator/src/main/resources/scripts/startNodeManager.sh
tbarnes-us: How is ISTIO_ENABLED getting passed to the pod scripts (which in turn make use of AdminURL)?  I don't see it getting set by the Operator, and I don't see the sample's domain resource template specifying the value.   The former is preferable.  For an example of the Operator setting env vars for the introspector job and for wl pods, search for "LOG_HOME" in the Operator java source.

davidzl: In the first phase of Istio support, let's support just plain ports now. Will research the secure ports in the future as separate tasks.

ISTIO_ENABLED is passed via "kubernetes/samples/scripts/common/utility.sh"

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: I suspect public-address will not work under Istio. In Istio, it will implicitly forward payloads to pod name to 127.0.0.1. Not sure how it will deal with public address. I think this can be investigated in the next stage.

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
davidzl: This is required in the create-domain step when the domain resource has not been generated.

kubernetes/charts/weblogic-operator/values.yaml
davidzl: ok

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
davidzl: ok

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
davidzl: ok

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
davidzl: that might be branch/merge issue.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
davidzl: added

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: added

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: fixed

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/create-domain-script.sh
davidzl: The wlst exception will be thrown to the pod logs if there is any.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Istio.java
davidzl: fixed

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-inputs.yaml
davidzl: fixed.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Istio.java
davidzl: fixed

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
davidzl: this is forced by the code style checking of mvn plugin.

operator/src/main/resources/scripts/introspectDomain.py
davidzl: fixed

operator/src/main/resources/scripts/startNodeManager.sh
tbarnes-us: > ISTIO_ENABLED is passed via "kubernetes/samples/scripts/common/utility.sh"

Instead, it should be set by the operator for all jobs/pods when the operator detects the domain is istio-enabled.  Follow 'LOG_HOME' in the java source as an example...

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
tbarnes-us: The 'toDNS1123Legal' is still required regardless.

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
tbarnes-us: Will this have a negative impact on domains that don't set the experimental istio flags?

markxnelson:
Tiburon team are trying out this branch and will provide detailed feedback. 
@mgianatagh @markxnelson

tbarnes-us:
> Tiburon team are trying out this branch and will provide detailed feedback.
> @mgianatagh @markxnelson

Note that specialized testing will eventually needed for paths that particularly depend on 'public addresses' that are being directly set in this branch's sample WL config.   In particular RMI smart stub reconnects and recovering interrupted multi-srerver JTA transactions.  If I understand David correctly, this testing has not been completed and he anticipates there may be problems in this area.

markxnelson:
Got it - and we are not assuming this is "done done"

kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/wlst/istio-fix-domain.py
davidzl: fixed.

davidzl: I assume toDNS1123Legal is also applied to the pod names, right? Otherwise, the pod name and wls public address does not match, that will be a problem.

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
davidzl: It is harmless. The property istioEnabled is defined in "kubernetes/charts/weblogic-operator/values.yaml" and its value is false by default.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Istio.java
rjeberhard: My intention was that istioEnabled here defaults to true...  The idea was that Istio is disabled by default, but enabled by default once the "istio: {}" element is included in domain.yaml.  So...

1. Default
Istio enablement is disabled

2. Istio element present
spec:
  istio: {}
Istio enablement is enabled

3. Istio element present, but disabled
spec:
  istio:
    istioEnabled = false
    ... other settings preserved
Istio enablement is disabled.

Also, I think I would rename "istioEnabled" to "enabled"

rjeberhard: Spacing got lost after I posted the comment...  If it's confusing, let me know.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
rjeberhard: I removed them, but I can move this to a different PR.

rjeberhard: Long ago, we thought we would use these labels to know when to update the CRD... Now, we use the CRD's version field, so we don't need these labels.  They just cause confusion.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
rjeberhard: This is a surprise.  I thought the CheckStyle rule expected braces.

operator/src/main/resources/scripts/startNodeManager.sh
davidzl: LOG_HOME is also provided in the create-domain-inputs.yaml:

# The in-pod location for domain log, server logs, server out, and node manager log files
# If not specified, the value is derived from the domainUID as /shared/logs/<domainUID>
logHome: /shared/logs/domain1

So I think we need include istio values in it too:

# experimental
# istioEnabled specifies whether or not it is deployed under Istio.
istioEnabled: false

# experimental
# istioReadinessPort specifies the WebLogic port for readiness probe which is used under Istio.
istioReadinessPort: 8888

PR #1166, opened 2019-07-30
Fix introspector to handle WL config templates that don't reference a cluster.

tbarnes-us:
- Fix introspector so that it properly skips WL config templates that don't reference a cluster (currently it throws an exception), as requested by internal customer via internal bug OWLS-76014 
- Expand introspector 'mock' test to include such templates
- Echo config.xml to introspector log as a FINE (to aid in debugging), this will also show up in the operator log if operator log level is FINE/FINER/FINEST.
- Remove use of unknown jython name 'UndeclaredThrowableException'

### DIFF
--- a/src/integration-tests/introspector/wl-create-domain-pod.pyt
+++ b/src/integration-tests/introspector/wl-create-domain-pod.pyt
@@ -245,7 +245,7 @@ else:
 
   templateName = '${CLUSTER_NAME}' + "-template-dummy1"
   print('Creating Server Template: %s' % templateName)
-  st1=create(templateName, 'ServerTemplate')
+  st2=create(templateName, 'ServerTemplate')
   print('Done creating Server Template: %s' % templateName)
   cd('/ServerTemplates/%s' % templateName)
   cmo.setListenPort(${MANAGED_SERVER_PORT})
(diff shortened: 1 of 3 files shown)


src/integration-tests/introspector/wl-create-domain-pod.pyt
```diff
@@ -243,6 +243,25 @@ else:
...
+ 
+  templateName = '${CLUSTER_NAME}' + "-template-dummy2"
+  print('Creating Server Template: %s' % templateName)
+  st1=create(templateName, 'ServerTemplate')
+  print('Done creating Server Template: %s' % templateName)
+  cd('/ServerTemplates/%s' % templateName)
+  cmo.setListenPort(${MANAGED_SERVER_PORT})
+  #cmo.setListenAddress('${DOMAIN_UID}-${MANAGED_SERVER_NAME_BASE}${id}') # subst-ignore-missing
+  #cmo.setCluster(cl)
+  print('Done setting attributes for Server Template: %s' % templateName);
```
alai8: does introspectTest.sh need to be updated to check for these additions?

alai8:
LGTM

src/integration-tests/introspector/wl-create-domain-pod.pyt
tbarnes-us: No.  These 'unused template' additions are simply a way to test the failure that this pull is fixing, and the change to introspectDomain.py teaches the introspector to handle the unused templates.

PR #1178, opened 2019-08-08
Update domain status for image fetch or instrospector job failures

russgold:
Initial domain status updating:

- any severe error logged by the introspector
- a failure to pull an image

### DIFF
--- a/operator/pom.xml
+++ b/operator/pom.xml
@@ -305,6 +305,12 @@
       <artifactId>simplestub</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.fatboyindustrial.gson-jodatime-serialisers</groupId>
+      <artifactId>gson-jodatime-serialisers</artifactId>
+      <version>1.7.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/operator/pom.xml
+++ b/operator/pom.xml
@@ -305,12 +305,6 @@
       <artifactId>simplestub</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.fatboyindustrial.gson-jodatime-serialisers</groupId>
-      <artifactId>gson-jodatime-serialisers</artifactId>
-      <version>1.7.1</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
 </project>
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -4,11 +4,6 @@
 
 package oracle.kubernetes.operator.helpers;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.StringReader;
-import java.util.List;
-
 import io.kubernetes.client.models.V1DeleteOptions;
 import io.kubernetes.client.models.V1EnvVar;
 import io.kubernetes.client.models.V1Job;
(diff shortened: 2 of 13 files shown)


operator/pom.xml
```diff
@@ -305,6 +305,12 @@
       <artifactId>simplestub</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.fatboyindustrial.gson-jodatime-serialisers</groupId>
```
markxnelson: there is no PLS third party approval for this library.

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -149,4 +151,24 @@ private static String getOperatorCreatedLabel(V1ObjectMeta metadata) {
...
+        patchBuilder.replace("/status/message", message);
+      } else {
+        patchBuilder.add("/status/reason", reason);
+        patchBuilder.add("/status/message", message);
+      }
+      new CallBuilder()
+          .patchDomain(
+              domain.getDomainUid(), domain.getMetadata().getNamespace(), patchBuilder.build());
+    } catch (ApiException ignored) {
+      /* extraneous comment to fool checkstyle into thinking that this is not an empty catch block. */
```
rjeberhard: That's hilarious; usually, I just put "no-op."

rjeberhard:
@russgold, when I try to build I get CheckStyle violations.  Did you disable them when building and testing locally?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -303,64 +303,115 @@ public NextAction apply(Packet packet) {
...
+
+    private void logToOperator() {
+      if (logMessage.length() == 0) return;
+
+      String logMsg = logMessage.toString();
+      switch (getLogLevel(logMsg)) {
+        case "SEVERE":
+          addSevereStatus(logMsg); // fall through
+        case "ERROR":
+          LOGGER.severe(logMsg);
```
tbarnes-us: I realize this pull has already been merged, but ERROR should also cause an 'addSevereStatus' since an ERROR is converted to a severe.   (I think I removed all errors, so this is not strictly needed - but this path exists just in case someone adds one back in later.)

tbarnes-us: If you do this, then you'll need to search for last-index-of ERROR in addition to SEVERE in the 'addSevereStatus'  method.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -303,64 +303,115 @@ public NextAction apply(Packet packet) {
...
+    //  - ignores all lines in the log up to the first line that starts with '@['
+    private void convertJobLogsToOperatorLogs(String jobLogs) {
+      for (String line : jobLogs.split(EOL_PATTERN)) {
+        if (line.startsWith("@[")) {
+          logToOperator();
+          logMessage = new StringBuilder(INTROSPECTOR_LOG_PREFIX).append(line.trim());
+        } else if (logMessage.length() > 0) {
+          logMessage.append(System.lineSeparator()).append(line.trim());
+        }
+      }
```
tbarnes-us: I assume the two 'trim()' statements were added to pretty-print logs that are converted into a status?   Unfortunately, this messes up the pretty-printing of multi-line log messages - some of which depend on prepending spaces to make them readable.

PR #1192, opened 2019-08-16
Fix issues in the usage of Istio in the sample scripts

rjeberhard:


kubernetes/samples/scripts/common/domain-template.yaml
```diff
@@ -75,8 +93,4 @@ spec:
     replicas: %INITIAL_MANAGED_SERVER_REPLICAS%
   # The number of managed servers to start for unlisted clusters
   # replicas: 1
-  # 
-  experimental:
```
davidzl: How about commenting these out instead of removing, so that it can be experimented easier?

kubernetes/samples/scripts/common/utility.sh
```diff
@@ -366,10 +366,12 @@ function createFiles {
     # extra entries for FMW Infra domains
     sed -i -e "s:%RCU_CREDENTIALS_SECRET_NAME%:${rcuCredentialsSecret}:g" ${createJobOutput}
     sed -i -e "s:%CUSTOM_RCUPREFIX%:${rcuSchemaPrefix}:g" ${createJobOutput}
-    sed -i -e "s|%CUSTOM_CONNECTION_STRING%|${rcuDatabaseURL}|g" ${createJobOutput}
+    sed -i -e "s:%CUSTOM_CONNECTION_STRING%:${rcuDatabaseURL}|g" ${createJobOutput}
     sed -i -e "s:%EXPOSE_T3_CHANNEL_PREFIX%:${exposeAdminT3Channel}:g" ${createJobOutput}
-    sed -i -e "s|%ISTIO_ENABLED%|${istioEnabled}|g" ${createJobOutput}
-    sed -i -e "s|%ISTIO_READINESS_PORT%|${istioReadinessPort}|g" ${createJobOutput}
+    # entries for Istio
+    sed -i -e "s:%ISTIO_PREFIX%:${istioPrefix}:g" ${createJobOutput}
```
davidzl: Does istioPrefix have been assigned value here?

rjeberhard: Thanks!  I had the assignment before the later use, but not before this one.  I'll move it up.

PR #1196, opened 2019-08-19
Pass admin server info as env vars to additional runs of the introspector

tbarnes-us:
This change passes ADMIN_NAME, ADMIN_PORT, ADMIN_PORT_SECURE, and AS_SERVICE_NAME env vars to an introspector job when the job is being rerun on an existing domain (the original job is what gathers this information).


tbarnes-us:
Hi @jshum2479 .  I haven't tested this change yet, but feel free to merge it early to the domain_in_image_poc branch - I don't think it'll have side effects...

PR #1202, opened 2019-08-21
Validate domain resource before running introspector

russgold:
Adds semantic validation of the domain resource itself. If it fails, updates the domain status. If it succeeds, allows the introspection to run.

### DIFF
--- a/model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
+++ b/model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
@@ -4,18 +4,24 @@
 
 package oracle.kubernetes.weblogic.domain.model;
 
+import java.io.File;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.validation.Valid;
 
 import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.SerializedName;
 import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.models.V1SecretReference;
+import io.kubernetes.client.models.V1VolumeMount;
 import oracle.kubernetes.json.Description;
 import oracle.kubernetes.operator.LabelConstants;
 import oracle.kubernetes.operator.VersionConstants;
@@ -348,25 +354,33 @@ public class Domain {
     return Optional.ofNullable(spec.getDomainUid()).orElse(getMetadata().getName());
   }
 
-  public String getLogHome() {
+  /**
+   * Returns the path to the log home to be used by this domain. Null if the log home is disabled.
+   * @return a path on a persistent volume, or null
+   */
+  public @Nullable String getEffectiveLogHome() {
+    return isLogHomeEnabled() ? getLogHome() : null;
+  }
+
+  @Nonnull String getLogHome() {
     return Optional.ofNullable(spec.getLogHome())
         .orElse(String.format(LOG_HOME_DEFAULT_PATTERN, getDomainUid()));
   }
 
-  public boolean getLogHomeEnabled() {
-    return spec.getLogHomeEnabled();
+  boolean isLogHomeEnabled() {
+    return spec.isLogHomeEnabled();
   }
 
   public boolean isIncludeServerOutInPodLog() {
     return spec.getIncludeServerOutInPodLog();
   }
 
-  public boolean isDomainHomeInImage() {
+  boolean isDomainHomeInImage() {
     return spec.isDomainHomeInImage();
   }
 
-  public boolean istioEnabled() {
-    return spec.istioEnabled();
+  public boolean isIstioEnabled() {
+    return spec.isIstioEnabled();
   }
 
   public int getIstioReadinessPort() {
@@ -399,7 +413,7 @@ public class Domain {
    *
    * @return a list of names; may be empty
    */
-  public List<String> getAdminServerChannelNames() {
+  List<String> getAdminServerChannelNames() {
     return getEffectiveConfigurationFactory().getAdminServerChannelNames();
   }
 
@@ -459,5 +473,58 @@ public class Domain {
         .append(spec, rhs.spec)
         .append(status, rhs.status)
         .isEquals();
+  }
+
+  public List<String> getValidationFailures() {
+    return new Validator().getValidationFailures();
+  }
+
+  class Validator {
+    private List<String> failures = new ArrayList<>();
+    private Set<String> clusterNames = new HashSet<>();
+    private Set<String> serverNames = new HashSet<>();
+
+    List<String> getValidationFailures() {
+      addDuplicateNames();
+      addUnmappedLogHome();
+
+      return failures;
+    }
+
+    private void addDuplicateNames() {
+      getSpec().getManagedServers().stream().map(ManagedServer::getServerName).forEach(this::checkDuplicateServerName);
+      getSpec().getClusters().stream().map(Cluster::getClusterName).forEach(this::checkDuplicateClusterName);
+    }
+
+    private void checkDuplicateServerName(String s) {
+      if (serverNames.contains(s))
+        failures.add(String.format("More than one server is named '%s'", s));
+      else
+        serverNames.add(s);
+    }
+
+    private void checkDuplicateClusterName(String s) {
+      if (clusterNames.contains(s))
+        failures.add(String.format("More than one cluster is named '%s'", s));
+      else
+        clusterNames.add(s);
+    }
+
+    private void addUnmappedLogHome() {
+      if (!isLogHomeEnabled()) return;
+
+      if (getSpec().getAdditionalVolumeMounts().stream().map(V1VolumeMount::getMountPath).noneMatch(this::mapsLogHome))
+        failures.add("No volume mount contains path for log home");
+    }
+
+    private boolean mapsLogHome(String mountPath) {
+      return getLogHome().startsWith(separatorTerminated(mountPath));
+    }
+
+    private String separatorTerminated(String path) {
+      if (path.endsWith(File.separator)) return path;
+      else return path + File.separator;
+    }
+
   }
 }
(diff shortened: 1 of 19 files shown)


model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -460,4 +474,57 @@ public boolean equals(Object other) {
...
+    }
+
+    private void addDuplicateNames() {
+      getSpec().getManagedServers().stream().map(ManagedServer::getServerName).forEach(this::checkDuplicateServerName);
+      getSpec().getClusters().stream().map(Cluster::getClusterName).forEach(this::checkDuplicateClusterName);
+    }
+
+    private void checkDuplicateServerName(String s) {
+      if (serverNames.contains(s))
+        failures.add(String.format("More than one server is named '%s'", s));
```
tbarnes-us: How could this ever happen unless there was a bug in our topology generator?   
Regardless 'server' --> 'WebLogic Server in your WebLogic domain'

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -460,4 +474,57 @@ public boolean equals(Object other) {
...
+      else
+        serverNames.add(s);
+    }
+
+    private void checkDuplicateClusterName(String s) {
+      if (clusterNames.contains(s))
+        failures.add(String.format("More than one cluster is named '%s'", s));
```
tbarnes-us: How could this ever happen unless there was a bug in our topology generator?
Regardless 'cluster' --> 'WebLogic Cluster in your WebLogic domain'

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -460,4 +474,57 @@ public boolean equals(Object other) {
...
+      else
+        clusterNames.add(s);
+    }
+
+    private void addUnmappedLogHome() {
+      if (!isLogHomeEnabled()) return;
+
+      if (getSpec().getAdditionalVolumeMounts().stream().map(V1VolumeMount::getMountPath).noneMatch(this::mapsLogHome))
+        failures.add("No volume mount contains path for log home");
```
tbarnes-us: for log home --> 'for the logHome '%s' specified in the domain resource' (%s should be this::mapsLogHome).

tbarnes-us: Is this a valid check?  IIRC, the startup code does a 'mkdir -p'?  If it does, then maybe only check that the first directory in each path matches.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -379,7 +380,13 @@ String getLogHome() {
...
 
   public void setLogHome(String logHome) {
-    this.logHome = logHome;
+    this.logHome = Optional.ofNullable(logHome).map(this::validatePath).orElse(null);
+  }
+
+  private String validatePath(String s) {
+    if (s.isBlank()) return null;
+    if (s.endsWith(File.separator)) return s;
+    return s + File.separator;
```
tbarnes-us: Should we require that the path start with a separator, or just add one if its not there?

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: Note that it _is_ possible for two servers to have the same name if they both happen to resolve to same DNS1123 form (all lower cases, underscores converted to hyphens).  There's a pending JIRA to check this in the introspector, but maybe it'd fine to check it here instead.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: Note that it _is_ possible for two cluster to have the same name if they both happen to resolve to same DNS1123 form (all lower case, underscores converted to hyphens).  There's a pending JIRA to check this in the introspector, but maybe it'd fine to check it here instead.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
russgold: This is checking the domain resource, not the WLS domain.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
russgold: This is checking the domain resource, not the WLS domain.

tbarnes-us: ok

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: ok

tbarnes-us: Actually, not 100% OK.  It'd be helpful if the error message clearly stated that this is a domain resource issue, and maybe which specific part of the schema was messed up.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: Actually, not 100% OK. It'd be helpful if the error message clearly stated that this is a domain resource issue, and maybe which specific part of the schema was messed up.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
russgold: this::mapsLogHome is a boolean. Do you mean this::logHome ?

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
russgold: I suppose we could apply a general "is this a valid path" logic to all such fields. That would include the leading separator and probably more.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: Yes.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
russgold: When you say that it is possible, do you mean that it is allowed? Or that we have to reject it?

That is, is it legal to define one server as "My_Server" and another as "my-server"? Doing so would cause a problem with pod and service names.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
russgold: Doing a `mkdir -p` shouldn't matter in this case, as we are not looking at the actual file system, here. The key is that:

`/my/path/to/log` is not a parent of `/my/path/to/logfiles/mydomain` even though the "startsWith" check without adding the trailing separator would pass. That's why we have to ensure that the trailing separator is there.

model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
tbarnes-us: Since this particular error message is about the user-supplied domain resource, that should be made clear.

As for My_Server/my-server that's a separate issue.  If operator isn't in the picture, it is legal to define My_Server and my-server.  If the operator is in the picture, this is not legal due to the service name issue as you point out, but we don't have any checks for this currently.  We could do the checks in the introspector but maybe it'd be better to do them in the Operator? (full details in pending JIRA OWLS-74954, which assumes the introspector should have the checks).

model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
russgold: After experimenting with some validation, I'm not sure that we can count on knowing what the start of a valid path has to be; it will depend on the mapped resource. Windows using a different initial string than does Linux. The only thing we can be sure of is that the trailing separator is significant.

tbarnes-us: Does this help?
```File file = new File(path);
if (file.isAbsolute()) {```

File: model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java (hunk 3 above)
Comment (tbarnes-us) [reply model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java]: Ah.  Makes sense.  Please update the log message regardless.

File: model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java (hunk 4 above)
Comment (russgold) [reply model/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java]: Yes, that works. New test and validation added.

File: model/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java [hunk 5]
```diff
@@ -480,41 +480,82 @@ public boolean equals(Object other) {
...
   class Validator {
+    static final String DUPLICATE_SERVER_NAME_FOUND
+          = "More than one item under spec.managedServers in the domain resource has DNS-1123 name '%s'";
+    static final String DUPLICATE_CLUSTER_NAME_FOUND
+          = "More than one item under spec.clusters in the domain resource has DNS-1123 name '%s'";
+    static final String LOG_HOME_PATH_NOT_MOUNTED
+          = "No volume mount contains path for log home '%s', %s in the domain resource";
+    static final String BAD_VOLUME_MOUNT_PATH
+          = "The mount path '%s', in entry '%s' of domain resource additionalVolumeMounts, is not valid";
+
```
tbarnes-us: I18N these?

russgold: What's our standard way for doing that?

rjeberhard: The Operator.properties resource bundle

russgold: That's in the operator sub-module, to which this code has no access. Maybe we should merge the model project into the operator subproject?

tbarnes-us:
Approving.   Pending issues for a future pull:  
- All status messages should be logged (not just show up in status), and logged as 'severe' if it's critical.   
- I18N

PR #1212, opened 2019-08-23
Fix secure protocol in AdminURL

qvdang:
I picked up the work from Paul's https://github.com/oracle/weblogic-kubernetes-operator/pull/1130.
This change makes sure the AdminURL is constructed with the correct protocol if the admin port is enabled or the admin server has an SSL port enabled.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -141,11 +141,30 @@ public abstract class PodStepContext extends StepContextBase {
         .getLocalAdminProtocolChannelPort();
   }
 
+  /**
+   * Check if the server is listening on a secure port. NOTE: If the targetted server is a managed
+   * server, this method is overridden to check if the managed server has a secure listen port rather
+   * than the admin server. See PodHelper.ManagedPodStepContext
+   *
+   * @return true if server is listening on a secure port
+   */
   boolean isLocalAdminProtocolChannelSecure() {
     return domainTopology
         .getServerConfig(domainTopology.getAdminServerName())
         .isLocalAdminProtocolChannelSecure();
   }
+
+  /**
+   * Check if the admin server is listening on a secure port.
+   *
+   * @return true if admin server is listening on a secure port
+   */
+  private boolean isAdminServerProtocolChannelSecure() {
+    return domainTopology
+        .getServerConfig(domainTopology.getAdminServerName())
+        .isLocalAdminProtocolChannelSecure();
+  }
+
 
   Integer getLocalAdminProtocolChannelPort() {
     return domainTopology
@@ -602,6 +621,13 @@ public abstract class PodStepContext extends StepContextBase {
     addEnvVar(vars, "ADMIN_PORT", getAsPort().toString());
     if (isLocalAdminProtocolChannelSecure()) {
       addEnvVar(vars, "ADMIN_PORT_SECURE", "true");
+    }
+    if (isAdminServerProtocolChannelSecure()) {
+      // The following env variable determines whether to set a secure protocol(https/t3s) in the "AdminURL" property 
+      // in NM startup.properties.
+      // WebLogic Node Manager then sets the ADMIN_URL env variable(based on the "AdminURL") before starting
+      // the managed server
+      addEnvVar(vars, "ADMIN_SERVER_PORT_SECURE", "true");
     }
     addEnvVar(vars, "SERVER_NAME", getServerName());
     addEnvVar(vars, "DOMAIN_UID", getDomainUid());
--- a/operator/src/main/resources/scripts/utils.sh
+++ b/operator/src/main/resources/scripts/utils.sh
@@ -395,3 +395,26 @@ checkWebLogicVersion()
   fi
   return 0
 }
+
+
+#
+# getAdminUrl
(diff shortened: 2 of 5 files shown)


qvdang:
Jenkins job: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2380/testReport/

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -141,12 +141,31 @@ Integer getAsPort() {
...
+   *
+   * @return true if admin server is listening on a secure port
+   */
+  private boolean isAdminServerProtocolChannelSecure() {
+    return domainTopology
+        .getServerConfig(domainTopology.getAdminServerName())
+        .isLocalAdminProtocolChannelSecure();
+  }
+
+
```
tbarnes-us: This looks exactly the same as isLocalAdminProtocolChannelSecure().  Does that mean there's a bug in that method?

operator/src/main/resources/scripts/utils.sh
```diff
@@ -395,3 +395,26 @@ checkWebLogicVersion()
...
 }
+
+
+#
+# getAdminUrl
+#   purpose: Get the ADMIN_URL used to connect to the admin server internally, e.g. when starting a managed server
+#   sample:
+#     ADMIN_URL=$(getAdminUrl)
+#
+function getAdminUrl() {
```
tbarnes-us: maybe change name to 'getAdminServerUrl' to help make clear this function returns url specific to the admin server...

operator/src/main/resources/scripts/utils.sh
```diff
@@ -395,3 +395,26 @@ checkWebLogicVersion()
   fi
   return 0
 }
+
+
+#
+# getAdminUrl
+#   purpose: Get the ADMIN_URL used to connect to the admin server internally, e.g. when starting a managed server
```
tbarnes-us: change word 'ADMIN_URL' to 'admin URL'

operator/src/main/resources/scripts/utils.sh
```diff
@@ -395,3 +395,26 @@ checkWebLogicVersion()
...
+#   sample:
+#     ADMIN_URL=$(getAdminUrl)
+#
+function getAdminUrl() {
+  admin_protocol="http"
```
tbarnes-us: add keyword 'local' before admin_protocol

tbarnes-us: (only needed for the first  use of the local variable, other functions should do this with local variables too but I wouldn't be surprised if we haven't caught them all)

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
tbarnes-us: My guess is that isLocalAdminProtocolChannelSecure should be calling 'getServerName()' instead of 'domainTopology.getAdminServerName()'.  

I know that's not directly related to this pull, but since the patient is open, maybe you can investigate?

operator/src/main/resources/scripts/utils.sh
qvdang: Ok.

operator/src/main/resources/scripts/utils.sh
qvdang: Noted.  Will make the change.

operator/src/main/resources/scripts/utils.sh
qvdang: Sure.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
qvdang: I don't think isLocalAdminProtocolChannelSecure() is buggy.  If you look at its method comment, it is overriden by PodHelper.ManagedPodStepContext([.isLocalAdminProtocolChannelSecure()]).  So the reported bug manifests when the domain wide admin port is not enabled and the admin server has an SSL port enabled which then results in the "http://admin_server_name:admin_server_ssl_port" AdminURL.

I agree having isLocalAdminProtocolChannelSecure call getServerName() seems to flow more naturally.  The managed server pod will call its overriding version anyway, and the admin server pod will get its own config.  This is an existing method and it works so I did not focus on that aspect.  I will double check this.

qvdang: And one advantage of having isLocalAdminProtocolChannelSecure call getServerName() is that the method then does not need to be overridden in PodHelper.ManagedPodStepContext.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -603,6 +622,13 @@ void overrideContainerWeblogicEnvVars(List<V1EnvVar> vars) {
     if (isLocalAdminProtocolChannelSecure()) {
       addEnvVar(vars, "ADMIN_PORT_SECURE", "true");
     }
+    if (isAdminServerProtocolChannelSecure()) {
```
TheFrogPad: As follow on to the comment above about the method and being overriden, can you add a comment for line 622 about this as well or better yet make it clear the difference between ADMIN_PORT_SECURE and ADMIN_SERVER_PORT_SECURE in that one is for the local servers connection and the other is for the connection to the admin server of the domain.

PR #1215, opened 2019-08-23
Validate specified environment variables

russgold:
This PR adds internationalization for domain validation messages and adds validation of the environment variables.

Some points to note:

- While domain validation is not logging, Tom B has suggested that we might want to log those issues as well as add them to the status message. I have therefore added message keys.
- I used a different message prefix for these keys, as they seem separate to me. Possibly they should use the same prefix and just start a higher sequence number
- to handle grammatical issues such as "is"/"are" I added some non-sequence keys
- environment variables that affect the admin servers cannot use some additional values reserved by the introspection job

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainValidationMessages.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainValidationMessages.java
@@ -1,0 +1,90 @@
+// Copyright 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Licensed under the Universal Permissive License v 1.0 as shown at
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.weblogic.domain.model;
+
+import java.text.ChoiceFormat;
+import java.text.Format;
+import java.text.MessageFormat;
+import java.util.List;
+import java.util.ResourceBundle;
+import javax.annotation.Nonnull;
+
+import io.kubernetes.client.models.V1VolumeMount;
+import oracle.kubernetes.operator.logging.MessageKeys;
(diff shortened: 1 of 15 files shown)


russgold:
Jenkins failed... investigating.

russgold:
I cannot reproduce the jenkins failure locally...

russgold:
This looks suspicious... 
```
[ERROR] whenReservedEnvironmentVariablesSpecifiedAtServerLevel_reportError(oracle.kubernetes.weblogic.domain.model.DomainValidationTest)  Time elapsed: 0.016 s  <<< ERROR!
java.util.MissingResourceException: Can't find bundle for base name operator, locale en_US
	at oracle.kubernetes.weblogic.domain.model.DomainValidationTest.whenReservedEnvironmentVariablesSpecifiedAtServerLevel_reportError(DomainValidationTest.java:139)
```

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainValidationMessages.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private static String getMessage(String key, Object... parameters) {
+    MessageFormat formatter = new MessageFormat("");
+    formatter.applyPattern(getBundleString(key));
+    return formatter.format(parameters);
+  }
+
+  private static String getBundleString(String key) {
+    return ResourceBundle.getBundle("operator").getString(key);
```
rjeberhard: Shouldn't this be "Operator"?

russgold: yes. I forgot that the Mac file system is case-insensitive.

Issue #1227, opened 2019-08-27
Support k8s service account for wls domain/cluster/server


alan-cao:
enhancement request for wls operator to set k8s service account for the wls domain or even cluster/server level.

The service account can be used to establish identity and enforce authorization for service-to-service calls via Istio authorization https://istio.io/docs/concepts/security/#authorization

 The lower level value will override. So that it is possible to make access control on different level. like Cluster-A can do something but Cluster-B of same domain can't.

PR #1235, opened 2019-08-31
K8s service account

alan-cao:
https://github.com/oracle/weblogic-kubernetes-operator/issues/1227

enhancement request for wls operator to set k8s service account for the wls domain or even cluster/server level.

The service account can be used to establish identity and enforce authorization for service-to-service calls via Istio authorization https://istio.io/docs/concepts/security/#authorization


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -515,6 +515,7 @@ protected V1PodSpec createSpec(TuningParameters tuningParameters) {
             .affinity(getAffinity())
             .nodeSelector(getServerSpec().getNodeSelectors())
             .nodeName(getServerSpec().getNodeName())
+            .serviceAccountName(getServerSpec().getServiceAccountName())
```
rjeberhard: We need to also add this to the pod spec in JobStepContext.  We've said that the introspector job runs as much as possible just like the admin server.

docs/domains/Domain.json
```diff
@@ -499,6 +499,10 @@
             "$ref": "https://github.com/garethr/kubernetes-json-schema/blob/master/v1.13.5/_definitions.json#/definitions/io.k8s.api.core.v1.PodReadinessGate"
           }
         },
+        "serviceAccountName": {
+          "description": "ServiceAccountName is the name of the ServiceAccount to use to run this pod. If it is not set, default ServiceAccount will be used. The ServiceAccount has to exist at the time the pod is created.",
```
rjeberhard: Since this is already the description, it doesn't need to be self-referential.  "ServiceAccountName is the name of the ServiceAccount..." -> "Name of the ServiceAccount..."

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
alan-cao: updated. please help to review

docs/domains/Domain.json
alan-cao: fixed

rjeberhard:
Need a second review... @ddsharpe, do you have time to please look at another today?

PR #1241, opened 2019-09-05
added the ability properly process the property file with commented out values, fixed typo

marinakog:
added the ability properly process the property file with commented out values, fixed typo in the README.md for terraform


markxnelson:
merged into develop also in commit 8a59a3474

PR #1245, opened 2019-09-06
Unprocessable Entity (422) status

russgold:
When Unprocessable Entity (422) status received from Kubernetes, record the message in the Domain status.

Jenkins job http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2515/


docs/charts/index.yaml
```diff
@@ -1,46 +1,46 @@
 apiVersion: v1
```
russgold: I didn't change this. Is there an easy way to remove it from the commit, or should I just ignore it? (Ditto the tgz)

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -273,21 +274,13 @@ public NextAction apply(Packet packet) {
                             metadata.getNamespace(),
                             new ResponseStep<V1Job>(null) {
                               @Override
-                              public NextAction onFailure(
-                                  Packet packet,
-                                  ApiException e,
-                                  int statusCode,
-                                  Map<String, List<String>> responseHeaders) {
-                                return super.onFailure(packet, e, statusCode, responseHeaders);
+                              public NextAction onFailure(Packet packet,CallResponse<V1Job> callResponse) {
```
russgold: Now that I look at it... I can just remove the override, can't I?

russgold:
Not happy with this. More to test.

PR #1250, opened 2019-09-10
Update domain status on Unprocessable Error

russgold:
With this update, 422 (Unprocessable Error) statuses from Kubernetes now result in the explanation from the exception being recorded in the domain status. It also changes the update on Domain validation failure to run asynchronously, and handles cases where there is no existing status before the patch.


operator/src/main/java/oracle/kubernetes/operator/calls/unprocessable/Cause.java
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+// Licensed under the Universal Permissive License v 1.0 as shown at
+// http://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.calls.unprocessable;
+
+public class Cause {
+  private String reason;
+  private String message;
```
markxnelson: how would message or field ever get set?

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainStatusPatch.java
```diff
@@ -6,46 +6,110 @@
...
+      patchBuilder.add("/status/message", message);
+    } else {
+      setSubField(patchBuilder, "/status/reason", domain.getStatus().getReason(), reason);
+      setSubField(patchBuilder, "/status/message", domain.getStatus().getMessage(), message);
+    }
+    return patchBuilder;
+  }
+
+  private static void setSubField(JsonPatchBuilder patchBuilder, String path, String oldValue, String newValue) {
+    if (oldValue == null)
```
markxnelson: braces

operator/src/main/java/oracle/kubernetes/operator/calls/unprocessable/Cause.java
russgold: In this case, it wouldn't. This is one of a set of classes which map to the content of the ApiException's 422 status. So far, I have not found a use for those fields, but since they are in the structure, I mapped them.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainStatusPatch.java
russgold: You know I hate braces for one-line conditional bodies :)

At some point, we should really have a discussion on the advantages and disadvantages.

russgold: BTW, I have found that it is precisely the "smallest possible change" that leads to code and architecture degradation. It is for that reason that the Agile style of development urges aggressive refactoring. It keeps the code clean and flexible. That necessarily demands extensive unit tests, of course.

markxnelson: exactly, if we had that level of test coverage, then i would have a different opinion

operator/src/main/java/oracle/kubernetes/operator/calls/unprocessable/Cause.java
markxnelson: ok

rjeberhard:
@russgold, where does the message go?  I tried creating a serverPod.volumeMount with a bad volume name and the introspector job fails (can't be created).  The unprocessable entity message is in the operator's log but I don't see anything in the domain status.

russgold:
@rjeberhard It's supposed to go the domain status; did you get a 422? A SEVERE error in the introspection log? How do I reproduce it?

rjeberhard:
Yes, I got a SEVERE message in the log.  I added content like the following under serverPod:

```
spec:
  serverPod:
     volumeMounts:
     - mountPath: /tmp/bad
       name: bad-volume
```
Since the operator will try and create this volumeMount for the introspector job, it will fail.

russgold:
With the latest commit, `kubectl describe domain sample-domain1 -n sample-domain1-ns` shows, in part:
```
Status:
  Clusters:
  Conditions:
    Last Transition Time:  2019-09-11T00:07:25.355Z
    Reason:                InspectingDomainPrescence
    Status:                True
    Type:                  Progressing
  Message:                 Job.batch "sample-domain1-introspect-domain-job" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: "weblogic-domain-storage-volume"
  Reason:                  FieldValueNotFound
  Servers:
  Start Time:  2019-09-11T00:07:25.288Z
```

russgold:
Hmmm... some info appears to have been lost from the exception. Gotta see how to include it.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainStatusPatch.java
russgold: Overall, we have 67% line coverage with unit tests. I've been increasing the coverage as I work, and other developers have been doing so as well. The class that you're looking at has 96% line coverage; the only line not covered is a catch for an exception that can never be thrown (but Java requires it to be caught).

It's mostly the old code that has poor test coverage.

PR #1258, opened 2019-09-15
Update pre-requisites

markxnelson:
This PR updates the pre-requisites page as follows: 
- include CRI-O as a supported container runtime
- update the Helm release to 2.14.3, since we have found that 2.8 has some compatibility issues with recent k8s versions
- note that OpenShift 4 certification is still underway

### DIFF
--- a/docs-source/content/userguide/introduction/introduction.md
+++ b/docs-source/content/userguide/introduction/introduction.md
@@ -18,7 +18,7 @@ Detailed instructions are available [here]({{< relref "/userguide/managing-opera
 
 * Kubernetes 1.11.5+, 1.12.3+, 1.13.0+, and 1.14.0+  (check with `kubectl version`).
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
-* Docker 18.03.1.ce (check with `docker version`) *or* CRI-O 1.14.7 (check with `crictl version | grep RuntimeVersion`).
+* Docker 18.9.1 (check with `docker version`) *or* CRI-O 1.14.7 (check with `crictl version | grep RuntimeVersion`).
 * Helm 2.14.3+ (check with `helm version`).
 * Oracle WebLogic Server 12.2.1.3.0 with patch 29135930.
    * The existing WebLogic Docker image, `container-registry.oracle.com/middleware/weblogic:12.2.1.3 `,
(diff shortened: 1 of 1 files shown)


docs-source/content/userguide/introduction/introduction.md
```diff
@@ -18,8 +18,8 @@ Detailed instructions are available [here]({{< relref "/userguide/managing-opera
 
 * Kubernetes 1.11.5+, 1.12.3+, 1.13.0+, and 1.14.0+  (check with `kubectl version`).
 * Flannel networking v0.9.1-amd64 (check with `docker images | grep flannel`).
-* Docker 18.03.1.ce (check with `docker version`).
-* Helm 2.8.2+ (check with `helm version`).
+* Docker 18.03.1.ce (check with `docker version`) *or* CRI-O 1.14.7 (check with `crictl version | grep RuntimeVersion`).
+* Helm 2.14.3+ (check with `helm version`).
```
rjeberhard: What's the version installed with OKE?  Similarly, should we bump the minimum Docker version?  I'm somewhat concerned this is requiring too recent a version of Helm.

markxnelson: I will check on the helm version.  We could say minimum X, recommended Y. 
Docker version in OKE is 18.9.1

markxnelson: OKE installs Helm 2.14.3

markxnelson: I know for sure 2.8 just plain will not work with k8s 1.15.   Helm 2.14 updated their k8s library to 1.14. I can't find a nice neat version compatibility matrix.  Unless you want to go test all the combinations, seems to me like saying we support a pretty recent version is a reasonable thing to do.

rjeberhard: Let's update the Docker runtime too and then LGTM.

markxnelson: what do you want to update it to?

rjeberhard: @mriccell, any objection to just saying requires Docker 18.9.1?  This would be my preference.  Given that OKE has also moved to 18.9.1, I don't see the need to introduce a requires/recommends split.

PR #1261, opened 2019-09-18
OWLS-73891 part2 - clean up stale states about introspector jobs and domain presence infos

doxiao:
More clean up of stale objects (domain presence infos and introspector jobs) for a deleted and recreated namespace. 
This helps a previously existing domain to be successfully recreated after its namespace is deleted/recreated, and deleted from and then added back to operator's domainNamespaces list.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -97,6 +97,9 @@ public class DomainProcessorImpl implements DomainProcessor {
     Map<String, DomainPresenceInfo> map = DOMAINS.get(ns);
     if (map != null) {
       map.remove(domainUid);
+      if (map.isEmpty()) {
+        DOMAINS.remove(ns);
+      }
     }
   }
 
@@ -195,7 +198,7 @@ public class DomainProcessorImpl implements DomainProcessor {
   }
 
   public void stopNamespace(String ns) {
-    Map<String, DomainPresenceInfo> map = DOMAINS.get(ns);
+    Map<String, DomainPresenceInfo> map = DOMAINS.remove(ns);
     if (map != null) {
       for (DomainPresenceInfo dpi : map.values()) {
         Domain dom = dpi.getDomain();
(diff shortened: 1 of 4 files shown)


doxiao:
The behaviors with and without the changes have been verified with manual testing. IMHO, it is probably more practical to verify the fix with integration test cases than unit test cases.

russgold:
I disagree. When you verify fixes with unit tests, they are verified on every build in milliseconds, which prevents regressions. When you verify them with integration tests, they only get run when those integrations tests are run - and if they are not added to QUICKTEST, thus adding minutes to that run, they only get verified very occasionally.

Can you describe the purpose of each change you've made?

doxiao:
The issue that the changes here address only happens when a namespace (with some weblogic domains)  that the operator manages is deleted, removed from the operator's domainNamespaces list, recreated, added back to the domainNamespaces list, and then the same domains are recreated. 

The changes just clean up the stale objects from the data structure that the operator maintains for the deleted domains so that the new objects can be populated and used by the operator. Without the changes, the recreated domains did not come up, and with the changes, they did.

russgold:
Can you tell me how you tested this? Looking at the code, it seems to me that the important change is shutting down the JobWatcher; if you don't, when a new namespace is created, the old JobWatcher won't see it. If that's the case, it is not actually important to clear the DOMAINS entries. 

And we can verify this with a unit test that confirms that, after a namespace is deleted, the JobWatcher for it is also deleted (or we could enhance the Watcher code to be resilient to namespace recreation, but this is simpler).

doxiao:
Manual testing steps:
1) install the operator using the helm charts with namespace default and ns1 in domainNamespaces.
2) delete the namespace ns1
3) use "helm upgrade" to remove "ns1" from domainNamespaces
4) recreate namespace ns1
5) use "helm upgrade" to add ns1 back to domainNamespaces
6) create domain1 in ns1 using one of the samples, and the server pods came up
7) repeat steps 2)-6)

Without the changes, domain1 did not come up at the end of step 7); no server pod started.
With the changes, the domain (all server pods) successfully started at the end of step7).

russgold:
I have added a unit test to verify the changed behavior: that removing a namespace from the operator's target namespaces and then re-adding it creates a new watcher. We don't actually need to clear the map of domains, as far as I can tell, @doxiao maybe you can verify that.

Ultimately, the namespace processing is all-over-the-place and should be cleaned up. The unit test will prevent us from breaking this.

doxiao:
Just to confirm - the JobWatcher clean up is the key for fixing the issue, and the clean up of the DOMAINS entries is nice to have. 
Thanks for the unit test cases; you beat me to it:). 
By the way, I filed a task to add integration test cases for this too (OWLS-76871).

russgold:
I would say that the DOMAINS entry cleanup is irrelevant, as no code cares whether the entries are already present.

tbarnes-us:
>I would say that the DOMAINS entry cleanup is irrelevant, as no code cares whether the entries are already present.
Seems worth keeping the cleanup.

PR #1262, opened 2019-09-18
OWLS-76806 (Resolves #1231) - scripts such as livenessProbe.sh should use JAVA_HOME if set

### LINKED ISSUE #1231: livenessProbe.sh should use JAVA_HOME if set
livenessProbe.sh script mounted via the operator generated config map at /weblogic-operator calls jps, which is assumed to be in the PATH.  Instead, this should use the JAVA_HOME environment variable if it is set.

Reference...
https://github.com/oracle/weblogic-kubernetes-operator/blob/master/operator/src/main/resources/scripts/livenessProbe.sh

alai8:
Append PATH with $JAVA_HOME/bin if $JAVA_HOME is set before calling jps

Tested on modified WebLogic image where java binaries are not found in PATH env var, but only in $JAVA_HOME/bin, and pods are started and remained in ready state.

Passed jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2605/testReport/

### DIFF
--- a/operator/src/main/resources/scripts/utils.sh
+++ b/operator/src/main/resources/scripts/utils.sh
@@ -418,3 +418,14 @@ function getAdminServerUrl() {
   fi
   echo ${admin_protocol}://${AS_SERVICE_NAME}:${ADMIN_PORT}
 }
+
+#
+# adjustPath
+#   purpose: Prepend $PATH with $JAVA_HOME/bin if $JAVA_HOME is set
+#
+function adjustPath() {
+  if [ ! -z ${JAVA_HOME} ]; then
+    export PATH="${JAVA_HOME}/bin:$PATH"
+  fi
+}
+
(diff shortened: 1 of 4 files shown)


operator/src/main/resources/scripts/utils.sh
```diff
@@ -418,3 +418,14 @@ function getAdminServerUrl() {
   fi
   echo ${admin_protocol}://${AS_SERVICE_NAME}:${ADMIN_PORT}
 }
+
+#
+# adjustPath
+#   purpose: Prepend $PATH with $JAVA_HOME/bin if $JAVA_HOME is set
+#
+function adjustPath() {
+  if [ ! -z ${JAVA_HOME} ]; then
```
qvdang: Maybe add a condition here to see if $JAVA_HOME/bin has already been added to the PATH?
Something like
if ! [[ $PATH =~ "${JAVA_HOME}/bin" ]]
then
  \# change the PATH
fi

PR #1265, opened 2019-09-20
Reduced execution time, added support to execute against different versions of Grafana, Prom, WDT, monitoring exporter

marinakog:
Reduced execution time, added support to execute against different versions of Grafana, Prom, WDT, monitoring exporter. Run test against certified versions and check against the latests chart versions of Grafana and Prometheus


marinakog:
jerkin run passed http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2638/

marinakog:
> 1. I see a lot of cases of code like this:
> 
> ```
>      Assume.assumeTrue(FULLTEST);
>      String testMethodName = new Object() {}.getClass().getEnclosingMethod().getName();		            	      
>      logTestBegin(testMethodName);
> ```
this is the style we use for all integration tests

> 
> That seems like needless duplication. Is that style used throughout the integration tests?
> 
> 1. I see that you are still git clone'ing and building the exporter. We now have instructions for configuring a released version, which will not only be faster, but will also prevent changes to the exporter from breaking this test. That is, you'd need to explicitly switch to a new release to get new
>    behavior. It would also allow me to simplify the way the exporter is built.

Changed the code to download released version instead of building

marinakog:
Updated to set versions via env and Jenkins

PR #1272, opened 2019-09-27
Update license format

rjeberhard:
Avi let me know that the official test of the UPL and the approved short format has changed.


rjeberhard:
Just find and replace in my IDE ;)

rjeberhard:
... but I also updated the Checkstyle rule so I know I got all the files.

PR #1277, opened 2019-09-27
Simplify WaitForReady step

russgold:
Massive refactoring of the WaitForReadyStep functionality in the Domain and Pod watchers, expanding test coverage, simplifying the code, and extracting some common functionality. This will create a firmer foundation for the next change, which will actually monitor certain introspection job failures not currently being detected.

Kudos to @alai for getting better code coverage in the JobWatcher than I originally had in the PodWatcher.

The main test and functionality changes are related to the fact that the steps have to consider three cases in which the resource is found to be ready:

It is ready when the step is created
It is detected as a ready on a callback, and
It became ready between the time we created the step and the time we registered the callback
All "isReady" functionality must operate in all three scenarios.

Problem with previous PR was that I had ignored the case of multiple pod MODIFIED events with the pod not ready, followed by one where it became ready. I have added a unit test for that case, now.

Will declare ready for merge once the Jenkins run passes.


pfmackin:
Change Simply to Simplify in title.

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
```diff
@@ -199,107 +171,54 @@ public Step waitForDelete(V1Pod pod, Step next) {
...
   private class WaitForPodReadyStep extends WaitForPodStatusStep {
+
     private WaitForPodReadyStep(V1Pod pod, Step next) {
       super(pod, next);
     }
 
+    // A pod is ready if it is not being deleted and has the ready status.
+    @Override
+    protected boolean isReady(V1Pod result) {
+      return result != null && !PodHelper.isDeleting(result) && PodHelper.isReady(result);
```
pfmackin: curious why you have to check for not deleting.

russgold: That's unchanged from the old code. isDeleting returns true if there is a deletionTimestamp defined on the pod's metadata, which indicates that Kubernetes is shutting it down. Such a pod is never deemed "ready," even if the other conditions hold.

rjeberhard: This catches the case where we are deleting the current pod and then recreating a new pod with the same name for which we then want to wait for ready.  If the code to wait is too fast, then we would detect as "still" ready the pod that is being deleted.  This code catches that case.

russgold:
It passes on Jenkins now http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2752/

markxnelson:
@rjeberhard please review, if you approve, i can do the merge

PR #1283, opened 2019-10-02
Owls 73891 retry

russgold:
Re-add the changes for Owls 73891, this time with the unit test corrected.

The problem seems to be that, as the code in question was originally added without tests, it was not designed to be unit-testable, which made retrofitting a test more complex than it should have been.

1. The code pulls Helm values using System.setenv, which does not provide any way to set within a process; for some reason, I read it as using _system properties_, which is incorrect. I have supplied a means of stubbing out the call, but since we use these in multiple places, it would be nice to find a consistent interface in the operator.

2. The life-cycle control of namespaces is controlled by multiple variables _isNamespaceStarted_ and _isNamespaceStopping_, both of which are maps. It seems to be more complex than necessary; possibly the functionality could be made to work with a single map to a state (but that would affect the watchers).

This PR will be enabled once the Jenkins test passes.


russgold:
And it passes Jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2756/

markxnelson:
@rjeberhard please review, if you approve, i can do the merge

russgold:
@rjeberhard Conflict resolved

russgold:
And revised change passes Jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2818/

.gitignore
```diff
@@ -45,3 +45,5 @@ dependency-reduced-pom.xml
 *.iws
 docs-source/docs
 .git/logs
+docs/charts/index.yaml
```
rjeberhard: Why did you add this?  We set up the build to keep the helm chart repository files up-to-date.

russgold: I had noticed that every time you do a build for testing, it wants to push your changes to this file to git, even if you make no changes to the chart. A fair number of PRs have included what seem to be extraneous changes like this. 

I think we have a philosophy difference, here. The standard way of using Maven is to distinguish between source and build. You check in source code, and do a build for test and release. We're combining them, and maybe I don't understand your approach, as it seems error-prone. Can you explain?

rjeberhard: These files are built from the chart, so the file is rebuilt anytime that files in the chart change.  I'm happy to hear ideas to improve the build process, but not as part of an unrelated PR.

PR #1295, opened 2019-10-09
add missing information to the domain home in image doc

markxnelson:


markxnelson:
also merged to develop in 9130e60fe

PR #1319, opened 2019-10-31
OWLS-76689 - SOA:ADMINSERVER TAKING TIME TO RESTART USING SERVERSTARTPOLICY(NEVER TO IF_NEEDED) ON OKE CLUSTER 

alai8:
Minor changes to preCreateServerServices behavior by removing precreated services for managed servers when both admin server and all managed servers in the cluster are being brought down.                      

Issue fixed (OWLS-76689):
In SOA domains, the startup of admin server involves JNDI lookup of managed servers, restarting the admin server was found to be very slow, when the preCreateServerServices feature was enabled. This is due to each managed server host name was able to resolve due to the existence of precreated server services, so the JNDI lookup will attempt to connect to each managed server until the configured timeout value in the JNDI lookup, since the managed servers are not running. With this change, each lookup of the managed server will result in a UnknownHostException, and failed fast, which was the same behavior as when the admin server was started up the first time when the domain was first started.

Filer verified that the change reduced the admin server startup time from 22 minutes to around 2 minutes in the restart scenario. Filer also verified that the
issue in OWLS-75155, which was why the preCreateServerService feature was introduced, did not regress with this change.

jenkins http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/2990/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
@@ -156,7 +156,7 @@ public class ManagedServersUpStep extends Step {
         servers.add(serverName);
         addStartupInfo(new ServerStartupInfo(serverConfig, clusterName, server));
         addToCluster(clusterName);
-      } else if (server.isPrecreateServerService()) {
+      } else if ((!domain.isShuttingDown() || server.shouldStart(0)) && server.isPrecreateServerService()) {
         servers.add(serverName);
         addStartupInfo(new ServerStartupInfo(serverConfig, clusterName, server, true));
       }
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
@@ -143,6 +143,21 @@ public class ManagedServersUpStep extends Step {
       this.domain = domain;
     }
 
+    /**
+     * Checks whether we should pre-create server service for the given server.
+     *
+     * @param server ServerSpec for the managed server
+     * @return True if we should pre-create server service for the given managed server, false
+     *         otherwise.
+     */
+    boolean shouldPrecreateServerService(ServerSpec server) {
+      if (server.isPrecreateServerService()) {
+        // skip pre-create if admin server and managed server are both shutting down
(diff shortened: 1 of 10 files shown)


operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
```diff
@@ -477,6 +489,46 @@ public void whenShuttingDown_allowAdminServerNameInListOfServers() {
...
   }
 
+  @Test
+  public void whenClusterStartupDefinedWithPreCreateServerService_addAllToServers() {
+    configureCluster("cluster1").withPrecreateServerService(true);
+    addWlsCluster("cluster1", "ms1", "ms2");
+
+    invokeStep();
+
+    assertThat(getServers(),
```
russgold: Not sure why you have this on multiple lines. 

allOf( hasItem("ms1"), hasItem("ms2") )

doesn't take up that much horizontal space, and seems a bit easier to read, I think. In fact, I suspect that the entire assert would fit comfortably on a single line.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -156,7 +156,7 @@ void addServerIfNeeded(@Nonnull WlsServerConfig serverConfig, WlsClusterConfig c
         servers.add(serverName);
         addStartupInfo(new ServerStartupInfo(serverConfig, clusterName, server));
         addToCluster(clusterName);
-      } else if (server.isPrecreateServerService()) {
+      } else if ((!domain.isShuttingDown() || server.shouldStart(0)) && server.isPrecreateServerService()) {
```
russgold: This condition is a bit complex. What is the intent? Can you describe it in a simple phrase that could be a method name?

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
russgold: I don't see a change or response to this comment.

alai8: Oops. I missed some occurrences. Should be fixed now. Thanks!

PR #1323, opened 2019-11-07
OWLS-77556: replace complete env dump in pods/job with curated listin…

tbarnes-us:
…g of env var values

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.sh
+++ b/operator/src/main/resources/scripts/introspectDomain.sh
@@ -67,7 +67,8 @@ exportInstallHomes
 
 # check if prereq env-vars, files, and directories exist
 
-checkEnv DOMAIN_UID \
+checkEnv -q \
+         DOMAIN_UID \
          NAMESPACE \
          DOMAIN_HOME \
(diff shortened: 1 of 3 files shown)


alai8:
LGTM. A minor comment: does checkEnv currently print the env var? Should we modify checkEnv to not print the env var value if the env var is set, so we don't print the values an extra time?

tbarnes-us:
> @alai8 wrote: A minor comment: does checkEnv currently print the env var? Should we modify checkEnv to not print the env var value if the env var is set, so we don't print the values an extra time?

There is some redundancy, but checkEnv has varied uses

tbarnes-us:
> @alai8 wrote: A minor comment: does checkEnv currently print the env var? Should we modify checkEnv to not print the env var value if the env var is set, so we don't print the values an extra time?

Looking into it.

tbarnes-us:
> > @alai8 wrote: A minor comment: does checkEnv currently print the env var? Should we modify checkEnv to not print the env var value if the env var is set, so we don't print the values an extra time?
> 
> Looking into it.

Done.

PR #1329, opened 2019-11-15
Added back MonitoringExporter, fixed db issues in Jenkins run

marinakog:
clean run: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3339/


marinakog:
> If we are not building the MonitoringExporter Project in our test infra, can we remove the git clone command in buildMonitoringExporter.sh script

We still need to clone monitoring project, we  build coordinator image and use endtoend  and deployments scripts to run our integration tests

vanajamukkara:
If someone wants to run the tests with lower version of monitoring exporter , we should provide the option to give mon exporter branch value as well. You can work on a new PR for this..merging this one..

PR #1332, opened 2019-11-15
Prepare file permissions on persistent volume using an init container

rjeberhard:
Several customers have reported difficulty creating a domain on a persistent volume because default file permissions on the volume are set to only give permission to the root user.  This change uses an initContainer to give permission to the "oracle" (1000) user.


markxnelson:
oh, i jut did that in my PR too :)

PR #1333, opened 2019-11-15
Operator version 2.4.0

rjeberhard:


markxnelson:
I think you should exclude the path `/docs` from this PR.  There are a lot of updates in there that are changing archived files, and even third party copyright notices.  I do not believe those changes are "correct".

PR #1339, opened 2019-11-18
Model in image merge

rjeberhard:


rjeberhard:
@jshum2479, I created this so I could see the delta -- I can now see that the merge eliminated some of your model-in-image work.  I'll try to recover it.  If you want to look too, then ignore the thousands of files and just focus on files you know you changed for mii.

PR #1343, opened 2019-11-20
Use patch to update status

russgold:
Moves status update logic into the status class (effect of adding new domain conditions).
When a 'failed' condition is added, reflects the reason and message to the domain status
Uses patches to update the status


russgold:
Passed Jenkins test http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3397/

PR #1344, opened 2019-11-20
Added request-timeout jenkins-ignore

hzhao-github:
the test passed ib both 3391 and 3392.


vanajamukkara:
can you submit 4 more jobs and see

PR #1369, opened 2019-12-12
Verify configured secrets before running introspector

russgold:
Queries available secrets from Kubernetes before validating the domain and reports a failure if the referenced weblogicCredential secret is not defined.


tbarnes-us:
The pull title says it's checking the WL credentials secret.  We also need to verify any other referenced secrets.  There can be override secrets, and I recall there's also an image pull secret.

russgold:
We have three different types of secrets, and each is specified in a different way. We should really fix that to be consistent.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -595,6 +603,39 @@ private void checkReservedIntrospectorVariables(BaseConfiguration configuration,
       new EnvironmentVariableCheck(IntrospectorJobEnvVars::isReserved).checkEnvironmentVariables(configuration, prefix);
     }
 
+    private void addMissingSecrets(KubernetesResourceLookup resourceLookup) {
+      verifySecretExists(resourceLookup, getWebLogicCredentialsSecretName(), SecretType.WebLogicCredentials);
+      for (V1LocalObjectReference reference : getImagePullSecrets())
+        verifySecretExists(resourceLookup, reference.getName(), SecretType.ImagePull);
+      for (String secretName : getConfigOverrideSecrets())
+        verifySecretExists(resourceLookup, secretName, SecretType.ConfigOverride);
```
tbarnes-us: FYI - More secrets are coming 'soon' for model-in-image, and config-override-secrets is going to be deprecated in favor of a new more generic secret list. I suspect it'll be easy to add a new secret and forget to wire it into this check.  Is it possible (and reasonably easy!) to come up with an approach that checks all secrets on the domain generically (perhaps discovering them via introspection?)   Or somehow have a unit test or assert that does the same?   Or come up  with a coding pattern that makes it less likely to forget - for example 'wraps' all secrets in the same helper in way that makes it more obvious that all secrets need to go through the same path?

russgold: The first problem is that we are not at all consistent in how we record secrets:

- the webLogicCredentialsSecret is represented as a V1SecretReference
- the imagePullSecrets are represented as V1LocalObjectReferences
- the configOverrideSecrets are represented as strings.

because all we really need is the name of the secret. We'd probably be better off picking one single style and sticking to it. 

Even then, it's not so clear how to recognize them. Best is just that, as fields are added, we also add unit tests to show that we are checking them.

PR #1376, opened 2019-12-16
Clean up ConfigMap watcher map and log messages

doxiao:
1) Fix log messages for config map events. The CM_CREATED, CM_REPLACED and CM_EXISTS log messages expect three parameter but the calling code only supplied two parameters. This is an old bug; not related to namespace events related work.

2) Fix some of the security check warning messages; the check() method needs to pass in the namespace when log a warning.

3) Clean up config map watcher map when a namespace that is deleted or removed from the operator's domainNamespaces.


doxiao:
Internal Jenkins test results:
1) Quick test passed: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3735/
2) Full test suite parallel run had 4 failures: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3736/. But the failures are different from the failures in the full suite against the recent develop branch: http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3738/.

Looking into the failures in the full suite parallel run.

doxiao:
As far as I can tell, the failures should not be related to my change. I have picked off re-runs against both develop and my branch and see if the failures persist.

doxiao:
The re-run had three failures (http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3748/), which are different from the previous run. So it looks like there are intermittent failures in the full test suite run.

doxiao:
Just had a clean full run. http://wls-jenkins.us.oracle.com/job/weblogic-kubernetes-operator-javatest/3750/

PR #1379, opened 2019-12-21
Opupgradefix

sankarpn:
The image property is not properly set when running in shared cluster for upgrade tests. This fix looks for the presence of operatorImageName, and operatorImageTag regardless of where the test is running and then sets the image property in operatorMap.


sankarpn:
Here is the external Jenkins test run results.
All tests
http://build.weblogick8s.org:8080/job/weblogic-kubernetes-operator-quicktest/797/testReport/

ItOperatorUpgrade tests alone.
http://build.weblogick8s.org:8080/job/weblogic-kubernetes-operator-quicktest/796/testReport/

PR #1381, opened 2020-01-07
add FAQs for mapping PVC and ConfigMap to instances

russgold:
Added FAQ entries to explain how to add per-server persistent volume claims and config maps.

I could not find information on what is actually required to get the new entries added to the index page, or the reason for the headers on other such pages that include dates and weight. Is that documented somewhere?

Note: this requires creating both a volume and a volumeMount entry. If we aren't attempting to support the full range of persistent volumes, but only these two, we might consider simplifying to a single entry. As it is, we should add some validations, including the existing of a matching volume for every volume mount, and possibly also some validation that the specified resource (PVC or ConfigMap) actually exists.

### DIFF
--- a/docs-source/content/faq/configmaps.md
+++ b/docs-source/content/faq/configmaps.md
@@ -37,7 +37,7 @@ in your [domain resource](https://github.com/oracle/weblogic-kubernetes-operator
 
 ```
 This provides access to two files, found at paths `/weblogic-operator/my/first` and `/weblogic-operator/my/second`. 
-Both a volume and a volumeMount entry are required, and must have the same name. The name of the ConfigMap is 
+Both a `volume` and a `volumeMount` entry are required, and must have the same name. The name of the `ConfigMap` is 
 specified in the `name` field under the `configMap` entry. The `items` entry is an array,
-in which each entry maps a ConfigMap key to a file name under the directory specified as `mountPath` under a volumeMount.
+in which each entry maps a `ConfigMap` key to a file name under the directory specified as `mountPath` under a `volumeMount`.
 
--- a/docs-source/content/faq/configmaps.md
+++ b/docs-source/content/faq/configmaps.md
@@ -1,0 +1,33 @@
+> I need to provide an instance with access to a Config Map
+
+Configuration information in Kubernetes may be provided by a 
+[ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-a-configmap), 
+which consists of a set of key-value pairs. Each entry may be accessed by one or more operator-managed nodes
+as a read-only text file. Access can be provided across the domain, within a single cluster, or for a single server.
+In each case, the access is configured within the ``ServerPod`` element of the configuration. For example, given
+a ConfigMap named `my-map` with entries `key-1` and `key-2`, you can provide access to both values as separate files
+in the same directory within the `cluster-1` cluster with the following: 
+
+```
+  clusters:
+  - clusterName: cluster-1
+    serverPod:
+      volumes:
+      - name: my-volume-1
+        configMap:
+          name: my-map
+          items: 
+          - key: key-1
+            path: first
+          - key: key-2
+            path: second
+      volumeMounts:
+      - name: my-volume-1
+        mountPath: /weblogic-operator/my
+
+```
+This provides access to two files, found at paths `/weblogic-operator/my/first` and `/weblogic-operator/my/second`. 
+Both a volume and a volumeMount entry are required, and must have the same name. The name of the ConfigMap is 
+specified in the `name` field under the `configMap` entry. The `items` entry is an array,
+in which each entry maps a ConfigMap key to a file name under the directory specified as `mountPath` under a volumeMount.
+
--- a/docs-source/content/faq/configmaps.md
+++ b/docs-source/content/faq/configmaps.md
@@ -1,12 +1,22 @@
-> I need to provide an instance with access to a Config Map
+---
+title: "Providing access to a Config Map"
+date: 2020-01-07T15:02:28-05:00
+draft: false
+weight: 11
+---
+> I need to provide an instance with access to a Config Map.
 
-Configuration information in Kubernetes may be provided by a 
+Configuration files can be supplied to Kubernetes pods and jobs by a 
 [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-a-configmap), 
 which consists of a set of key-value pairs. Each entry may be accessed by one or more operator-managed nodes
 as a read-only text file. Access can be provided across the domain, within a single cluster, or for a single server.
-In each case, the access is configured within the ``ServerPod`` element of the configuration. For example, given
+In each case, the access is configured within the `serverPod` element of the desired scope. 
+
+For example, given
 a ConfigMap named `my-map` with entries `key-1` and `key-2`, you can provide access to both values as separate files
-in the same directory within the `cluster-1` cluster with the following: 
+in the same directory within the `cluster-1` cluster with the following
+in your [domain resource](https://github.com/oracle/weblogic-kubernetes-operator/blob/master/docs/domains/Domain.md):
+ 
 
 ```
   clusters:
--- a/docs-source/content/faq/volumes.md
+++ b/docs-source/content/faq/volumes.md
@@ -1,11 +1,20 @@
-> I need to provide an instance with access to a Persistent Volume Claim
+---
+title: "Providing access to a Persistent Volume Claim"
+date: 2020-01-07T15:02:28-05:00
+draft: false
+weight: 10
+---
+> I need to provide an instance with access to a Persistent Volume Claim.
 
 Some applications need access to a file, either to read data or to provide additional logging beyond what is 
 built into the operator. One common way of doing that within Kubernetes is to create a 
 [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) (PVC) and map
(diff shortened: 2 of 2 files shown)


tbarnes-us:
> I could not find information on what is actually required to get the new entries added to the index page, or the reason for the headers on other such pages that include dates and weight. Is that documented somewhere?

I assume there's documentation somewhere in the hugo docs, but, when I added a FAQ entry recently, I experimentally found that (a) the weight determines the location of the FAQ entry relative to the other FAQ entries - the lower the weight the earlier the FAQ entry, and (b) the FAQ index is automatically updated when new FAQ entries are added.

tbarnes-us:
FWIW I recently took a stab at setting the weights of each FAQ entry - they were all at '1' previously.  Feel free to change the weights around if you think the ordering is wacky.

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
...
+  accessModes:
+    - ReadOnlyMany
+  volumeMode: Filesystem
+  resources:
+    requests:
+      storage: 8Gi
+  storageClassName: slow
+```
+
+To provide access to this claim within the `cluster-1` cluster, specify
```
tbarnes-us: specify --> specify the following in your domain resource:

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
+> I need to provide an instance with access to a Persistent Volume Claim
+
+Some applications need access to a file, either to read data or to provide additional logging beyond what is 
+built into the operator. One common way of doing that within Kubernetes is to create a 
+[PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) (PVC) and map
+it to a file. The domain configuration can then be used to provide access to the claim across the domain, 
+within a single cluster, or for a single server.
+In each case, the access is configured within the ``ServerPod`` element of the configuration. For example, here is
```
tbarnes-us: "In each case, the access is configured within the ``ServerPod`` element of the configuration."

looks incorrect.  (1) Element name is 'serverPod' not 'ServerPod'., (2) The cluster configuration is in 'clusters', not 'ServerPod', as per your example below.

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
...
+a read-only PersistentVolumeClaim specification. Note that its name is `myclaim`.
```
tbarnes-us: Consider moving "For example, ..." to it's own new paragraph.

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
...
+        persistentVolumeClaim:
+          claimName: myclaim
+      volumeMounts:
+      - name: my-volume-1
+        mountPath: /weblogic-operator/my/volume1
+
+``` 
+Note the use of the claim name in the `claimName` field of the `volume` entry. Both a volume and a 
+volumeMount entry are required, and must have the same name. The volume entry associates that name with the claim,
+while the volumeMount entry defines the path to it that the application can use to access the file.
```
tbarnes-us: Back-quote all occurrences of volume, volumeMount.

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
...
+
+NOTE: if the PVC is mapped either across the domain or to a cluster, 
+its access mode must be either `ReadOnlyMany` or `ReadWriteMany`.
+
+[ConfigMaps](https://kubernetes.io/docs/concepts/storage/volumes/#configmap). In both cases, all you need
+is the name of the corresponding Kubernetes resource.
```
tbarnes-us: This sentence looks like it was dangling.

docs-source/content/faq/configmaps.md
```diff
@@ -0,0 +1,33 @@
+> I need to provide an instance with access to a Config Map
+
+Configuration information in Kubernetes may be provided by a
```
tbarnes-us: It'd be more clear IMO to change

Configuration information in Kubernetes may be provided by a

to

Configuration files can be supplied to Kubernetes pods and jobs by a

docs-source/content/faq/configmaps.md
```diff
@@ -0,0 +1,33 @@
+> I need to provide an instance with access to a Config Map
```
rosemarymarano: Both FAQs need a title. The date is also required but the weight is not (unless you want to specify ordering). The FAQ are not "added to the index page" but appear under the category where you locate the MD files.

docs-source/content/faq/configmaps.md
```diff
@@ -0,0 +1,33 @@
...
+
+Configuration information in Kubernetes may be provided by a 
+[ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-a-configmap), 
+which consists of a set of key-value pairs. Each entry may be accessed by one or more operator-managed nodes
+as a read-only text file. Access can be provided across the domain, within a single cluster, or for a single server.
+In each case, the access is configured within the ``ServerPod`` element of the configuration. For example, given
```
tbarnes-us: "In each case, the access is configured within the ServerPod element of the configuration."

looks incorrect. (1) Element name is 'serverPod' not 'ServerPod'., (2) The cluster configuration is in 'clusters', not 'ServerPod', as per your example below.

How about "Access is configured within a domain resource using its 'serverPod', 'adminServer', 'managedServers', or 'clusters' elements. (I'm not sure if all 4 have access - please confirm.)

Also, please turn 'domain resource' into a link that references the "Domain resource" doc page.

docs-source/content/faq/configmaps.md
```diff
@@ -0,0 +1,33 @@
...
+a ConfigMap named `my-map` with entries `key-1` and `key-2`, you can provide access to both values as separate files
+in the same directory within the `cluster-1` cluster with the following:
```
tbarnes-us: Put 'For example ...' into its own paragraph.

docs-source/content/faq/volumes.md
```diff
@@ -0,0 +1,49 @@
...
+      volumeMounts:
+      - name: my-volume-1
+        mountPath: /weblogic-operator/my/volume1
+
+``` 
+Note the use of the claim name in the `claimName` field of the `volume` entry. Both a volume and a 
+volumeMount entry are required, and must have the same name. The volume entry associates that name with the claim,
+while the volumeMount entry defines the path to it that the application can use to access the file.
+
+NOTE: if the PVC is mapped either across the domain or to a cluster,
```
rosemarymarano: if -> If

docs-source/content/faq/volumes.md
rosemarymarano: PersistentVolumeClaim -> `PersistentVolumeClaim` (code font, place inside single back ticks)

docs-source/content/faq/configmaps.md
rosemarymarano: Also, the problem statements in both FAQs should end with a period. I see that we missed that detail in the Cannot Pull Image FAQ; that should be fixed, not propagated.

docs-source/content/faq/volumes.md
russgold: (1) Yes, it should be 'serverPod'
(2) I don't understand your second comment. The 'volumes' and 'volumeMounts' fields are under serverPod which is inherited by Cluster from BaseConfiguration.

russgold: You can configure either the admin server or a managed server. The latter is configured under the managedServers: array. I am assuming that the users know _something_ about the domain resource.

docs-source/content/faq/volumes.md
russgold: Oops - thought I'd deleted that

docs-source/content/faq/volumes.md
tbarnes-us: I don't think the inheritance hierarchy is obvious to users - if one is directly working with domain resource XML there's nothing there to tell you that 'clusters' inherits the same type as 'serverPod'.  It's much more clear IMO just to directly state the names of the parent stanzas that support volume and volumeMount child stanzas.  MBeans in WL work very similarly - two beans may inherit the same fields, but we typically treat them independently in documentation.

docs-source/content/faq/volumes.md
tbarnes-us: Done. Thanks.

docs-source/content/faq/volumes.md
tbarnes-us: done

docs-source/content/faq/volumes.md
tbarnes-us: done

docs-source/content/faq/configmaps.md
```diff
@@ -0,0 +1,43 @@
...
+      volumeMounts:
+      - name: my-volume-1
+        mountPath: /weblogic-operator/my
+
+```
+This provides access to two files, found at paths `/weblogic-operator/my/first` and `/weblogic-operator/my/second`. 
+Both a volume and a volumeMount entry are required, and must have the same name. The name of the ConfigMap is 
+specified in the `name` field under the `configMap` entry. The `items` entry is an array,
+in which each entry maps a ConfigMap key to a file name under the directory specified as `mountPath` under a volumeMount.
+
```
tbarnes-us: quote all occurrences of volume and volumeMount (just as you did for volumes.md)

docs-source/content/faq/volumes.md
russgold: > Did you disagree that PersistentVolumeClaim should be in code font?

No, just missed it.

PR #1382, opened 2020-01-07
Nm mem args

lennyphan:
NM_MEM_ARGS environment variable will, by default, include the java security property '-Djava.security.egd=file:/dev/./urandom', if not already defined.  This is to help speed up the Node Manager startup, especially when the system has low entropy.   Also updated 'Domain Resource' section of user guide documentation to include defaulting information about 'NODEMGR_MEM_ARGS' environment variable.

### DIFF
--- a/operator/src/main/resources/scripts/startNodeManager.sh
+++ b/operator/src/main/resources/scripts/startNodeManager.sh
@@ -296,14 +296,7 @@ fi
 
 if [ -z "${NODEMGR_MEM_ARGS}" ]; then
   # Default JVM memory arguments for Node Manager
-  NODEMGR_MEM_ARGS="-Xms64m -Xmx100m"
-fi
-
-# If not already defined, set 'java.security.egd' option to /dev/urandom for
-# Node Manager JVM for low entropy.  Speeds up boot process of Node Manager.
-if [[ ! "${NODEMGR_MEM_ARGS}" =~ .*urandom* ]]
-then
-  NODEMGR_MEM_ARGS="${NODEMGR_MEM_ARGS} -Djava.security.egd=file:/dev/./urandom "
+  NODEMGR_MEM_ARGS="-Xms64m -Xmx100m -Djava.security.egd=file:/dev/./urandom "
 fi
 
 # We prevent USER_MEM_ARGS from being applied to the NM here and only pass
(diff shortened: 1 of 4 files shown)


operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -299,6 +299,13 @@ if [ -z "${NODEMGR_MEM_ARGS}" ]; then
   NODEMGR_MEM_ARGS="-Xms64m -Xmx100m"
 fi
 
+# If not already defined, set 'java.security.egd' option to /dev/urandom for
+# Node Manager JVM for low entropy.  Speeds up boot process of Node Manager.
+if [[ ! "${NODEMGR_MEM_ARGS}" =~ .*urandom* ]]
+then
+  NODEMGR_MEM_ARGS="${NODEMGR_MEM_ARGS} -Djava.security.egd=file:/dev/./urandom "
```
tbarnes-us: It's my understanding that urandom isn't considered to be the most secure option for all situations, so we should provide a way to setup the NM without it.   Instead of appending urandom if it's missing, how about making the default for NODEMGR_MEM_ARGS "-Xms64m -Xmx100m -Djava.security.egd=file:/dev/./urandom"?  Plus updating the doc above to accordingly?

lennyphan: Instead of searching for 'urandom' specifically, should we search for 'java.security.egd' security property is specified and set to 'urandom' if the security property isn't set?

tbarnes-us: Why not simply remove the new if statement, and include urandom in the NODEMGR_MEM_ARGS default that's initialized at line 299?

lennyphan: I was thinking about the case where NODEMGR_MEM_ARGS is specified but doesn't define 'java.security.egd'.  Thus, we would boot Node Manager without any security random generator specified.

tbarnes-us: I think we don't need worry about that as long as we clearly document that customizing NODEMGR_MEM_ARGS implicitly remove's the NM's (often helpful) urandom. It becomes the admin's responsibility to specifically set it when they customize NM mem args.  This is very similar to the approach that we're already taking for the WLS command line arguments - you can see we document this just a few lines above in the same documentation.

lennyphan: Sounds good.  Thanks for the valuable feedback.  Will finish the changes and document.

tbarnes-us: The changes look good to me.

PR #1383, opened 2020-01-09
Make getAdminService idempotent

russgold:
The getAdminService method was changing the AdminServer object in a way that changed the behavior of equals(), causing repeated domain updates. This corrects that.


russgold:
Successful Jenkins run http://build.weblogick8s.org:8080/job/weblogic-kubernetes-operator-quicktest/822/

PR #1387, opened 2020-01-13
Introspect Test Fix: integration test PV directory cleanup now expect…

tbarnes-us:
…s tests to be one directory deeper

This change puts this standalone test's PV directories one level deeper so that the new version of 'cleanup.sh' can find them.

### DIFF
--- a/src/integration-tests/introspector/createTestRoot.sh
+++ b/src/integration-tests/introspector/createTestRoot.sh
@@ -15,9 +15,9 @@
 
 DOMAIN_UID=${1?}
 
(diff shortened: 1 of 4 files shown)


vanajamukkara:
LGTM. @anpanigr @tbarnes-us we are not running these tests in our integration tests, we should include them.

tbarnes-us:
> @anpanigr wrote:  Do we need to run these tests as part of the integration test on jenkin

Would be nice. I think there's already a JIRA tracking this.  Not a super high priority.

PR #1388, opened 2020-01-13
Improve node manager debugging

tbarnes-us:
Generate a NM thread dump if NM times out. Also fix NM .out dump to stdout.


tbarnes-us:
> Was it a bug that the code was doing "cat ${NODEMGR_OUT_FILE}" instead of "cat ${nodemgr_out_file}"?

@alai8 Yes.  At some point someone flipped the case of the variable - not sure why and I don't know when.

PR #1407, opened 2020-01-29
Owls 79377

anpanigr:
Modified Sample Script (samples/charts/util/setup.sh) to support Helm Version 2.x and 3.x


anpanigr:
Run the QUICKTESTS on internal Jenkin ( with Helm 2.x )
http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/4033/   ( All WLS Tests Passed )

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
-# Copyright (c) 2018, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+# Copyright (c) 2018,2019,2020 Oracle Corporation and/or its affiliates.  All rights reserved.
```
rjeberhard: This is incorrect.  It should be <begin year>,<space><most recent year>, so "2018, 2020"

kubernetes/samples/charts/util/setup.sh
```diff
@@ -130,12 +154,12 @@ function deleteVoyager() {
...
-    echo "Delete Traefik operator." 
-    helm delete --purge $TNAME
-    kubectl delete ns traefik
+  if [ "$(helm list ${t_list_args}| grep $TNAME |  wc -l)" = 1 ]; then
+    echo "Uninstall Traefik operator." 
+    helm uninstall $TNAME ${t_uninstall_args}
+    kubectl delete ns ${TSPACE}
   else
-    echo "Traefik operator has already been deleted." 
+    echo "Traefik operator has already been unistalled"
```
rjeberhard: uninstalled

kubernetes/samples/charts/util/setup.sh
```diff
@@ -130,12 +154,12 @@ function deleteVoyager() {
 }
 
 function deleteTraefik() {
-  if [ "$(helm list | grep $TNAME |  wc -l)" = 1 ]; then
-    echo "Delete Traefik operator." 
-    helm delete --purge $TNAME
-    kubectl delete ns traefik
+  if [ "$(helm list ${t_list_args}| grep $TNAME |  wc -l)" = 1 ]; then
+    echo "Uninstall Traefik operator."
```
rjeberhard: Uninstalling?

kubernetes/samples/charts/util/setup.sh
```diff
@@ -111,13 +135,13 @@ function purgeCRDs() {
 }
 
 function deleteVoyager() {
-  if [ "$(helm list | grep $VNAME |  wc -l)" = 1 ]; then
-    echo "Delete Voyager Operator. "
-    helm delete --purge $VNAME 
-    kubectl delete ns voyager
+  if [ "$(helm list ${v_list_args} | grep $VNAME |  wc -l)" = 1 ]; then
+    echo "Uninstall Voyager Operator. "
```
rjeberhard: Uninstalling?

bsmoyers:

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
...
+VSPACE=voyager # NameSpace for Voyager
+TSPACE=traefik   # NameSpace for Traefik
+
+helm version --short | grep v3
+[[ $? == 0 ]] && HELM_VERSION=V3
+[[ $? == 1 ]] && HELM_VERSION=V2
+
+echo "Detected Helm Version [$HELM_VERSION]"
+
+if [ "$HELM_VERSION" == "V3" ]; then
```
tbarnes-us: Reverse the logic here and check for V2 instead of V3 (more future proofing).

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
...
 
 MYDIR="$(dirname "$(readlink -f "$0")")"
 VNAME=voyager-operator  # release name of Voyager
 TNAME=traefik-operator  # release name of Traefik
+VSPACE=voyager # NameSpace for Voyager
+TSPACE=traefik   # NameSpace for Traefik
+
+helm version --short | grep v3
+[[ $? == 0 ]] && HELM_VERSION=V3
+[[ $? == 1 ]] && HELM_VERSION=V2
```
tbarnes-us: This logic looks like it will return V2 if the helm version actually returns V4 or later.  Can you grep for v2 instead?  That will future proof this code (assuming V4 and V3 have the same syntax).

kubernetes/samples/charts/util/setup.sh
anpanigr: Resolved in latest update

kubernetes/samples/charts/util/setup.sh
anpanigr: Now using delete command instead of uninstall.  Uninstall does not work in 2.x version
3.x -  helm delete --keep-history 
2.x -  helm delete --purge   ( purge not supported )

kubernetes/samples/charts/util/setup.sh
anpanigr: Resolved

kubernetes/samples/charts/util/setup.sh
anpanigr: Resolved

kubernetes/samples/charts/util/setup.sh
anpanigr: Yes, I reversed the logic in recent commit

kubernetes/samples/charts/util/setup.sh
anpanigr: Yes, I reversed the logic in recent commit

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
-# Copyright (c) 2018, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+# Copyright (c) 2018,2020 Oracle Corporation and/or its affiliates.  All rights reserved.
```
markxnelson: fix copyright line please

`# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates. `

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
...
 
 #  This script is to create or delete Ingress controllers. We support two ingress controllers: traefik and voyager.
 
 MYDIR="$(dirname "$(readlink -f "$0")")"
 VNAME=voyager-operator  # release name of Voyager
 TNAME=traefik-operator  # release name of Traefik
+VSPACE=voyager # NameSpace for Voyager
+TSPACE=traefik   # NameSpace for Traefik
+
+helm version --short | grep v2
```
markxnelson: check specifically for the string you want instead of assuming that the lack of presence means it is the other version

markxnelson: this could also fail if the server version string has v2 or v3 in it.

markxnelson: add `--client`

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,12 +1,36 @@
...
+   v_list_args=""
+   t_list_args=""
+   v_helm_delete="helm delete --purge"
+   t_helm_delete="helm delete --purge"
+   v_helm_install="helm install appscode/voyager --name $VNAME  "
+   t_helm_install="helm install stable/traefik --name $TNAME "
+else
+   v_list_args="--namespace $VSPACE "
+   t_list_args="--namespace $TSPACE "
+   v_helm_delete="helm delete --keep-history --namespace $VSPACE "
```
markxnelson: why `--keep-history` and not `--purge` here?

kubernetes/samples/charts/util/setup.sh
```diff
@@ -21,11 +45,11 @@ function createVoyager() {
   fi
   echo
 
-  if [ "$(helm list | grep $VNAME |  wc -l)" = 0 ]; then
-    echo "Ihstall voyager operator."
+  if [ "$(helm list ${v_list_args} | grep $VNAME |  wc -l)" = 0 ]; then
+    echo "Install voyager operator."
```
markxnelson: install -> installing

kubernetes/samples/charts/util/setup.sh
```diff
@@ -21,11 +45,11 @@ function createVoyager() {
...
     
-    helm install appscode/voyager --name $VNAME --version 7.4.0 \
-      --namespace voyager \
+    ${v_helm_install} --version 7.4.0 \
```
markxnelson: probably a good idea to make the voyager version a variable

kubernetes/samples/charts/util/setup.sh
tbarnes-us: @markxnelson 

> "check specifically for the string you want instead of assuming that the lack of presence means it is the other version".   

The code assumes v3 or later if v2 is not present so that future version updates beyond v3 don't necessarily break the code.   What do you advise?   Fail if it's not v2 or v3?

kubernetes/samples/charts/util/setup.sh
anpanigr: --purge is not supported in 3.x

kubernetes/samples/charts/util/setup.sh
anpanigr: Done

kubernetes/samples/charts/util/setup.sh
anpanigr: We are suggesting this as sample on development environment not on production.  Certification Matrix will grow if we verify each version of Voyager which is not in our control. We have already big matrix of K8s Version/Helm Version/Prometheous Version

kubernetes/samples/charts/util/setup.sh
anpanigr: From V3 and above there is no tiller deployment. 

[2.x]  helm version --short --client
Client: v2.16.1+gbbdfe5e
[3.x]  helm version --short --client
v3.0.2+g19e47ee

I have added --client parameter to verify the client only version  as follows ...
helm version --short --client | grep v2

kubernetes/samples/charts/util/setup.sh
anpanigr: Provide a new Optional Positional Parameter to take Custom Voyager Version
e.g.
 $ sh setup.sh
 Detected Helm Version [V3]
 [ERROR] Requires atleast 2 positional parameters
 usage: setup.sh create|delete traefik|voyager [voyager version]

kubernetes/samples/charts/util/setup.sh
```diff
@@ -1,18 +1,55 @@
...
 
 MYDIR="$(dirname "$(readlink -f "$0")")"
 VNAME=voyager-operator  # release name of Voyager
 TNAME=traefik-operator  # release name of Traefik
+VSPACE=voyager # NameSpace for Voyager
+TSPACE=traefik   # NameSpace for Traefik
+DefaultVoyagerVersion=7.4.0
+
+helm version --short --client | grep v2
+[[ $? == 0 ]] && HELM_VERSION=V2
```
markxnelson: this is still broken 
you should explicitly check for both V2 and V3, not just check for one and assume it is the other if it fails

anpanigr: Here V3 represent any version  above  v3 so that we donot have to change the script fro other version

TheFrogPad:
@anpanigr @mriccell Do we want to have this PR#1407 pickup the README.md changes from PR#1406 and resolve both issues using develop branch?

rjeberhard:
> @anpanigr @mriccell Do we want to have this PR#1407 pickup the README.md changes from PR#1406 and resolve both issues using develop branch?

This is what I've been asking for :)

kubernetes/samples/charts/util/setup.sh
anpanigr: Updated the script with explicit  Helm Version check
===== 
helm version --short --client | grep v2
[[ $? == 0 ]] && HELM_VERSION=V2
helm version --short --client | grep v3
[[ $? == 0 ]] && HELM_VERSION=V3
======

anpanigr:
@mriccell  @rjeberhard  Should we keep v12.0.0-rc.1 as default Voyager Version or 7.4.0 as  default Voyager Version

markxnelson:
I'm with Craig on this - the default should not be an RC. It should be the latest one that we know works reliably.

kubernetes/samples/charts/util/setup.sh
markxnelson: I'm sorry, this is still broken.  What it is not v2 or v3?  what if it is v1? or v4? Your logic will apply the "v3" settings in any case that is not v2.

anpanigr: We are NOT supporting V1 and V4 is not available ( how to check )  We will add another condition when V4 is available and it has a different syntax.  If there is not syntax change in future version of Helm, the current logic for V3 will hold good.

I am open to alternate logic/solution  if you do not agree with the current one.

anpanigr:
Should we bump the Default Voyager version to 10.0.0  from 7.4.0   since 11.0.1  does not work.
Ref https://github.com/appscode/voyager#supported-versions

markxnelson:
Does 10.0.0 work? Or the latest 10.x.x

anpanigr:
yes, version 10.0.0 works

kubernetes/samples/charts/util/setup.sh
tbarnes-us: Sorry about that Mark - I'd advised Pani to 'future proof' by assuming anything that wasn't v2 was v3.  You've advised Pani that we should outright fail if we can't be certain we are v2 or v3.  I'm fine either way.   (FWIW The reason I advised the former is that I thought this matched the spirit we use throughout the operator. We don't tend to have explicit version checks for various other dependencies throughout and consequently just assume the current syntax/API is supported regardless of version.  This 'future' proofs in the sense that we'll continue to work as long as API doesn't change out from underneath us.)

rjeberhard:
Agreed that we should use the latest released version of Voyager... whatever 10.x that is.

markxnelson:
yes it appears 10.0.0 is the latest 10,x release, I am happy to go with that release.

kubernetes/samples/charts/voyager/README.md
```diff
@@ -106,20 +128,26 @@ Then deploy the TLS Ingress.
...
+$ helm uninstall voyager-operator --namespace voyager
+```
+Using Helm 2:
 ```
 $ helm delete --purge voyager-operator
 ```
 
 ## Install and uninstall the Voyager operator with setup.sh
-Alternatively, you can run the helper script `setup.sh`, under the `kubernetes/samples/charts/util` folder, to install and uninstall Voyager.
+Alternatively, you can run the helper script `setup.sh` when using Helm 2, under the `kubernetes/samples/charts/util` folder, to install and uninstall Voyager.
```
TheFrogPad: Pani, you should update the text "when using Helm 2" -> "when using Helm 2 or Helm 3" as your updated to `setup.sh` works with Helm 2 or Helm 3. 

It would also be fine to remove this additional text in whole if you prefer....

This text was added for PR#1406 since the script on `master` only works with Helm 2.

anpanigr: setup.sh automatically take care both Helm Version, so no need to mention about Helm Version specific usage.  
Additionally  now setup.sh take an optional 3rd parameter to download  specific version of Voyager. Let me update the PR

rjeberhard:
@markxnelson, are you satisfied that your concerns have been addressed?  If so, please update your review.

markxnelson:
@rjeberhard apart from the incorrect copyright notice, all of my other concerns appear to have been addressed satisfactorily. thanks

rjeberhard:
Okay, @markxnelson, copyright is correct in the latest merge.

PR #1408, opened 2020-01-29
Remove the build badge 

markxnelson:
Remove the build badge  which is not using https, does not have a valid cert, and which is not being used 'right' anyway - not pointing to a consistent branch.


bsmoyers:

PR #1425, opened 2020-02-12
Add FAQ entry for persistent volume using OCI FSS

TheFrogPad:
Add documenation FAQ item about updating the `initContainers` with a different command for OCI FSS persistent volume handling as the domain-on-pv example fails when using OKE and a PV from OCI FSS.

This is the first cut at the information and I'm using the command that was initially discussed. I'm happy to change this command along with wording or the basic content layout I have started here.

The FAQ is trying to present the basic issue with some references and more details about how the problem shows up and where to go make the change.

Thanks, -Craig

### DIFF
--- a/docs-source/content/faq/oci-fss-pv.md
+++ b/docs-source/content/faq/oci-fss-pv.md
@@ -2,6 +2,7 @@
 title: "Using OCI File Storage (FSS) for Persistent Volumes"
 date: 2020-02-12T12:12:12-05:00
 draft: false
+weight: 6
 ---
 
 If you are running your Kubernetes cluster on Oracle Container Engine 
--- a/docs-source/content/faq/oci-fss-pv.md
+++ b/docs-source/content/faq/oci-fss-pv.md
@@ -79,7 +79,7 @@ spec:
       initContainers:
         - name: fix-pvc-owner
           image: %WEBLOGIC_IMAGE%
-          command: ["sh", "-c", "find %DOMAIN_ROOT_DIR%/. -maxdepth ! -name '.snapshot' ! -name '.' -print0 | xargs -0 chown 1000:1000"]
+          command: ["sh", "-c", "find %DOMAIN_ROOT_DIR%/. -maxdepth 1 ! -name '.snapshot' ! -name '.' -print0 | xargs -r -0 chown 1000:1000"]
           volumeMounts:
           - name: weblogic-sample-domain-storage-volume
             mountPath: %DOMAIN_ROOT_DIR%
--- a/docs-source/content/faq/oci-fss-pv.md
+++ b/docs-source/content/faq/oci-fss-pv.md
@@ -79,7 +79,7 @@ spec:
       initContainers:
         - name: fix-pvc-owner
           image: %WEBLOGIC_IMAGE%
-          command: ["sh", "-c", "find %DOMAIN_ROOT_DIR% ! -name '.snapshot' | xargs chown 1000:1000"]
+          command: ["sh", "-c", "find %DOMAIN_ROOT_DIR%/. -maxdepth ! -name '.snapshot' ! -name '.' -print0 | xargs -0 chown 1000:1000"]
           volumeMounts:
           - name: weblogic-sample-domain-storage-volume
             mountPath: %DOMAIN_ROOT_DIR%
--- a/docs-source/content/faq/oci-fss-pv.md
+++ b/docs-source/content/faq/oci-fss-pv.md
@@ -6,7 +6,7 @@ draft: false
 
 If you are running your Kubernetes cluster on Oracle Container Engine 
 for Kubernetes (commonly known as OKE), and you use OCI File Storage (FSS)
-for persistent volumes to store the WebLogic domain home then the file system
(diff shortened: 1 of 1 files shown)


docs-source/content/faq/oci-fss-pv.md
```diff
@@ -0,0 +1,96 @@
+---
+title: "Using OCI File Storage (FSS) for Persistent Volumes"
+date: 2020-02-12T12:12:12-05:00
+draft: false
```
tbarnes-us: This should have a weight in order to choose where it ends up in the FAQ contents list.  See other FAQs for their weights.

docs-source/content/faq/oci-fss-pv.md
```diff
@@ -0,0 +1,96 @@
...
+---
+
+If you are running your Kubernetes cluster on Oracle Container Engine 
+for Kubernetes (commonly known as OKE), and you use OCI File Storage (FSS)
+for persistent volumes to store the WebLogic domain home then the file system
+handling demonstrated in the operator persistent volume sample will require
+an update to properly initialize the file ownership on the persistent volume
+when the domain is initially created.
+
```
tbarnes-us: Long sentence.  Maybe at least put a comma before 'then the file system'.

docs-source/content/faq/oci-fss-pv.md
TheFrogPad: I didn't and don't see a weight on any FAQ file?

docs-source/content/faq/oci-fss-pv.md
TheFrogPad: Sure, @rosemarymarano, can you provide your usual input?

TheFrogPad: BTW, this is very similar to the FAQ of OCI Load Balancer so I can update there if needed...

docs-source/content/faq/oci-fss-pv.md
rosemarymarano: You're right. The FAQs do not have weights, not sure why other than perhaps we have not been particular about their order.

docs-source/content/faq/oci-fss-pv.md
```diff
@@ -0,0 +1,96 @@
...
+### Failure during domain creation with persistent volume sample
+
+The existing sample for [creation of a domain home on persistent volume](https://github.com/oracle/weblogic-kubernetes-operator/tree/master/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv)
+uses a Kubernetes job to create the domain. The sample uses an
+`initContainers` section to change the file ownership which will
+fail for OCI FSS created volumes used with an OKE cluster.
+
+The OCI FSS volume contains some files that are not modifiable thus
+causing the Kubernetes job to fail. The failure is see from the
```
rosemarymarano: see from the -> seen in the

docs-source/content/faq/oci-fss-pv.md
```diff
@@ -0,0 +1,96 @@
...
+      Finished:     Wed, 12 Feb 2020 18:28:53 +0000
+    Ready:          False
+    Restart Count:  0
+    Environment:    <none>
+
+   :
+```
+
+### Updating the domain on persistent volume sample
+In the snippet of the [create-domain-job-template.yaml](https://github.com/oracle/weblogic-kubernetes-operator/blob/master/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain-job-template.yaml)
```
rosemarymarano: In the snippet of the create-domain-job-template.yaml below, you will see -> In the following snippet of the create-domain-job-template.yaml, you can see

docs-source/content/faq/oci-fss-pv.md
```diff
@@ -0,0 +1,96 @@
...
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
+      containers:
+        - name: create-weblogic-sample-domain-job
+          image: %WEBLOGIC_IMAGE%
+
+  :
+```
+The new `command` shown above should be made to your copy of this template file
```
rosemarymarano: The new command shown above should be made to your copy of this template file and will result in the ownership updated only for the expected files before the WebLogic domain is created on the persistent volume. -> 
Use this new command in your copy of this template file. This will result in the ownership being updated only for the expected files before the WebLogic domain is created on the persistent volume.

rosemarymarano: Actually, the last part of this second sentence should be, "being updated for the expected files only, before the WebLogic domain is created on the persistent volume."
(Position of only: Keep only close to that which it modifies to improve the clarity and accuracy of the sentence.)

docs-source/content/faq/oci-fss-pv.md
tbarnes-us: Not a show-stopper, but the weight entry is still missing.  The FAQ files in the 'develop' branch look like they all have 'weight:' entries.

markxnelson: fixed

PR #1434, opened 2020-02-18
OWLS-77253 Support cluster without name in <dynamic-servers> config

alai8:
Minor change to fix error in introspectDomain.py while starting a domain with a dynamic-servers configuration that does not contain a name element, when using WebLogic image with patch for bug 30419039 applied.


markxnelson:
can you explain what this is about please?  isn't a cluster without a name going to break the operator anyway?

alai8:
> can you explain what this is about please? isn't a cluster without a name going to break the operator anyway?

Sorry about the confusion. It is not the cluster name, but the name of the <dynamic-servers> element inside a dynamic cluster configuration, such as:

  <cluster>
    <name>cluster-2</name>
    <cluster-messaging-mode>unicast</cluster-messaging-mode>
    <dynamic-servers>
      <server-template>c2-Template</server-template>
      <maximum-dynamic-server-count>2</maximum-dynamic-server-count>
      <calculated-listen-ports>false</calculated-listen-ports>
      <calculated-machine-names>true</calculated-machine-names>
      <server-name-prefix>c2-</server-name-prefix>
      <dynamic-cluster-size>2</dynamic-cluster-size>
      <max-dynamic-cluster-size>5</max-dynamic-cluster-size>
    </dynamic-servers>
  </cluster>

markxnelson:
ok thanks for the explanation @alai8 - go ahead with merge

PR #1443, opened 2020-02-20
Fix what is broken

rjeberhard:
1. Cherry-pick Marina's pom change to disable the monitoring exporter test that is hanging
2. No longer skip CRD operations based on the pre-auth, health check results, but stick with the pattern of trying the operation and having the operator handle failures
3. Clean-up some of the unnecessary pre-auth, access checks given the reduction in privileges.

Quick test is clean; waiting on full test, but looking good.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Main.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Main.java
@@ -39,7 +39,6 @@ import io.kubernetes.client.openapi.models.V1ServiceList;
 import io.kubernetes.client.openapi.models.V1SubjectRulesReviewStatus;
 import io.kubernetes.client.util.Watch;
 import oracle.kubernetes.operator.calls.CallResponse;
-import oracle.kubernetes.operator.helpers.AuthorizationProxy;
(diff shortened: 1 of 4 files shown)


operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -616,13 +613,18 @@ protected Step action(String ns) {
...
+
     NamespaceRulesReviewStep(String ns) {
       this.ns = ns;
     }
 
     @Override
     public NextAction apply(Packet packet) {
-      NamespaceStatus nss = namespaceStatuses.computeIfAbsent(ns, (key) -> new NamespaceStatus());
+      NamespaceStatus nss = namespaceStatuses.computeIfAbsent(
+          ns != null ? ns : operatorNamespace, (key) -> new NamespaceStatus());
```
markxnelson: i feel this syntax could lead to maintenance issues as people wonder what exactly your intent was. 
not a big deal, but i would prefer to either add a comment, or just make it a little more explicit

rjeberhard: I'll add a comment

PR #1445, opened 2020-02-21
Domain secret MD5 checksum.

tbarnes-us:
This is a check to make sure that the domain secret found by the introspector job matches the domain secret found by WebLogic pods.

Ad hoc testing passed - including manually forcing MD5 failures.

Integration sanity tests pass.


markxnelson:
I think I am missing where the original md5 file comes from?

rjeberhard:
@markxnelson introspectDomain.sh line 127

markxnelson:
that puts in in `/tmp/DomainSecret.md5` but later, it is looking for it in `/weblogic-operator/introspector/DomainSecret.md5`

rjeberhard:
@tbarnes-us, let's walk this through end-to-end, but the later path, Mark, is where the contents of the cm generated during introspection are mounted on a regular server-instance pod.

markxnelson:
if you say its ok, then i'll buy that :)

tbarnes-us:
@markxnelson 
introspectDomain.sh calculates original  MD5 and puts it in /tmp

PR #1456, opened 2020-03-02
Boot identity not valid FAQ

rjeberhard:


docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+draft: false
+weight: 1
+---
+
+> One or more WebLogic Server instances in my domain will not start and I see errors in the server log like this:
+>
+> `<Feb 6, 2020 12:05:35,550 AM GMT> <Critical> <Security> <BEA-090402> <Authentication denied: Boot identity not valid. The user name or password or both from the boot identity file (boot.properties) is not valid. The boot identity may have been changed since the boot identity file was created. Please edit and update the boot identity file with the proper values of username and password. The first time the updatedboot identity file is used to start the server, these new values are encrypted.>`
+
+When you see these kinds of errors, it means that either the user name and password provided in the `weblogicCredentialsSecret` are invalid
+or that the WebLogic domain directory's security configuration files have changed in an incompatible way between when the Operator scanned
```
rosemarymarano: Operator -> operator (we use lower case except when referring to its full official title)

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+
+This can happen in a variety of ways, depending on the [model selected](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/choosing-a-model/):
+
+#### Domain in a Docker image and rolling to an image containing new or unrelated domain directory
+
+The domain is using the "domain in a Docker image" model and the error occurs while rolling to a new Docker image that contains an entirely new or unrelated domain directory.
+
+The problem is that WebLogic cannot support server instances being part of the same WebLogic domain if the server instances do 
+not all share the same domain-specific encryption key and other bootstrapping security details. Additionally, operator introspection
+currently only happens when starting servers following a total shutdown. Therefore, the "boot.properites" generated from
```
rosemarymarano: currently only happens -> currently happens only (Keep _only_ close to that which it modifies to improve the clarity and accuracy of the sentence.)

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+introspecting the image containing the original domain directory will be invalid when used with a container started with
+the updated Docker image containing the new or unrelated domain directory.
+
+The solution is to either follow the recommended [CI/CD guidelines](https://oracle.github.io/weblogic-kubernetes-operator/userguide/cicd/) so that the original and new Docker images contain domain directories
```
rosemarymarano: is to either follow -> is to follow either

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+
+The simplest solution is to set `imagePullPolicy` to `Always`; however, the better solution would be to design your development
+pipeline to generate new Docker image tags on every build and to never reuse an existing tag.
+
+#### Domain on a persistent volume and completely replacing the domain directory
+
+The domain is using the "domain on a persistent volume" model and the error occurs while starting servers when the domain directory
+change was made while other servers were still running.
+
+If completely replacing the domain directory then you must [stop all running servers](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/domain-lifecycle/startup/#starting-and-stopping-servers).
```
rosemarymarano: directory then -> directory, then

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+
+Since all servers will already be stopped, there is no requirement that the new contents of the domain directory be related to
```
rosemarymarano: Since -> Because (since – don't use to mean for the reason of; use because. Use since only for matters of time, for example, since the last time the program was run.)

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+secured value can be injected as the "boot.properties" contents when starting server instances.
+
+When the domain directory is changed such that the domain-specific encryption key is different, the "boot.properties" entries
+generated during introspection will now be invalid.
+
+This can happen in a variety of ways, depending on the [model selected](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/choosing-a-model/):
+
+#### Domain in a Docker image and rolling to an image containing new or unrelated domain directory
+
+The domain is using the "domain in a Docker image" model and the error occurs while rolling to a new Docker image that contains an entirely new or unrelated domain directory.
```
rosemarymarano: "while rolling to a new Docker image" I would not have understood this phrase, perhaps, developers will. 
Are referring to using a rolling restart with a new Docker image?

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "Boot Identity Not Valid"
+date: 2020-03-02T08:08:19-04:00
+draft: false
+weight: 1
```
tbarnes-us: Minor complaint:  There are already two other FAQ entries at weight 1. It'd be nice if the ordering of FAQ entries was more predictable.

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+---
+
+> One or more WebLogic Server instances in my domain will not start and I see errors in the server log like this:
+>
+> `<Feb 6, 2020 12:05:35,550 AM GMT> <Critical> <Security> <BEA-090402> <Authentication denied: Boot identity not valid. The user name or password or both from the boot identity file (boot.properties) is not valid. The boot identity may have been changed since the boot identity file was created. Please edit and update the boot identity file with the proper values of username and password. The first time the updatedboot identity file is used to start the server, these new values are encrypted.>`
+
```
tbarnes-us: Theoretically, one won't see the above error if it came about because the domain secret changed.   If the domain secret changed, a different error is generated first and  the operator won't even attempt to start WLS.  The other error is: " trace SEVERE "Domain secret mismatch. The domain secret in 'DOMAIN_HOME/security/SerializedSystemIni.dat' where DOMAIN_HOME='$DOMAIN_HOME' does not match the domain secret found by the introspector job. WebLogic requires that all WebLogic servers in the same domain share the same domain secret. See 'Boot Identity Not Valid' in the operator FAQ (https://oracle.github.io/weblogic-kubernetes-operator/faq/boot-identity-not-valid/). MD5 checksum diff:"
"...

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+The problem is that WebLogic cannot support server instances being part of the same WebLogic domain if the server instances do 
+not all share the same domain-specific encryption key and other bootstrapping security details. Additionally, operator introspection
+currently only happens when starting servers following a total shutdown. Therefore, the "boot.properites" generated from
+introspecting the image containing the original domain directory will be invalid when used with a container started with
+the updated Docker image containing the new or unrelated domain directory.
+
+The solution is to either follow the recommended [CI/CD guidelines](https://oracle.github.io/weblogic-kubernetes-operator/userguide/cicd/) so that the original and new Docker images contain domain directories
+with consistent domain-specific encryption keys and bootstrapping security details or to [perform a total shutdown](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/domain-lifecycle/startup/#starting-and-stopping-servers) of the domain so
```
tbarnes-us: and a "details or to" --> "details, or to".  (Makes it easier to understand what you meant by 'either' earlier in the sentence).

docs-source/content/faq/boot-identity-not-valid.md
```diff
@@ -0,0 +1,71 @@
...
+that introspection reoccurs as servers are restarted.
+
+#### Domain in a Docker image after a full domain shutdown and restart
+
+The domain is using the "domain in a Docker image" model and the error occurs while starting servers after a full domain shutdown. 
+
+For this type of failure, it is even more probable that the user name and password specified in the Kubernertes secret referenced by
```
tbarnes-us: You have two causes and one solution here all kind of jammed together.  Make it easier on the reader by making a bullet list of two bullets, each with one cause and its associated solution. Perhaps:
-------
The domain is using the "domain in a Docker image" model and the error occurs while starting servers after a full domain shutdown. For this type of failure, there are two potential causes:

* It is most probable that the user name and password specified in the Kubernertes secret referenced by `weblogicCredentialsSecret` are wrong. 

* If your development model generates new Docker images with new and unrelated domain directories and then tags those images with the same tag, then different Kubernetes worker nodes may have different images under the same tag in their local Docker repositories. The simplest solution is to set `imagePullPolicy` to `Always`; however, the better solution would be to design your development pipeline to generate new Docker image tags on every build and to never reuse an existing tag.

docs-source/content/faq/boot-identity-not-valid.md
TheFrogPad: Doesn't the fact that the server startup detecting the MD5 checksum has changed now change some of the FAQ content?

In other words, if you see the MD5 checksum error that means do X but if you still see Authentication Denied then do Y?

docs-source/content/faq/boot-identity-not-valid.md
TheFrogPad: Along these same lines for easier read and reference, what about having sections and sub section similar to:

* Domain In Image
  - rolling to an image
    > A
  - a full domain shutdown and restart
    > A
    > B
* Domain on a persistent volume
  - replacing the domain directory
    > A

TheFrogPad:
Ryan, did you want to list the case where the managed server does not boot with the authentication or authorization error but MD5 is good? This case is likely because the managed server cannot reach the admin server and the manged server enters MSI mode but without any security data.

docs-source/content/faq/boot-identity-not-valid.md
rjeberhard: Reformated to two different FAQ entries.

rjeberhard:
@TheFrogPad @tbarnes-us @rosemarymarano, I've refactored this into two different FAQ's to capture the different kinds of failures now that we have the validation.  The earlier review comments have been resolved in the moved content.  Take a look, please?

TheFrogPad:
I realized you merged already (sorry to much context switching for me) but this looks good...

PR #1471, opened 2020-03-16
Use timestamp and date in image name for domain in image tests

vanajamukkara:
Use timestamp and date in image name for domain in image tests
delete unused properties file

Full test suite - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1118/ - failures are existing intermittent failures


rjeberhard:
I don't know of any code that deletes images from machines local Docker repositories; however, we do have a global retention policy on OCIR that deletes images not pulled in the last 45 days.

vanajamukkara:
Domain in image tests delete the domain images from the local repository at the end of the test. We don't have any script or cron job to delete the images from the remote k8s cluster nodes.

PR #1479, opened 2020-03-18
Update fluentd example

markxnelson:
This PR updates the fluentd sample that we have in the docs.  The current sample passes through the `timestamp` as a `string` which means you cannot do time-based queries like "before x" or "between x and y" or "in the last 15 minutes".

The PR updates the sample so that the special `@timestamp` field will be added, with type `time` and with the same value as the `string timestamp`.  This will allow users to do time based queries.

### DIFF
--- a/docs-source/content/samples/simple/elastic-stack/weblogic-domain/_index.md
+++ b/docs-source/content/samples/simple/elastic-stack/weblogic-domain/_index.md
@@ -138,7 +138,7 @@ data:
       key_name timestamp 
       types timestamp:time
       # inject the @timestamp special field (as type time) into the record
-      # so you will can do time based queries.
+      # so you will be able to do time based queries.
       # not to be confused with timestamp which is of type string!!!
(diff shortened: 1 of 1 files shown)


docs-source/content/samples/simple/elastic-stack/weblogic-domain/_index.md
```diff
@@ -131,6 +135,12 @@ data:
       index_name "#{ENV['DOMAIN_UID']}"
       scheme https
       ssl_version TLSv1_2
+      key_name timestamp 
+      types timestamp:time
+      # inject the @timestamp special field (as type time) into the record
+      # so you will can do time based queries.
```
rosemarymarano: so you will can -> so you can

markxnelson: fixed

hzhao-github:
Looks good to me

PR #1484, opened 2020-03-20
OWLS-78058 Add logs/status for domain refs to non-existant WL clusters & servers

alai8:
1) If a domain resource references a non-existent WL cluster or server, then (a) log a warning in the operator, and (b) update the domain status with a similar warning.

2) Added "Replicas Goal" to cluster status which is the replicas specified in the domain spec for the cluster

3) Added "desired state" to server status which is the server state specified in the domain spec for the server

4) Cleaned up some unused code in wlsconfig package

Example warning messages:

{"timestamp":"03-20-2020T22:49:17.183+0000","thread":23,"fiber":"engine-operator-thread-5-fiber-5","domainUID":"domain1","level":"WARNING","class":"oracle.kubernetes.operator.helpers.DomainValidationSteps$ValidateDomainTopologyStep","method":"logAndAddWarning","timeInMillis":1584744557183,"message":"Managed Server 'ghost-server' specified in the domain resource does not exist in the WebLogic domain","exception":"","code":"","headers":{},"body":""}

Example domain status:

Status:
  Clusters:
    Cluster Name:      cluster-1
    Maximum Replicas:  3
    Ready Replicas:    1
    Replicas:          1
    Replicas Goal:     1    <--- new
  Conditions:
    Reason:  ServersReady
    Status:  True
    Type:    Available
  Message:   Cluster 'ghost-cluster' specified in the domain resource does not exist in the WebLogic domain
Managed Server 'ghost-server' specified in the domain resource does not exist in the WebLogic domain     <-- new
  Replicas:  1
  Servers:
    Cluster Name:   cluster-1
    Desired State:  RUNNING
    Server Name:    managed-server2
    State:          SHUTDOWN
    Cluster Name:   cluster-1
    Desired State:  RUNNING         <--- new


lennyphan:
Should 'docs/charts/weblogic-operator-2.6.0.tgz ' be updated as develop is not version 3.0.0?  e.g. docs/charts/weblogic-operator-3.0.0.tgz

rjeberhard:
@lennyphan, my plan is that develop will be 3.0.0.

operator/src/main/resources/Operator.properties
```diff
@@ -182,6 +182,9 @@ WLSDO-0003=No volume mount contains path for log home ''{0}''
 WLSDO-0004=The mount path ''{0}'', in entry ''{1}'' of domain resource ''additionalVolumeMounts'', is not valid
 WLSDO-0005=Environment {0} {1}, specified under ''{2}'', {3} reserved for use by the operator
 WLSDO-0006=Bad namespace for weblogicCredentialsSecret: ''{0}''. If specified, must be same as domain namespace.
+WLSDO-0007=Cluster ''{0}'' specified in the domain resource does not exist in the WebLogic domain
+WLSDO-0008=Managed Server ''{0}'' specified in the domain resource does not exist in the WebLogic domain
```
tbarnes-us: maybe change "WebLogic domain" --> "WebLogic domain home configuration" in both logs to help disambiguate a little

tbarnes-us:
General comment: Maybe fix the following in this JIRA, or do it another JIRA:IIRC the status report is in a semi-random order.  It should be admin-server first, then grouped by cluster, then alpha-num within cluster.  The numeric portion of alpha-num requires finding and 0 filling (padding) numbers within names for sorting purposes so my2 sorts before my12 (eg. my2x111f --> my000002x000111f for sorting purposes).   The same problem occurs with rolling upgrade ordering, but there's already a JIRA for you.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
```diff
@@ -38,6 +38,11 @@
   @Range(minimum = 0)
   private Integer maximumReplicas;
 
+  /** The requested number of cluster members. Required. */
+  @Description("The requested number of cluster members. Required.")
```
rjeberhard: Let's discuss the use of "Required" here.  I think it's okay because we are going to version 3.0.0, so there can be non-backward compatible changes. However, just having "Required" in the description doesn't make the field required.  If we want the field to be required in the schema, then tag the field with @NonNull.

tbarnes-us: I don't think it should be required - seems too prescriptive. I suspect what Lenny really meant was that the default is 0, which surprises some folks.  How about something like "This value must be set above 0 (the default) to ensure cluster members start."?   Not sure if that's good wording, but maybe something like that?

tbarnes-us:
Overall this LGTM, apart from the 'bonus' work I suggested and apart from Ryan's comment on the 'Required' reading.

operator/src/main/resources/Operator.properties
alai8: Updated to:
WLSDO-0007=Cluster ''{0}'' specified in the domain resource does not exist in the WebLogic domain home configuration.
WLSDO-0008=Managed Server ''{0}'' specified in the domain resource does not exist in the WebLogic domain home configuration.

alai8:
Regarding the bonus work. I will work on it as a different JIRA to minimize potential for merge conflicts since Lenny is working on a JIRA that touches some of the same source files.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
alai8: Updated replicasGoal to:

  /** The requested number of cluster members from the domain spec. */
  @Description("The requested number of cluster members from the domain spec. "
      + "Cluster members will be started by the operator if this value is larger than zero.")
  @Range(minimum = 0)
  private Integer replicasGoal;

alai8:
About docs/charts/weblogic-operator-2.6.0.tgz. The branch probably started before the version was changed to 3.0. Not sure. But this file has been removed from this branch.

PR #1485, opened 2020-03-21
OWLS-80458  New Upgrade Usecase 

anpanigr:
Here are highlights of changes  addressed on this PR 

(a) Added new Upgrade usecas from Release 2.5.0 
(b) Removed the old upgrade Usecases  form Releases ( 2.0, 2.0.1, 2.1, 2.2.0, 2.2.1, 2.3.0)
      Above older release does not support Helm v3+. Those release only  support Helm v2+ 
      However Opeartor Release v 3.0 will ONLY support v3, 
      So only supported upgrade Usecase is from Op 2.5.0 to any Release above ( e.g. 3.0 ) 
(c) Remove apiVersion of Domain Object Checking after Upgrade  
      It takes 7-8 min to update the apiVersion.  Instead added check for Domain CRD version
      upgrade to the latest release.   
(d)  Added method to save the logs for K8s resources after test is complete.

Helm 2.0 run  (internal jenkin)  
http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/4476/

Helm 3.0  (internal jenkin)  
http://wls-jenkins.us.oracle.com/view/weblogic-operator/job/weblogic-kubernetes-operator-javatest/4479/

Test tun on External Jenkins 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1199/


anpanigr:
> I don't think this is consistent with what we discussed

Here is the detailed steps for the usecase 

 a.  Start the Domain in Older Release  ( using Quick Start Guide ) 
 b.  Update the Operator Helm Chart to latest Release 
 c.  Check the Domain CRD version is updated to latest Release 
 d.  Make sure non of the server get re-started
 e.  Scale up/down the Domain with New Operator
 f.  Shutdown the Domain
 g. Remove the domain CRD 

Am I missing any steps.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -322,15 +272,11 @@ private void setupOperatorAndDomain(String operatorGitRelease, String operatorRe
     domNs.add(DOM_NS);
     operatorMap.put("domainNamespaces", domNs);
     operator = TestUtils.createOperator(operatorMap, Operator.RestCertType.LEGACY);
-    // TestUtils.exec("kubectl get all --all-namespaces", true);
-
-    // Map<String, Object> wlstDomainMap = TestUtils.loadYaml(DOMAININIMAGE_WLST_YAML);
-    Map<String, Object> wlstDomainMap = createDomainInImageMap(getNewSuffixCount(), false, testClassName);
+    Map<String, Object> wlstDomainMap = createDomainMap(getNewSuffixCount(),testClassName);
```
ddsharpe: Why are we switching from Domain in Image to Domain in PV?  If we need to test Domain in PV here, we should ADD it (or add a copy of this test for PV), not remove Domain in Image.

anpanigr: The purpose of the Upgrade Usecase is to verify the ability of the Upgraded Operator to manage the existing Weblogic Domain. Will the upgrade process takes a different code path based on the domain image is generated.  

Another Reason :   I have modified the domain-in-image sample script  ( in develop) to skip the domain image creation if the image exits to reduce the test execution time.  But this change is not available in master ( 2.5.0 branch), so I can not use this  feature while creating domain in older release to start the Domain

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -293,6 +229,20 @@ private void checkDomainRollingRestarted() throws Exception {
...
+   * not restarted due to Operator Upgrade
+   *
+   * @throws Exception If restarted
+   */
+  private void checkDomainNotRestarted() throws Exception {
+    TestUtils.checkPodReady(DUID + "-" + domain.getAdminServerName(), DOM_NS);
+    for (int i = 2; i >= 1; i--) {
+      LoggerHelper.getLocal().log(Level.INFO,
+          "Checking if managed server pod(" + DUID + "--managed-server" + i + ") is RUNNING");
+      TestUtils.checkPodReady(DUID + "-managed-server" + i, DOM_NS);
```
markxnelson: the comment says we want to check they have not been restarted, but this just checks if they are running, not if they wer restarted.  is the comment wrong? or does the code need to be improved to check for restart?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
+   *
+   * @throws Exception when version does not match
+   */
+  private void checkCrdVersion() throws Exception {
+    boolean result = false;
+    LoggerHelper.getLocal().log(
+        Level.INFO,
+        "Checking for the CRD Version "
+            + getDomainApiVersion()
+            + " in a loop for up to 15 minutes");
```
markxnelson: didn't we discuss this and realize that the 15 minute wait was only needed for a domain to be updated?  and that we only need to check the CRD apiVersion is udpated, not a CR (instance)?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
+      if (exec.stdout().contains(getCrdVersion())) {
+        LoggerHelper.getLocal().log(Level.INFO, "Got expected CRD Version");
         result = true;
         break;
       }
       Thread.sleep(1000 * 10);
     }
     if (!result) {
-      throw new RuntimeException("FAILURE: Didn't get the expected operator version");
+      throw new RuntimeException("FAILURE: Didn't get expected CRD Version");
```
markxnelson: does this need to be a RuntimeException? not just a regular Exception?

integration-tests/src/test/resources/OperatorIT.properties
```diff
@@ -16,6 +16,7 @@ oracledbImageTag = 12.2.0.1-slim
 oracledbImageName = container-registry.oracle.com/database/enterprise
 OCR_SERVER=container-registry.oracle.com
 DOMAIN_API_VERSION = weblogic.oracle/v7
+CRD_VERSION = v7
```
vanajamukkara: can you get the version from develop crd instead of adding a property here, we will have to change this value for every release.

anpanigr: Unless I start the domain in develop branch, is it possible to get the domain CRD offline.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
+   * In Jenkins it takes nearly 8 minutes to show the updated value 
+   * 
+   * The .apiVersion of the domain will not be updated until the domain is 
+   * written again (e.g. by the operator when updating status)
+   * This contributes to long time period. We can skip this check 
    *
    * @throws Exception when version does not match
    */
-  private void checkOperatorVersion() throws Exception {
+  private void checkDomainApiVersion() throws Exception {
```
vanajamukkara: remove this method as its not used

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
+   * @throws Exception when version does not match
+   */
+  private void checkCrdVersion() throws Exception {
+    boolean result = false;
+    LoggerHelper.getLocal().log(
+        Level.INFO,
+        "Checking for the CRD Version "
+            + getDomainApiVersion()
+            + " in a loop for up to 15 minutes");
+    for (int i = 0; i < 900; i = i + 10) {
```
vanajamukkara: use BaseTest.getMaxIterationsPod() and BaseTest.getWaitTimePod() - these are controlled from properties file or env var. we use these across the tests if we check for something in loop.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -293,6 +229,20 @@ private void checkDomainRollingRestarted() throws Exception {
...
 
+  /**
+   * Check whether the weblogic server instances are still RUNNING 
+   * not restarted due to Operator Upgrade
+   *
+   * @throws Exception If restarted
+   */
+  private void checkDomainNotRestarted() throws Exception {
+    TestUtils.checkPodReady(DUID + "-" + domain.getAdminServerName(), DOM_NS);
+    for (int i = 2; i >= 1; i--) {
```
vanajamukkara: if rolling restart is triggered, do u see it immediately or will there be little lag? checking once immediately is enough to make sure rolling restart is not triggered?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: @rjeberhard ^^ ?

markxnelson: oh i see you removed the call on 148 to this method -- so can this method be deleted now? not needed?

integration-tests/src/test/resources/OperatorIT.properties
rjeberhard: I don't understand the question, Pani.  Vanaja, I'm okay with having a separate test value here so that the test code doesn't depend on symbols from the production code.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -293,6 +229,20 @@ private void checkDomainRollingRestarted() throws Exception {
     }
   }
 
+  /**
+   * Check whether the weblogic server instances are still RUNNING
```
markxnelson: none of the javadoc in this file is javadoc compliant.  can you please run javadoc and fix the errors?

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh
```diff
@@ -34,6 +34,7 @@ function usage {
   echo "  -v Validate the existence of persistentVolumeClaim, optional."
   echo "  -k Keep what has been previously from cloned https://github.com/oracle/docker-images.git, optional. "
   echo "     If not specified, this script will always remove existing project directory and clone again."
+  echo "  -s Skip the domain image build optional. "
```
markxnelson: fix this sentence

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh
```diff
@@ -171,8 +175,13 @@ function getDockerSample {
 
 #
 # Function to build docker image and create WebLogic domain home
-#
+# Image build is skipped when -s option is specified OR image is not available
+# e.g. If -s option is specified script will skip the image build only when 
+# image is available else build the image
+# If -s option is NOT specified script will ALWYAS build the image
```
markxnelson: typo

integration-tests/src/test/resources/OperatorIT.properties
vanajamukkara: Ryan, still the value here is whatever is used in production code right.

rjeberhard: Yes, that's right.  They would need to be kept in sync.  Here's another option.  We make another module in maven with shared symbols like this that the operator runtime and integration tests could share.

vanajamukkara: That's a good idea.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
rjeberhard: Combining here since @markxnelson made the same comment below... Checking that the operator doesn't restart pods is a bit tricky.  The operator will respond very quickly to check all existing domains after it starts, so if the operator hasn't started the deletion of any pods within a few seconds then it isn't rolling the servers.  You can check if pods are in the process of being deleted by checking the .metadata.deletionTimestamp.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,145 +104,33 @@ public void cleanupOperatorAndDomain() throws Exception {
         operator.destroy();
       }
       TestUtils.exec("rm -rf " + Paths.get(opUpgradeTmpDir).toString());
+      TestUtils.exec("kubectl delete crd domains.weblogic.oracle --ignore-not-found");
```
markxnelson: should check this actually worked by trying to get the crd again

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
    * @param restart boolean parameter used to determine if a restart of domain is checked
```
markxnelson: there is not restart param in the method signature

integration-tests/src/test/resources/OperatorIT.properties
rjeberhard: Okay, I can make that happen but not in the PR.

vanajamukkara: Ok. should we leave it hard coded for now or get the value from develop crd?

rjeberhard: Let's leave it hardcoded now please because I would strongly like to not have integration-test depend on the operator module.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Keeping it if we can decrease the time by revisiting the product code in future.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Done

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
vanajamukkara: @rjeberhard confirmed we don't need to verify domain resource apiVersion, checking crd apiVersion is enough. Ryan, can you please confirm again.

integration-tests/src/test/resources/OperatorIT.properties
vanajamukkara: Will just starting the operator in develop branch create crd domains.weblogic.oracle?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
rjeberhard: Yes, correct.  We do not need to confirm that the operator updates the .apiVersion of a Domain resource.  I'm just interested in the CRD version and then the checks that the pods don't roll and that the domain remains manageable.

integration-tests/src/test/resources/OperatorIT.properties
anpanigr: We will keep the value hard-coded till we get a shared module for crd version on the github repository

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh
anpanigr: I replicate the sentences from other existing parameters  such as 
echo "  -e Also create the resources in the generated YAML files, optional."
echo "  -v Validate the existence of persistentVolumeClaim, optional."

Any Suggestion like this 
   echo "  -s Skip the domain image build  (optional)  "

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
vanajamukkara: @anpanigr you can remove the method

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Does the style check plugin does not check for javadoc compatibility. When I run "javaoc  ItOperatorUpgrade.java" It only complains about the missing package 
--
ItOperatorUpgrade.java:19: error: package org.junit.jupiter.api does not exist
import org.junit.jupiter.api.AfterEach;
---

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh
markxnelson: comma or parens are ok - but the way it is now is not correct

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Modified as follows  ..
  throw new Exception("FAILURE: Didn't get expected CRD Version");

kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh
anpanigr: I modified the sentence as follows with an additional comma 
echo "  -s Skip the domain image build, optional. "

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Here should I expect a NULL value for {.metadata.deletionTimestamp}
$ kubectl get po/operatordomain250-managed-server1 -n weblogic-domain250 -o  jsonpath={.metadata.deletionTimestamp}

rjeberhard: Yes, that's right.  Oh, and for completeness you should check that the .metadata.creationTimesamp.  I don't think it's an or...

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: i get this error: 
```
/home/mark/jdk-11.0.6/bin/javadoc -private -splitindex -d /home/mark/operator/weblogic-kubernetes-operator/integration-tests/target/site @/tmp/javadoc_args
Loading source file /home/mark/operator/weblogic-kubernetes-operator/integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java...
Constructing Javadoc information...
Standard Doclet version 11.0.6
Building tree for all the packages and classes...
Generating /home/mark/operator/weblogic-kubernetes-operator/integration-tests/target/site/oracle/kubernetes/operator/ItOperatorUpgrade.html...
/home/mark/operator/weblogic-kubernetes-operator/integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java:141: error: @param name not found
   * @param restart boolean parameter used to determine if a restart of domain is checked
            ^
```

the ide shows it too:
Screenshot from 2020-03-23 11-44-31

markxnelson: Screenshot from 2020-03-24 12-38-35

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: well that's great that you changed it, but the question was does it **need** to be a RE?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: removed the method to check DomainApi version

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: I added the logic to check the creationTimesamp before and after upgrade to make sure that the server pod has not been re-stared due to upgrade.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: I added the logic to check the creationTimesamp before and after upgrade to make sure that the server pod has not been re-stared due to upgrade.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Junit fails when the method throws any Exception, it does not matter it is a RE or simple Exception.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: This is the cleanup section, when we delete the CRD from default NameSpace after the test is completed, so that it will not interfere with other test

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: @ddsharpe does this resolve your concern?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
-        LoggerHelper.getLocal().log(Level.INFO, "Got the expected apiVersion");
+        int ttime = 10 * (i - 1); 
+        LoggerHelper.getLocal().log(Level.INFO, "Got the expected API Version after [" + ttime + "] Seconds");
+        result = true;
+        break;
+      }
+      Thread.sleep(1000 * 10);
+    }
+    if (!result) {
+      throw new RuntimeException("FAILURE: Didn't get the expected API Version");
```
markxnelson: RuntimeException again?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -238,41 +141,74 @@ public void testOperatorUpgradeFrom2_3_0() throws Exception {
...
           TestUtils.exec(
               "kubectl get domain -n " + DOM_NS + "  " + DUID + " -o jsonpath={.apiVersion}", true);
       if (exec.stdout().contains(getDomainApiVersion())) {
-        LoggerHelper.getLocal().log(Level.INFO, "Got the expected apiVersion");
+        int ttime = 10 * (i - 1); 
+        LoggerHelper.getLocal().log(Level.INFO, "Got the expected API Version after [" + ttime + "] Seconds");
+        result = true;
+        break;
+      }
+      Thread.sleep(1000 * 10);
```
markxnelson: i could be wrong, but i think a `Thread.sleep()` needs to be in a `try { .. } catch (InterruptedException) {} `
@rjeberhard ?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: CRD's are not namespaced, they are at the cluster scope.  So deleting a CRD by definition would almost certainly interfere with other tests.
The original question stands though - you are no checking that it was deleted.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,190 +106,82 @@ public void cleanupOperatorAndDomain() throws Exception {
...
    */
-  private void upgradeOperator(boolean restart) throws Exception {
+  private void upgradeOperator() throws Exception {
     operator.callHelmUpgrade("image=" + OP_TARGET_RELEASE);
-    if (restart) {
-      checkDomainRollingRestarted();
-    }
-    checkOperatorVersion();
+    checkCrdVersion();
+    System.out.println("Before Upgrade M1CTS[" + M1_CTS + "]");
```
markxnelson: log? not sop?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,190 +106,82 @@ public void cleanupOperatorAndDomain() throws Exception {
...
-      if (exec.stdout().contains(getDomainApiVersion())) {
-        LoggerHelper.getLocal().log(Level.INFO, "Got the expected apiVersion");
+              "kubectl get crd domains.weblogic.oracle -o jsonpath='{.spec.versions[?(@.storage==true)].name}'", true);
+      if (exec.stdout().contains(getCrdVersion())) {
+        LoggerHelper.getLocal().log(Level.INFO, "Got expected CRD Version");
         result = true;
         break;
       }
-      Thread.sleep(1000 * 10);
+      Thread.sleep(BaseTest.getWaitTimePod() * 1000);
```
markxnelson: enclose in try? @rjeberhard ?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: Please spell "WebLogic" right too

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -293,6 +202,29 @@ private void checkDomainRollingRestarted() throws Exception {
...
+   */
+  private void checkDomainNotRestarted() throws Exception {
+    TestUtils.checkPodReady(DUID + "-" + domain.getAdminServerName(), DOM_NS);
+    for (int i = 2; i >= 1; i--) {
+     TestUtils.checkPodReady(DUID + "-managed-server" + i, DOM_NS);
+    }
+    String m1 = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-managed-server1");
+    String m2 = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-managed-server2");
+    String admin = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-admin-server");
+    System.out.println("After Upgrade M1CTS[" + m1 + "]");
```
markxnelson: log not sop?

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -1980,4 +1980,36 @@ public static ExecResult kubectlpatch(String domainUid, String domainNS, String
     return exec(cmd, true);
   }
 
+  /**
```
markxnelson: missing javadoc

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -1980,4 +1980,36 @@ public static ExecResult kubectlpatch(String domainUid, String domainNS, String
...
+   * @return creationTimestamp of the Pod 
+   * @throws Exception on failure
+   */
+  public static String getCreationTimeStamp(String namespace, String pod) throws Exception {
+    String kcmd = "kubectl get pod "
+              + pod 
+              + " --namespace "
+              + namespace
+              + " -o jsonpath='{.metadata.creationTimestamp}'";
+    System.out.println("KUBECTL CMD " + kcmd);
```
markxnelson: log not sop

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -1980,4 +1980,36 @@ public static ExecResult kubectlpatch(String domainUid, String domainNS, String
...
+    ExecResult result = ExecCommand.exec(kcmd);
+    return result.stdout().trim();
+  }
+
+  /**
+   * @param namesapce
```
markxnelson: missing javadoc.
typo in param name

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: what's an M1CTS?  think of the person reading this log message, who is probably not you. how do they know what this means?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: same comment as below - what does M1CTS mean to the person who has to read this log message?

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,190 +106,82 @@ public void cleanupOperatorAndDomain() throws Exception {
...
+    OP_NS = "weblogic-operator250";
+    DOM_NS = "weblogic-domain250";
+    namespaceList.append(OP_NS);
+    namespaceList.append(" ").append(DOM_NS);
+    OP_DEP_NAME = "operator-upgrade250";
+    OP_SA = "operator-sa250";
+    DUID = "operatordomain250";
+    setupOperatorAndDomain("release/2.5.0", "2.5.0");
+
+    M1_CTS = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-managed-server1");
```
markxnelson: its kinda weird to get these here, and then print them in the upgradeOperator() method

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: $ grep RuntimeException  ItOperatorUpgrade.java 
$ 
May be something wring with Review board

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Modified the Log Message as follows ...
println("After Upgrade CreateTimeStamp for managed-server1 pod [" + m1 + "]");
println ("After Upgrade CreateTimeStamp for managed-server2 pod [" + m2 + "]");
println("After Upgrade CreateTimeStamp for admin-server pod [" + admin + "]");

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Move the information to testOperatorUpgradeFrom2_5_0() with detail info 

// Save the CreateTimeStamp for the server pod(s) to compare with CreateTimeStamp after upgrade to make sure the pod(s) are not re-stated
    Managed1_CreateTimeStamp = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-managed-server1");
    Managed2_CreateTimeStamp = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-managed-server2");
    Admin_CreateTimeStamp = TestUtils.getCreationTimeStamp(DOM_NS,DUID + "-admin-server");
    System.out.println("Before Upgrade CreateTimeStamp for managed-server1 pod [" + Managed1_CreateTimeStamp + "]");
    System.out.println("Before Upgrade CreateTimeStamp for managed-server2 pod [" + Managed2_CreateTimeStamp + "]");
    System.out.println("Before Upgrade CreateTimeStamp for admin-server pod [" + Admin_CreateTimeStamp + "]");

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Move the CreateTimeStamp Info into @Test Method

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
anpanigr: Removed the Debug print statement.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
anpanigr: fixed the typo

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Replaced  WebLogic --> Weblogic

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Upgrade test will not be running in parallel with other Tests, so deleting a CRD will not affect any other tests.  I have added an Assertion to make sure CRD is deleted completely form K8s. 

TestUtils.exec("kubectl delete crd domains.weblogic.oracle --ignore-not-found");
// Make sure domain CRD is deleted form k8s 
ExecResult result = TestUtils.exec("kubectl get crd domains.weblogic.oracle");
Assertions.assertEquals(1, result.exitValue());

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Added InterruptedException to the Thread.sleep
 try {
        Thread.sleep(BaseTest.getWaitTimePod() * 1000);
      } catch (InterruptedException e) {
        System.out.println("Got InterruptedException " + e);
      }

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Added InterruptedException

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
rjeberhard: @anpanigr, the product name has the capital L.  The correct name is "WebLogic"

markxnelson: i cant see the update - still appears to say `weblogic`

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -293,9 +209,31 @@ private void checkDomainRollingRestarted() throws Exception {
...
+    System.out.println("After Upgrade CreateTimeStamp for managed-server2 pod [" + m2 + "]");
+    System.out.println("After Upgrade CreateTimeStamp for admin-server pod [" + admin + "]");
+    Assertions.assertEquals(Managed1_CreateTimeStamp, m1);
+    Assertions.assertEquals(Managed2_CreateTimeStamp, m2);
+    Assertions.assertEquals(Admin_CreateTimeStamp, admin);
+  }
 
   /**
-   * Creates operator based on operatorRelease passed to it and then creates a WebLogic domain
+   * Creates operator based on operatorRelease passed to it and then creates a Weblogic domain
```
markxnelson: WebLogic spelled wrong

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
markxnelson: its still missing.....

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -1980,4 +1980,37 @@ public static ExecResult kubectlpatch(String domainUid, String domainNS, String
...
+              + pod 
+              + " --namespace "
+              + namespace
+              + " -o jsonpath='{.metadata.creationTimestamp}'";
+    ExecResult result = ExecCommand.exec(kcmd);
+    return result.stdout().trim();
+  }
+
+  /**
+   * @param namespace
```
markxnelson: write javadoc

vanajamukkara:
@anpanigr you need to commit your changes to github for others to see the changes/comments you resolved. May be that's why Mark is saying he can't see the changes.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Modified in latest commit

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -49,11 +54,24 @@
...
+    namespaceList = new StringBuffer();
     testClassName = new Object() {
     }.getClass().getEnclosingClass().getSimpleName();
     initialize(APP_PROPS_FILE, testClassName);
   }
 
+  /**
+   * Releases k8s cluster lease, archives result, pv directories.
+   *
+   * @throws Exception exception
```
markxnelson: yeah, i know it is an exception, you are meant to explain why

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,211 +106,110 @@ public void cleanupOperatorAndDomain() throws Exception {
...
-    checkOperatorVersion();
+    checkCrdVersion();
+    checkDomainNotRestarted();
     testClusterScaling(operator, domain, false);
   }
 
   /**
-   * checks the expected version of the upgraded operator in a loop. In Jenkins it takes nearly 8
-   * minutes to show the updated value of the domain CRD.
+   * Checks the expected Upgraded Version of CustomResourceDefintion (CRD)
```
markxnelson: CRD's are not namespaced resoruces, so this doc is wrong.

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -89,211 +106,110 @@ public void cleanupOperatorAndDomain() throws Exception {
...
-      Thread.sleep(1000 * 10);
+      try { 
+        Thread.sleep(BaseTest.getWaitTimePod() * 1000); 
+      } catch (InterruptedException e) {
+        LoggerHelper.getLocal().log(Level.INFO,"Got InterruptedException " + e);
+      } 
     }
     if (!result) {
-      throw new RuntimeException("FAILURE: Didn't get the expected operator version");
+      throw new Exception("FAILURE: Didn't get expected CRD Version");
```
markxnelson: include in the message what you did get, and what you expected to get

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: @ddsharpe

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: This is standard junit lifecycle method being used across all test classes.   What you want me to explain more here.

markxnelson: which is better:
``` @throws Exception exception```
or 
``` @throws Exception if unable to collect diagnostics and clean up resources.```

Just because it is done wrong elsewhere, does not make it right to do it wrong here.

markxnelson: or, you could just say `@throws Exception @see BaseTest#tearDown`

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Remove the line containing default ns

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: Added 
@throws Exception if tearDown() fails
@see BaseTest#tearDown

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
anpanigr: done

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
anpanigr: done

integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
markxnelson: confirmed derek is ok with the response

PR #1488, opened 2020-03-22
Make cleanup script more orderly in its deletes.

tbarnes-us:
Various updates to cleanup script:

- Additional tracing throughout

- Added a 'dry run' mode. This is useful for debugging the script itself. If you pass '-dryrun' on the command line, then the script will proceed without deleting anything - it will just echo potential delete commands - plus it will reduce it's wait-for-pods timeouts (since we know the pods aren't going away).

- Add timestamps to trace statements

- Remove its delete of tiller.

- Remove its direct delete of namespaces.

- Add new deletes before attempting its 'usual' deletes:
  - Delete domains and wait up to 60 seconds for their pods to go away (assumes operators will do this)
  - Delete operator deployments
  - Delete any leftover WLS and introspector pods and wait up to 60 seconds for them to go away

- Modify the 'old' path a bit. Here's the 'old' path in order of occurrence:
  - Uninstall every helm install as before
  - Delete by type-and-label is now updated to delete each type individually in a specific order, instead of all types in one command
  - Note that the final catch-all 'generic' delete step is still random in ordering, but hopefully the new and ordered deletes above will mitigate issues with this...
  - Delete hostpath directory as before

### DIFF
--- a/src/integration-tests/bash/cleanup.sh
+++ b/src/integration-tests/bash/cleanup.sh
@@ -65,6 +65,10 @@
 #             (See optional LEASE_ID env var above.)
 #
 
+function timestamp {
+  echo -n [`date '+%m-%d-%YT%H:%M:%S'`]
+}
+
 SCRIPTPATH="$( cd "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
 PROJECT_ROOT="$SCRIPTPATH/../../.."
 RESULT_ROOT=${RESULT_ROOT:-/scratch/$USER/wl_k8s_test_results}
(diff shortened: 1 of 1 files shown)


src/integration-tests/bash/cleanup.sh
```diff
@@ -20,21 +20,34 @@
 #                   given lease on a failure.
 #
 #   SHARED_CLUSTER  Set this to true if you want cleanup to delete tiller
+#                   TBD tiller delete is disabled
```
rjeberhard: Commenting so that we don't forget about this.  We need to see how else this flag is used.  If it really is all about deleting tiller than it was a strange name :)

markxnelson:
@rjeberhard i have a good run here https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1176/
i would like to merge this now if you are good

PR #1490, opened 2020-03-23
fix helm2 delete charts

vanajamukkara:
fix helm2 chart deletion

### DIFF
--- a/src/integration-tests/bash/cleanup.sh
+++ b/src/integration-tests/bash/cleanup.sh
@@ -524,31 +524,34 @@ if [ -x "$(command -v helm)" ]; then
   echo "Detected Helm Version [$(helm version --short --client)]"
   echo @@ `timestamp` Deleting installed helm charts
   namespaces=`kubectl get ns | grep -v NAME | awk '{ print $1 }'`
-  for ns in $namespaces
-  do 
-   if [ ! "$DRY_RUN" = "true" ]; then
-     (
-     set -x
-     helm list --short --namespace $ns | while read helm_name; do
-       if [ "$HELM_VERSION" == "V2" ]; then
-         helm delete --purge  $helm_name
-       else 
-         helm uninstall $helm_name -n $ns 
-       fi
-     done
-     )
-   else
-     (
-     helm list --short --namespace $ns | while read helm_name; do
-       if [ "$HELM_VERSION" == "V2" ]; then
-         echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
-       else 
-         echo @@ `timestamp` Info: DRYRUN: helm uninstall $helm_name -n $ns 
-       fi
-     done
-     )
-   fi
-  done
+
+  if [ "$HELM_VERSION" == "V2" ]; then
+    helm list --short | while read helm_name; do
+      if [ ! "$DRY_RUN" = "true" ]; then
+        (
+        set -x
+        helm delete --purge  $helm_name
+        )
+      else
+        echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+      fi
+    done
+  else
+    namespaces=`kubectl get ns | grep -v NAME | awk '{ print $1 }'`
+    for ns in $namespaces
+    do
+        helm list --short --namespace $ns | while read helm_name; do
+          if [ ! "$DRY_RUN" = "true" ]; then
+            (
+            set -x
+            helm uninstall $helm_name -n $ns
+            )
+          else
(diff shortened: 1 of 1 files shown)


src/integration-tests/bash/cleanup.sh
```diff
@@ -524,31 +524,34 @@ if [ -x "$(command -v helm)" ]; then
...
+        set -x
+        helm delete --purge  $helm_name
+        )
+      else
+        echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+      fi
+    done
+  else
+    namespaces=`kubectl get ns | grep -v NAME | awk '{ print $1 }'`
+    for ns in $namespaces
```
markxnelson: still don't need this loop....
`helm list --all-namespaces`

vanajamukkara: yes, I haven't fixed helm3 ..may be @anpanigr or I can look into helm3. I have only fixed helm2 for now, external jenkins is using helm2. Have something to work on mii, will get back to helm3 later..

anpanigr: Helm 3 need an explicit namespace to delete helm chart. So we need to loop thru the same space.  If we can find the list of name spaces dynamically created for the testing, we can avoid looping for all namesapces

vanajamukkara: @anpanigr how about do `helm list --all-namespaces` and get the ns and use that for deletion instead of looping through all the existing namespaces..

tbarnes-us: Why is  this such a big worry? The script takes several seconds to run, and this would only shave off milliseconds.

tbarnes-us: (Just curious if I'm misunderstanding.)

markxnelson: yesterday i watched it take over 15 minutes as it ran helm list 483 times

markxnelson: sure it will run in a few seconds if there is nothing to delete!

vanajamukkara: more namespaces more time ..

rjeberhard: You should be able to get the namespace from the table.  If that is hard to parse, `helm list` supports output formats of json and yaml.

tbarnes-us: Oh. That's an impressive namespaces count!

vanajamukkara: yes. helm3 chart deletion needs to be reviewed and fixed. In any case, this PR is just for helm2 fix for now. I have submitted a job on external jenkins with my branch. @markxnelson @rjeberhard you can merge this PR if you are comfortable after seeing the jenkins results.

tbarnes-us:
I realize it's unrelated to this change, but we can also make the 'genericDelete' a little safer in terms of its ordering.   Change:

genericDelete "all,cm,pvc,roles,rolebindings,serviceaccount,secrets,ingress" "crd,pv,ns,clusterroles,clusterrolebindings" "logstash|kibana|elastisearch|weblogic|elk|domain|traefik|voyager|apache-webtier|mysql"

into two calls to genericDelete.  Where:

(1) we generic delete the first time with  "all" "" "logstash|kibana|elastisearch|weblogic|elk|domain|traefik|voyager|apache-webtier|mysql"

(2) we generic delete the second time with the complete list as before...

vanajamukkara:
Quick test ran clean on external jenkins - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1179/consoleFull

src/integration-tests/bash/cleanup.sh
rjeberhard: Why are we spending time on fixing helm2 when we dropped helm2 support for 3.0.0?  For this reason, I'm only interested in fixing helm3 and having our external Jenkins use helm3.

vanajamukkara: @anpanigr changed the integration tests for helm3, but the changes were not tested on external jenkins using helm3. That needs to be done along with the helm3 fix in cleanup.sh. I can close this PR if we don't need helm2 fix.

vanajamukkara:
Closing the PR. New PR will be created after testing helm3 on external jenkins, with helm3 fix in cleanup script.

src/integration-tests/bash/cleanup.sh
markxnelson: ok - well let's make sure the test are in fact only using helm 3 then please :)

PR #1494, opened 2020-03-24
Debug it

bhavaniravichandran:


integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
```diff
@@ -228,6 +243,12 @@ private void createDomainAndDeployApp(Map<String, Object> domainMap, String doma
     domain = null;
     domain = TestUtils.createDomain(domainMap);
     domain.verifyDomainCreated();
+ 
+    // The above command only tests that the pod from the 1st cluster is Running (1/1)
+    // So, checking the MS pod of the other cluster inidvidually here.
```
markxnelson: typo

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
```diff
@@ -238,6 +259,13 @@ private void createDomainAndDeployApp(Map<String, Object> domainMap, String doma
       domainUid + "-new-managed-server2",
     };
     verifyServersStatus(domain, pods, domainUid);
+    Thread.sleep(60 * 1000);
```
markxnelson: does this need to be in a try/catch @rjeberhard ?

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
```diff
@@ -373,4 +405,62 @@ private ExecResult clearCache() throws Exception {
...
+
+  }
+
+  private void verifyServersIngressRunning() throws Exception {
+    LoggerHelper.getLocal().log(Level.INFO, "Verifying server status");
+    domain.verifyServersReady();
+    LoggerHelper.getLocal().log(Level.INFO, "Verifying traefik and ingress created for the domain");
+    LoggerHelper.getLocal().log(Level.INFO,
+        TestUtils.exec("helm status traefik-operator").stdout(), true);
+    LoggerHelper.getLocal().log(Level.INFO,
```
markxnelson: feels like we should be checking the returned values in this block of code, not just printing them out

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
```diff
@@ -25,7 +25,7 @@ PORT=$2
 USER=${3:-weblogic}
 PASSWORD=${4:-welcome1}
 APP_DIR_INPOD=$5
-//APP_NAME=$6 - ignore this for coherence since the dir name and the app name are different.
+#APP_NAME=$6 - ignore this for coherence since the dir name and the app name are different.
 EAR_DEPLOY_TARGET=$7
```
markxnelson: ok, but do you need to rename the others by one number less?? or are you just ignoring that value altogether, not removing it from the call?

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
```diff
@@ -62,24 +62,28 @@ cp ${APP_DIR_INPOD}/builddir/${APP_NAME}.gar ${APP_DIR_INPOD}/.
...
-curl --noproxy '*' --silent  --user ${USER}:${PASSWORD} -H X-Requested-By:MyClient -H Accept:application/json -H Content-Type:multipart/form-data -F "model={ name: '${APP_NAME}GAR', targets: [ { identity: [ clusters, '${GAR_DEPLOY_TARGET}' ] } ] }" -F "sourcePath=@${ARCHIVE_FILE_GAR}" -H "Prefer:respond-async" -X POST http://${HOST}:${PORT}/management/weblogic/latest/edit/appDeployments -o ${APP_DIR_INPOD}/deployGAR.out
+
+echo -e "curl --noproxy '*' -v --user ${USER}:${PASSWORD} -H X-Requested-By:MyClient -H Accept:application/json -H Content-Type:multipart/form-data -F \"model={ name: '${APP_NAME}GAR', targets: [ { identity: [ clusters, '${GAR_DEPLOY_TARGET}' ] } ] }\" -F \"sourcePath=@${ARCHIVE_FILE_GAR}\"  -H \"Prefer:respond-async\" -X POST http://${HOST}:${PORT}/management/weblogic/latest/edit/appDeployments -o ${APP_DIR_INPOD}/deployGAR.out\n"
```
markxnelson: the command printed out and the command being run are not the same - seems it would be better to put the command in a variable, print it and then exec it, to avoid this issue

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
```diff
@@ -62,24 +62,28 @@ cp ${APP_DIR_INPOD}/builddir/${APP_NAME}.gar ${APP_DIR_INPOD}/.
...
+
+curl --noproxy '*' -v --user ${USER}:${PASSWORD} -H X-Requested-By:MyClient -H Accept:application/json -H Content-Type:application/json -d "{name: '${APP_NAME}GAR', sourcePath='${ARCHIVE_FILE_GAR}', targets: [ { identity: [ 'clusters', '${GAR_DEPLOY_TARGET}' ] } ] }" -X POST http://${HOST}:${PORT}/management/weblogic/latest/edit/appDeployments -o ${APP_DIR_INPOD}/deployGAR.out
 sleep 10
-grep -q "STATE_RUNNING" ${APP_DIR_INPOD}/deployGAR.out
+
```
markxnelson: feels like a loop might be needed here instead of only checking once? unless we are sure it cannot possibly take more than 10 sec

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
rjeberhard: There must be a try/catch at a higher level... yes, the method signature throws Exception.  This is okay if on InterruptedException it's correct that the test should fail.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
rjeberhard: Does TestUtils.exec check the return value?  I agree that it's not obvious how the test code handles a failed helm command.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
```diff
@@ -190,7 +190,12 @@ private String getKubernetesNamespaceToUpdate(String domainNamespace) throws Exc
         String trimmed = debracketed.replaceAll("\\s+", "");
         // now have an ArrayList containing "foo", "bar" and "baz"
         ArrayList<String> list = new ArrayList<String>(Arrays.asList(trimmed.split(",")));
-        list.add(domainNamespace);
+        if (!list.contains(domainNamespace)) {
```
rjeberhard: You could just make the list a set.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
vanajamukkara: Is the sleep really needed here? If not remove or reduce? we usually don't put hard sleep's like this in test, instead wait for a condition in a loop with small sleep time.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
```diff
@@ -373,4 +405,62 @@ private ExecResult clearCache() throws Exception {
     return result;
   }
 
+  private void restartCluster(String domainUid) throws Exception {
```
vanajamukkara: can you make this method generic, so that if we have to use from other test classes it can be used? does the other multi cluster test you are debugging needs to restart the cluster as well?

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
vanajamukkara: This method was added while I was debugging the test failure initially, later Bhavani took over the branch. It was not meant to get it merged as is.

bhavaniravichandran: I left it there thinking that more debug is good. Was going to delete this when I edit this file to remove the debug flags after we finally resolve the issue. I will remove it now.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
bhavaniravichandran: restarting the cluster resolves the issue, but not sure it is the right solution. Not calling this method now, but did not want to delete it in case we decide that is what needs to be done.

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
bhavaniravichandran: The script is being called from a utility method. Other tests are also using the same method to call their shell scripts to build and deploy. This argument is not needed for my script, so I just ignore it.

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
bhavaniravichandran: I don't need a sleep here since it is a synchronous call. I will change the other curl command to a synchronous call as well and remove sleep from it as well.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
bhavaniravichandran: This file was not part of my PR. I don't want to make changes to this file without running all the tests as it is a common file used by all the tests - worried about breaking tests. I can maybe take it up as separate work to fix this after discussing with Vanaja.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
bhavaniravichandran: Will remove the sleep. Will not add try/catch. reason - what Ryan said.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
bhavaniravichandran: Already checking the status of all the servers before trying to deploy the app.

integration-tests/src/test/java/oracle/kubernetes/operator/ItManagedCoherence.java
bhavaniravichandran: Deleted the method. Will add it later if need to.

integration-tests/src/test/resources/apps/buildDeployCoherenceAppInPod.sh
```diff
@@ -75,11 +74,11 @@ cres=$?
 
 echo "Deploy ${APP_NAME} ear using curl:"
 
-echo -e "curl --noproxy '*' --silent  --user ${USER}:${PASSWORD} -H X-Requested-By:MyClient -H Accept:application/json -H Content-Type:multipart/form-data -F \"model={ name: '${APP_NAME}', targets: [ { identity: [ clusters, '${EAR_DEPLOY_TARGET}' ] } ] }\" -F \"sourcePath=@${ARCHIVE_FILE_EAR}\" -H \"Prefer:respond-async\" -X POST http://${HOST}:${PORT}/management/weblogic/latest/edit/appDeployments -o ${APP_DIR_INPOD}/deployear.out\n"
+echo -e "curl --noproxy '*' -v --user ${USER}:${PASSWORD} -H X-Requested-By:MyClient -H Accept:application/json -H Content-Type:application/json -d "{name: '${APP_NAME}', sourcePath='${ARCHIVE_FILE_EAR}', targets: [ { identity: [ 'clusters', '${EAR_DEPLOY_TARGET}' ] } ] }" -X POST http://${HOST}:${PORT}/management/weblogic/latest/edit/appDeployments -o ${APP_DIR_INPOD}/deployear.out\n"
```
markxnelson: it is my strong preference that you do not duplicate this string (the command to be run) -- it will just lead to future mistakes or difficulty understanding the intent

bhavaniravichandran: you would like me to remove the echo statements?

markxnelson: no, those are useful, i just want you to do something like this pseudocode:
```
COMMAND="whatever the command is with all its arguments and stuff"
echo "Executing $COMMAND"
$COMMAND
```

bhavaniravichandran: got it. Will fix it.

bhavaniravichandran:
external jenkins run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1198/
There was 1 failure - oracle.kubernetes.operator.ItJrfPvWlst.testJrfDomainOnPvUsingWlst
unrelated to my change and another intermittent test failur

PR #1495, opened 2020-03-25
two automated test case for Wdt Config Update w single Model File jenkins-ignore

hzhao-github:
Local test passed:

oracle.kubernetes.operator.ItModelInImageOverride verifyJdbcOverride JDBC DS info from server pod: MyDataSource-0625-jdbc.xml:    <jndi-name>jdbc/generic1</jndi-name>
...
oracle.kubernetes.operator.ItModelInImageOverride testMiiOverrideNonExistJdbc SUCCESS
...
[INFO] Reactor Summary:
[INFO] 
[INFO] Build Tools 1.0 .................................... SUCCESS [  1.253 s]
[INFO] weblogic-kubernetes-operator 3.0.0 ................. SUCCESS [  2.608 s]
[INFO] operator-swagger 3.0.0 ............................. SUCCESS [  7.289 s]
[INFO] json-schema 3.0.0 .................................. SUCCESS [  9.848 s]
[INFO] jsonschema-maven-plugin Maven Mojo 3.0.0 ........... SUCCESS [  7.919 s]
[INFO] operator-runtime 3.0.0 ............................. SUCCESS [01:01 min]
[INFO] operator-integration-tests 3.0.0 ................... SUCCESS [13:12 min]
[INFO] installation-tests 3.0.0 ........................... SUCCESS [  2.187 s]
[INFO] Project Reports 3.0.0 .............................. SUCCESS [  0.077 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14:45 min
[INFO] Finished at: 2020-03-25T08:48:57-07:00
[INFO] ------------------------------------------------------------------------

kicking off Jenkins jobs now...


hzhao-github:
Are you talking about
public static void createConfigMap(String cmName, String fileOrDirPath,
                                     String namespace, String label) throws Exception {

I saw this new method after synced up with latest version. here is the thing , the new method doesn't delete an existing cm and it will fail re-creating cm. I am not sure I should change yours. So I keep and mine own.

vanajamukkara:
Yes, please make the required changes, we can review them. 
Also take a look at the utility methods in https://github.com/oracle/weblogic-kubernetes-operator/blob/model-in-image/integration-tests/src/test/java/oracle/kubernetes/operator/MiiBaseTest.java and if you can use them.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+  private static String testClassName;
+  private static StringBuffer namespaceList;
+  private static final String configMapSuffix = "-mii-config-map";
+
+  /**
+   * This method gets called only once before any of the test methods are executed. It does the
+   * initialization of the integration test properties defined in OperatorIT.properties and setting
+   * the resultRoot, pvRoot and projectRoot attributes.
+   *
+   * @throws Exception exception
```
markxnelson: explain why an exception might be thrown, we already know it is an exception

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+import oracle.kubernetes.operator.utils.TestUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Simple JUnit test file used for testing Model in Image.
```
markxnelson: no need to say "simple JUnit test file" -- just say what it does

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+  public static void staticPrepare() throws Exception {
+    namespaceList = new StringBuffer();
+    testClassName = new Object() {
+    }.getClass().getEnclosingClass().getSimpleName();
+    // initialize test properties and create the directories
+    initialize(APP_PROPS_FILE, testClassName);
+  }
+
+  /**
+   * This method gets called before every test. It creates the result/pv root directories
```
markxnelson: no need to sat the method is called before every test, we know that already :)

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+   */
+  @BeforeEach
+  public void prepare() throws Exception {
+    createResultAndPvDirs(testClassName);
+
+    // create operator1
+    if (operator == null) {
+      Map<String, Object> operatorMap = createOperatorMap(getNewSuffixCount(),
+          true, testClassName);
+      operator = TestUtils.createOperator(operatorMap, RestCertType.SELF_SIGNED);
```
markxnelson: do we really  need to create operator before each test? seems a little heavy weight

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+
+  /**
+   * Create a domain using model in image and having configmap in the domain.yaml
+   * before deploying the domain. After deploying the domain crd,
+   * re-create the configmap with a model file that define a JDBC DataSource
+   * and update the domain crd to new config map and change domain
+   * restartVersion to reload the model, generate new config and initiate a
+   * rolling restart.
+   *
+   * @throws Exception exception
```
markxnelson: why?

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+    tearDown(new Object() {
+    }.getClass().getEnclosingClass().getSimpleName(), namespaceList.toString());
+
+    LoggerHelper.getLocal().info("SUCCESS");
+  }
+
+  /**
+   * Create a domain using model in image and having configmap in the domain.yaml
+   * before deploying the domain. After deploying the domain crd,
+   * re-create the configmap with a model file that define a JDBC DataSource
```
markxnelson: recreate? or update?

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+  public void testMiiOverrideNonExistJdbc() throws Exception {
+    Assumptions.assumeTrue(QUICKTEST);
+    String testMethodName = new Object() {
+    }.getClass().getEnclosingMethod().getName();
+    logTestBegin(testMethodName);
+    LoggerHelper.getLocal().log(Level.INFO,
+        "Creating Domain & waiting for the script to complete execution");
+    boolean testCompletedSuccessfully = false;
+    try {
+      // create Domain using the image created by MII
```
markxnelson: this is a really nice test, i like this, i can read it understand the intent easily.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+    // Apply the new yaml to update the domain crd
+    LoggerHelper.getLocal().log(Level.INFO, "kubectl apply -f {0}", path.toString());
+    ExecResult exec = TestUtils.exec("kubectl apply -f " + path.toString());
+    LoggerHelper.getLocal().log(Level.INFO, exec.stdout());
+    LoggerHelper.getLocal().log(Level.INFO, "Verifying if the domain is restarted");
+    domain.verifyDomainRestarted();
+  }
+
+  private void verifyJdbcOverride() throws Exception {
+    // get domain name
```
markxnelson: this is a really nice method, logs what it is doing, puts the command in a string, which it prints and execs, so no chance of someone coming along later and updating one but not the other, the assertion is specific and has a reasonable message.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,246 @@
...
+        .append(domainName)
+        .append("/config/jdbc/")
+        .append(" && grep -R jdbc/generic1")
+        .append("'");
+
+    LoggerHelper.getLocal().log(Level.INFO, "Command to exec: " + cmdKubectlSh);
+    result = TestUtils.exec(cmdKubectlSh.toString());
+    LoggerHelper.getLocal().log(Level.INFO, "JDBC DS info from server pod: " + result.stdout());
+
+    Assumptions.assumeTrue(result.stdout().contains("<jndi-name>jdbc/generic1</jndi-name>"),
```
markxnelson: should this be an assumption or an assertion?

integration-tests/src/test/resources/model-in-image/model.jdbc.properties
```diff
@@ -0,0 +1,9 @@
+# Copyright 2019, Oracle Corporation and/or its affiliates. All rights reserved.
```
markxnelson: copyright statement wrong

integration-tests/src/test/resources/model-in-image/model.jdbc.properties
```diff
@@ -0,0 +1,9 @@
...
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+#
+
+MANSVR_PORT=8001
+dbcs1.url=slcai724.us.oracle.com:1583:w03ys12c
```
markxnelson: no oracle internal system names to be checked into public repositories

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: changed to:

 * This JUnit test is used for testing
 * Wdt Config Override with Model File(s) to existing MII domain

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: change to:

  * This method creates the result/pv root directories for the test.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: change to:

  * @throws Exception exception if initializing the application properties 
   *          and creates directories for results fails.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: Let me ask Vanaja this question

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: I removed and change the doc to "to new config map" because cm name remains same.

   * Create a domain using model in image and having configmap in the domain.yaml
   * before deploying the domain. After deploying the domain crd,
   * re-create the configmap with a model file that define a JDBC DataSource
   * and update the domain crd to change domain restartVersion
   * to reload the model, generate new config and initiate a rolling restart.
 
The purpose of changing restartVersion is to trigger Inspector to do rolling restart per Tom and Johnny

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: Thanks. I just added a python file to verify dsResource values like readTimeout, DS name, etc. Ready for the second test case :)

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: I delete it first and then create it again. Should it be re-created?

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: Thanks,Mark for encouragement :)

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: You are right, should be assertion so that the failures can be registered as a failed test. fixed

integration-tests/src/test/resources/model-in-image/model.jdbc.properties
hzhao-github: fixed

integration-tests/src/test/resources/model-in-image/model.jdbc.properties
hzhao-github: oh my word. I did this. Sorry. Was for testing purpose. forgot to change. Good Catch!

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: Hi Mark, here is Vanaja's response. Hope that it answers your question. If you think we still could improve here, just let QA know, we can discuss.

Vanajakshi Mukkara  8:11 AM
@huizhao Current tests create one operator for each test class, same operator is used for all the test methods in the class. We do this so that each test class can run independent of other tests while running in parallel.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,348 @@
...
+
+    // Re-create config map after deploying domain crd
+    final String domainUid = domain.getDomainUid();
+    final String cmName = domainUid + configMapSuffix;
+    final String label = "weblogic.domainUID=" + domainUid;
+
+    TestUtils.createConfigMap(cmName, destDir, domainNS, label);
+  }
+
+  private void modifyDomainYamlWithRestartVersion()
```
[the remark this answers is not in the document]
hzhao-github: modifyDomainYamlWithNewConfigMapAndDomainRestartVersion in MiiBaseTest.java does two things:

1. replace cm name with a new one
2. add restartVersion: v1.1

In my test, 
1. I don't need to create a brand new cm, I just re-create the existing one, so "replacing cm with a new name" doesn't apply to my test
2. restartVersion is hard coded, I need it to be flex for later test case. 

I can move modifyDomainYamlWithRestartVersion() in my test suite to MiiBaseTest.java though, if you like. Just let me know.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,348 @@
...
+    } finally {
+      if (domain != null && (JENKINS || testCompletedSuccessfully)) {
+        TestUtils.deleteWeblogicDomainResources(domain.getDomainUid());
+      }
+    }
+
+    LoggerHelper.getLocal().log(Level.INFO, "SUCCESS - " + testMethodName);
+  }
+
+  private void createDomainUsingMii() throws Exception {
```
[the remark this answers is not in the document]
hzhao-github: When I started coding, this method doesn't existing and I need the model files to be passed as params. Good we have it now. Will change to use it. Thanks

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,348 @@
...
+    try {
+      // create Domain using the image created by MII
+      createDomainUsingMii();
+
+      // override config
+      wdtConfigOverride();
+
+      // update domain yaml with restartVersion and
+      // apply the domain yaml, verify domain restarted
+      modifyDomainYamlWithRestartVersion();
```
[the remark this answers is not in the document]
hzhao-github: Ok. Will move my to MiiBaseTerst.java. Thanks

vanajamukkara:
@hzhao-github please update the PR with external jenkins results

integration-tests/src/test/java/oracle/kubernetes/operator/MiiBaseTest.java
```diff
@@ -116,4 +116,57 @@ public void modifyDomainYamlWithNewConfigMapAndDomainRestartVersion(
...
   }
+
+  /**
+   * Modify the domain yaml to change domain-level restart version.
+   * @param domain the domain
+   * @param versionNo version number of domain
+   *
+   * @throws Exception on failure
+   */
+  protected void modifyDomainYamlWithRestartVersion(
```
anpanigr: In the method modifyDomainYamlWithRestartVersion can't we use  kubectl patch command to update the Restart Version instead of modifying the domain.yaml file. We can get the current version and add +1 to it.  No need to get VersionNumber parameter. It will be easier to port to Junit5 where we will be using kubectl API.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,405 @@
...
+import oracle.kubernetes.operator.utils.TestUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * This JUnit test is used for testing
```
markxnelson: delete line 30 - we know junit tests are used for testing

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,405 @@
...
+  private static final String dsName = "MyDataSource";
+  private static final String readTimeout_1 = "30001";
+  private static final String readTimeout_2 = "30002";
+
+  /**
+   * This method does the initialization of the integration test
+   * properties defined in OperatorIT.properties and setting
+   * the resultRoot, pvRoot and projectRoot attributes.
+   *
+   * @throws Exception exception if initializing the application properties
```
markxnelson: you don't  need type & name on @throws 
just say @throws exception if initializing........

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
markxnelson: right, but you have this in a @BeforeEach not a @BeforeAll 
so this will do if for every test method, which is not what vanaja said -one per class

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
```diff
@@ -0,0 +1,405 @@
...
+      domainNS = ((ArrayList<String>) operatorMap.get("domainNamespaces")).get(0);
+      namespaceList.append((String)operatorMap.get("namespace"));
+      namespaceList.append(" ").append(domainNS);
+    }
+  }
+
+  /**
+   * Releases k8s cluster lease, archives result, pv directories.
+   *
+   * @throws Exception exception
```
markxnelson: fix this

markxnelson: stopping here - resubmit when you fix the basic stuff..

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: done

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: Exception is removed

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: fixed all :(

hzhao-github: fixed all :(

integration-tests/src/test/java/oracle/kubernetes/operator/MiiBaseTest.java
hzhao-github: solved

hzhao-github:
Passed on External Jenkins

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1220/consoleFull

markxnelson:
you still have basic problems that need to be fixed.
no point submitting a jenkins run of bad code. 
fix the problems first.

integration-tests/src/test/java/oracle/kubernetes/operator/ItModelInImageOverride.java
hzhao-github: it may cause extra works if we wanted to merge "static staticPrepare()" and "non-static prepare()" methods, because:
1. in staticPrepare() , it calls a static method in its parent class, so this method has to be a static one
2. Creating Operator method calls non-static method in Operator.java and BaseTest.java, so I can't move it to staticPrepare() 
3. Even though beforeEach annotation is used, Operator is still created once due to the condition check if)operator != null)
4. If we wanted to get rid of this beforeEach annotation when creating Operator, it need changes in our test framework.

after discussing with Mark, we will keep as it is for now

tbarnes-us:
I'd prefer to see this renamed 'mii-update-tests' or 'mii-runtime-update-tests' or even better 'mii-config-update-tests'.   

Also any use of the term override within the test itself should be converted.   

The reason is that we're not using overrides here - we're using model update files.   Overrides is an overloaded term - it's already in use for config-overrides feature which is not the same thing.  It gets doubly confusing since we are using overrides internally still for things like the listen address, but that's not a model update.

hzhao-github:
removed the work=d "override" from the test suite

PR #1503, opened 2020-03-26
OWLS-80522 Domain status ordering, OWLS-80144 Ordered rolling upgrade

alai8:
-   List domain status by admin server first, then alpha-num across non-clustered servers, then grouped by cluster alpha-numerically, then alpha-num within cluster, ie, server1, server2, ..., server10, server11... etc.
-    Perform rolling upgrade of servers alpha-numerically within same cluster when restartVersion value is updated.


operator/src/main/java/oracle/kubernetes/utils/OperatorUtils.java
```diff
@@ -19,4 +19,62 @@ public static String joinListGrammatically(final List<String> list) {
...
+    ret += getSortingWord(word);
+    return ret;
+  }
+
+  private static String getSortingWord(String word) {
+    if (word.length() == 0) {
+      return word;
+    }
+    if (Character.isDigit(word.charAt(0))) {
+      for (int i = word.length(); i < 6; i++) {
```
tbarnes-us: How about something more generous than 6 - like 20?  The numbers we're padding come from the customer and there's not much need to be stingy here.

tbarnes-us:
I have a couple of general questions.  Sorry, I should have checked on them before approving:

1: How do non-clustered servers fit into the sort pattern?    I suspect it'd be best to list them right after the admin server.

2:  And do cluster names also sort alpha-numerically?

Assuming the answer to both 1 & 2 are yes, please update the pull description accordingly: 

Change "List domain status by admin server first, then grouped by cluster, then alpha-num within cluster, ie, server1, server2, ..., server10, server11... etc.".  to something like  "List domain status by admin server first, then alpha-num across non-clustered servers, then grouped by cluster alpha-numerically, then alpha-numerically within cluster, ie, server1, server2, ..., server10, server11... etc."

rjeberhard:
@alai8, this looks really good to me.  Thanks for consistently having detailed unit-tests.  Is this really "work in progress"?

alai8:
@rjeberhard yes, WIP. please don't merge yet.

operator/src/main/java/oracle/kubernetes/utils/OperatorUtils.java
alai8: done. Yes, we should be more generous.

alai8:
> I have a couple of general questions. Sorry, I should have checked on them before approving:
> 
> 1: How do non-clustered servers fit into the sort pattern? I suspect it'd be best to list them right after the admin server.
> 
> 2: And do cluster names also sort alpha-numerically?
> 
> Assuming the answer to both 1 & 2 are yes, please update the pull description accordingly:
> 
> Change "List domain status by admin server first, then grouped by cluster, then alpha-num within cluster, ie, server1, server2, ..., server10, server11... etc.". to something like "List domain status by admin server first, then alpha-num across non-clustered servers, then grouped by cluster alpha-numerically, then alpha-numerically within cluster, ie, server1, server2, ..., server10, server11... etc."

Yes to both with the latest change, for domain status.
But this does not applies to rolling upgrade because:
    - rolling upgrade of non-clustered servers, other than admin servers, could be performed concurrently and not one after another. This is different from servers in a cluster.
    - rolling upgrade of multiple clusters are not performed in any specific ordering, but can be performed concurrently. For example, restart of servers cluster1-1 in cluster1 and server cluster2-1 in cluster2 can be initiated at the same time by the operator.

PR #1506, opened 2020-03-27
Latest updates for SOA

sbattagi:
Hi @markxnelson @rjeberhard ,

This PR has two changes:
1) Updated the documentation, to use only operator 2.4.0 version for SOA.
2) Fix for bug 31011443, where in domain.yaml spec level the "serverService" was under serverPod, which was causing the validation error. As per Ryan's inputs both the serverService and serverPod should be at same level as per the schema.
We have commented this change as we don't need serverService at spec level (for us having this only at cluster level should be fine).
   serverPod:
    serverService:
      precreateService: true

Please review and approve these changes.

Thanks
Samba


docs-source/content/userguide/managing-fmw-domains/soa-suite/_index.md
```diff
@@ -64,7 +64,7 @@ following limitations currently exist for SOA Suite domains:
 * Only configured clusters are supported.  Dynamic clusters are not supported for
   SOA Suite domains.  Note that you can still use all of the scaling features,
   you just need to define the maximum size of your cluster at domain creation time.
-* Deploying and running SOA Suite domains is supported only in operator versions 2.4.0 and later.
+* Deploying and running SOA Suite domains is supported only in operator version 2.4.0.
```
rjeberhard: @rosemarymarano, should this be "supported only with" or "supported only on"?  It sounds odd to say "only in" with just one version.

rosemarymarano: @rjeberhard Yep, "supported only with" is correct.

markxnelson: @rdas0405 @sbattagi can you help me understand why you are saying only with 2.4.0, rather than 2.4.0 or later please

rjeberhard: I've been thinking about this more...  We need to get on the same page.  Since this is an open-source project with a short release cycle, the intention was to always move the releases forward -- that is, customers would receive bug fixes by upgrading to newer versions of the operator.  We may need to designate certain versions as having long term support, but that would only be practical for the last release in a given major version line (so, not 2.4.0).  If a SOA customer finds a bug while using 2.4.0, my expectation is that this customer would use 2.5.0.

sbattagi: Hi @markxnelson 

This is for SOA early access customers we are suggesting to use operator 2.4.0 only, as with 2.5.0 we have a schema validation failure, which needs a code fix.

Thanks
Samba

markxnelson:
Can you please provide details of the issue?

rjeberhard:
In 2.5.0, we fixed an issue with CRD creation that was previously disabling Kubernetes schema validation.  This means that you had the same schema validation failure in 2.4.0; it just wasn't reporting the error.  Assuming this is the serverService problem then this should be a one-line fix for you.

sbattagi:
As @rjeberhard rightly mentioned, the issue is with "serverService" as reported in bug 31011443.

rjeberhard:
Is this one bug fix the only issue blocking support for 2.5.0?

sbattagi:
> 
> 
> Is this one bug fix the only issue blocking support for 2.5.0?

Yes. This is only for Early access customers, as all our testing and certification is done based on 2.4.0.
Later for the production support we will support 2.5.0+.

sbattagi:
Kindly add @rdas0405 and @vivek-sam to this review.

sbattagi:
Hi @rjeberhard  @markxnelson,

Kindly review and approve this pull request. 
Also please add @rdas0405 and @vivek-sam to this review.

Thanks
Samba

PR #1512, opened 2020-03-27
Install wit actions interface

doxiao:
Add actions for installer and WebLogicImageTool operations. The review goal is to make sure that style and interface is okay. The full impl will be added once the interface/style is in agreement.

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
@@ -58,22 +58,35 @@ public class TestActions {
     	return 
             new WebLogicImageTool()
                 .with(new WITParams()
-                     .baseName(baseImageName).tag(imageTag).modelFiles(modelFiles)
-                     .modelVariableFiles(modelVariableFiles).modelArchiveFiles(modelArchiveFiles)
-                     .wdtVersion(wdtVersion).domainType(domainType)).updateImage();
+                     .baseName(baseImageName)
+                     .tag(imageTag)
+                     .modelFiles(modelFiles)
+                     .modelVariableFiles(modelVariableFiles)
+                     .modelArchiveFiles(modelArchiveFiles)
+                     .wdtVersion(wdtVersion)
+                     .domainType(domainType))
+                .updateImage();
     }
  
  
     // ------------------------ Installer  -------------------------
 
     public static boolean verifyAndInstallWIT(String version, String location) {
-        return new Installer().with(new InstallParams().type("WIT").version(version).location(location))
-        		      .download();
+        return new Installer()
+            .with(new InstallParams()
+                  .type("WIT")
+                  .version(version)
+                  .location(location))
+            .download();
     }
  
     public static boolean verifyAndInstallWDT(String version, String location) {
-        return new Installer().with(new InstallParams().type("WDT").version(version).location(location))
-  		      .download();
+        return new Installer()
+            .with(new InstallParams()
+                  .type("WDT")
+                  .version(version)
+                  .location(location))
+            .download();
 
     }
  
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
@@ -46,25 +46,10 @@ public class TestActions {
 
     // ------------------------ Docker image  -------------------------
 
-    public static boolean createMIIImage(
-        String baseImageName, 
-        String imageTag,
-        List<String> modelFiles,
-        List<String> modelVariableFiles,
-        List<String> modelArchiveFiles,
-        String wdtVersion,
-        String domainType) {
-  
+    public static boolean createMIIImage(WITParams params) {
     	return 
             new WebLogicImageTool()
-                .with(new WITParams()
-                     .baseName(baseImageName)
-                     .tag(imageTag)
-                     .modelFiles(modelFiles)
-                     .modelVariableFiles(modelVariableFiles)
-                     .modelArchiveFiles(modelArchiveFiles)
-                     .wdtVersion(wdtVersion)
-                     .domainType(domainType))
+                .with(params)
                 .updateImage();
     }
  
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
@@ -59,8 +59,8 @@ public class TestActions {
             new WebLogicImageTool()
                 .with(new WITParams()
                      .baseName(baseImageName).tag(imageTag).modelFiles(modelFiles)
-    	             .modelVariableFiles(modelVariableFiles).modelArchiveFiles(modelArchiveFiles)
-    		     .wdtVersion(wdtVersion).domainType(domainType)).updateImage();
+                     .modelVariableFiles(modelVariableFiles).modelArchiveFiles(modelArchiveFiles)
+                     .wdtVersion(wdtVersion).domainType(domainType)).updateImage();
     }
  
  
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
@@ -3,9 +3,15 @@
 
 package oracle.weblogic.kubernetes.actions;
 
+import java.util.List;
+
 import oracle.weblogic.kubernetes.actions.impl.Domain;
 import oracle.weblogic.kubernetes.actions.impl.Namespace;
 import oracle.weblogic.kubernetes.actions.impl.Operator;
+import oracle.weblogic.kubernetes.actions.impl.primitive.Installer;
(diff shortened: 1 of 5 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
```diff
@@ -38,6 +44,37 @@ public static String createUniqueNamespace() {
...
+        List<String> modelFiles,
+        List<String> modelVariableFiles,
+        List<String> modelArchiveFiles,
+        String wdtVersion,
+        String domainType) {
+  
+    	return 
+            new WebLogicImageTool()
+                .with(new WITParams()
+                     .baseName(baseImageName).tag(imageTag).modelFiles(modelFiles)
```
markxnelson: this is just a nit pick, i prefer to put start a new line before every dot/period

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
```diff
@@ -38,6 +44,37 @@ public static String createUniqueNamespace() {
...
+        String baseImageName, 
+        String imageTag,
+        List<String> modelFiles,
+        List<String> modelVariableFiles,
+        List<String> modelArchiveFiles,
+        String wdtVersion,
+        String domainType) {
+  
+    	return 
+            new WebLogicImageTool()
```
markxnelson: this is ok for now, but i might ask to make it statics rather than passing around objects when i see you use this in  a test

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
tbarnes-us: I assume baseImageName/Tag will itself eventually come from an 'ImageParams' type.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
```diff
@@ -38,6 +44,37 @@ public static String createUniqueNamespace() {
...
+                .with(new WITParams()
+                     .baseName(baseImageName).tag(imageTag).modelFiles(modelFiles)
+                     .modelVariableFiles(modelVariableFiles).modelArchiveFiles(modelArchiveFiles)
+                     .wdtVersion(wdtVersion).domainType(domainType)).updateImage();
+    }
+ 
+ 
+    // ------------------------ Installer  -------------------------
+
+    public static boolean verifyAndInstallWIT(String version, String location) {
```
tbarnes-us: IMO, ideally version/location would itself come from a global type - the net idea is that at the beginning of a suite, all such settings are defined once (and only once) and then passed around.  Like the 'packet' object in the operator code.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
tbarnes-us: I guess I was thinking of something different.  I would have thought createMIIImage would take a WITParams() as a parameter.   It wouldn't have a signature with all of the params listed out - instead WITParams would be built in some central location, or generated from some central set of settings for the suite, and passed in.

markxnelson: yes, some reusable object with a fluent api would make a nice param to this method

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
doxiao: @markxnelson it does make the logic clearer to start a new line before every dot.

doxiao: @tbarnes-us We could add an "ImageParams" if/when the need comes.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
doxiao: I think the problem may be the additional layer at the WebLogicImageTool level. One solution is to remove the WITParams, and use the WebLogicImageTool as the reusable object with the fluent api that is in WITParams now, as well as the updateImage method that WebLogicImageTool currently has.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
doxiao: I hope we only need to do download/install of WIT and WDT binary once for the entire integration test suite. We could make them static constants, but parameterizing them makes it a little easier for the test suite to test a newer version (when the time comes).

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
doxiao: Then the TestActions will have a with() method that takes a WebLogicImageTool object. A test code of creating a MII image would look like the following.

WebLogicImageTool wit = new WebLogicImageTool()
                                                 .imageName("xxx)
                                                 .tag("yyy")
                                                 and etc.
new TestActions().with(wit).createMIIImage().

doxiao: I can give it a try and see if it looks better.

doxiao: I think it is easier to just stick with the current approach and change the parameters into a WITParams.

markxnelson: no - we would not instantiate TestActions - all of the methods are statics, so you would `import static TestActions` and then say:
`with(wit).createMIIImage()`

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
markxnelson: i think it makes sense to download all of the prereqs up front (in the jenkins job) before we start running the tests - e.g. WIT, WDT, whatever installer/patches we need, helm, kubectl, etc.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
doxiao: I think what I eventually did is similar to what you suggested - TestActions does the following.
new WebLogicImageTool().with(wit).updateImage().

ddsharpe: if I understand correctly, you two are close.  But, I think Mark is suggesting an overloaded "with()" method where the argument decides the operation.  In this case "with(WebLogicImageTool)" finds the ImageTool with statement, and since the TestActions is a static import and all "with()" methods are static, it looks more like with(wit).createMIIImage().  I think you are mostly on the same page here, so approving.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
ddsharpe: Let's put the prerequisite downloads in the Jenkins file.  I think it will be much easier to maintain.  And, if we switch to another environment like running in Docker containers with the software pre-positioned, the change is very minor.

doxiao: Agreed. We'll need to have a way for the test code to get the location of WIT and WDT binary; the MII use case needs to run WIT to generate a nee image that contains the WDT binary.

ddsharpe: Its easy for the Jenkinsfile to download the bits, extract them, and then put the folder/path in an environment variable.  Will that work?

tbarnes-us: @ddsharpe I think we shjould avoid depending on Jenkins for any critical part of the flow.  It should be possible to get a test up and running, _with a single mvn call_, in any basic WKO env. We shouldn't need to install Jenkins to get it working IMO - unless installing Jenkins and setting it up is itself one step?

tbarnes-us: So would relying on a Jenkinsfile also rely on Jenkins being installed?

tbarnes-us: Derek and I have agreed to disagree off-line - and are looking for other opinions.  

Minimally I'd like to see (1) single or at most two places that 'own' setting up WDT/WIT, (2) target/scope install location of WDT/WIT to the the test suite (cleaned up when test suite is deleted - saved if along with test results if test suite results are saved), (3) it must be very easy to get a test running stand-alone (if missing the test infra should get it - or at least a single command for setting up prior to test suite).

anpanigr: If we offload the WIT/WDT download  responsibly to Jenkin Script, then to run the test locally we need to add  a similar script to the operator github project. There may be out-of-sync  issue between  Jenkin and a local copy

PR #1514, opened 2020-03-30
K8s client api initial impl for feedback on usage

lennyphan:
PR for feedback to validate usage of Kubernetes client API is correct and that I'm on the same page as everyone else on direction.

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
@@ -3,7 +3,7 @@
 
 package oracle.weblogic.kubernetes;
 
-import io.kubernetes.client.openapi.ApiException;
+import oracle.weblogic.kubernetes.actions.TestActions;
 import oracle.weblogic.kubernetes.annotations.IntegrationTest;
 import oracle.weblogic.kubernetes.annotations.tags.Slow;
 import oracle.weblogic.kubernetes.extensions.LoggedTest;
@@ -13,10 +13,10 @@ import org.junit.jupiter.api.Test;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static oracle.weblogic.kubernetes.actions.TestActions.createDomainCustomResource;
-import static oracle.weblogic.kubernetes.actions.TestActions.createUniqueNamespace;
 import static oracle.weblogic.kubernetes.actions.TestActions.deleteNamespace;
 import static oracle.weblogic.kubernetes.assertions.TestAssertions.domainExists;
 import static org.awaitility.Awaitility.with;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @DisplayName("Simple validation of basic domain functions")
@@ -33,17 +33,9 @@ class ItSimpleDomainValidation implements LoggedTest {
 
     // get a new unique namespace
     String namespace = null;
-    try {
-      namespace = createUniqueNamespace();
-      logger.info(String.format("Got a new namespace called %s", namespace));
-    } catch (ApiException e) {
-      // TODO: test in the calling method where we say something like
-      //
-      //  assert.DoesNotThrow(whatever(), ApiException.class, e,
-      //    String.format("could not do whatever, got exception %s", e))
-      //
-      //  so we have the exception and we can print out a meaningful error message
-    }
+    namespace = assertDoesNotThrow(TestActions::createUniqueNamespace,
+        "Failed to create unique namespace due to ApiException");
+    logger.info(String.format("Got a new namespace called %s", namespace));
 
     // create the domain CR
     boolean success = createDomainCustomResource(domainUID, namespace, domainYAML);
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
@@ -13,6 +13,7 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static oracle.weblogic.kubernetes.actions.TestActions.createDomainCustomResource;
 import static oracle.weblogic.kubernetes.actions.TestActions.createUniqueNamespace;
+import static oracle.weblogic.kubernetes.actions.TestActions.deleteNamespace;
 import static oracle.weblogic.kubernetes.assertions.TestAssertions.domainExists;
 import static org.awaitility.Awaitility.with;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -52,6 +53,10 @@ class ItSimpleDomainValidation implements LoggedTest {
         // wait for the admin server pod to exist
 
         // wait for the managed servers to exist
+
+        // Delete namespace
+        deleteNamespace(namespace);
+        logger.info("Deleted namespace: " + namespace);
 
     }
 
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
@@ -32,8 +32,7 @@ class ItSimpleDomainValidation implements LoggedTest {
     String domainYAML = "something";
 
     // get a new unique namespace
-    String namespace = null;
-    namespace = assertDoesNotThrow(TestActions::createUniqueNamespace,
+    String namespace = assertDoesNotThrow(TestActions::createUniqueNamespace,
         "Failed to create unique namespace due to ApiException");
     logger.info(String.format("Got a new namespace called %s", namespace));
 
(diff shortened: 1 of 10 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -32,7 +78,18 @@ public static List listDeployments() {
...
 
     public static boolean createNamespace(String name) {
-        return true;
+        V1ObjectMeta meta = new V1ObjectMeta();
+        meta.name(name);
+        V1Namespace namespace = new V1Namespace();
+        namespace.metadata(meta);
+        try {
+            namespace = coreV1Api.createNamespace(namespace, pretty, null, null);
+            System.out.println("Kubernetes.createNamespace namespace: " + namespace);
```
markxnelson: let's use log not sop

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -32,7 +78,18 @@ public static List listDeployments() {
...
+            return true;
+        } catch (ApiException e) {
+            e.printStackTrace();
```
markxnelson: probably a good idea to return the exception

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -54,4 +111,83 @@ public static String createUniqueNamespace() {
...
+        String writeValueAsString = jsonWriter.writeValueAsString(yamlObj);
+        System.out.println("Kubernetes.convertYamlToJson writeValueAsString: " + writeValueAsString);
+        JsonNode root = new ObjectMapper().readTree(writeValueAsString);
+        return root;
+    }
+
+    public static List<String> listDomains(String namespace) {
+        ArrayList<String> domains = new ArrayList<>();
+        try {
+            Map response = (Map) customObjectsApi.listNamespacedCustomObject(DOMAIN_GROUP, DOMAIN_VERSION, namespace, DOMAIN_PLURAL, null,
```
markxnelson: nit pick - put each param on a new line - maybe put a comment after them to say what they are, since the six nulls seem a bit magical :)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Namespace.java
```diff
@@ -1,10 +1,12 @@
-// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
+// Copyright 2020, Oracle Corporation and/or its affiliates.
```
ddsharpe: Why did this get changed? The (c) is part of the check-style.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1,10 +1,29 @@
-// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
+// Copyright 2020, Oracle Corporation and/or its affiliates.
```
ddsharpe: same issue, (c)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
rjeberhard: Yeah, it's too bad they don't have a builder for requests (I'll validate that this is still correct).

rjeberhard:
@lennyphan, just a general heads-up that I'm separately working on a fluent Domain api for use by the test code.  This will be just like the code for V1Pod, etc., so won't be a copy of the product code.  When that's ready we'll switch the YAML String parameters to these types.  We'll still have to use CustomObjectsApi or GenericKubernetesApi.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Namespace.java
lennyphan: Ah, Thanks.  this is a copy over issue.  I created a different branch for the PR and pulled over my modified file that has an older copyright. Thanks.

lennyphan:
@rjeberhard Thank you for the heads up.  I realized this based on reading your slack comments about what you were doing over the weekend.  This early PR review helps to clarify how I'm working through the Kubernetes java client api.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
tbarnes-us: @markxnelson Should we catch Throwable throughout ?

markxnelson: I prefer to catch the most specific object and handle it

tbarnes-us: I realize that's SOP for system programming in Java to narrowly only check for the checked exceptions.  But I have two counter thoughts to this.  One, we're talking about testing here - so 'happy path' programming is far more acceptable.   Two, if there's an insistence not to catch Runtime/Throwable throughout, then there will be many cases where an exception will float out of a library all the up to the highest level.  But I thought the goal you set out was to prevent this from happening.  For example, there's a (sad IMO) school of thought these days that even `normal` paths can normally throw runtimes - even some JDBC drivers do that unfortunately - and we've no clue in a callout to external library what the hell it's going to throw...

tbarnes-us: Even in system programming catch Throwable tends to be useful.  For calls to the 'outside' we've found we had to do that in WL in many places.  Calling customer code.  Calling shaky external libraries.

lennyphan: Hi Mark, can you clarify here?  I assume you want to just declare the exception in the throws clause and have it propagate up?   Just a little confusion on my part relative to the discussion about handling the types of exceptions.  Thanks for clarifying.

markxnelson: yeah, i am thinking about the test in the calling method where we say something like ```
```assert.DoesNotThrow(whatever(), ApiException.class, e,
  String.format("could not do whatever, got exception %s", e))```

File: new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java [hunk 6]
```diff
@@ -30,8 +32,19 @@ public void testCreatingDomain() {
...
+            logger.info(String.format("Got a new namespace called %s", namespace));
+        } catch (ApiException e) {
+            // TODO: test in the calling method where we say something like
+            //
+            //  assert.DoesNotThrow(whatever(), ApiException.class, e,
+            //    String.format("could not do whatever, got exception %s", e))
+            //
+            //  so we have the exception and we can print out a meaningful error message
+        }
+
```
lennyphan: For the moment, updated the test to surround with try-catch until we learn how to handle exceptions with JUnit 5 assertions, as noted from Mark.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleDomainValidation.java
```diff
@@ -53,6 +66,16 @@ public void testCreatingDomain() {
...
+        // Delete namespace
+        try {
+            deleteNamespace(namespace);
+            logger.info("Deleted namespace: " + namespace);
+        } catch (Exception e) {
+            // TODO: Fix as there is a known bug that delete can return either the object
+            //  just deleted or a status.  We can workaround by either retrying or using
+            //  the general GenericKubernetesApi client class and doing our own type checks
+        }
+
```
lennyphan: Same as above comment.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
lennyphan: Changed to use logger.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
lennyphan: Added throws clause to method signature.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
lennyphan: split params to individual lines and added comments to explain parameter meaning.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Namespace.java
lennyphan: Corrected copyright.

lennyphan:
Updated with review comments.  Merged checkstyle changes, added kubernetes api calls for PV/PVC, secrets, config maps.

markxnelson:
This looks really good Lenny.
@rjeberhard @ddsharpe I think this is ready to merge.

PR #1517, opened 2020-03-30
Enhance Helm chart deletion Logic in cleanup.sh Script  

anpanigr:
Modified the clean up scripts to loop thru the NameSpace which has active Helm Deployment for the deletion of Helm Deployment.  The new logic will take care of  both Helm 2.x/3.x

### DIFF
--- a/src/integration-tests/bash/cleanup.sh
+++ b/src/integration-tests/bash/cleanup.sh
@@ -537,23 +537,15 @@ if [ -x "$(command -v helm)" ]; then
   fi
 
   if [ "$HELM_VERSION" == "V3" ]; then
-   namespaces=`helm list --all-namespaces | grep -v NAME | awk '{print $2}'`
-   for ns in $namespaces
-   do 
-     echo "Looking for helm lists in namesapce $ns"
-     helm_names=`helm list --short -n $ns`
-     for helm_name in $helm_names
-     do 
-        if [ ! "$DRY_RUN" = "true" ]; then
-        (
-          set -x
-          helm uninstall $helm_name -n $ns
-        )
-       else 
-         echo @@ `timestamp` Info: DRYRUN: helm uninstall $helm_name -n $ns
-       fi
-     done
-   done
+    if [ ! "$DRY_RUN" = "true" ]; then
+    (
+      set -x
+      helm list --all-namespaces | grep -v NAME | awk '{system("helm uninstall " $1 " --namespace " $2)}'
+    )
+    else 
+      echo @@ `timestamp` Info: DRYRUN: helm uninstall 
+      helm list --all-namespaces | grep -v NAME | awk '{print $1 "\t" $2)}'
+   fi
   fi
 
   # cleanup tiller artifacts
--- a/src/integration-tests/bash/cleanup.sh
+++ b/src/integration-tests/bash/cleanup.sh
@@ -523,32 +523,38 @@ if [ -x "$(command -v helm)" ]; then
   [[ $? == 1 ]] && HELM_VERSION=V3
   echo "Detected Helm Version [$(helm version --short --client)]"
   echo @@ `timestamp` Deleting installed helm charts
-  namespaces=`kubectl get ns | grep -v NAME | awk '{ print $1 }'`
-  for ns in $namespaces
-  do 
+  if [ "$HELM_VERSION" == "V2" ]; then
+   helm list --short | while read helm_name; do
    if [ ! "$DRY_RUN" = "true" ]; then
-     (
+   (
      set -x
-     helm list --short --namespace $ns | while read helm_name; do
-       if [ "$HELM_VERSION" == "V2" ]; then
-         helm delete --purge  $helm_name
-       else 
-         helm uninstall $helm_name -n $ns 
-       fi
-     done
-     )
+     helm delete --purge  $helm_name
+    )
    else
-     (
-     helm list --short --namespace $ns | while read helm_name; do
-       if [ "$HELM_VERSION" == "V2" ]; then
-         echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+     echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+   fi
+   done
+  fi
+
+  if [ "$HELM_VERSION" == "V3" ]; then
+   namespaces=`helm list --all-namespaces | grep -v NAME | awk '{print $2}'`
+   for ns in $namespaces
+   do 
+     echo "Looking for helm lists in namesapce $ns"
+     helm_names=`helm list --short -n $ns`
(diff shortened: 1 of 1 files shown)


src/integration-tests/bash/cleanup.sh
```diff
@@ -523,32 +523,38 @@ if [ -x "$(command -v helm)" ]; then
...
+     echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+   fi
+   done
+  fi
+
+  if [ "$HELM_VERSION" == "V3" ]; then
+   namespaces=`helm list --all-namespaces | grep -v NAME | awk '{print $2}'`
+   for ns in $namespaces
+   do 
+     echo "Looking for helm lists in namesapce $ns"
```
markxnelson: The loop is still here

src/integration-tests/bash/cleanup.sh
```diff
@@ -523,32 +523,38 @@ if [ -x "$(command -v helm)" ]; then
...
-     helm list --short --namespace $ns | while read helm_name; do
-       if [ "$HELM_VERSION" == "V2" ]; then
-         echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+     echo @@ `timestamp` Info: DRYRUN: helm delete --purge  $helm_name
+   fi
+   done
+  fi
+
+  if [ "$HELM_VERSION" == "V3" ]; then
+   namespaces=`helm list --all-namespaces | grep -v NAME | awk '{print $2}'`
```
tbarnes-us: I suspect this should be:      

helm list --all-namespaces | grep -v NAME | awk '{print $2}' | sort -u

We only want to iterate through each namespace once.

src/integration-tests/bash/cleanup.sh
anpanigr: namespaces variable will store only the namesapce(s) that has a helm deploymet

tbarnes-us: The loop here is OK IMO, as long as we only iterate through each NS once per my previous comment.

tbarnes-us: But if you want to further optimize, iterate through the output of the first all namespace command in a single "big loop" that parses its lines, which which parses out the ns from the line and parses out the helm name from the same line.

src/integration-tests/bash/cleanup.sh
anpanigr: nice suggestion. will implement change in next commit

tbarnes-us: i am assuming that the above command does not yield a ns unless there's at least one helm release in the ns

tbarnes-us:
> The list gives you namespace and release, awk print them both and delete

This is more efficient than my 'sort -u' suggestion above.

anpanigr:
I tried the following 
helm list --all-namespaces | grep -v NAME | awk '{system("helm uninstall $1 --namespace $2 ")}'
But it throws helm usage  ( Error: flag needs an argument: --namespace )
Am I doing something wrong ?  I am not a awk expert

markxnelson:
2020-03-30 20:12:33:~ 
$ helm version
version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}

2020-03-30 20:12:36:~ 
$ helm list --all-namespaces
NAME                            	NAMESPACE                     	REVISION	UPDATED                                	STATUS	CHART                  	APP VERSION
wls-operator-bobs-books-binding1	verrazzano-bobs-books-binding1	1       	2020-03-30 20:16:06.628931236 +0000 UTC	failed	weblogic-operator-2.5.0

markxnelson:
what version of helm do you have? i have 3.1.2

markxnelson:
$ helm list --all-namespaces | grep -v NAME | awk ' { print "helm uninstall " $1 " --namespace " $2 }' 
helm uninstall wls-operator-bobs-books-binding1 --namespace verrazzano-bobs-books-binding1

anpanigr:
I am using following helm 
version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}

markxnelson:
$ helm list --all-namespaces | grep -v NAME | awk ' { system( "helm uninstall " $1 " --namespace " $2) }' 
release "wls-operator-bobs-books-binding1" uninstalled

markxnelson:
ok let me know when you are ready @anpanigr 
i will run a jenkins job to confirm it works, and then merge it tonight before the nightly test run

tbarnes-us:
> $ helm list --all-namespaces | grep -v NAME | awk ' { system( "helm uninstall " $1 " --namespace " $2) }'

Here's a modification of the above that's a bit harder to read (code looks more complex), but should be easier to debug if something goes wrong at runtime:

awk ' { system( "echo -n About to helm uninstall " $1 " --namespace " $2 ": ; helm uninstall " $1 " --namespace " $2 " ) }' 

It might be that the set -x that's in there already will do the above.  But if it doesn't...

tbarnes-us:
> > $ helm list --all-namespaces | grep -v NAME | awk ' { system( "helm uninstall " $1 " --namespace " $2) }'
> 
> Here's a modification of the above that's a bit harder to read (code looks more complex), but should be easier to debug if something goes wrong at runtime:
> 
> awk ' { system( "echo -n About to helm uninstall " $1 " --namespace " $2 ": ; helm uninstall " $1 " --namespace " $2 " ) }'
> 
> It might be that the set -x that's in there already will provide sufficient debug output. But if it doesn't...

Or even better:

awk ' { system( "set -x ; helm uninstall " $1 " --namespace " $2 " ) }'

(The set -x scopes inside the system call.)

src/integration-tests/bash/cleanup.sh
```diff
@@ -537,23 +537,15 @@ if [ -x "$(command -v helm)" ]; then
...
-        )
-       else 
-         echo @@ `timestamp` Info: DRYRUN: helm uninstall $helm_name -n $ns
-       fi
-     done
-   done
+    if [ ! "$DRY_RUN" = "true" ]; then
+    (
+      set -x
+      helm list --all-namespaces | grep -v NAME | awk '{system("helm uninstall " $1 " --namespace " $2)}'
```
tbarnes-us: Change "helm uninstall" to "set -x ; helm uninstall".  In my testing it turns out that the outer scoped set -x is suppressed by teh the system call scope, so you need one inside the system call scope too.

markxnelson:
@rjeberhard cherry picked this into frozen branch model-in-image as this is an important fix for nightly tests

PR #1522, opened 2020-03-31
Adding impl in Operator

vanajamukkara:
fluent Namespace 
adding TraefikParams
adding impl in Operator.. more to do


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
```diff
@@ -3,35 +3,78 @@
...
+   * @return true if the operator is successfully installed, false otherwise.
+   * @throws ApiException - if Kubernetes client API call fails
+   */
+  public static boolean install(OperatorParams params) throws ApiException {
+
+    String namespace = params.getNamespace();
+    String serviceAccount = params.getServiceAccount();
+
+    // assertions for required parameters
+    assertNotNull(namespace, "Operator namespace cannot be null");
```
markxnelson, line 38: use assertj - will make this look like
```   
import static org.assertj.core.api.Assertions.*;

 assertThat(namespace)
            .as("make sure namespace is not empty or null")
            .isNotEqualTo("")
            .isNotNull();
```
which is a lot clearer

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
```diff
@@ -3,35 +3,78 @@
...
+    assertNotNull(params.getReleaseName(), "Operator release name cannot be null");
+    assertNotEquals("",namespace.trim(), "Operator namespace cannot be empty string");
+    assertNotEquals("",params.getReleaseName().trim(), "Operator release name cannot be empty string");
+
+    //delete the namespace if it exists
+    if (Namespace.exists(namespace)) {
```
markxnelson, line 44: why delete the namespace? seems a bit extreme

vanajamukkara: what if the namespace already exists? just proceed with it?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
vanajamukkara: looking for something similar to isNotNullOrEmpty() as it takes care of multiple white spaces..

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
markxnelson: yeah - the method is meant to install an operator --- if doing so deleted the target namespace, that would be unexpected and surprising behavior.  we should not do that

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
rjeberhard: We have Guava as a dependency in other modules.  You could add it here and use: assertFalse(Strings.isNullOrEmpty(text));

vanajamukkara: added assertj core dependency and doing ..
```
    assertThat(namespace)
        .as("make sure namespace is not empty or null")
        .isNotNull()
        .isNotEmpty();
```

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
vanajamukkara: ok, will modify the code accordingly

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
```diff
@@ -58,19 +59,26 @@ public void testInstallingOperator() {
...
-    OperatorParams opParams
-        = new OperatorParams().image("weblogic-kubernetes-operator:test_itsimpleoperator")
-          .domainNamespaces(Arrays.asList("domainns1", "domainns2"))
-          .serviceAccount("sa-opns1");
-    boolean success = installOperator("weblogic-operator", opns, opParams);
+    OperatorParams opParams =
+        new OperatorParams().releaseName("weblogic-operator")
+            .namespace(opns)
+            .image("weblogic-kubernetes-operator:test_itsimpleoperator")
+            .domainNamespaces("domainns1, domainns2")
```
markxnelson, line 67: i'm not sure about this being a string like this - i liked it better the way you had it before, where is was clear it was a list of things, this is just a string that has some kind of special meaning that i don't understand

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
```diff
@@ -3,35 +3,73 @@
...
+    String serviceAccount = params.getServiceAccount();
+
+    // assertions for required parameters
+    assertThat(namespace)
+        .as("make sure namespace is not empty or null")
+        .isNotNull()
+        .isNotEmpty();
+
+    assertThat(params.getReleaseName())
+        .as("make sure namespace is not empty or null")
```
markxnelson, line 44: namespace?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
```diff
@@ -3,35 +3,73 @@
...
+        .isNotNull()
+        .isNotEmpty();
+
+    // create namespace
+    if (!Namespace.exists(namespace)) {
```
markxnelson, line 49: again, this is surprising/unexpected.  the method is installOperator - i would not expect that to create a namespace. if you try to deploy something into a non-existent namespace in k8s, it will give an error.  that is what i would expect this to do too. we should try to stick to the standard/predictable/expected behavior

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
```diff
@@ -3,35 +3,73 @@
...
+      logger.info(String.format("Creating namespace %s", namespace));
+      Namespace ns = new Namespace().name(namespace);
+    }
+
+    // create service account
+    if (serviceAccount != null && !serviceAccount.equals("default")) {
```
markxnelson, line 55: same here

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
vanajamukkara: ok ..I will create the namespace, service account in the test class ..

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
vanajamukkara: changed it back

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
vanajamukkara: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Operator.java
vanajamukkara: removed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
```diff
@@ -59,18 +65,37 @@ public void testInstallingOperator() {
...
+    // get a new unique namespace
+    final String namespace = assertDoesNotThrow(TestActions::createUniqueNamespace,
+        "Failed to create unique namespace due to ApiException");
+    logger.info(String.format("Got a new namespace called %s", namespace));
 
-    // we can use a standard JUnit assertion to check on the result
-    assertEquals(true, success, "There MUST be a descriptive message here");
+    // Create a service account for the unique namespace
+    final String serviceAccountName = namespace + "-sa";
+    assertDoesNotThrow(
```
markxnelson, line 75: wow, that's a bit awkward.  is there an assertJ assertion we can use instead of this one, that makes the code a bit nicer?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
```diff
@@ -59,18 +65,37 @@ public void testInstallingOperator() {
...
+            .domainNamespaces(Arrays.asList("domainns1", "domainns2"))
+            .serviceAccount("opns1-sa");
 
-    // we can use the injected logger like this:
-    logger.info("hello");
+    //ToDO: use Junit 5 assertions
+    try {
+      installOperator(opParams);
+    } catch (ApiException e) {
+      e.printStackTrace();
```
markxnelson, line 91: put it in the log, not SOP

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
vanajamukkara: will see ...

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItSimpleOperatorValidation.java
vanajamukkara: will do

PR #1524, opened 2020-04-01
Additional mii integration test support for K8s resources

lennyphan:
1. Added support for create/delete of service accounts
2. Changed Kubernetes.deleteNamespace() to use GenericKubernetesApi
3. Added create and delete service account calls to ItSimpleDomainValidation.testCreatingDomain() test.
4. Added dependency on 'io.kubernetes:client-java-extended' module for use of GenericKubernetesApi

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
@@ -155,11 +155,24 @@ public class Kubernetes implements LoggedTest {
     return nameSpaces;
   }
 
+  /**
+   * Delete a namespace for the given name
+   *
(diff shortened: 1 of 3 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -534,7 +544,93 @@ public static boolean deletePvc(String pvcName, String namespace) throws ApiExce
 
     return false;
   }
-  // --------------------------
+
+  // --------------------------- service account ---------------------------
+
+  /**
+   * Create a service account for a given namespace
```
markxnelson, line 551: tiny nitpick - the javadoc suggest that we are providing a namespace as a param, but it is actually embedded in the SA object, i would remove the "for a given namespace" just to avoid confusion

PR #1525, opened 2020-04-02
Missing secret dir fix

jshum2479:
This is a temporary work around for WDT returning 1 instead of 0,  when the model has only credential secrets but no other custom secret, in this case the operator will not mount the directory ```/weblogic-operator/config-overrides-secrets``` and WDT checks whether the directory exists and return 1 for warning if not there.

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -53,6 +53,11 @@ RCU_PASSWORD_CHANGED=5
 SCRIPT_ERROR=255
 
 export WDT_MODEL_SECRETS_DIRS="/weblogic-operator/config-overrides-secrets"
+if [ ! -d ${WDT_MODEL_SECRETS_DIRS} ] ; then
+  # Temporary work around for WDT to avoid non zero return code if the directory doesn't exist
+  mkdir -p /tmp/nosuchdir
+  export WDT_MODEL_SECRETS_DIRS="/tmp/nosuchdir"
+fi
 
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -53,6 +53,11 @@ RCU_PASSWORD_CHANGED=5
 SCRIPT_ERROR=255
 
 export WDT_MODEL_SECRETS_DIRS="/weblogic-operator/config-overrides-secrets"
+if [ ! -d ${WDT_MODEL_SECRETS_DIRS} ] ; then
+  # Temporary work around for WDT to avoid non zero return code if the directory doesn't exist
+  mkdir -p /tmp/nosuchdir
+  export WDT_MODEL_SECRETS_DIRS="/tmp/nosuchdir"
+fi
```
tbarnes-us, line 60: Does it suppress the WARNING if you skip the export all-together?

Does this need to be a temporary work-around?  The  WARNING is distracting in the first place, even if WDT is fixed to return non-zero.   Maybe just remove the comment. 

If answer to both is yes, the change is as sample as:
[ ! -d ${WDT_MODEL_SECRETS_DIRS} ]  && unset WDT_MODEL_SECRETS_DIRS

jshum2479: I requested a fix on WDT to make it INFO instead of WARNING and remove the default `weblogic-operator/config-overrides-secrets`  Once WDT fixes and release, we shouldn't need this all together.

tbarnes-us: OK. That's fine.

PR #1526, opened 2020-04-02
Create Domain types for use by tests

rjeberhard:
These do not depend on product code, but were created looking at that code.

- All non-bean functionality has been removed, e.g. code that figures out default values for a field potentially based on other fields.
- Our custom schema-related annotations have been replaced with standard annotations.
- Several errors in the product code were found and fixed in these copies. :(

Note: I'll do a separate PR against the product code to feed these fixes back.


new-integration-tests/pom.xml
```diff
@@ -73,6 +75,12 @@
             <version>2.10.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.swagger</groupId>
+            <artifactId>swagger-annotations</artifactId>
+            <version>1.6.0</version>
```
ddsharpe, line 81: please use dependency management section for version numbers (like operator parent pom). Please fix the other version number in here while you are there (that slipped through reviews).

new-integration-tests/src/test/java/oracle/weblogic/domain/ClusterStatus.java
```diff
@@ -0,0 +1,156 @@
...
+
+  public Integer getReplicas() {
+    return replicas;
+  }
+
+  public void setReplicas(Integer replicas) {
+    this.replicas = replicas;
+  }
+
+  public ClusterStatus readyReplicas(Integer readyReplicas) {
```
ddsharpe, line 61: Do you want to use fluent API or not?  readyReplicas is, but setReplicas is not.  Please explain, maybe this is intended, and I am just confused.

markxnelson:
is this the generated stuff? seems inconsistent with what we are looking for

rjeberhard:
These aren't generated.  They're just java beans.  Compare with what we have in the product code.

new-integration-tests/src/test/java/oracle/weblogic/domain/ClusterStatus.java
rjeberhard: The types had both, so I kept them.  I'm happy to delete set*

new-integration-tests/src/test/java/oracle/weblogic/domain/AdminServer.java
```diff
@@ -0,0 +1,131 @@
...
+      "If present, every time this value is updated the operator will restart"
+          + " the required servers.")
+  private String restartVersion;
+
+  public AdminServer adminService(AdminService adminService) {
+    this.adminService = adminService;
+    return this;
+  }
+
+  public AdminService getAdminService() {
```
ddsharpe, line 47: Sorry if I am off here, but I am new to the fluent API approach.  From what I read, I don't think they use "get" or "set" for fluent API.  For example, given a field "pepper", the getter for pepper is just pepper(), and the setter is pepper(T newValue).  So in this case, adminService(AdminService x), and adminService().

rjeberhard: oh, I see... okay.  thanks.

PR #1530, opened 2020-04-03
Mii introspect test

jshum2479:
Added coverage for MII in introspectTest.sh


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -576,6 +576,8 @@ function generateMergedModel() {
   ${WDT_BINDIR}/validateModel.sh -oracle_home ${ORACLE_HOME} ${model_list} \
     ${archive_list} ${variable_list}  -domain_type ${WDT_DOMAIN_TYPE}  > ${WDT_OUTPUT}
   ret=$?
+  trace "RETURNING $ret"
+  sleep 5
```
markxnelson, line 580: why sleep?

src/integration-tests/bash/cleanup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
+# Copyright (c) 2018, 2019, Oracle Corporation and/or its affiliates. All rights reserved.
```
markxnelson, line 2: copyright wrong

src/integration-tests/bash/cleanup.sh
```diff
@@ -77,282 +60,128 @@ RESULT_DIR="$RESULT_ROOT/acceptance_test_tmp"
...
-# usage: doDeleteByName [-n foobar] kind name
-function doDeleteByName {
-
-  local tmpfile="/tmp/$(basename $0).doDeleteByName.$PPID.$SECONDS"
-
-  kubectl get "$@" -o=jsonpath='{.items[*]}{.kind}{" "}{.metadata.name}{" -n "}{.metadata.namespace}{"\n"}' > $tmpfile
-
-  # exit silently if nothing to delete
-  [ `cat $tmpfile | wc -l` -eq 0 ] && return
+#!/bin/bash
```
markxnelson, line 70: why a hashbang line?

tbarnes-us: is this a bad merge?

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -123,7 +138,7 @@ function cleanupMajor() {
 
   rm -fr /tmp/introspect
   mkdir -p $test_home || exit 1
-
+  rm -fr ${SCRIPTPATH}/mii/workdir
```
tbarnes-us, line 141: This should use ${test_home}/mii/workdir throughout - same location as the rest of the tests temp files...

tbarnes-us:
I suspect there should not be a `src/integration-tests/introspector/mii/models/archive.zip` .  We're generally only allowed to check in source - not zips.  Plus a zip makes it harder to maintain the source there-in.

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: Oops, removing it

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -99,6 +99,21 @@ export ALLOW_DYNAMIC_CLUSTER_IN_FMW=${ALLOW_DYNAMIC_CLUSTER_IN_FMW:-false}
 # whether this test run is expecting a domain validation error
 export EXPECT_INVALID_DOMAIN=${EXPECT_INVALID_DOMAIN:-false}
 
+DOMAIN_SOURCE_TYPE=${DOMAIN_SOURCE_TYPE:-Image}
```
tbarnes-us, line 102: This should be 'PersistentVolume'.  It's interesting that 'Image' works.  The test does not create an Image in its default path.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -99,6 +99,21 @@ export ALLOW_DYNAMIC_CLUSTER_IN_FMW=${ALLOW_DYNAMIC_CLUSTER_IN_FMW:-false}
...
+export DOMAIN_SOURCE_TYPE=${DOMAIN_SOURCE_TYPE}
+WDT_DOMAIN_TYPE=${WDT_DOMAIN_TYPE:-WLS}
+export WDT_DOMAIN_TYPE=${WDT_DOMAIN_TYPE}
```
tbarnes-us, line 105: WDT_DOMAIN_TYPE only needed for FromModel path.

src/integration-tests/introspector/introspectTest.sh
jshum2479: changed

src/integration-tests/bash/cleanup.sh
jshum2479: actually, it's in the model-in-image branch, but removing it now.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -304,6 +320,11 @@ function deployCustomOverridesConfigMap() {
      ${SCRIPTPATH}/util_subst.sh -g ${filname} ${cmdir}/${bfilname}  || exit 1
   done
 
+  # We don't use overrides for MII
```
tbarnes-us, line 323: If we don't use custom overrides for me, than this function should never be called in the first place.

src/integration-tests/introspector/introspectTest.sh
jshum2479: I changed the yaml template file to have the environment variable

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +420,48 @@ function deployCreateDomainJobPod() {
...
+  export WORKDIR=${SCRIPTPATH}/mii/workdir || exit 1
+  export MODEL_IMAGE_TAG=it || exit 1
+  export MODEL_IMAGE_NAME=model-in-image || exit 1
+
+  docker rmi ${MODEL_IMAGE_NAME}:${MODEL_IMAGE_TAG} --force > /dev/null 2>&1
+
+  trace "Info: Downloading WDT and WIT..."
+
+  ${SOURCEPATH}/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build_download.sh \
+   > ${test_home}/miibuild_download.out 2>&1 || exit 1
```
tbarnes-us, line 441: If this fails, cat the output to stdout, and then exit after that.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +420,48 @@ function deployCreateDomainJobPod() {
...
+
+  trace "Info: Launching WIT to build the image..."
+
+  ${SOURCEPATH}/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build_image_model.sh \
+   > ${test_home}/miibuild_image.out  2>&1  || exit 1
```
tbarnes-us, line 446: If this fails, cat the output to stdout, and then exit after that.

src/integration-tests/introspector/introspectTest.sh
jshum2479: It won't work if it is not called. The operator has default config overrides which requires version.txt.  That's why I only clean up others to avoid duplicating the block and if conditions ..

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -955,7 +1030,9 @@ checkWLVersionChecks
 # Check admin-server pod log and also call on-line WLST to check if
 # automatic and custom overrides are taking effect in the bean tree:
 
-checkOverrides
+if [ "${DOMAIN_SOURCE_TYPE}" != "FromModel" ] ; then
+  checkOverrides
```
tbarnes-us, line 1034: Several of the tested overrides are internal overrides - all of the 'ListenAddress' and 'Log' related ones.

Maybe we should test those on the model path?  

If so, the test needs to be made a little smarter so that internal overrides are tested whilst custom overrides are not.    Perhaps split them into two different "inputt" files - one for internal and one for external.  The combine the files into a temporary file for the PV path, and only copy one of the files over for the Model path.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +420,48 @@ function deployCreateDomainJobPod() {
...
+function createMII_Image() {
+  trace "Info: Create MII Image"
+
+  mkdir -p ${SCRIPTPATH}/mii/workdir/models || exit 1
+  cp ${SCRIPTPATH}/mii/models/*  ${SCRIPTPATH}/mii/workdir/models || exit 1
+  cd ${SCRIPTPATH}/mii/workdir || exit 1
+  export WORKDIR=${SCRIPTPATH}/mii/workdir || exit 1
+  export MODEL_IMAGE_TAG=it || exit 1
+  export MODEL_IMAGE_NAME=model-in-image || exit 1
+
```
tbarnes-us, line 435: When setting up env vars that are very specific to external scripts, consider encasing the entire set of calls into a "(" ... script calls ... ")".  This will make it clear (self documenting) they're scoped only for the external script and will make sure they don't accidentally leak to other calls to the external script in the global scope.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +420,48 @@ function deployCreateDomainJobPod() {
...
+  docker rmi ${MODEL_IMAGE_NAME}:${MODEL_IMAGE_TAG} --force > /dev/null 2>&1
+
+  trace "Info: Downloading WDT and WIT..."
+
+  ${SOURCEPATH}/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build_download.sh \
+   > ${test_home}/miibuild_download.out 2>&1 || exit 1
+
+  trace "Info: Launching WIT to build the image..."
+
+  ${SOURCEPATH}/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build_image_model.sh \
```
tbarnes-us, line 445: Right now, by default, this always builds even if a local image with the given image-name/tag already exists - it should probably set MODEL_IMAGE_BUILD to 'when-missing'.  It's my understanding this is the general directive for all tests.

Separately, I've been asked to change the default of this script to 'when-missing' - but I won't get to that until at least next week.

src/integration-tests/bash/cleanup.sh
jshum2479: changing it

src/integration-tests/introspector/introspectTest.sh
jshum2479: no problem, added now.

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/introspectTest.sh
jshum2479: This will take a little bit more time, I need to take a look of what it is being tested first.

src/integration-tests/bash/cleanup.sh
```diff
@@ -20,34 +20,21 @@
...
 #
 #   FAST_DELETE     Set to "--grace-period=1 --timeout=1" to speedup
 #                   deletes and skip phase 2.
 #
-# Dry run option: 
-#
-#   To show what the script would do without actually doing
-#   any deletes pass "-dryrun" as the first parameter.
-#
 # --------------------
```
tbarnes-us, line 30: Definitely looks like a bad merge or some other git issue.   It looks like your diff here is vs. develop instead of MII.    Develop has  the 'Dry run' stuff - it's the latest and newer than what's in MII.  Where-as the version you changed does not.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: More on this.  It looks like the test is not testing 'model updates'.   E.g. the custom overrides from the test  should be passed in as config map updates _and_ tested to make sure that they actually occurred.   So one alternative would be:

1) Modify the check bean program's 'inputt' file format so that fields that are from internal overrides are tagged to indicate they always come from an internal override.
2) Modify the mbean checker so that it acts differently in FromModel mode and reacts to the tag differently accordingly.   E.g. the tag makes no difference in PV mode, but in FromModel mode, it expects things that don't have the 'IO' internal-override tag to have the expected value in both the live and the static mbean trees, instead of just in the live tree.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: i don't understand.  The operator's 'internal overrides' have no version.txt.  The custom overrides config map has that file, and custom overrides config map should not be set up for the MII path.  In fact, if you add that to the MII domain resource, the domain deployment is expected to fail (since custom overrides aren't supported with MII).

src/integration-tests/introspector/introspectTest.sh
jshum2479: The current test is only one pass, i.e. no update test case.  This is one area that I agreed we should spend some effort into.

src/integration-tests/bash/cleanup.sh
jshum2479: Not sure why, it shouldn't have those now.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: I'm not advocating a second pass.  Just a single pass that includes verification that the settings from the config map took effect, and that verifies the internal overrides took effect.

tbarnes-us:
@markxnelson wrote

> i need to understand why you are proposing such significant changes to cleanup.sh which is used all over the place..

I think this is a bad merge of some kind.  This diff is somehow against the newest version whilst he made changes to an older version.

src/integration-tests/bash/cleanup.sh
tbarnes-us: The diff still looks bad.   'Dry run' should not have been removed.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: Note that the yamlt template files can handle default macros - you can modify them so that `${WDT_DOMAIN_TYPE}` becomes `${WDT_DOMAIN_TYPE:-WLS}`.

src/integration-tests/introspector/introspectTest.sh
jshum2479: Yes, we do `override--version.txtt`  If I don't call this it won't creates it in the location and the introspect failed. 

[2020-04-03T16:27:43.390 UTC][][SEVERE] Error: Exiting test due to domainValid from introspecting domain is false!

@[2020-04-03T16:27:39.862 UTC][introspectDomain.py:277][FINE] Printing file /tmp/introspect/domain1/topology.yaml
>>>  /tmp/introspect/domain1/topology.yaml
domainValid: false
validationErrors:
- "Error, Required file, '/weblogic-operator/config-overrides/version.txt', does not exist"

>>> EOF

src/integration-tests/bash/cleanup.sh
jshum2479: I am comparing with origin/model-in-image and I don't see Dry Run

tbarnes-us: But the diff is showing dryRun - so something is definitely wrong.   This is somehow comparing to the develop branch version instead of the MII version.   Note that the develop branch version is newer.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: Once you do that, you don't need to set WDT_DOMAIN_TYPE globally to something else unless going through the MII path...

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -123,7 +138,7 @@ function cleanupMajor() {
 
   rm -fr /tmp/introspect
   mkdir -p $test_home || exit 1
-
+  rm -fr ${test_home}/mii/workdir
```
tbarnes-us, line 141: You don't need this line now because test_home is /tmp/introspect, and /tmp/introspect is removed two lines above.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: OK. It turns out the test always mounts the config map even though it isn't used, and the introspector takes a different action when it sees this mount.  This is a side effect of the fact that the test doesn't actually run the operator.  I guess this is a fine work-around - we're not trying to emulate the operator perfectly.

tbarnes-us: Resolving.

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/introspectTest.sh
jshum2479: Yes, it's already there.  The configmap is actually coded in a way that works for both non mii case

tbarnes-us: Where's the verification that the listen address, etc, are as expected?

jshum2479: Not for listening address, but the data sources are there.  We can change the override to check the internal ones only for mii

tbarnes-us: Please do.  See my original suggestion at the beginning of this thread.

src/integration-tests/introspector/mii/models/domain1.yaml
```diff
@@ -0,0 +1,105 @@
+domainInfo:
```
markxnelson, line 1: you need copyright and license statements in yaml files
unless there is some technical reason they cannot be here
@ddsharpe i assume WDT can handle comments, right? i believe i have WDT models with comments in them

src/integration-tests/introspector/mii/wdtconfigmap/domain1.10.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
```
markxnelson, line 1: need copyright

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: I don't see any cat after the failure.  It looks like it just exits immediately without a cat of miibuild_download.out

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +418,65 @@ function deployCreateDomainJobPod() {
...
+
+  ${SOURCEPATH}/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build_image_model.sh \
+   > ${test_home}/miibuild_image.out  2>&1
+
+  if [ $? -ne 0 ] ; then
+    trace "Error: createMII_Image: build image failed"
+    cat ${test_home}/miibuild_image.out
+    exit 1
+  fi
+  )
```
tbarnes-us, line 462: Forgot to tell you to add an '|| exit 1' after the close paren. (An exit in a paren scope only exits the paren scope.)

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -393,6 +418,65 @@ function deployCreateDomainJobPod() {
...
+  kubectl -n $NAMESPACE create secret generic  ${DOMAIN_UID}-runtime-encryption-secret \
+        --from-literal=password=welcome1 | tracePipe "Info: kubectl output: "
+
+  kubectl -n $NAMESPACE label secret ${DOMAIN_UID}-runtime-encryption-secret weblogic.domainUID=$DOMAIN_UID || exit 1
+
+}
+
 #############################################################################
 #
 # Run introspection "JobPod", parse its output to files, and put files in a cm
```
tbarnes-us, line 482: Please change ")" to ") || exit 1" at line 646 - this is an old bug unrelated to your current change. Otherwise, if the yaml template substitution fails, then the script will keep on truckin' and fail later rather than at the point of the original failure...

e.g. change:

```
  ( # Generate svc yaml from template
    export SERVER_NAME="${server_name}"
    export SERVICE_INTERNAL_PORT="${internal_port}"
    export SERVICE_EXTERNAL_PORT="${external_port}"
    export SERVICE_NAME=${service_name}
    ${SCRIPTPATH}/util_subst.sh -g wl-nodeport-svc.yamlt ${target_yaml} || exit 1
  )

```

to 

```
  ( # Generate svc yaml from template
    export SERVER_NAME="${server_name}"
    export SERVICE_INTERNAL_PORT="${internal_port}"
    export SERVICE_EXTERNAL_PORT="${external_port}"
    export SERVICE_NAME=${service_name}
    ${SCRIPTPATH}/util_subst.sh -g wl-nodeport-svc.yamlt ${target_yaml} || exit 1
  ) || exit 1
```

tbarnes-us: Thanks for doing that.

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -26,7 +26,16 @@
...
-
+# Usage:
+#     For non Model In Image test
+#
+#         introspectTest.sh
+#
+#     For Model In Image test
+#
+#         export DOMAIN_SOURCE_TYPE=FromModel
+#         intropsectTest.sh
```
tbarnes-us, line 37: typo - introspectTest.sh

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -439,23 +448,27 @@ function createMII_Image() {
 
   docker rmi ${MODEL_IMAGE_NAME}:${MODEL_IMAGE_TAG} --force > /dev/null 2>&1
 
-  trace "Info: Downloading WDT and WIT..."
-
+  trace "Info: Downloading WDT and WIT"
```
tbarnes-us, line 451: trace --> tracen

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -439,23 +448,27 @@ function createMII_Image() {
...
+  local rc=$?
+  printdots_end
+  if [ $rc -ne 0 ] ; then
     trace "Error: createMII_Image: download tools failed"
     cat ${test_home}/miibuild_download.out
     exit 1
   fi
 
-  trace "Info: Launching WIT to build the image..."
+  trace "Info: Launching WIT to build the image"
```
tbarnes-us, line 463: trace -> tracen

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -26,7 +26,16 @@
 #   IMAGE_NAME, etc, all have defaults, or can be passed in.  See the 'export'
 #   calls in the implementation below for the complete list.
 #
-
+# Usage:
```
tbarnes-us, line 29: Please update README file with this information too.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: Looks good now.

src/integration-tests/introspector/mii/models/domain1.yaml
jshum2479: Fixed missing copyrights

src/integration-tests/introspector/mii/wdtconfigmap/domain1.10.yaml
jshum2479: Fixed missing copyrights

src/integration-tests/introspector/introspectTest.sh
jshum2479: changed

src/integration-tests/introspector/introspectTest.sh
jshum2479: changed

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/README
```diff
@@ -19,8 +19,11 @@ Usage:
     One way to do this when running locally is to run the integration tests.
     (Note that this test doesn't depend on Operator image as of 2018/10/16, but may in the future.)
 (2) Optionally specify values for input env vars (see introspectTest.sh for the list).
-(3) Run introspectTest.sh
-
+(3) For non Model In Image domain type.
+        Run introspectTest.sh
+    For Model In Image domain type.
+    export DOMAIN_SOURCE_TYPE=FromModel
+    Run introspectTest.sh
```
tbarnes-us, line 26: Indent lines 25 and 26 four more spaces.  Right now they line up right under 'For Model'.   Change 'type.' to 'type:'.   Change 'For ' to 'For a '.

src/integration-tests/introspector/introspectTest.sh
jshum2479: done

src/integration-tests/introspector/README
jshum2479: done

src/integration-tests/introspector/checkMIIBeans.inputt
```diff
@@ -0,0 +1,27 @@
+/Log/base_domain,FileName,logs/base_domain.log,/shared/logs/base_domain.log
```
markxnelson, line 1: is this file able to have a copyright notice added?

src/integration-tests/introspector/mii/wdtconfigmap/domain1.10.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
```
markxnelson, line 1: is this a new file? if so, just 2020

src/integration-tests/introspector/mii/models/model1.properties
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
```
markxnelson, line 1: is this a new file? if so, just 2020

src/integration-tests/introspector/mii/models/domain1.yaml
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
```
markxnelson, line 1: is this a new file? if so, just 2020

src/integration-tests/introspector/mii/wdtconfigmap/domain1.10.yaml
jshum2479: updated

src/integration-tests/introspector/mii/models/model1.properties
jshum2479: updated

src/integration-tests/introspector/mii/models/domain1.yaml
jshum2479: updated

src/integration-tests/introspector/checkMIIBeans.inputt
jshum2479: This one is  not same as the checkBeans.inputt

markxnelson: sorry, i don't understand, can it have a copyright notice added please? if not, why not?

jshum2479: This data file is a copy of the original one (also without copyrights) but looking at the checkBeans.py actually it can take comments, so I am adding copyrights to both

PR #1533, opened 2020-04-03
removing commented/incomplete test code

vanajamukkara:


markxnelson:
why do we not want this test? i might have missed something

PR #1534, opened 2020-04-03
Add more contents into MII image plus cleanup

doxiao:
Add WDT binary and a WDT model into the first test MII image, and tested the generated image using the sample.
Moved the constants in the install and WIT actions into a new constants class ; the impl class static import them.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
```diff
@@ -3,24 +3,45 @@
...
   @DisplayName("Create a MII image")
   public void testCreatingMIIImage() {
 
-    // create the MII image
-    // TODO add model files and other contents to the image once we have those resources
-    boolean success = TestActions.createMIIImage(withWITParams());
+    logger.info("WDT model directory is " + MODEL_DIR);
+
+    // build the model file list
+    List<String> modelList = Arrays.asList(MODEL_DIR + "/" + WDT_MODEL_FILE);
```
markxnelson, line 32: @ddsharpe is this a WDT thing, that it needs to be a forward slash, or should this be File.separator ?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
```diff
@@ -3,24 +3,45 @@
...
 
+    // build an image using WebLogic Image Tool
+    boolean success = TestActions.createMIIImage(
```
markxnelson, line 35: please static import TestActions so you can just say `createMIIImage()` instead of `TestActions.createMIIImage()`

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
```diff
@@ -3,24 +3,45 @@
...
+        withWITParams()
+            .modelImageName(IMAGE_NAME)
+            .modelImageTag(IMAGE_TAG)
+            .modelFiles(modelList)
+            .wdtVersion("latest")
+            .redirect(false));
+ 
     assertEquals(true, success, "Failed to create the image using WebLogic Deploy Tool");
+    TestAssertions.imageExists(IMAGE_NAME, IMAGE_TAG);
```
markxnelson, line 44: pleas static import so you don't need `TestAssertions.` here

markxnelson: `imageExists()` could probably use a better name too - maybe `assertImageExists`

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.weblogic.kubernetes.actions;
+
+public interface ActionConstants {
+
+  // Work directory for the integration test suite
+  public static final String WORK_DIR 
+      = System.getProperty("java.io.tmpdir") + "/it-results";
```
markxnelson, line 10: needs a null check

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public static final String DEFAULT_MODEL_IMAGE_NAME = "test-mii-image";
+  public static final String DEFAULT_MODEL_IMAGE_TAG  = "v1";
+  
+  // ------------ WebLogic Image Tool constants----------------------------
+  public static final String WIT = "WIT";
+  public static final String WDT = "WDT";
+
+  public static final String WIT_DOWNLOAD_URL 
+      = "https://github.com//oracle/weblogic-image-tool";
```
markxnelson, line 37: typo - two slashes

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String WIT_VERSION    = "release-1.8.3";
+  public static final String WIT_FILE_NAME  = "imagetool.zip";
+
+  public static final String WDT_DOWNLOAD_URL 
+      = "https://github.com//oracle/weblogic-deploy-tooling";
```
markxnelson, line 42: typo

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallParams.java
```diff
@@ -53,7 +51,7 @@ public InstallParams version(String version) {
 
   public String version() {
     if (version == null) {
-      return WIT_TYPE.equals(type) ? DEFAULT_WIT_VERSION : DEFAULT_WDT_VERSION;
+      return WIT.equals(type) ? WIT_VERSION : WDT_VERSION;
```
markxnelson, line 54: this is ok, but i had to stop and think to work out what you meant here.
i know this is uglier:
``` 
if (type == WIT) {
  return WIT_VERSION;
} else if (type == WDT) {
  return WDT_VERSION;
} else {
 // do something!!!
}
```

but it is much easier to read and immediately understand what you meant, plus it is more explicit about what do to if type is not WIT or WDT ..

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
```diff
@@ -42,17 +36,17 @@ protected void checkDirectory(String dir) {
   }
 
   /**
-   * Check if the required directoies exist.
-   * Currently the directies will be created if missing. We may remove this function
-   * once we have all required working directies pre-created.
+   * Check if the required directories exist.
+   * Currently the directories will be created if missing. We may remove this function
```
markxnelson, line 40: the code does not appear to create any directories?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
```diff
@@ -42,17 +36,17 @@ protected void checkDirectory(String dir) {
...
+   * once we have all required working directories pre-created.
    *
    * @param dir the directory that needs to be checked
    */
   protected void checkFile(String fileName) throws FileNotFoundException {
     File file = new File(fileName);
     if (!file.exists()) {
```
markxnelson, line 47: do we need to check that it is a file, not a directory or something else?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
```diff
@@ -42,17 +36,17 @@ protected void checkDirectory(String dir) {
...
-      logger.warning("The expected file \" + fileName + \" not found.");
-      throw new FileNotFoundException("The expected file \" + fileName + \" not found.");
+      logger.warning("The expected file " + fileName + " not found.");
```
markxnelson, line 48: "was not found"

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -5,17 +5,19 @@
...
 import static oracle.weblogic.kubernetes.extensions.LoggedTest.logger;
 
+
 /**
  *  Implementation of actions that download/install tools for the uses to use.
  *  NOTE: This class is a temporary solution, and may go away once we eventually
  *  install everything before the Java test starts to run.
  */
 
 public class Installer extends InstallWITCommon {
```
markxnelson, line 19: i think we should try to avoid inheritance as much as possible

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -35,25 +37,25 @@ public boolean download() {
...
     if (params.verify()
-        && new File(downloadDir, params.fileName()).exists()) {
+        && new File(DOWNLOAD_DIR, params.fileName()).exists()) {
       logger.info("File " + params.fileName() + " already exists.");
     } else {
       checkDirectory(WORK_DIR);
       downloadResult = executeAndVerify(buildDownloadCommand(), params.direct());
     }
-    if (!(new File(downloadDir + "../imagetool/imagetool.sh").exists()) 
+    if (!(new File(IMAGE_TOOL).exists())
```
markxnelson, line 46: i am not sure what this is meant to do.
are you trying to say - if the file DOES exist, then unzip it?
or if it DOES NOT exist, and you asked me to unzip, then unzip it?
seems we could write this with code that is longer, but much clearer

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -35,25 +37,25 @@ public boolean download() {
...
         && params.unzip()) {
-      unzipResult = unzip(downloadDir, params.fileName(), downloadDir + "/..");
+      unzipResult = unzip();
     }
     return downloadResult && unzipResult;
   }
 
-  private boolean unzip(String path, String fileName, String targetDir) {
+  private boolean unzip() {
     String command =
```
markxnelson, line 54: I think `String.format()` would make this easier to read

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -67,6 +69,6 @@ private String buildDownloadCommand() {
     
     return "curl -fL "
```
markxnelson, line 70: i think this whole method could be better written as 
```
String command = String.format("...", a, b, c, d);
return command;
```

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
```diff
@@ -55,7 +59,21 @@ public boolean updateImage() {
     } 
 
     try {
+      // check if the WIT binary exists, throws FileNotFoundException if the file does not exist
```
markxnelson, line 62: didn't you literally just download it and check it was there a few lines ago? 
do we need another check?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
```diff
@@ -124,4 +142,35 @@ private String buildList(List<String> list) {
...
+   * @return true if the command succeeds 
+   */
+  public boolean addInstaller() throws FileNotFoundException {
+    try {
+      checkFile(WDT_ZIP_PATH);
+    } catch (FileNotFoundException fnfe) {
+      logger.warning("Failed to create an image due to Exception: " + fnfe.getMessage());
+      throw fnfe;
+    }
+    return executeAndVerify(
```
markxnelson, line 157: much better to build the string first, then return it - easier to debug

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
```diff
@@ -124,4 +142,35 @@ private String buildList(List<String> list) {
...
+        + " --version " + params.wdtVersion()
+        + " --path " + WDT_ZIP_PATH);
+  }
+  
+  /**
+   * Delete the WDT installer cache entry from the WebLogic Image Tool
+   * @return true if the command succeeds
+   */
+  public boolean deleteEntry() {
+    return executeAndVerify(
```
markxnelson, line 170: same

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -112,5 +113,15 @@ public static boolean adminT3ChannelAccessible(String domainUID, String namespac
...
     return Domain.adminNodePortAccessible(domainUID, namespace);
   }
+  
+  /**
+   * Check if a Docker image exists.
+   * @param imageName - the name of the image to be checked
+   * @param imageTag  - the tag of the image to be checked
+   * @return true if the image does exist, false otherwise
+   */
+  public static boolean imageExists(String imageName, String imageTag) {
```
markxnelson, line 123: maybe call this `doesDockerImageExist` so it is clearer what it does

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/WITAssertion.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return true if the image does exist, false otherwise
+   */
+  public static boolean imageExists(String imageName, String imageTag) {
+    logger.info("Checking if image " + imageName + ":" + imageTag + " exists.");
+    // verify the docker image is created
+    try {
+      ExecResult result = ExecCommand.exec(
+          "docker images | grep " 
+          + imageName
+          + " | grep imageTag "
```
markxnelson, line 30: imageTag is a variable right? not a string literal?

markxnelson: why not grep for `imageName + ":" + imageTag`

markxnelson: or even better:
`docker images -q imageName:imageTag` which will return a single line with the image hash tag if it exists - no nothing if the image does not exist

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: this should be File.separator.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
doxiao: I assumed this system property would never be null. I don't know what to do if it actually returns null. 
I am using this as a temporary solution for a to-be-added script that runs upfront to setup necessary environment for the integration test suite. That script would set the WORK_DIR to an appropriate directory according to the platform where the integration test runs. Once we have that script in place, we don't need this anymore.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
doxiao: good point. I need to check if it is indeed a file.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: This is try not to unzip it more than once.  I'll make it clearer.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: String.format does make it easier to read and less error-prone.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
doxiao: Although we have java code doing the download and unzip for now, the plan is to use a script to download all installers upfront before the Java test starts. We'll probably need this check then.  I could remove this check now,  and add it back when/if we get rid of the java Installer code.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallParams.java
doxiao: will change to 
return WIT.equals(type) ? WIT_FILE_NAME : WDT_FILE_NAME;

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: yes. Changed the code to use String.format().

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: Done.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: Made the change to use static import and rename the method.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
doxiao: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
doxiao: removed the check

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
doxiao: Done

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
doxiao: done

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
doxiao: changed the method name.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/WITAssertion.java
doxiao: changed the code to use docker images -q imageName:imageTag.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
doxiao: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
doxiao: fixed the javadoc

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallWITCommon.java
doxiao: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: Moved the common methods into FileUtils and a new action: Command and CommandParams.

 InstallWITCommon class is removed. We can also get rid of ActionImplCommon once Vanaj's code in PR#1535 is updated.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: Modified the code.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: done

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
ddsharpe: Line 32 doesn't seem correct.  string concatenation (so one string), wrapped with Arrays.asList(), why?  If its just one model and not a list, wouldn't you prefer to use Collections.singletonList()?  And yes, this slash should be File.separator.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
ddsharpe: I am okay without having a null check here.  If this fails, the build should fail.  Because this is in the test code, the NPE stacktrace is sufficient to debug it, in my opinion.  Shouldn't this "/" be File.separator also?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String DEFAULT_MODEL_IMAGE_NAME = "test-mii-image";
+  public static final String DEFAULT_MODEL_IMAGE_TAG  = "v1";
+  
+  // ------------ WebLogic Image Tool constants----------------------------
+  public static final String WIT = "WIT";
+  public static final String WDT = "WDT";
+
+  public static final String WIT_DOWNLOAD_URL 
+      = "https://github.com/oracle/weblogic-image-tool";
+  public static final String WIT_VERSION    = "release-1.8.3";
```
ddsharpe, line 38: I am okay to leave this here for now, but I think it would be easier to manage these versions from a property in the maven pom.xml.  Like <properties> <wit-version>release-1.8.3</wit-version> </properties> in the pom.xml for this test module.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String WIT_FILE_NAME  = "imagetool.zip";
+
+  public static final String WDT_DOWNLOAD_URL 
+      = "https://github.com/oracle/weblogic-deploy-tooling";
+  public static final String WDT_VERSION    = "weblogic-deploy-tooling-1.7.2";
```
ddsharpe, line 43: same comment as wit version (above)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallParams.java
ddsharpe: Where there are this many "if" checks on type, it makes me think this should be two separate classes (that could extend the same base/interface).  It would also make it easier to read :)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: yes, right now, the test only has one model. but the code tries to show how to have one or more models in one image.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
rjeberhard: java.io.tmpdir is guaranteed to have a value by the System.getProperty javadoc.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/ActionImplCommon.java
```diff
@@ -4,9 +4,8 @@
 package oracle.weblogic.kubernetes.actions.impl;
 
 // import java.io.BufferedReader;
-
-import java.io.IOException;
 // import java.io.InputStreamReader;
```
rjeberhard, line 7: Can you please remove commented out imports?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -5,24 +5,42 @@
...
   /**
    * Set up the installer with given parameters
    * @return the installer instance 
    */
-  public Installer with(InstallParams params) {
+  public static Installer withParams(InstallParams params) {
+    return new Installer().with(params);
+  }
+  
+  private Installer with(InstallParams params) {
```
rjeberhard, line 43: Fluent style would have this method just be named params

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
```diff
@@ -55,30 +65,45 @@ public boolean updateImage() {
     } 
 
     try {
-      checkFile(IMAGE_TOOL);
+      // delete the old cache entry for the WDT installer
+      if (!deleteEntry()) {
+        logger.warning("Failed to delete cache entry in WebLogic Image Tool");
```
rjeberhard, line 70: Why warnings?  These should either be asserts that fail the test or they aren't important.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
```diff
@@ -124,4 +149,46 @@ private String buildList(List<String> list) {
...
+  
+  /**
+   * Add WDT installer to the WebLogic Image Tool cache
+   * @return true if the command succeeds 
+   */
+  public boolean addInstaller() throws FileNotFoundException {
+    try {
+      checkFile(WDT_ZIP_PATH);
+    } catch (FileNotFoundException fnfe) {
+      logger.warning("Failed to create an image due to Exception: " + fnfe.getMessage());
```
rjeberhard, line 161: Same.  If the file isn't found then this should fail the test.  I don't want to have to search the test logs for warnings to figure out why something later failed.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -152,5 +153,15 @@ public static boolean adminT3ChannelAccessible(String domainUID, String namespac
   public static boolean adminNodePortAccessible(String domainUID, String namespace) {
     return Domain.adminNodePortAccessible(domainUID, namespace);
   }
+  
+  /**
+   * Check if a Docker image exists.
+   * @param imageName - the name of the image to be checked
```
rjeberhard, line 159: Don't need dashes in javadoc.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -152,5 +153,15 @@ public static boolean adminT3ChannelAccessible(String domainUID, String namespac
...
+   * @param imageTag  - the tag of the image to be checked
+   * @return true if the image does exist, false otherwise
+   */
+  public static boolean assertDockerImageExists(String imageName, String imageTag) {
```
rjeberhard, line 163: Don't need "assert" in name.  This can just be doesDockerImageExist()

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/WITAssertion.java
```diff
@@ -0,0 +1,48 @@
...
+      if (Integer.parseInt(result.stdout().trim()) != 1) {
+        return false;
+      }
+    } catch (Exception e) {
+      logger.warning("Failed to check if Docker image " 
+          + imageName 
+          + ":" 
+          + imageTag
+          + "exists due to Exception: " 
+          + e.getMessage());
```
rjeberhard, line 43: This loses the stack trace. @ddsharpe, do we have a pattern for this?  That should at least include using the variant of warning() that takes the exception and not just the message.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
markxnelson: ok - ignore me :)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/ActionImplCommon.java
doxiao: It is supposed to go with the commented out ProcessBuilder code below. I'll remove them together.

doxiao: Note that this whole class will be removed once other parts of the code sync up with this PR.  The methods here have been moved into Command action.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: I named it with() to indicate that this returns an Installer instance with some parameters. Name params() might give readers an impression that this returns a params instance. But I am new to fluent style and my understanding might be wrong.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/WebLogicImageTool.java
doxiao: This is part of test infra, not a test. The pattern we have been following is for the infra part to return success or not. The test code, such as ItWITValication, will then fail the test with assertions. But my understanding might be wrong here.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
doxiao: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
doxiao: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/ActionImplCommon.java
doxiao: done

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/WITAssertion.java
doxiao: converted to an assertion.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
doxiao: changed the code to get the version from pom.xml.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/ActionConstants.java
doxiao: done

doxiao:
@markxnelson @rjeberhard @ddsharpe I believe all comments have been addressed. Please let me know if I missed anything.

PR #1536, opened 2020-04-03
OWLS-80829 - Fix ConcurrentModificationException in JobWatcher and DomainStatus

alai8:
Fixes `ConcurrentModificationException` from `JobWatcher.getOrCreateFor()` as described in OWLS-80829

Also fixes `ConcurrentModificationException` from `DomainStatus.getServers()` and `getClusters()` in the `Array sort()` method calls.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
@@ -275,11 +275,12 @@ public class DomainStatus {
   /**
    * Status of WebLogic servers in this domain.
    *
-   * @return servers
+   * @return servers sorted list of ServerStatus containing status of WebLogic servers in this domain
    */
   public List<ServerStatus> getServers() {
-    servers.sort(Comparator.naturalOrder());
-    return servers;
+    List<ServerStatus> sortedServers = new ArrayList<>(servers);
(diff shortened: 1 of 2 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -275,11 +275,12 @@ public DomainStatus withReplicas(Integer replicas) {
   /**
    * Status of WebLogic servers in this domain.
    *
-   * @return servers
+   * @return servers sorted list of ServerStatus containing status of WebLogic servers in this domain
```
markxnelson, line 278: nitpick: you don't name the return value (which is what i assume "servers") is, you just explain it

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -326,9 +327,15 @@ public DomainStatus addServer(ServerStatus server) {
     return this;
   }
 
+  /**
+   * Status of WebLogic clusters in this domain.
+   *
+   * @return clusters sorted list of ClusterStatus containing status of WebLogic clusters in this domain
```
markxnelson, line 333: same

PR #1537, opened 2020-04-04
WIP: Mark's sample tests (DO NOT MERGE)

markxnelson:
This is just a fake PR to hold some sample tests to demonstrate various patterns.

### DIFF
--- a/new-integration-tests/pom.xml
+++ b/new-integration-tests/pom.xml
@@ -44,6 +44,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.15.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
             <version>4.0.2</version>
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
@@ -38,7 +38,7 @@ class ItMark implements LoggedTest {
       result = createSecret("bob", // name
           "weblogic", // username
           "welcome1", // password
-          "default");
+          "default"); // namespace
     } catch (Exception e) {
       logger.info("result is " + result);
       assertThat(e)
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
@@ -11,7 +11,6 @@ import org.junit.jupiter.api.Test;
 
 import static oracle.weblogic.kubernetes.actions.TestActions.createSecret;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.catchThrowable;
 
 @DisplayName("Simple validation of basic domain functions")
@@ -34,14 +33,24 @@ class ItMark implements LoggedTest {
   public void testThereWasNoException() {
     // this test will fail because it does throw an exception!
 
-    assertThatCode(() ->
-        createSecret("bob", // name
-            "weblogic", // username
-            "welcome1", // password
-            "default")) //namespace
-        .as("Test that createSecret does not throw an exception")
-        .withFailMessage("OMG! createSecret() threw an unexpected exception")
-        .doesNotThrowAnyException();
+    boolean result = false;
+    try {
+      result = createSecret("bob", // name
+          "weblogic", // username
+          "welcome1", // password
+          "default");
+    } catch (Exception e) {
+      logger.info("result is " + result);
+      assertThat(e)
+          .as("Test that createSecret does not throw an exception")
+          .withFailMessage("OMG! createSecret() threw an unexpected exception")
+          .isNotInstanceOf(Exception.class);
+    }
+    assertThat(result)
+        .as("Test that createSecret returns true")
+        .withFailMessage("OMG! createSecret() did not return true!")
+        .isTrue();
+
   }
 
   @Test
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
@@ -16,7 +16,7 @@ import static org.assertj.core.api.Assertions.catchThrowable;
 
 @DisplayName("Simple validation of basic domain functions")
 @IntegrationTest
-class Mark implements LoggedTest {
+class ItMark implements LoggedTest {
 
   @Test
   @DisplayName("Test something")
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/Mark.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/Mark.java
@@ -1,0 +1,70 @@
+// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.weblogic.kubernetes;
+
+import io.kubernetes.client.openapi.ApiException;
+import oracle.weblogic.kubernetes.annotations.IntegrationTest;
+import oracle.weblogic.kubernetes.extensions.LoggedTest;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static oracle.weblogic.kubernetes.actions.TestActions.createSecret;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+@DisplayName("Simple validation of basic domain functions")
+@IntegrationTest
+class Mark implements LoggedTest {
+
+  @Test
+  @DisplayName("Test something")
+  public void testSomeStuff() {
+    int x = 1 + 5;
+    logger.info("log message");
+
+    assertThat(true).as("Test that true is true")
+        .withFailMessage("OMG! True was not true")
+        .isEqualTo(true);
+  }
+
+  @Test
+  @DisplayName("Test something else")
+  public void testThereWasNoException() {
+    // this test will fail because it does throw an exception!
+
+    assertThatCode(() ->
+        createSecret("bob", // name
+            "weblogic", // username
+            "welcome1", // password
+            "default")) //namespace
+        .as("Test that createSecret does not throw an exception")
+        .withFailMessage("OMG! createSecret() threw an unexpected exception")
+        .doesNotThrowAnyException();
+  }
+
+  @Test
+  @DisplayName("Test catching an exception")
+  public void testCatchingAnException() {
+    // this test will pass because it does throw an exception of
+    // the expected type, and with the expected message
+
+    // when i call createSecret()
+    Throwable thrown = catchThrowable(() ->
+        createSecret("bob", // name
+            "weblogic", // username
+            "welcome1", // password
+            "default"));
+
+    // then
+    assertThat(thrown)
+        .as("Test that createSecret does not throw an APIException")
+        .withFailMessage("OMG! createSecret() threw an APIException")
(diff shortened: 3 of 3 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/Mark.java
```diff
@@ -0,0 +1,70 @@
...
+            "welcome1", // password
+            "default")) //namespace
+        .as("Test that createSecret does not throw an exception")
+        .withFailMessage("OMG! createSecret() threw an unexpected exception")
+        .doesNotThrowAnyException();
+  }
+
+  @Test
+  @DisplayName("Test catching an exception")
+  public void testCatchingAnException() {
```
tbarnes-us, line 49: Looks like Junit 5's default display name generator can be replaced so as to make method/class names look pretty in the logging.  [Here's a camel case one](https://www.baeldung.com/junit-custom-display-name-generator) that parses the method name and would generate something that closely matches the custom display name you had to type in by hand here.

tbarnes-us: (That said I'd personally prefer just to see the method name and a timestamp.)

markxnelson: That's a fair point @tbarnes-us -- the Jenkins plugins that I use ignore the display name anyway:

Screenshot from 2020-04-03 23-22-00


Though the output does have both: 
```
[INFO] Running oracle.weblogic.kubernetes.ItMark
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest beforeEachTest> <About to execute [Test catching an exception] in oracle.weblogic.kubernetes.ItMark.testCatchingAnException()>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.Timing afterTestExecution> <Method [testCatchingAnException] took 223 ms.>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest afterEachTest> <Finished executing [Test catching an exception] in oracle.weblogic.kubernetes.ItMark.testCatchingAnException()>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest beforeEachTest> <About to execute [Test something] in oracle.weblogic.kubernetes.ItMark.testSomeStuff()>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.ItMark testSomeStuff> <log message>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.Timing afterTestExecution> <Method [testSomeStuff] took 2 ms.>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest afterEachTest> <Finished executing [Test something] in oracle.weblogic.kubernetes.ItMark.testSomeStuff()>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest beforeEachTest> <About to execute [Test something else] in oracle.weblogic.kubernetes.ItMark.testThereWasNoException()>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.Timing afterTestExecution> <Method [testThereWasNoException] took 170 ms.>
<04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.LoggedTest afterEachTest> <Finished executing [Test something else] in oracle.weblogic.kubernetes.ItMark.testThereWasNoException()>
[ERROR] Tests run: 3, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 0.417 s <<< FAILURE! - in oracle.weblogic.kubernetes.ItMark
[ERROR] oracle.weblogic.kubernetes.ItMark.testCatchingAnException  Time elapsed: 0.226 s  <<< FAILURE!
java.lang.AssertionError: [Test that createSecret does not throw an APIException] OMG! createSecret() threw an APIException
	at oracle.weblogic.kubernetes.ItMark.testCatchingAnException(ItMark.java:65)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
     ...
```

Honestly, I just put them in there in the first place because I was trying out various JUnit features, and then they got copied.  

In another project, where we use different tools, those types of names are more useful.. I dropped an image in slack to show you what I mean,

I am ok to drop them here, or leave them in, as the team feels best.  I don't think they hurt any - you get both the class/method and the display name in the output.

tbarnes-us: Fine by me. It's nice to see the default is generally to show the timestamp/class-name/method-name _plus_ the display name, if any...

A couple of side nits:

1.  I find it a little annoying that the exception reported by [ERROR] is not also included in-line in the logging exactly when it happens, and the exception that is reported doesn't include a timestamp.  This makes it more work to figure out exactly where and when a failure occurred.

2. It'd be nice if the 'took 2 ms' log message included the full path.  
  - Right now it's:
     - <04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.Timing afterTestExecution> <Method [testSomeStuff] took 2 ms.>
  - but would be nicer if it was:
     - <04-04-2020 03:20:55> <INFO> <oracle.weblogic.kubernetes.extensions.Timing afterTestExecution> <Method [oracle.weblogic.kubernetes.ItMark.testSomeStuff] took 2 ms.>
  - or even better, reduce the logging verbosity and append the 'took 2ms' at the end of the 'Finished executing' log message instead of having an entire separate log message for 'took'.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+@DisplayName("Simple validation of basic domain functions")
+@IntegrationTest
+class ItMark implements LoggedTest {
+
+  @Test
+  @DisplayName("Test something")
```
tbarnes-us, line 22: Another random nit, unrelated to this check-in, but still useful, would be a built-in 'fail fast' option.  This is useful for debugging.

The old operator tests don't seem to have one - at least I couldn't get one from QA, or just plain forgot, and maybe was a bit too lazy to spend much time researching one.  This forced me to hand code a temporary System.exit on failure, or to sit there, watch the test run, and ^C in order to stop the test and leave all resources up/running and debuggable at the point of failure.

Maybe JUnit 5 has one?  I couldn't find one.  I did find [this](http://www.jeeatwork.com/?p=246) after some googling, which lets you replace the JUnit failure handler and have it call System.exit.  If JUnit5 doesn't have something built in, we could modify this solution slightly and only have it activate itself if some easy-to-pass-in global option comes from the command line.

markxnelson: That's an interesting point- and one where i think a big global switch would make sense.  I understand if you are running the suite on your machine, and it fails, you would like it to stop immediately so you can go look at the environment.  But at the same time, when I am running it in Jenkins, I want it to go all the way through and run all the tests no matter what. 
Interestingly Ginkgo, which we use on other projects, has a nice global switch called `-keepGoing` that provides exactly that behavior!

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
```diff
@@ -0,0 +1,70 @@
...
+        .isEqualTo(true);
+  }
+
+  @Test
+  @DisplayName("Test something else")
+  public void testThereWasNoException() {
+    // this test will fail because it does throw an exception!
+
+    assertThatCode(() ->
+        createSecret("bob", // name
```
vanajamukkara, line 38: if createSecret returns a string, how to capture the return value?

vanajamukkara: I am looking to write assertions for createUniqueNamespace() call, I need to capture the return value.

markxnelson: I pushed an update that does that.  It is not quite so "pretty" - but it does work, and it is easy enough to read it and understand what it means

vanajamukkara: currently I have a mix of Junit assertions and assertJ in ItSimpleOperatorValidation.java, thought I would use assertJ for all, but this looks like adding more code if I want to replace assertDoesNotThrow calls.

markxnelson: that's ok - whatever is the most readable is fine

vanajamukkara: Ok. I modified the code, I have like this 
```
    String opNamespace = null;
    try {
      opNamespace = createUniqueNamespace();
    } catch (Exception e) {
      assertThat(e)
          .as("Test that createUniqueNamespace does not throw an exception")
          .withFailMessage("createUniqueNamespace() threw an unexpected exception")
          .isNotInstanceOf(Exception.class);
    }
```
later I use this variable to delete the namespace using lambda - 
```
      assertThatCode(
          () -> deleteNamespace(opNamespace))
          .as("Test that deleteNamespace doesn not throw an exception")
          .withFailMessage("deleteNamespace() threw an exception")
          .doesNotThrowAnyException();
```
 It gives error 
```
local variables referenced from a lambda expression must be final or effectively final
```
so, I can't use lambda here also as I have non-final variable but use try catch block.

markxnelson: don't you need to name the variable before the `-?` to pass it in?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
```diff
@@ -0,0 +1,79 @@
...
+    // this test will fail because it does throw an exception!
+
+    boolean result = false;
+    try {
+      result = createSecret("bob", // name
+          "weblogic", // username
+          "welcome1", // password
+          "default"); // namespace
+    } catch (Exception e) {
+      logger.info("result is " + result);
```
tbarnes-us, line 43: Why not 'catch Throwable?'

markxnelson: scroll down...

tbarnes-us: Which line number in this method handles a runtime/error?

new-integration-tests/pom.xml
```diff
@@ -43,6 +43,12 @@
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.15.0</version>
```
rjeberhard, line 49: We are using dependency management lower in the pom.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
rjeberhard, line 43: Two problems: result can't be set at this line because we've caught an exception and setting result was the last step in the try block, and the stack trace of the exception is lost.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMark.java
```diff
@@ -0,0 +1,79 @@
...
+      assertThat(e)
+          .as("Test that createSecret does not throw an exception")
+          .withFailMessage("OMG! createSecret() threw an unexpected exception")
+          .isNotInstanceOf(Exception.class);
```
TheFrogPad, line 47: Unless I'm reading something incorrectly, you are asserting the Exception is not an Exception to cause a failure when the test action itself throws an exception?

If this is the case why can't the test writer simply assert a failure directly? I did a quick search and found that from https://assertj.github.io/doc/#assertj-core-assertions-guide there is:
```
import static org.assertj.core.api.Assertions.fail;
import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
```
So here one could write:
```
  fail("OMG! createSecret() threw an unexpected exception", e)
```
I think this may also address Ryan's point and potentially Tom's since the exception is a failure?

PR #1550, opened 2020-04-09
Add app archive to MII image 

doxiao:
Add AppBuilder action and related params class for building app archives.
Add the sample_app to the existing mii image generation and tested it using the sample environment.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
```diff
@@ -12,20 +12,28 @@
...
 
 @DisplayName("Simple validation of basic WIT functions")
 class ItWITValidation implements LoggedTest {
   private static final String WDT_MODEL_FILE = "model1-wls.yaml";
   private static final String IMAGE_NAME = "test-mii-image-2";
   private static final String IMAGE_TAG = "v1";
   
+  // Name of the directory under resources/apps for an application
+  // Note: use "_" instead of "-" in app directories names because
+  // WDT does not like "-" to be in the name of an archive file
```
rjeberhard, line 34: @ddsharpe, is there a bug for this?  Generally, I don't want to obscure problems by having the tests workaround something that doesn't work.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
```diff
@@ -34,7 +42,21 @@ public void testCreatingMIIImage() {
 
     // build the model file list
     List<String> modelList = Collections.singletonList(MODEL_DIR + "/" + WDT_MODEL_FILE);
-
+    
+    // build an application archive using what is in resources/apps/APP_NAME
+    boolean archiveBuilt = buildAppArchive(
+        withAppParams()
```
rjeberhard, line 48: It feels like the use of withAppParams() is disguising what is being built and tested.  I'll look below to see what this method does, but it's not clear what these statements do.

doxiao: withAppParams() method is a way of getting an instance of params without using the impl directly in the test code. The particular one just returns a new instance.

doxiao: Will it help if I add more explanation in the javadoc?

markxnelson: this feels too magical - i assume it is setting some defaults, we should either not do that, or change the name to something like `withDefaultAppParams()` so it is clear what is happening

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+        () -> { 
+        cleanupDirectory(ARCHIVE_DIR);
+        checkDirectory(ARCHIVE_SRC_DIR);
+        copyFolder(
+            APP_DIR + "/" + params.srcDir(), 
+            ARCHIVE_SRC_DIR);
+        });
+        
+    assertThat(throwable)
+        .as("Create directory " + ARCHIVE_DIR)
```
markxnelson, line 65: these are not accurate -- that closure is doing more than this
please make the descriptions as accurate as possible

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+      boolean jarBuilt = buildJarArchive(jarPath, ARCHIVE_SRC_DIR);
+      assertThat(jarBuilt)
+          .as("Create app ear file " + jarPath)
+          .withFailMessage("Failed to create the app ear file " + jarPath) 
+          .isTrue();
+    } catch (Exception e) {
+      exception = e;
+    }
+    
+    assertThat(exception)
```
markxnelson, line 82: include the exception in the fail message, so we know why it failed

markxnelson: applies to multiple places

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
```diff
@@ -55,4 +63,55 @@ public static boolean doesFileExist(String fileName) {
...
+   */
+  public static void cleanupDirectory(String dir) throws IOException {
+    File file = new File(dir);
+    logger.info("Cleaning up directory " + dir);
+    if (!file.exists()) {
+      // nothing to do
+      return;
+    }
+
+    assertThat(file.isDirectory())
```
markxnelson, line 80: this feel like it could be a simple check and a throw IllegalArgsException - assertions are really for tests

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
```diff
@@ -55,4 +63,55 @@ public static boolean doesFileExist(String fileName) {
...
+   */
+  public static void copyFolder(String srcDir, String destDir) throws IOException {
+    Path srcPath = Paths.get(srcDir);
+    Path destPath = Paths.get(destDir);
+    try (Stream<Path> stream = Files.walk(srcPath)) {
+      stream.forEach(source -> {
+        try {
+          copy(source, destPath.resolve(srcPath.relativize(source)));
+        } catch (IOException e) {
+          // cannot throw non runtime exception. the caller checks throwable
```
markxnelson, line 107: nice comment

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: I'll change the method name to defaultAppParams to be a little bit more clear. this method also hide things that a test writer does not need to know, for example, the version of the installers and etc. All parameters that a test writer cares needs to be set once he gets an AppParams instance.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
doxiao: good catch. fixed.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
doxiao: Changed the assertion to fail in two places.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
doxiao: done.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
ddsharpe: @rjeberhard it is not a known issue.  I just ran a quick test and do not see any issues.
./createDomain.sh -oracle_home ~/Oracle/Middleware12.2.1.3.0 -domain_parent . -model_file ~/Desktop/demo/my_domain.yaml -archive_file ~/Desktop/demo/my-domain.zip  -variable_file ~/Desktop/demo/my_domain1.10.properties

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    Throwable throwable = catchThrowable(
+        () -> { 
+        cleanupDirectory(ARCHIVE_DIR);
+        checkDirectory(ARCHIVE_SRC_DIR);
+        copyFolder(
+            APP_DIR + "/" + params.srcDir(), 
+            ARCHIVE_SRC_DIR);
+        });
+        
+    assertThat(throwable)
```
markxnelson, line 65: assertions are for tests, when implementing an action, i think we can just use the normal try/catch 
@rjeberhard thoughts?

markxnelson: i don't like tests relying on assertions hidden away in actions/helper methods

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
```diff
@@ -55,4 +62,54 @@ public static boolean doesFileExist(String fileName) {
...
+    if (!file.exists()) {
+      // nothing to do
+      return;
+    }
+
+    if (!file.isDirectory()) {
+      throw new IllegalArgumentException("The parameter " + dir + " should be a directory.");
+    }
+
+    Files.walk(Paths.get(dir))
```
markxnelson, line 83: we have commons-io in the pom, so you could do this with just one call to `FileUtils.deleteDirectory(file)`

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
ddsharpe: Agreed.  The test should deal with the exception.  Let "build()" throw it.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+            ARCHIVE_DIR))
+        .withFailMessage("Failed to get the directory " + ARCHIVE_DIR + " ready")
+        .isNull();
+
+    // build the app archive 
+    Exception exception = null;
+    String jarPath = String.format("%s.ear", params.srcDir());
+    try {
+      boolean jarBuilt = buildJarArchive(jarPath, ARCHIVE_SRC_DIR);
+      assertThat(jarBuilt)
```
ddsharpe, line 77: Same, throw an exception if something fails in an action (not a run-time exception, something that the test must catch)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: I recall that before I changed "-" to "_" in the sourcePath in my model file, the following caused an error.
appDeployments:
    Application:
        myear:
            SourcePath: wlsdeploy/applications/sample_app.ear
            ModuleType: ear
            Target: 'cluster-1'

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
doxiao: That is good to know. I'll change it.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/FileUtils.java
```diff
@@ -55,4 +62,50 @@ public static boolean doesFileExist(String fileName) {
...
+    logger.info("Cleaning up directory " + dir);
+    if (!file.exists()) {
+      // nothing to do
+      return;
+    }
+
+    if (!file.isDirectory()) {
+      throw new IllegalArgumentException("The parameter " + dir + " should be a directory.");
+    }
+    cleanDirectory(file);
```
markxnelson, line 82: um what is going on here? this is an infinite recursion?

doxiao: hmm, I don't see recursiveness here. cleanDirectory is common-io FileUtils.cleanDirectory().  The method calls this is cleanupDirectory().

doxiao: And I tested the change:).

markxnelson: oh sorry, my bad -- those names were so close :)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: Somehow WDT failed to parse the model yams file when the resourcePath contains "-". It seems that WDT considered -a as an option token and printed out this error:
####<Apr 10, 2020 6:00:40 PM> <INFO> <WebLogicDeployToolingVersion> <logVersionInfo> <WLSDPLY-01750> <The WebLogic Deploy Tooling validateModel version is 1.7.2:master.10eaf6e:Mar 20, 2020 23:24 UTC>
line 31:53 token recognition error at: '-a'
####<Apr 10, 2020 6:00:40 PM> <SEVERE> <ParsingErrorListener> <syntaxError> <WLSDPLY-18018> <Parse error for file /u01/wdt/models/model1-wls.yaml at line 31 position 55: extraneous input 'pp.ear' expecting {COMMENT, NEWLINE}>
####<Apr 10, 2020 6:00:40 PM> <SEVERE> <validate> <__perform_model_file_validation> <WLSDPLY-20009> <validateModel failed to load the model file /u01/wdt/models/model1-wls.yaml: Unable to parse model from file /u01/wdt/models/model1-wls.yaml : The YAML parser encountered 1 error(s) while parsing /u01/wdt/models/model1-wls.yaml>
####<Apr 10, 2020 6:00:40 PM> <SEVERE> <validate> <main> <WLSDPLY-20000> <validateModel encountered an unexpected validation error: Unable to parse model from file /u01/wdt/models/model1-wls.yaml : The YAML parser encountered 1 error(s) while parsing /u01/wdt/models/model1-wls.yaml>

doxiao: We could handle this by converting "-" to "_" in the AppBuilder action, but it is easier just not to have "-" in the app name.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/AppBuilder.java
doxiao: Removed all assertion in the actions.

doxiao:
@markxnelson @rjeberhard @ddsharpe All comments have been addressed. Please review.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
doxiao: Once I added "" around the value of sourcePath, the "-" in the application ear file path worked.

new-integration-tests/pom.xml
```diff
@@ -92,6 +92,11 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+          <version>2.6</version>
```
rjeberhard, line 98: Check the pattern... we're using dependency management.  The dependency goes here without the version number, but with <scope>test</scope>.  Then, there is a matching dependency in the dependency management section below where the version comes from a property.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItWITValidation.java
rjeberhard: This comment is still here :).  Let's change "sample_app" to use "sample-app" just to prove that it is fixed.

new-integration-tests/pom.xml
doxiao: fixed.

PR #1552, opened 2020-04-10
delete image only for domain in image tests

vanajamukkara:
Note: I don't have any of the multi node clusters free to test the change, will test on internal jenkins(single node) for now.

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
@@ -76,11 +76,13 @@ public class SitConfig extends BaseTest {
    *
    * @throws Exception when domain destruction fails
    */
-  protected static void destroySitConfigDomain(Domain domain) throws Exception {
+  protected static void destroySitConfigDomain(Domain domain, boolean domainInImage) throws Exception {
     if (domain != null) {
       LoggerHelper.getLocal().log(Level.INFO, "Destroying domain...");
       domain.destroy();
-      domain.deleteImage();
+      if (domainInImage) {
+        domain.deleteImage();
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
```diff
@@ -76,11 +76,13 @@ protected  Domain createSitConfigDomain(boolean domainInImage, String domainScri
    *
```
rjeberhard, line 76: Please fix the javadoc

integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
```diff
@@ -76,11 +76,13 @@ protected  Domain createSitConfigDomain(boolean domainInImage, String domainScri
...
    * @throws Exception when domain destruction fails
    */
-  protected static void destroySitConfigDomain(Domain domain) throws Exception {
+  protected static void destroySitConfigDomain(Domain domain, boolean domainInImage) throws Exception {
```
rjeberhard, line 79: Actually, I'm not sure hiding this here helps.  The caller has the domain object and could just call domain.deleteImage().

integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
vanajamukkara: yes, working on it :-)

integration-tests/src/test/java/oracle/kubernetes/operator/SitConfig.java
vanajamukkara: same goes for domain.destroy() also and don't even need this method destroySitConfigDomain(). I think the method was added here as there are two sit config test classes, so that they both inherit these methods.

vanajamukkara: destroySitConfigDomain() not really doing much, I can put those two lines in the test class itself.

PR #1557, opened 2020-04-11
cleanup.sh:  add fast WL pod exit, --force phase, backgrounding, and --timeout

tbarnes-us:
__The problem:__

The QA Jenkins runs are intermittently leaving k8s zombies behind for unknown reasons and consequently 'cleanup.sh' delete commands can hang.  

In addition, cleanup.sh can take a relatively long time to finish even when it's running fine.

__The fixes:__

(1) Add '--timeout=60s' to all initial 'nice' deletes.

(2) Speedup WL pod termination by (a) modifying stopServer.sh to exit without first calling WLST shutdown if '/tmp/diefast' exists, and (b) having the cleanup.sh script create '/tmp/diefast' on all running wl pods before calling delete

(3) Speedup deletes by running deletes in background, plus kill background jobs if they hang around more than 15 seconds.

(4) Add a final phase to cleanup.sh that uses the '--force' delete option. This should be OK because cleanup.sh is only used in test and dev environments.


tbarnes-us:
> I don't think this will work Tom. The problem is line 120 will not return in the failure case, so it will never get to your new code.
> 
> https://github.com/oracle/weblogic-kubernetes-operator/blob/c78b4641c88c6fe2739b4f10a475d98f4a5cb5c7/src/integration-tests/bash/cleanup.sh#L120

tbarnes-us:
@markxnelson wrote:

> I don't think this will work Tom. The problem is line 120 will not return in the failure case, so it will never get to your new code.
> 

I've pushed a change that sets a global timeout.  Still testing it.

operator/src/main/resources/scripts/stopServer.sh
```diff
@@ -48,6 +48,14 @@ function check_for_shutdown() {
     return 0
   fi
 
+  if [ -e /tmp/diefast ]; then
```
rjeberhard, line 51: This scares me 😨

tbarnes-us: This can cut several seconds off of shutdown, and we don't need a (slow) clean shutdown for the test cleanup case.  Would you like a different name that's more clear?  `bypassSafeShutdown` or some-such?

rjeberhard: I'll keep it as is... I just said that it was scary.

PR #1558, opened 2020-04-13
Log level review and change status from using patch to using endpoint

rjeberhard:
This work started after my observation that the operator logs were cluttered with too much "stream of consciousness" that obscured important behavior.  Similarly, failures were hard to triage since messages about async responses were not correlated with the requests. 

Therefore, my initial intention was simply to adjust the log level of messages -- that is, move some log messages from INFO to FINE and vice-versa. However, I quickly found that more was needed as domain status updates were consistently failing and (as noted above) async response failures didn't include information about the request.

In order to fix the domain status update, I've decided to propose a non-backward compatible change...  I've modified our CRD to enable the status subresource: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource.  This means that the k8s REST API will expose a "/status" subresource for our domain resources specifically for updating the status stanza of the document. This is non-backward compatible because if an operator of version 2.5.0 was running in the same Kubernetes cluster as an operator of version 3.0.0, then the version 2.5.0 operator would be unable to update the status of domains.  I'm proposing we make this change now because 1) we wanted to do this at some point, so 3.0.0 makes sense, and 2) the domain status updating using the patch operation was working so inconsistently that it could not be relied upon.

This change includes markers for two "TODO's" that I thought are too large for this PR.


russgold:
If you’re going to a status subresource, would it make sense to dispense with the status patching now?

rjeberhard:
Yup, @russgold.  I've done that.  Calls to update the domain status are going to the status endpoint and are not patches.

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFacade.java
```diff
@@ -651,6 +652,36 @@ public String getFormattedMessage(String msg, Object... args) {
...
+    throw new AssertionError(formatMessage(MessageKeys.RESOURCE_BUNDLE_NOT_FOUND));
+  }
+
+  /**
+   * Formats message based on string loaded from the resource bundle backing this logger.
+   * @param msgId Message id
+   * @param params Parameters to message formatting
+   * @return Formatted message
+   */
+  public String formatMessage(String msgId, Object... params) {
```
alai8, line 675: There is a getFormattedMessage() method in the same class that does the same thing, but with less error checking. We can get rid of that method and replace usages with this new method instead.

rjeberhard: Oh, thanks!  I missed it.  Let me check.

rjeberhard: I've decided to keep the formatMessage() method and remove the existing getFormattedMessage() method.

operator/src/main/java/oracle/kubernetes/operator/calls/RequestParams.java
```diff
@@ -46,4 +51,31 @@ public RequestParams(
...
+    if (namespace != null) {
+      sb.append(' ');
+      sb.append(LOGGER.formatMessage(MessageKeys.REQUEST_PARAMS_IN_NS, namespace));
+    }
+    if (name != null) {
+      sb.append(namespace != null ? ", " : " ");
+      sb.append(LOGGER.formatMessage(MessageKeys.REQUEST_PARAMS_FOR_NAME, name));
+    }
+    if (includeBody && body != null) {
+      sb.append(namespace != null || name != null ? ", " : " ");
```
markxnelson, line 76: what does this mean? which is evaluated first the or, or the elvis?
is that what you intended? 
seems like some parens here would help future generations greatly

PR #1572, opened 2020-04-17
Modified test infra to run integration test in helm3 

anpanigr:
Modify ItUsabilityOperatorHelmChart 

a) to support Helm 3.0. Few use cases were failing as the Exception description is different in Helm 2.0 and Helm 3.0. Need to change the expected exception String to a common string to support both Helm2/Helm3
b) Removed Duplicate Usecase
c) Removed commented Tests which can not be run on the parallel mode

Modified setupenv.sh to exclude tiller set up in case of Helm 3.0
Modified  sample/charts/util/setup.sh to modify logic verify pre-existing Chart
Modified LoadBalancer.java to check the logic to check helm version and added --debug option

External Jenkin ( quicktest cluster )  All passed 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1318/

Cluster 6 ( with helm 3 ) 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-integration-test-cluster-6/125/


rjeberhard:
Comment on the PR description... I still don't want to support Helm 2 & 3.  The description mentions a common set of exception strings so that we can support both versions of Helm.  We are explicitly dropping support for Helm 2 with the 3.0.0 operator version because the industry now regards the use of a "Tiller" pod as unsafe.

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
```diff
@@ -151,50 +160,32 @@ public void testCreateSecondOperatorUsingSameOperatorNsNegativeInstall() throws
...
-      }
-      if (BaseTest.HELM_VERSION.equals("V3")) {
-        cmdLb = "helm list --failed --namespace " + opnamespace + "  | grep " + oprelease;
-      }
-      LoggerHelper.getLocal().log(Level.INFO, "Executing cmd " + cmdLb);
-      ExecResult result = ExecCommand.exec(cmdLb);
-      if (result.exitValue() != 0) {
-        throw new RuntimeException(
-            "FAILURE: Helm installs second operator with same namespace as the first one ");
+          .contains("weblogic-operator-secrets")) {
```
rjeberhard, line 170: I don't understand this and it seems brittle.  If the Helm install is supposed to fail, I would expect a non-zero exit code from invoking the `helm` binary and that the list of installed Helm charts does not include the chart we just tried to install.

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
```diff
@@ -215,7 +206,7 @@ public void testNotPreCreatedOpNsCreateOperatorNegativeInstall() throws Exceptio
     TestUtils.exec(command);
     try {
       operator.callHelmInstall();
-      throw new RuntimeException("FAILURE: Helm install operator with not preexisted namespace ");
+      throw new Exception("FAILURE: Helm install operator with not preexisted namespace ");
```
rjeberhard, line 209: pre-existing

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
```diff
@@ -224,10 +215,12 @@ public void testNotPreCreatedOpNsCreateOperatorNegativeInstall() throws Exceptio
   }
 
   /**
-   * Negative test : Helm will install the operator with not preexisted operator service account,
-   * deployment will not start until service account will be created.
+   * Install the operator with non existing operator service account.
```
rjeberhard, line 218: "not existing" or "non pre-existing"

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
```diff
@@ -292,10 +285,15 @@ public void testNotPreexistedOpServiceAccountCreateOperatorNegativeInstall() thr
   }
 
   /**
-   * Negative test : Helm will install the second operator with same target domain namespace as the
-   * first operator.
+   * Install operator usab-1 with target DomainNameSpace [usab-domainns-1].
+   * Install operator usab-2 with same target DomainNamesapce [usab-domainns-1].
+   * Second operator should fail to install with following exception
```
rjeberhard, line 290: Same comment... I'm less concerned with the fact that Helm failed with a certain message than that the Helm install failed and that the chart was not installed.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
```diff
@@ -37,7 +37,7 @@ public LoadBalancer(Map lbMap) throws Exception {
 
     if (lbMap.get("loadBalancer").equals("TRAEFIK")) {
       String cmdLb = "";
-      if (! BaseTest.HELM_VERSION.equals("V2")) {
+      if (BaseTest.HELM_VERSION.equals("V2")) {
```
rjeberhard, line 40: Remove support for Helm 2.x

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
```diff
@@ -146,6 +146,7 @@ private void upgradeTraefikNamespace() throws Exception {
     StringBuffer cmd = new StringBuffer("helm upgrade ");
     cmd.append(" traefik-operator")
        .append(" stable/traefik ")
+       .append(" --debug ")
```
rjeberhard, line 149: Why are you adding the "--debug" flag here?  Is it because we are going to assert on debug output?

integration-tests/src/test/resources/setupenv.sh
```diff
@@ -179,7 +179,15 @@ if [ "$SHARED_CLUSTER" = "true" ]; then
 	fi
     	
   #fi
-  setup_shared_cluster
+  helm version --short --client | grep v2
+  [[ $? == 0 ]] && HELM_VERSION=V2
+  [[ $? == 1 ]] && HELM_VERSION=V3
+  if [ "$HELM_VERSION" == "V2" ]; then
```
rjeberhard, line 185: I'd recommend failing the test if you find Helm 2.x.  This will help us upgrade the clusters quickly.

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
anpanigr: I want to know the reason if the helm command fails so that it will be easier to debug when the test fails

kubernetes/samples/charts/util/setup.sh
```diff
@@ -2,19 +2,20 @@
...
+TNAME=traefik-operator  # release name for Traefik
 VSPACE=voyager # NameSpace for Voyager
 TSPACE=traefik   # NameSpace for Traefik
 DefaultVoyagerVersion=10.0.0
 
 HELM_VERSION=$(helm version --short --client)
 
 if [[ "$HELM_VERSION" =~ "v2" ]]; then
-   echo "Detected Helm Version [${HELM_VERSION}]"
+   echo "Detected helm version [${HELM_VERSION}]"
```
rjeberhard, line 18: Detect Helm 2 and fail

rjeberhard:
This PR is so much better... Just a few questions.  I think the only substantive item is that I'd like to not assert on particular error messages as this is likely brittle.

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
anpanigr: done

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
anpanigr: domne

integration-tests/src/test/java/oracle/kubernetes/operator/utils/LoadBalancer.java
anpanigr: done

integration-tests/src/test/resources/setupenv.sh
anpanigr: done

kubernetes/samples/charts/util/setup.sh
anpanigr: done

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
anpanigr: I removed the Helm version check

integration-tests/src/test/java/oracle/kubernetes/operator/ItUsabilityOperatorHelmChart.java
anpanigr: Here the callHelmInstall() expects to fail with a proper exception with the following message

Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: kind: Secret, namespace: ons, name: weblogic-operator-secrets

PR #1574, opened 2020-04-17
Always use the latest release of WDT and Image Tool

doxiao:
Find out the release number of the latest WDT/ImageTool, and use it as the version to use.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Command.java
```diff
@@ -46,12 +46,24 @@ public boolean execute() {
...
           params.redirect(),
           params.env());
-      if (result.exitValue() != 0) {
-        logger.warning("The command execution failed with the result: {0}", result);
+      if (params.saveStdOut()) {
+        params.stdOut(result.stdout());
+      }
+
+      // check both exitValue and stderr since sometimes the exitValue was 0 even when
+      // the test execution failed.
```
rjeberhard, line 54: Dongbo, can you give details of when WIT returned a 0 exit code even though image creation failed?  Our tests should not work around bugs in the product.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -80,13 +85,23 @@ private Installer params(InstallParams params) {
...
       logger.info("File {0} already exists.", params.fileName());
     } else {
+      // check and make sure DOWNLOAD_DIR exists; will create it if it is missing
       checkDirectory(DOWNLOAD_DIR);
+      
+      // we check if we get the version of the tool correctly first
+      if (params.version() == null 
+          || params.version().length() == 0
+          || params.version().equalsIgnoreCase("latest")) {
+        logger.severe("Failed to get the latest version of {0}", params.type());
```
rjeberhard, line 102: I don't understand what this check is ensuring.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallParams.java
```diff
@@ -49,6 +49,7 @@ public InstallParams version(String version) {
   }
 
   public String version() {
+    version = Installer.getActualVersionIfNeeded(location, type, version);
```
rjeberhard, line 52: This looks like a simple fluent getter but is doing a lot more.  Let's make the method more descriptive of what is going on.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -122,4 +137,45 @@ private String buildDownloadCommand() {
         params.fileName());
     return command;
   }
+
+  // Figure out the actual release number of the latest
```
rjeberhard, line 141: javadoc

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -122,4 +137,45 @@ private String buildDownloadCommand() {
...
+    if (version == null || version.equalsIgnoreCase("latest")) {
+      String command = String.format(
+          "curl -fL %s/releases/latest -o %s/%s-%s", 
+          location,
+          DOWNLOAD_DIR,
+          type,
+          TMP_FILE_NAME);
+      
+      if (!Command.withParams(defaultCommandParams().command(command)).execute()) {
+        return null;
```
rjeberhard, line 156: What does it mean to return null here?  If the command fails isn't that a bigger issue?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
```diff
@@ -122,4 +137,45 @@ private String buildDownloadCommand() {
...
+      CommandParams params = 
+          defaultCommandParams()
+          .command(command)
+          .saveStdOut(true)
+          .redirect(true);
+
+      if (Command.withParams(params).execute()) {
+        return params.stdOut();
+      } else {
+        return null;
```
rjeberhard, line 176: Same comment

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Command.java
doxiao: This is not only applicable to the WIT commands This method is used for all command execution such as running a curl command to download the tool, or creating a zip file. I tested a couple of error conditions such as wrong download url, and somehow the exitValue was zero. So I added the additional checks to make sure the code catches the error condition. I myself found this behavior strange/unexpected. Let me test those error conditions one more time tomorrow and see if I did something wrong last week when I did the testing.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: When the params.version() method is called the first time,  it (unexpectedly, as you indicated below) performs some additional operations to find out the release number of the "latest" by calling getRealVersionIfNeeded(). Method getRealVersionIfNeeded may return null. This checks ensure that we did get the real version, and the code returns false to indicate a failure if finding the real version somehow failed.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/InstallParams.java
doxiao: Okay

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: As I mentioned in the earlier responses, returning null is a way to tell the caller that this failed, and the caller will log a severe message, and return false to indicate a failure condition. The failure usually does not accompanied by an exception. We could manufacture an exception and throw here, but I am not sure what exception to throw.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Command.java
doxiao: Jus to be clear - I have not seen that a WIT command returned zero exitValue when there was a problem in command execution.

doxiao: I did more testing. In the most of the cases, the exitValue does reflect the execution status. Only in the new getRealVersionIfNeeded method, there were cases that the exitValue is still zero when there was a problem in one of the commands that pipelined together. In those cases, the later check in Installer.java line#99 will catch it and log a severe message. I'll change the code back to only check exitValue.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
ddsharpe: Keep in mind we are trying to make it clear for the "next" person.  If I understand this correctly, I think throwing an exception is clearer that "things are not working as expected".  Create a new exception class if it helps make things more clear.

doxiao: Agreed. working on a change.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Installer.java
doxiao: fixed with a severe log message and throwing a RuntimeException. The caller catches the exception and returns false to fail the test case.

PR #1579, opened 2020-04-17
Setting WLSDEPLOY_PROPERTIES cause WDT failed to start

jshum2479:
CGBU discovered a problem, when setting WLSDEPLOY_PROPERTIES with -Doracle.jdbc.fanEnabled in the domain resource, it caused the WDT create fails. The reason is -skipWLSModuleScanning is not a JVM properties and it caused the the wlst failed to start.  It shouldn't be there in the first place, I am removing for now. In order to add this, it requires a WDT enhancement.


jshum2479:
Actually I don't remembering seeing a significant differences in start time from earlier performance results and casual observation. Plus since the flag is a parameter `-skipWLSModuleScanning' and not a java property, it must be placed exactly before the actual python script to be run, otherwise the JVM will complain that's the wrong jvm option.

PR #1581, opened 2020-04-21
POM updates for new integration tests and other clean-ups.

rjeberhard:
There are various clean-ups:

- Connect new integration tests module to the parent pom
- Fix issues and we are now generating coverage!!!
- Add dependency on JUnit 4 for modules with unconverted tests
- Remove build-tools module and replace with a parameter referencing parent location
- Remove unused files
- Fix spelling of "WebLogic" :)


build-tools/checkstyle/suppressions.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
```
markxnelson, line 1: should have copyright notice

rjeberhard: Thanks

PR #1584, opened 2020-04-22
update quickstart based on user feedback

markxnelson:


docs-source/content/quickstart/create-domain.md
```diff
@@ -25,6 +25,9 @@ weight: 6
 1.	Create a new image with a domain home by running the [create-domain](http://github.com/oracle/weblogic-kubernetes-operator/blob/master/kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain.sh) script. First, copy the sample [create-domain-inputs.yaml](http://github.com/oracle/weblogic-kubernetes-operator/blob/master/kubernetes/samples/scripts/create-weblogic-domain/domain-home-in-image/create-domain-inputs.yaml) file and update your copy with:  
        * `domainUID`: `sample-domain1`
        * `image`: Leave empty unless you need to tag the new image that the script builds to a different name.
+          For example if you are using a remote cluster that will need to pull the image from a Docker registry,
```
rosemarymarano, line 28: For example if you are using a remote cluster that will need to pull the image from a Docker registry, in which case you should set this value -> For example, if you are using a remote cluster that will need to pull the image from a Docker registry, then you should set this value

markxnelson: updated

markxnelson: done

PR #1585, opened 2020-04-22
Find parent pom directory using maven magic

rjeberhard:


build-tools/checkstyle/customized_google_checks.xml
```diff
@@ -41,7 +41,7 @@
     </module>
 
     <module name="RegexpHeader">
-        <property name="headerFile" value="build-tools/checkstyle/java.header"/>
+        <property name="header" value="^// Copyright \(c\) (\d\d\d\d, )+Oracle Corporation and\/or its affiliates\.$\n^// Licensed under the Universal Permissive License v 1\.0 as shown at https://oss\.oracle\.com/licenses/upl\.$"/>
```
ddsharpe, line 44: does this work better?

rjeberhard: Depending on whether you executed `mvn` at the top-level or in the directory of a child pom, this would fail to process the URI reference.  The folks working on the new integration tests have been cd'ing into new-integration-tests and Checkstyle was failing to find this file.

PR #1588, opened 2020-04-23
Model in image sample v2 merge

tbarnes-us:
### **Changes since 'v1':**

This is 'version 2' of the MII sample, which includes many updates, small and large.  Including, but not limited to:

   - The sample itself:
      - Is in directory `kubernetes/samples/scripts/create-weblogic-domain/model-in-image`.
      - Has been simplified to only include utility scripts, templates, and its web-app. 
      - Has new `utils/wl-pod-wait.sh` helper utility for waiting until pods reach their desired state.  In its verbose mode, it shows the target restartVersion and image which it obtains from the live domain resource, the corresponding fields in each pod, plus each pod's ready state. It also updates is output as pods start/roll. 
      - Uses configured cluster for JRF instead of dynamic.
      - Has updated sample web-app
         - App now queries mbeans to show existing configured datasources (useful for demo'ing the dynamic addition of a datasource for a configmap)
         - App now is 'exploded' (useful for testing updates to it on running pods)

  - Minor update to operator java:
     - The change makes domain resource 'restartVersion' defaulting less confusing. 
     - The operator was treating "not set" and "empty string" as different restart versions, which is messy because tooling like 'kubectl get' tends to return the same value for either.  
     - This change has been reviewed and approved by Ryan.

   - There's a new test for the sample: 
     - See `./src/integration-tests/model-in-image/README`.  
     - _This is not part of the sample itself - it's a test of the sample._
     - It's completely stand-alone, but should be straight forward to wire into the test framework.  
     - Even if we leave it 'unwired' for a while, it's useful as a stand-alone dev tool for quickly trying out dev changes to the MII sample/internals.

   - The original sample's 'env var' create, build, and stage scripts:
      - Have moved to `src/integration-tests/model-in-image/mii-sample-wrapper`.
      - _They are no longer part of the sample itself - they're solely used to help test the sample's utility scripts, apps, and templates, and to help generate its documentation._
      - These scripts are now even finer grained and have been expanded so that they cover all of the 'four use cases'.
      - Their generated files now get decorated with domain-uid and/or image-name as an aid in demonstrating the concurrent handling of multiple domains/images.
      - New scripts:
         - New script for testing model updates via a configmap.  Previously this was entirely within the old sample via doc (no helper scripts for testing same).
         - New script for staging and deploying ingresses (previously this was a 'doc only' step).
         - New editable `custom-env.sh` script containing description of all env vars and their defaults.  Copy the file to the tests  `WORKDIR` and edit if you want the wrapper scripts to load an edited version.

### **To understand the sample:**

- Read through existing doc for the sample 
  - Doc location in hugo menus: 
     - 'Samples -> Simple Samples -> Domains -> Model in Image'.
     - 'User Guide -> Manage WebLogic Domains -> Model in Image''.
  - Doc source :
     - Sample: `docs-source/content/samples/simple/domains/model-in-image` 
     - User guide: `docs-source/content/userguide/managing-domains/model-in-image`.
  - To view the docs, locally clone this pull request's branch and start a 'hugo' web server:
     - `cd docs-source ; ./run-local.sh`: Starts hugo accessible at `http://localhost:1313/weblogic-kubernetes-operator`.
     - `cd docs-source ; ./run-public.sh`: Starts hugo accessible at `http://your-host-machines-address:1313/weblogic-kubernetes-operator`.

### **To run the sample test itself (which in turn runs the test's sample wrapper scripts):**

To run a stand-alone test that does both a 'create image and deploy domain resource' and an 'add datasource configmap and roll' call `src/integration-tests/model-in-image/run.sh -clean`.

**WARNING:** The "-clean" options is needed to run the test in its simplest form, but it's destructive.  _It pre-deletes the entire WORKDIR (default `/tmp/$USER/model-in-image-sample-work-dir`), and it runs 'cleanup.sh'._

See the README in the test directory for details.

### **To generate fodder for the documentation:**

**WARNING!**:  _The generate script deletes directory `/tmp/miisample` and places/replaces files in this directory _and_ the sample source directory._

Run `src/integration-tests/model-in-image/mii-sample-wrapper/generate-sample-doc.sh`.

### **Misc notes:**

- **Not tested on Mac yet. Anyone with a Mac env want to give it a whirl?**

- The JRF/RestrictedJRF path through the sample hasn't been retested.  That's in progress. 

- There are existing sample tests from the QA team which will need to be updated.  I will work with them to do so (likely in a separate pull for convenience).


rjeberhard:
Okay, I'm happy about some things but there are also some disconnects from what I was expecting.  Good news first: I was expecting scripts for creating a model in image, creating a config map with a model, and creating a domain resource yaml and I see `build-model-image.sh` and `create-domain-resource.sh`.  Then the bad news: I was expecting to see scripts with arguments and the environment variables removed. It's not usable to expect customers to set environment variables ahead of calling a tool. The flow given above still includes an all-encompassing set of "run" scripts that obscure what is happening and include actions I would not want as a customer.  Instead, I just want to be able to make 3 or 4 calls that are specific to each of the steps. By having arguments to the scripts, specifically referencing the sample model and application archive, then it becomes very obvious how to use my own model and archive.  I'll put other specific comments inline.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+# Copyright (c) 2019, 2020, Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+# 
+#  Summary:
+#
+#    This script builds a model image using the WebLogic Image Tool. The
+#    tool pulls a base image if there isn't already a local base image.
+#    This script, by default, builds the model image with model files from
+#    './stage-model-image.sh' plus tooling downloaded by './stage-tooling.sh.'.
```
rjeberhard, line 11: Let's not have a default, but have a required argument for the location of the model file.  I'll look at stage-model-image.sh, but it's unclear why I should need to stage the model file.  At this point, this seems unnecessary.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+#
+#  Assumptions:
+#
+#    - The WebLogic Image Tool zip is 'WORKDIR/weblogic-image-tool.zip' and
```
rjeberhard, line 15: Image tool has the concept of a cache directory, which you then reference.  It feels like we are just adding a bunch of machinery on top of the image tool.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+#      the WebLogic Deploy Tool zip is 'WORKDIR/weblogic-deploy-tooling.zip'
+#      (see './stage-tooling.sh').
+#
+#    - Model files have been staged in the MODEL_DIR directory
+#      (by './stage-model-image.sh' or some custom process).
+#
+#  Optional environment variables:
+#
+#    WORKDIR
```
rjeberhard, line 24: Replace with arguments with defaults where the argument isn't required

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+#      Defaults to '/tmp/$USER/model-in-image-sample-work-dir'.
+#
+#    MODEL_DIR:
+#      Location of the model .zip, .properties, and .yaml files
+#      that will be copied to the model image.  Default is:
+#        'WORKDIR/model/image--$(basename $MODEL_IMAGE_NAME):$MODEL_IMAGE_TAG'
+#      which is usually populated by the './stage-model-image.sh' script.
+#
+#    MODEL_IMAGE_BUILD:
+#      Set to 'when-changed' (default) or 'always'. Default behavior is to
```
rjeberhard, line 35: This doesn't belong in this tool, but outside of it.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+#      'WLS' (default), 'RestrictedJRF', or 'JRF'.
+#
+#    BASE_IMAGE_NAME, BASE_IMAGE_TAG:
+#      The base image name defaults to 
+#         'container-registry.oracle.com/middleware/weblogic'
+#      for the 'WLS' domain type, and otherwise defaults to 
+#         'container-registry.oracle.com/middleware/fmw-infrastructure'. 
+#      The tag defaults to '12.2.1.4'.
+#
+#    MODEL_IMAGE_NAME, MODEL_IMAGE_TAG:
```
rjeberhard, line 49: Should just be a single argument.  Also, call to WIT below uses variable MODEL_IMAGE, which is not created from these variables in this script.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: The comment here documents the default, and it is being honored.

All of the scripts replace arguments with defaults.  

This script loads them up via 'env-init.sh'.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: What doesn't?

tbarnes-us: The default behavior to skip rebuild when image already exists was insisted on by Mark.

tbarnes-us: And I agree that it's helpful...

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: This can change - I agree it's confusing, but I'm following the tradition of the other samples.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: What machinery is added?  This particular script expects some staged files.   The image tool requires them.  The staged files are setup with two small scripts - stage-tooling and stage-model-image.  Are you saying that's too much?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: Staging them gives the opportunity to customize the source image on a per image basis. The demo is intended to help show the development of different images.   Different images will commonly want to different model files/archives.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+
+echo @@ Info: WDT_DOMAIN_TYPE=${WDT_DOMAIN_TYPE}
+echo @@ Info: MODEL_DIR=${MODEL_DIR}
+echo @@ Info: BASE_IMAGE_NAME=${BASE_IMAGE_NAME}
+echo @@ Info: BASE_IMAGE_TAG=${BASE_IMAGE_TAG}
+echo @@ Info: MODEL_IMAGE_NAME=${MODEL_IMAGE_NAME}
+echo @@ Info: MODEL_IMAGE_TAG=${MODEL_IMAGE_TAG}
+echo @@ Info: MODEL_IMAGE_BUILD=${MODEL_IMAGE_BUILD}
+
+if [ ! "$MODEL_IMAGE_BUILD" = "always" ] \
```
rjeberhard, line 69: This feature just doesn't make sense.  If I call a script called `build-model-image.sh`, then it should build the image.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
```diff
@@ -0,0 +1,144 @@
...
+# Run the image tool to create the model image. It will use the WDT binaries
+# in the local image tool cache marked with key 'myversion' (see 'cache' commands above).
+# Note: The "${macro:+text}" syntax expands to "" if 'macro' is empty, and to 'text' if it isn't
+#
+
+# TBD test empty cases, even all three empty
+
+set -x
+
+${IMGTOOL} update \
```
rjeberhard, line 130: On our "stagging" discussion, it seems down here when we call WIT that you can just list the models, properties files, and archives. This script could just take the same variables and be a lot simpler.  Then the doc. can show how to call it for the sample models and archives.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/create-domain-resource.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash
+# Copyright (c) 2019, 2020, Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+#
+# This script deploys the domain resoure in WORKDIR/mii-DOMAIN_UID.yaml.
```
rjeberhard, line 6: Oh, this script deploys the domain resource... I was thinking that this script creates the domain resource yaml based on the model.  I don't really need a script for the customer to `kubectl apply -f domain.yaml`.  Let me see what else is here.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: The script lets you specify a directory that contains the the models/props/archives.   How could it be simpler than that?  As for the staging, I realize it's a level of indirection, but staging gives the user a chance to customize same without obliterating their work on a different image.

tbarnes-us: Are you saying you want the option to specify specific files for each? I had that option for at one point during development, and I can add that back in, but note that that turned out to have it's own confusion in practice.  It added more 'options' which had a side effect of obscuring the overall intent of the sample.  It also discouraged a best practice of having all files needed for an image gathered in one place...

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/create-domain-resource.sh
tbarnes-us: The sample doc that walks through the sample will clearly include the exact apply command and won't emphasize this script.  In fact, the sample doc will do pretty much the same for all of the scripts. The script is mainly intended for automation/testing - which are core requirements that aren't directly related to using this branch as a sample.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: This was requested by Mark.  This script is intended to be closer to 'production' than 'sample' - so it has 'optimizations' like this.  One side effect is that it forces the user to think once they see the message, to see if there intent was to obliterate an existing image.  For example, maybe they need to change the image name/tag.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/util-patch-restart-version.sh
```diff
@@ -0,0 +1,83 @@
+# !/bin/sh
```
rjeberhard, line 1: This can be done with just a few steps:
```bash
#! /bin/bash
current=`kubectl get domain sample-domain1 -n sample-domain1-ns -o=jsonpath='{.spec.restartVersion}'`
next=$((current + 1))
echo Rolling domain to restartVersion=$next
kubectl patch domain sample-domain1 -n sample-domain1-ns --type='json' -p='[{"op": "replace", "path": "/spec/restartVersion", "value": "'${next}'" }]'
```

tbarnes-us:
> Okay, I'm happy about some things but there are also some disconnects from what I was expecting. Good news first: I was expecting scripts for creating a model in image, creating a config map with a model, and creating a domain resource yaml and I see `build-model-image.sh` and `create-domain-resource.sh`. Then the bad news: I was expecting to see scripts with arguments and the environment variables removed. It's not usable to expect customers to set environment variables ahead of calling a tool. The flow given above still includes an all-encompassing set of "run" scripts that obscure what is happening and include actions I would not want as a customer. Instead, I just want to be able to make 3 or 4 calls that are specific to each of the steps. By having arguments to the scripts, specifically referencing the sample model and application archive, then it becomes very obvious how to use my own model and archive. I'll put other specific comments inline.

I was forced to put in the variables for test automation and repeatability as a core requirement.
Command line args is a new requirement - I thought have putting them in but no one was asking for them in the original, and they conflict with another core requirements which is to try keep the scripts simple.  If I put them in, I could follow the pattern of the 'util' scripts,  which accept both env vars and command line args (the latter override the former).

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerSpecCommonImpl.java
```diff
@@ -228,17 +228,20 @@ public V1SecurityContext getContainerSecurityContext() {
 
   @Override
   public String getDomainRestartVersion() {
-    return domainSpec.getRestartVersion();
+    // treat null, unset, and "" as the same restartVersion ""
```
rjeberhard, line 231: Need to see if you did it, but if we normalize missing labels on existing pods before creating the hash, then we won't roll existing pods.

tbarnes-us: Isn't the hash generated using the above?   This change is for 3.0, so there won't be existing pods that aren't normalized.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: Are you saying that the WORKDIR env-var/argument isn't required?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: Note that I'm under the impression image tool 'cache' concept is mainly intended for caching installers, and that's how it's being used in the sample.  This is the first I've used of it being used for caching model files.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: I think the idea is that customers will commonly want to change the tag independent of the image name...

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/sample-archive/wlsdeploy/applications/myapp/sample_war/index.jsp
```diff
@@ -0,0 +1,44 @@
...
+       String dsName  = (String)mbs.getAttribute(dataSource, "Name");
+       String dsState = (String)mbs.getAttribute(dataSource, "State");
+       out.println("  Datasource '" + dsName + "': State='" + dsState +"'");
+    }
+
+    out.println();
+    out.println("**************************************************************");
+
+  } catch (Throwable t) {
+    t.printStackTrace(new PrintStream(response.getOutputStream()));
```
jshum2479, line 39: This won't show in the browser but goes to the stdout of the server

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/sample-model/JRF/model.10.yaml
```diff
@@ -0,0 +1,39 @@
...
+            Cluster: 'cluster-1'
+            ListenPort: '@@PROP:MANSVR_PORT@@'
+         'managed-server4':
+            Cluster: 'cluster-1'
+            ListenPort: '@@PROP:MANSVR_PORT@@'
+           
+appDeployments:
+    Application:
+        myear:
+            SourcePath: 'wlsdeploy/applications/myapp'
```
jshum2479, line 37: Are we packaging as exploded format or ear format?

src/integration-tests/model-in-image/README
```diff
@@ -0,0 +1,120 @@
...
+           which deletes almost everything in your k8s cluster.
+
+         - Deletes the current WORKDIR before proceeding.
+
+         - Implicitly cause the test to build/deploy the operator,
+           and to deploy traefik on node port 303035.
+
+       - The '-db' option will cause the test to deploy Oracle DB.
+
+       - Pass '-noupdate' to skip the multi-minute 'update' portion of the test.
```
jshum2479, line 60: What's the meaning of multi-minute update?

src/integration-tests/model-in-image/README
```diff
@@ -0,0 +1,120 @@
...
+
+  (2) This test assumes Kubernetes is running locally 
+      on a single machine. (It builds/uses local images
+      without pushing them.)
+
+-------------------
+To run stand alone:
+-------------------
+
+  1) Optionally customize the following:
```
jshum2479, line 22: Is this the first file user will look at (or the landing page is at the documentation)?  I am just wondering how the user knows where things occured

src/integration-tests/model-in-image/deploy-wl-operator.sh
```diff
@@ -0,0 +1,74 @@
...
+
+set -eu
+cd ${SRCDIR}
+
+helm install $OPER_NAME kubernetes/charts/weblogic-operator \
+  --namespace $OPER_NAMESPACE \
+  --set       image=$OPER_IMAGE \
+  --set       serviceAccount=$OPER_SA \
+  --set       "domainNamespaces={$DOMAIN_NAMESPACE}" \
+  --set       "javaLoggingLevel=FINEST" \
```
jshum2479, line 65: Should the default be FINEST?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/sample-model/JRF/model.10.yaml
tbarnes-us: Exploded.

src/integration-tests/model-in-image/README
tbarnes-us: Note that this directory is for test purposes, it's not part of the sample.

The reason for the comment is that it takes several minutes to do a roll.  So the option is there to skip that part.  When testing, or giving a 'simplified' form of the demo, it's often sufficient to skip the roll demo...

src/integration-tests/model-in-image/README
tbarnes-us: No. This is a test. It's not meant for users.    See the files in the sample directory for the sample.

Read the pull summary for details.

markxnelson:
I will be happy to review when the documentation updates are done.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/sample-archive/wlsdeploy/applications/myapp/sample_war/index.jsp
tbarnes-us: I think it does show in the browser - I've tested it.   Note that it's not going to 'stdout' but to 'response.getOutputStream()'.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/build-model-image.sh
tbarnes-us: Are this and your other comments resolved now?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/sample-model/JRF/model.10.yaml
tbarnes-us: Please resolve if your fine with this.

docs-source/content/userguide/managing-domains/model-in-image/usage.md
```diff
@@ -48,11 +48,11 @@ There are two methods for layering Model in Image artifacts on top of a base ima
 
 For more information about model file syntax, see [Model files]({{< relref "/userguide/managing-domains/model-in-image/model-files.md" >}}).
 
-#### 3. Optional WDT model config map
+#### 3. Optional WDT model configmap
```
markxnelson, line 51: i found line 40 confusing - it implied to me that i do not have to put the archive in the image, that i can put it somewhere else - maybe in the configmap, like this section goes on to describe for the model and properties.  but i think it actually means to say that is is optional because i might not have one at all? is that right?

docs-source/content/userguide/managing-domains/model-in-image/usage.md
```diff
@@ -105,8 +105,8 @@ The following domain resource attributes are specific to Model in Image domains.
...
-| `configuration.model.configMap`             | Optional. Set if you have stored additional models in a config map as per [Optional WDT model config map](#3-optional-wdt-model-config-map). |
-| `configuration.secrets`                      | Optional. Set this array if your image or config map models contain macros that reference custom Kubernetes secrets. For example, if your macros depend on secrets `my-secret` and `my-other-secret`, then set to `[my-secret, my-other-secret]`.|
+| `configuration.model.configMap`             | Optional. Set if you have stored additional models in a configmap as per [Optional WDT model configmap](#3-optional-wdt-model-config-map). |
+| `configuration.secrets`                      | Optional. Set this array if your image or configmap models contain macros that reference custom Kubernetes secrets. For example, if your macros depend on secrets `my-secret` and `my-other-secret`, then set to `[my-secret, my-other-secret]`.|
```
markxnelson, line 109: for line 154 (not sure why it won't let me add a comment there)  
i don't think this warning is quite accurate - they can always ask the dba to reset the password

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -2,142 +2,366 @@
...
+
+1. _Set up Ingresses that will redirect HTTP from Traefik port 30305 to the clusters in this sample's WebLogic domains._
 
-5. Deploy the operator and set up the operator to manage the namespace, `sample-domain1-ns`. Optionally, deploy a Traefik load balancer that manages the same namespace. For example, follow the same steps as the [Quick Start](https://oracle.github.io/weblogic-kubernetes-operator/quickstart/), up through the [Prepare for a domain]({{< relref "/quickstart/prepare.md" >}}) step.
+    Run one of the following commands:
 
-   Note that:
-   - Deploying the Traefik load balancer is optional, but is a prerequisite for testing the web application that's deployed to WebLogic as part of this sample.
-   - You can skip the Quick Start steps for obtaining a WebLogic image because you will be creating your own Docker image.
+    - To deploy the Ingresses, one option is to use your favorite editor to cut and paste the following YAML to a file called `/tmp/mii-sample/ingresses/myingresses.yaml` and then call `kubectl apply -f /tmp/mii-sample/ingresses/myingresses.yaml`:
```
markxnelson, line 173: i found this a bit magical... i think it needs to be explained more.  maybe move the note from 270 up to above this and expand it a little bit?   so i know before i get to this step what i am about to do, otherwise i am left wondering for a little while and might go off on a tangent trying to work it out only to come back later and find the answer on the next line :(

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -379,182 +1209,733 @@ If you run `kubectl get pods -n sample-domain1-ns --watch`, then you should see
...
+           JDBCConnectionPoolParams:
+               InitialCapacity: 0
+               MaxCapacity: 1
+               TestTableName: SQL ISVALID
+               TestConnectionsOnReserve: true
+
    ```
+   {{% /expand%}}
+
+   You can cut and paste the above model snippet to a file and then use it for your configmap, or alternatively, you can use the same data source that's already there for you in `/tmp/mii-domain/model-configmaps/model.20.datasource.yaml`.
```
markxnelson, line 1461: i got confused here -- what am i meant to do with this? how to i create the CM?  i assume that you will tell me later...

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -379,182 +1209,733 @@ If you run `kubectl get pods -n sample-domain1-ns --watch`, then you should see
...
-1. Find out the service name of the Administration Server and service port number.
+1. _Update your domain resource file to refer to the configmap and its secret, and apply it._
 
-   The service name follows the pattern `domainuid-adminservername`, all lower case, with a hyphen `-` substituted for each underscore `_`.
-   The port number for administration traffic is configured in your WebLogic configuration (your model files), where the default is `7001`.
+      Do one the following:
 
-   If your domain resource is deployed, then you can also find the information by getting the port number for the Administration Server pod:
+      {{%expand "Click here to expand." %}}
+  - Option 1: Update your current domain resource file from the Initial use case.
```
markxnelson, line 1524: i could have used a reminder here of what i called that file :)

docs-source/content/userguide/managing-domains/model-in-image/model-files.md
```diff
@@ -86,11 +86,11 @@ Some notes about the sample model file:
 
 Refer to this section if you need to control the order in which your model files are loaded.  The order is important when two or more model files refer to the same configuration, because the last model that's loaded has the highest precedence.
```
jshum2479, line 87: same configuration - probably need to elaborate more >  e.g.  Same jdbc datasource URL is referenced by two different model files or the same property is referenced by two different properties files

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -2,142 +2,366 @@
...
+     {{% /expand%}}
 
-A JRF domain requires an infrastructure database called an RCU database, requires initializing this database, and requires configuring your domain to access this database. All of these steps must occur before you first deploy your domain.
+1. _Download the latest WebLogic Deploying Tooling and WebLogic Image Tool installer ZIP files to your `/tmp/mii-sample/model-images` directory._
 
-Furthermore, if you want to have a restarted JRF domain access updates to the infrastructure database that the domain made at an earlier time, the restarted domain must be supplied a wallet file that was obtained from a previous run of the domain.
+   {{%expand "Click here for details." %}}
+
+   Both WDT and WIT are required to create your own Model in Image Docker images. Download the latest version of each tool's installer ZIP file to the `/tmp/mii-sample/model-images` directory. Make sure that the files are named `weblogic-deploy-tooling.zip` and `weblogic-image-tool.zip`.
```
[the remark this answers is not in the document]
tbarnes-us: Where's the rename?  Do you mean you have your own versions of these zip already and your own versions have different names? The curl commands use the same names as above. And the image tool command cache is using the names specified here too.

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -260,113 +485,718 @@ To reuse the wallet for subsequent redeployments or share the RCU tables between
...
+
+{{% notice note %}}
+If you are taking the `JRF` path through the sample, then substitute `JRF` for `WLS` in your image names and directory paths. Also note that the JRF-v1 model YAML differs from the WLS-v1 YAML file (it contains an additional `domainInfo -> RCUDbInfo` stanza).
+{{% /notice %}}
+
+#### Image creation - Introduction
+
+The goal of the initial use case 'image creation' is to demonstrate using the WebLogic Image Tool to create an image named `model-in-image:WLS-v1` from files that we will stage to `/tmp/mii-sample/model-images/model-in-image:WLS-v1/`. The staged files will contain a web application in a WDT archive, and WDT model configuration for a WebLogic Administration Server called `admin-server` and a WebLogic cluster called `cluster-1`.
+
+Overall, a Model in Image image must contain a WebLogic installation, a WebLogic Deploy Tooling installation in its `/u01/wdt/weblogic-deploy` directory, your WDT model YAML and properties files in its `/u01/wdt/models` directory, and your WDT model archive files in the same directory.
```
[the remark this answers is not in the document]
tbarnes-us: good catch - i've updated this and the related user doc

docs-source/content/userguide/managing-domains/model-in-image/usage.md
tbarnes-us: good catch - i've updated this and the related mention in the sample doc

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -260,113 +485,718 @@ To reuse the wallet for subsequent redeployments or share the RCU tables between
...
-If you intend to use a remote Docker registry, you need to tag and push the image to the remote Docker registry.
+  - Understanding a basic WDT archive's structure
+  - Creating a WDT archive ZIP file that contains a web application.
+  - Creating a WDT YAML model that references the web application, and defining a WDT `properties` file.
+  - Using the WebLogic Image Tool to create a model image from these files.
 
-1.  Tag the image for the remote Docker registry, for example:
+#### Understanding our first archive
+
+The sample includes a predefined archive directory in `/tmp/archives/archive-v1` that we will use to create an archive ZIP file for the image.
```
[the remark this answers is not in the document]
tbarnes-us: fixed

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -260,113 +485,718 @@ To reuse the wallet for subsequent redeployments or share the RCU tables between
...
 ```
-docker push my.remote.registry.com/model-in-image:v1
+# Delete existing archive.zip in case we have an old leftover version
+rm -f /tmp/mii-sample/model-images/model-in-image__WLS-v1/archive.zip
+
+# Move to the directory which contains the source files for our archive
+cd /tmp/mii-sample/archives/archive-v1
+
+# Zip the archive to the location will later use when we run the WebLogic Image Tool
+zip -q -r /tmp/mii-sample/model-images/model-in-image__WLS-v1/archive.zip wlsdeploy
```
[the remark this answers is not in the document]
tbarnes-us: done

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -260,113 +485,718 @@ To reuse the wallet for subsequent redeployments or share the RCU tables between
...
+Run the following commands to create the model image and verify that it worked:
+
+{{%expand "Click here to expand." %}}
+
+  ```
+  cd /tmp/mii-sample/model-images
+  ./imagetool/bin/imagetool.sh update \
+    --tag model-in-image:WLS-v1 \
+    --fromImage container-registry.oracle.com/middleware/weblogic:12.2.1.4 \
+    --wdtModel /tmp/mii-sample/model-images/model-in-image__WLS-v1/model.10.yaml \
```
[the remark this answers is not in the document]
tbarnes-us: done

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -379,182 +1209,733 @@ If you run `kubectl get pods -n sample-domain1-ns --watch`, then you should see
...
-1. Ensure Traefik has been installed and is servicing external port 30305, as per [Prerequisites for all domain types](#prerequisites-for-all-domain-types).
+Now that all the initial use case resources have been deployed, let's invoke the sample web application through the Traefik load balancer port. Note that the web application will report if it finds any data sources, but we don't expect it to find any because the model doesn't contain any at this point.
 
-2. Create a Kubernetes Ingress for the domain's WebLogic cluster in the domain's namespace by using the sample Helm chart:
+Run one of the following commands:
 
-   For Helm 3.x:
+   ```
+   curl -s -S -m 10 -H 'host: sample-domain1-cluster-cluster-1.mii-sample.org' \
+      http://localhost:30305/myapp_war/index.jsp
```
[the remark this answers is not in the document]
tbarnes-us: IIRC, I added a mention of the heuristic "kubectl cluster-info | grep KubeDNS" elsewhere, I could mention it here and anywhere else it isn't brought up.  Is that a good heuristic?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerSpecCommonImpl.java
rjeberhard: Do we still need these changes?

tbarnes-us: The changes were never absolutely needed.  They're a "nice to have". 

The issue is that the average user plus at least one public Kubernetes API view 0-length and null strings as the same, but the operator treats them differently.  This makes the attribute a little confusing at its "edges".

docs-source/content/samples/simple/domains/model-in-image/_index.md
tbarnes-us: done.

PR #1593, opened 2020-04-25
Fix exec command

doxiao:
Fix intermittent issues on Jenkins in exec command util.
1) revert the changes that perform mark and reset.
2) force the thread to join so that the copied input stream is ready to be read for getting the stdout.

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
@@ -64,13 +64,8 @@ public class ExecCommand {
     Thread out = null;
 
     try {
-      String stdout = null;
       if (isRedirectToOut) {
         InputStream i = in.getInputStream();
-        i.mark(0);
-        // read the stdout first, and reset before we redirect the output
-        stdout = read(i);
-        i.reset();
         @SuppressWarnings("resource")
         CopyingOutputStream copyOut = new CopyingOutputStream(System.out);
         // this makes sense because CopyingOutputStream is an InputStreamWrapper
@@ -88,12 +83,8 @@ public class ExecCommand {
       }
 
       p.waitFor();
-      
-      // if we have not read the stdout, we do it now
-      if (stdout == null) {
-        stdout = read(in.getInputStream());
-      }
-      return new ExecResult(p.exitValue(), stdout, read(p.getErrorStream()));
+
+      return new ExecResult(p.exitValue(), read(in.getInputStream()), read(p.getErrorStream()));
     
     } finally {
       if (out != null) {
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
@@ -84,9 +84,16 @@ public class ExecCommand {
 
       p.waitFor();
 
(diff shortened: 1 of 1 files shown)


markxnelson:
Can you please explain why you are doing this?  What is the background? Did you find a problem, a failing test?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
```diff
@@ -88,14 +83,17 @@ public static ExecResult exec(
       }
 
       p.waitFor();
-      
-      // if we have not read the stdout, we do it now
-      if (stdout == null) {
-        stdout = read(in.getInputStream());
+
```
markxnelson, line 86: curiously, i don't see this "remove code" in develop at all -- you may need to rebase or pull develop into your branch

doxiao:
The first part of the change is to revert a change that was made in PR#1587. That change was meant to fix a problem - the saved stdout of a command run was empty -  that occurred consistently on develops' machines. The change in PR#187 fixed the problem. But after we enabled the new integration test suite on the external Jenkins, we started to notice intermittent errors in the fix. During investigation, I realized that the fix was not needed. I went back and retested the original code locally and on Jenkins. The original code worked consistently on developers' machines where the original issue was reported. This is why I said on Friday's standup that the original error was a mystery.  Meanwhile, additional testing of the original code on Jenkins hit a different intermittent issue, which has not occurred locally so far. 

The symptom of the newly discovered intermittent issue on Jenkins is that the stdout sometimes was not fully saved, say the end of the content is missing. The second part of the fix is to make sure that the operation of copying the stdout input stream completes (by joining the thread that does the copy) before we read from the duplicated input stream for getting the stdout. With this change I have no longer seen the incomplete stdout problem on Jenkins.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ExecCommand.java
doxiao: I checked and the code is still there in develop branch. I am wondering if you were looking at a same-named class in the old integration test code.

markxnelson: oh i was. sorry, my mistake - the review view truncates the file name and i assumed :( :( :(

markxnelson:
Ok, that is interesting, I have also seen issues where it looks like the stdout is missing the end.  I have fixed one today with #1595 - but that was due to the method throwing a RuntimeException.

doxiao:
> Ok, that is interesting, I have also seen issues where it looks like the stdout is missing the end. I have fixed one today with #1595 - but that was due to the method throwing a RuntimeException.

I actually thought about if we should make the same fix in the current integration test suite as well. But we have not got a smoking gun there. Potentially this can cause problem there too, especially if the test depends on the stdout in the result. 

It is also interesting that I have not seen it happening locally; probably just timing.

vanajamukkara:
@doxiao please update the PR with Jenkins results

doxiao:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/400/console -  in which testCreateMiiDomain passed (which would fail if we hit the issue), although the test run hit a known failure in ItSimpleOperatorValidation.testInstallingOperator.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/397/ (all test case passed using mii-app-patching-test branch which also contains this fix).

PR #1600, opened 2020-04-27
It resilience

markxnelson:
fix code so it retries and does not die on RuntimeException on first iteration.

Observed a failure in this run:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1375/testReport/oracle.kubernetes.operator/ItElasticLogging/testWebLogicLogSearch/

testWebLogicLogSearch fails arund line 1114 of the log (out file).
The first fall to execSearchQuery is getting rc=6 from the curl
command and then throws an RE and prematurely aborts the test.

This PR changes it to use ExecCommand.exec() instead of TestUtils.exec()
so that the result is just returned to the called without killing
the test run.


markxnelson:
new test run, with fix: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1387

PR #1604, opened 2020-04-28
Support for Kubernetes exec command

lennyphan:
1. Added support for executing commands in a container of a pod using Kubernetes client 'exec' API.
2. Added a simple test that uses the 'exec' API to execute readState.sh script, to testCreateMiiDomain, to verify WebLogic server status.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
```diff
@@ -757,4 +767,23 @@ private void checkServiceCreated(String serviceName, String domNamespace) {
 
   }
 
+  private void checkServerReadyStatusByExec(String podName, String namespace) {
+    final V1Pod pod = assertDoesNotThrow(() -> oracle.weblogic.kubernetes.assertions.impl.Kubernetes
```
doxiao, line 771: In general we would not want the test code directly works with low-level Kubernetes objects. But I am not worried about this here because this method is a sample to demonstrate and test the new functionality, which eventually may not needed.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -491,9 +507,15 @@ public static boolean deleteNamespace(String name) {
...
 
     if (response.getObject() != null) {
-      logger.info(
-          "Received after-deletion status of the requested object, will be deleting namespace"
-              + " in background!");
+      V1Namespace namespace = (V1Namespace) response.getObject();
+      String phase = namespace.getStatus().getPhase();
+      logger.fine(
+          "Deletion of namespace " + namespace.getMetadata().getName() + " is in phase: " + phase);
+      if (!phase.equals("Terminating")) {
```
doxiao, line 514: I don't know how long it takes for the phase to be updated to "terminating", and how long it will stay in that phase. Wondering if this check may hit a window, and as a result, reports a false failure.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1327,5 +1352,149 @@ public static boolean deleteClusterRoleBinding(String name) {
...
+  public static ExecResult exec(V1Pod pod, String containerName, boolean redirectToStdout,
+      String... command)
+      throws IOException, ApiException, InterruptedException {
+
+    // Execute command using Kubernetes API
+    KubernetesExec kubernetesExec = createKubernetesExec(pod, containerName);
+    final Process proc = kubernetesExec.exec(command);
+
+    final CopyingOutputStream copyOut =
+        redirectToStdout ? new CopyingOutputStream(System.out) : new CopyingOutputStream(null);
```
doxiao, line 1379: It seems that if the redirectToStdout is false, the following code will still make a copy of the original  input stream while the original input stream can be directly used.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1327,5 +1352,149 @@ public static boolean deleteClusterRoleBinding(String name) {
...
+              () -> {
+                try {
+                  ByteStreams.copy(proc.getInputStream(), copyOut);
+                } catch (IOException ex) {
+                  logger.warning("Exception reading from input stream.", ex);
+                }
+              });
+      out.start();
+
+      // wait for the Process, which represents the executing command, to terminate
```
markxnelson, line 1395: process does not need capital all through here

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1327,5 +1352,149 @@ public static boolean deleteClusterRoleBinding(String name) {
...
+      if (proc != null) {
+        proc.destroy();
+      }
+    }
+  }
+
+  /**
+   * Create an object which can execute commands in Kubertenes container.
+   *
+   * @param pod The pod where the command is run
```
markxnelson, line 1420: same

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1327,5 +1352,149 @@ public static boolean deleteClusterRoleBinding(String name) {
...
+    kubernetesExec.setTty(tty);
+    return kubernetesExec;
+  }
+
   //------------------------
+
+  /**
+   * KubernetesExec object for executing a command in the container of a pod.
+   */
+  public static class KubernetesExecImpl extends KubernetesExec {
```
markxnelson, line 1437: an inline class that implements an abstract class defined in a different file seems really cumbersome

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
```diff
@@ -0,0 +1,40 @@
...
+/** A base class for an object which can execute a command in an Kubertenes containers. */
+public abstract class KubernetesExec {
+  private boolean stdin = true;
+  private boolean tty = true;
+
+  boolean isStdin() {
+    return stdin;
+  }
+
+  public void setStdin(boolean stdin) {
```
markxnelson, line 19: fluent pattern does not use "set" prefix
also this field name is bad - i read "stdin" and i assume you mean the stdin - but you really mean something like "do you want to have a stdin" or something?? choose a better name

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+
+import io.kubernetes.client.openapi.ApiException;
+
+/** A base class for an object which can execute a command in an Kubertenes containers. */
+public abstract class KubernetesExec {
+  private boolean stdin = true;
+  private boolean tty = true;
+
+  boolean isStdin() {
```
markxnelson, line 15: is -> has ??

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
```diff
@@ -0,0 +1,40 @@
...
+  boolean isTty() {
+    return tty;
+  }
+
+  public void setTty(boolean tty) {
+    this.tty = tty;
+  }
+
+  /**
+   * Execute a command in a container.
```
markxnelson, line 32: javadoc does not seem to describe what this abstract method is actually for

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExecFactory.java
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.weblogic.kubernetes.actions.impl.primitive;
+
+import io.kubernetes.client.openapi.models.V1Pod;
```
markxnelson, line 6: way to much OOP going on here -- use the simple functional pattern with static methods and fluent api instead of all this

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
```diff
@@ -757,4 +767,23 @@ private void checkServiceCreated(String serviceName, String domNamespace) {
...
+        .getPod(namespace, null, podName));
+
+    if (pod != null) {
+      ExecResult execResult = assertDoesNotThrow(
+          () -> execCommand(pod, null, true, READ_STATE_COMMAND));
+      if (execResult.exitValue() == 0) {
+        logger.info("execResult: " + execResult);
+        assertEquals("RUNNING", execResult.stdout(),
```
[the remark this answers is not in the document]
rjeberhard: The readState.sh script already trims the rest of the contents of the node manager's server status file, so this is equals.  Of course, the server could be in other statuses, such as "ADMIN". :)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
lennyphan: Yes, I used the low-level Kubernetes call because I know that Pani has a PR that has an action to get the Pod.  I'll change it to use the action in Pani's PR when it eventually gets merged or we can remove this example altogether. IMO we should consider consolidating the Kubernetes API's into one Kubernetes class from which both actions and assertions would then share.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
lennyphan: Instead of waiting for Process.waitFor() to read from the stream directly, I wanted to creating a separate thread (used for redirect) in order to consume the process's output stream to prevent the output stream buffer from becoming full and blocking.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
doxiao: sounds good.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
doxiao: modified code looks good.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Get the Kubernetes pod where the command is to be run.
+   *
+   * @return Kubernetes pod object
+   */
+  public V1Pod getPod() {
+    return pod;
+  }
+
+  public void setPod(V1Pod pod) {
```
ddsharpe, line 64: Why do we need multiple setters for "pod"?  And, are we doing fluent API or not?  This class seems to mix both.  For example line 50 and 64.

lennyphan: yes, it is a mix.  I was looking at the K8s api resource objects and our model objects and fashioned it after them but realized that they have set/get methods in order  to be serialized/de-serialized using the JSON/YAML libraries.  I can clean it up to have the more fluent api look.  Thanks

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
```diff
@@ -531,6 +535,27 @@ public static JsonObject createDockerConfigJson(String username, String password
     return Docker.createDockerConfigJson(username, password, email, registry);
   }
 
+  // ----------------------- Execute a Command   ---------------------------
+
+  /**
+   * Execute a command in a container.
+   *
+   * @param pod The pod where the command is run
```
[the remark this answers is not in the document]
lennyphan: Fixed this syntax changes in all the files.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/TestActions.java
```diff
@@ -531,6 +535,27 @@ public static JsonObject createDockerConfigJson(String username, String password
...
+   * @param containerName The container in the Pod where the command is run. If no container
+   *     name is provided than the first container in the Pod is used.
+   * @param redirectToStdout copy Process output to stdout
```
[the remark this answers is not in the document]
lennyphan: Removed capital from all files

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
lennyphan: Simplified the implementation as your suggestion below.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExecFactory.java
lennyphan: Simplified the implementation to be more functional.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
lennyphan: Removed set/get methods since not necessary.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
markxnelson: is run -> is to be run
same on next line

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -1327,5 +1368,127 @@ public static boolean deleteClusterRoleBinding(String name) {
...
+        .apiClient(apiClient) // the Kubernetes api client to dispatch the "exec" command
+        .pod(pod) // The pod where the command is to be run
+        .containerName(containerName) // the container in which the command is to be run
+        .passStdinAsStream() // pass a stdin stream into the container
+        .stdinIsTty(); // stdin is a TTY (only applies if stdin is true)
+  }
+
   //------------------------
+
+  private static String read(InputStream is) throws IOException {
```
markxnelson, line 1455: this method names seems too generic - what does `read` mean?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/KubernetesExec.java
```diff
@@ -0,0 +1,69 @@
...
+    this.containerName = containerName;
+    return this;
+  }
+
+  public KubernetesExec passStdinAsStream() {
+    this.passStdinAsStream = true;
+    return this;
+  }
+
+  public KubernetesExec stdinIsTty() {
```
ddsharpe, line 53: Javadoc in this class is VERY sparse.  Please increase the coverage

lennyphan: Updated with javadoc.

PR #1605, opened 2020-04-28
Attempt to clean up use of bad exec()

markxnelson:
This PR attempts to resolve a number of test failures that occur due
to the bad use of the TestUtils.exec() method.  This method throws a
RuntimeException if the return code of the executed command is non-zero.
This method is being used in many places (over 40).  Many of those do
not appear to need or want that behavior, and should instead be using
the ExecCommand.exec() method which executes a command and then returns
the result, without throwing an exception if the command fails.

In this PR, I have done a visual inspection of each use of
TestUtils.exec() and changed them to ExecCommand.exec() where it
appeared to make sense.  This includes every case where the call was
using the return value from the method, and those where it appears clear
from the subsequent logic that a failed command was not expected to
terminate the process.

There were some calls that I left because I could not be sure of the
intent, and these will require more study.

In order to make the code clearer, and make people think twice about
using TestUtils.exec(), which has caused so many failures, I renamed
that method to TestUtils.execOrAbortProcess().  Hopefully that will make
it clearer that that method has a dangerous and unexpected side effect.

Note: This PR does include unmerged changes from PR 1601

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/ItElasticLogging.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/ItElasticLogging.java
@@ -104,7 +104,7 @@ public class ItElasticLogging extends BaseTest {
                 .append("/")
                 .append(elasticStackYamlLoc);
         LoggerHelper.getLocal().log(Level.INFO, "Command to Install Elastic Stack: " + cmd.toString());
-        TestUtils.exec(cmd.toString());
+        TestUtils.execOrAbortProcess(cmd.toString());
 
         LoggerHelper.getLocal().log(Level.INFO, "Creating Operator & waiting for the script to complete execution");
         Map<String, Object> operatorMap = createOperatorMap(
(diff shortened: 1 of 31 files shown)


markxnelson:
Queued up a build https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1389/

integration-tests/src/test/java/oracle/kubernetes/operator/ItElasticLogging.java
```diff
@@ -514,12 +516,14 @@ private String execSearchQuery(String queryCriteria, String index)
             .append("'")
             .toString();
     LoggerHelper.getLocal().log(Level.INFO, "Command to search: " + cmd);
-    ExecResult result = TestUtils.exec(cmd);
+    ExecResult result = TestUtils.execOrAbortProcess(cmd);
```
vanajamukkara, line 519: we looked into this test yesterday, this should be ExecCommand.exec() ?

vanajamukkara:
LGTM except one I commented above and also some places replaced with ExecCommand.exec() where TestUtils.execOrAbortProcess() can also work, this is fine ASA the method throws Exception.

markxnelson:
test run has 81 passing, 12 failing
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1389/testReport/

PR #1608, opened 2020-04-28
added option to configure operator pod resource requests and limit

ankedia:
Make resource requests and limits configurable for operator pod. Resource requests and limits can be configured by specifying below 4 properties in values yaml file.  
cpuRequests  
memoryRequests
cpuLimits
memoryLimits

### DIFF
--- a/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
+++ b/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
@@ -58,8 +58,15 @@ spec:
         {{- end }}
         resources:
           requests:
-            cpu: "100m"
-            memory: "512Mi"
+            cpu: {{ .cpuRequests | default "100m" }}
+            memory: {{ .memoryRequests | default "512Mi" }}
+          limits:
+            {{- if .cpuLimits}}
+            cpu: {{ .cpuLimits }}
+            {{- end }}
+            {{- if .memoryLimits}}
+            memory: {{ .memoryLimits }}
+            {{- end }}
         volumeMounts:
         - name: "weblogic-operator-cm-volume"
           mountPath: "/operator/config"
--- a/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
+++ b/kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
@@ -58,7 +58,7 @@ spec:
         {{- end }}
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -58,8 +58,15 @@ spec:
         {{- end }}
         resources:
           requests:
-            cpu: "100m"
-            memory: "512Mi"
+            cpu: {{ .cpuRequests | default "100m" }}
+            memory: {{ .memoryRequests | default "512Mi" }}
+          limits:
```
tbarnes-us, line 63: Should the defaults be increased?  This question is for both Anil & @rjeberhard

ankedia: Yes, I agree that defaults should be increased at least for cpu request. With cpu request of "100m", JVM uses "defNew" and "SerialOld" GC (single threaded). Even when I configured JVM to use G1 GC, JVM ran with a parallel GC but with only one GC thread. However, I'm not aware of reason for previous default value of "100m". I wanted to get @rjeberhard Ryan's thoughts on the default values and I can increase them accordingly.

ankedia:
> nothing to add - come up with some sensible defaults, and then submit the PR

Ok, thanks. Please note that this PR is not changing the defaults, it's just making these values configurable. IMO a good default value will vary based on the use-case and environment. Tom and I discussed about this, one suggestion is to increase cpu request default value to "250m" (4 core laptop running 16 pods). Ryan and Mark, please weigh in with your thoughts.

markxnelson:
I need to read up on cpu limits to make a sensible comment

rjeberhard:
LGTM.  My thought is that we should set the default CPU to "250m" to get this in and then re-evaluate the default after all of the "synchronous calls from a fiber" bugs are fixed.

ankedia:
> LGTM. My thought is that we should set the default CPU to "250m" to get this in and then re-evaluate the default after all of the "synchronous calls from a fiber" bugs are fixed.

Thanks Ryan. I have changed the default CPU to "250m" and pushed the commit. I have re-requested review from Mark. As you suggested, we will re-evaluate defaults after "sync calls from fiber" bugs are fixed.

ankedia:
> I need to read up on cpu limits to make a sensible comment

I have changed the default CPU to "250m" as suggested by Ryan and will re-evaluate this after other changes are made. Thanks.

Issue #1610, opened 2020-04-29
Scripts under create-rcu-schema does not support different namespace, DB sys and schema password and uses scott demo user


ashageetha:
Scripts under create-rcu-schema has below issues and will not work in some use-cases as listed below:
1)  Uses namespace “default” which may not work if the user does not have access to it and need to provide specific namespace.
2) "rcu.yaml" always uses imagePullSecret even if docker-secret is not passed.
3) Both sys and schema database password is hard coded to “Oradoc_db1” and does not give option  to provide different value.
     echo “Oradoc_db1” > pwd.txt
    echo “Oradoc_db1" >> pwd.txt
4) Create and drop rcu schema scripts both uses scott demo user to test the connectivity before proceeding with create or drop schemas which may not be present in fresh install database. Hence need to either use system user.
5) "rcu.yaml" is directly edited and may not be usable to use the same repo for another scenarios.

tbarnes-us:
I'm assuming the issue is resolved by now, feel free to re-open the issue if I'm incorrect.

PR #1611, opened 2020-04-29
WILL NOT MERGE: Fix for create-rcu-schema scripts (issue 1610)

ashageetha:
Hi @markxnelson @rjeberhard  @mriccell  @rdas0405,

This PR has fixes for #1610 -
- Support for namespace
- Fix for image pull secret in rcu yaml
- Support for sys and schema passwords
- Use output dir for creating rcu.yaml based on the template rather updating existing common/rcu.yaml directly
- Use sys user instead of demo user scott which may not be available in all Database for connectivity check.
- Update README.md with latest changes.
  
Please review and approve these changes.

Thanks,
Asha


markxnelson:
this needs to be targeted at `develop` - we do not update things in master like this without comprehensive testing first

PR #1619, opened 2020-05-05
Fix assertion and add events log to the logs list

sankarpn:
Add namespaced events in the collected logs list
Fix the persistent volume claim log file name
Fix the isPodReady test assertion to check for pod readiness instead of Running status
Change the domain/operator wait time from 5 to 10 minutes.


sankarpn:
> 
> 
> LGTM. Similar change is needed to check the server pods are ready, Xian has the change in her PR#1614

Yes, I changed the server pods ready check to for pod readiness instead of Running status. Its in TestAssertion.java, line # 98

PR #1623, opened 2020-05-05
Updated versions of Grafana, Prometheus, added extra debugging

marinakog:
1. Updated to newest versions of Helm Chart for Grafana and Prometheus ( it incomparable with current k8s and helm version, used in current external Jen run
2. Split some shell scripts and added extra debugging 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1437/ - passed for MonExporter tests ( one test fail due generic connectivity issues)


integration-tests/src/test/java/oracle/kubernetes/operator/ItMonitoringExporter.java
```diff
@@ -179,35 +173,35 @@ public void prepare() throws Exception {
...
-
-          String image = System.getenv("REPO_REGISTRY")
-              + "/weblogick8s/"
-              + domainNS2
-              + "-image:" + currentDateTime;
-          String cmd = "docker rmi -f " + image;
-          TestUtils.execOrAbortProcess(cmd, true);
-        }
-        String cmd = "docker rmi -f " + domainNS2 + "-image:" + currentDateTime;
+    Assumptions.assumeTrue(QUICKTEST);
```
sankarpn, line 176: FULLTEST or QUICKTEST ?

marinakog: fixed , thank you

integration-tests/src/test/java/oracle/kubernetes/operator/ItMonitoringExporter.java
```diff
@@ -179,35 +173,35 @@ public void prepare() throws Exception {
...
-        uninstallWebHookPrometheusGrafanaMySql();
-      } catch (Exception ex) {
-        LoggerHelper.getLocal().log(Level.INFO,
-            "Exception caught while uninstalling webhook/prometheus/Grafana/Mysql " + ex.getMessage());
-      }
-
-      tearDown(new Object() {}.getClass().getEnclosingClass().getSimpleName(), namespaceList.toString());
-      LoggerHelper.getLocal().log(Level.INFO,"SUCCESS");
+      String cmd = "docker rmi -f " + domainNS2 + "-image:" + currentDateTime;
+      TestUtils.execOrAbortProcess(cmd, true);
```
sankarpn, line 190: If its a SHARED_CLUSTER both docker rmi -f commands get executed or line # 189 is supposed to be in a else block?

marinakog: yes, if shared cluster it also tags image to [ocipath]/imagename as well, so need to delete both images [ocipath]/imagename in else block for shared cluster and "imagename" in general

integration-tests/src/test/resources/deploywebapp.py
```diff
@@ -1,22 +1,12 @@
...
-#             curstate4=cmo.getCurrentState(appName.getName(),targetinst)
-#             print '-----------', curstate4, '-----------', appName.getName()
\ No newline at end of file
+appName=sys.argv[4]
+appPath=sys.argv[5]
+clusterName=sys.argv[6]
+print 'Deploying application[%s] to the Cluster[%s] located @ [%s] ' %(appName,clusterName,appPath)
+print 'Checking the server status on the Cluster[%s] ' %(clusterName)
+domainRuntime()
+state(clusterName,'Cluster')
```
sankarpn, line 10: is the check is only for debugging purposes? or deploy needs to be aborted when cluster state is not in the expected state.

integration-tests/src/test/resources/deploywebapp.py
```diff
@@ -1,22 +1,12 @@
...
+#deploy(sys.argv[4],sys.argv[5],sys.argv[6],upload='false',remote='false')
+deploy(appName=appName, path=appPath, targets=clusterName)
```
sankarpn, line 12: deploy works even when in domainRuntime? or you have to be in config tree. Just want to make sure.

integration-tests/src/test/resources/exporter/createGrafana.sh
```diff
@@ -5,23 +5,57 @@ monitoringExporterDir=$1
...
-
+CHARTNS=monitortestns
+CHARTNAME=grafana
 HELM_VERSION=$(helm version --short --client)
 
+function checkPod() {
+  max=20
+  count=0
+  echo "Checking pods $CHARTNAME"
+  grafanapod=$(kubectl get po -n $CHARTNS | grep $CHARTNAME | awk '{print $1 }')
```
sankarpn, line 16: At this point you are sure that the grafana pod at least created ?

integration-tests/src/test/resources/deploywebapp.py
marinakog: it is just for debugging purpose, I used code from Vanaja, she is investigating the deploy issue

integration-tests/src/test/resources/deploywebapp.py
marinakog: yes it is existed code I just modified for debug purposes

integration-tests/src/test/resources/exporter/createGrafana.sh
marinakog: the grafanapod value is used only for debugging of failure. There is another check for helm install success before calling this code. So at this point pod is created but might has failures

integration-tests/src/test/java/oracle/kubernetes/operator/ItMonitoringExporter.java
```diff
@@ -260,12 +254,19 @@ private static void executeShelScript(String srcLoc, String destLoc, String file
             + args
             + " | tee "
             + outLogFile;
-    TestUtils.execOrAbortProcess(crdCmd, true);
-    crdCmd = " cat " + destLoc + "/" + outLogFile;
-    ExecResult result = ExecCommand.exec(crdCmd);
-    assertFalse(
-        result.stdout().contains("BUILD FAILURE"), "Shell script failed: " + result.stdout());
+    ExecResult result = TestUtils.execOrAbortProcess(crdCmd, true);
+    if (result.exitValue() != 0) {
```
markxnelson, line 258: if the exit value was not zero, you will never get to this line.
if you need the result, you have to use `ExecCommand.exec()` instead

integration-tests/src/test/resources/deploywebapp.py
markxnelson, line 12: @ddsharpe @rjeberhard do we need some kind of `write()` or something at the end of this script?  or at least a `disconnect()` or whatever??

integration-tests/src/test/java/oracle/kubernetes/operator/ItMonitoringExporter.java
marinakog: replaced with execCommand.exec()

integration-tests/src/test/resources/deploywebapp.py
marinakog: reverted this file to original, it was added only for debugging

PR #1624, opened 2020-05-05
update helm ls

rosemarymarano:
OWLS-80278 - Helm 3 documentation needs to inform user that `helm list` is per namespace, part II. Edit `helm ls` command in Managing Domain Namespaces FAQ.

### DIFF
--- a/docs-source/content/faq/namespace-management.md
+++ b/docs-source/content/faq/namespace-management.md
@@ -51,10 +51,11 @@ suspendOnDebugStartup: false
 
 ```
 
-If you don't know the release name of the operator, you can use `helm ls` to list all the releases:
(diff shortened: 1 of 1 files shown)


docs-source/content/faq/namespace-management.md
```diff
@@ -51,10 +51,11 @@ suspendOnDebugStartup: false
 
 ```
 
-If you don't know the release name of the operator, you can use `helm ls` to list all the releases:
+If you don't know the release name of the operator, you can use `helm ls` to list all the releases for a specified namespace or all namespaces:
 
 ```
-$ helm ls
+$ helm ls --namespace <namespace>
```
markxnelson, line 57: should use `ls` or `list` in both

rosemarymarano: Absolutely; typo on my part.

PR #1626, opened 2020-05-05
Script to run new integration tests against a "kind" cluster

rjeberhard:
This PR introduces a ./kindtest.sh that creates a Kubernetes cluster running in Docker containers and then runs the new integration tests against that cluster.

Presently, the cluster has a master and one worker node and both are running Kubernetes 1.15.7.  It would be fairly straightforward to make the Kubernetes version configurable.

The worker node is configured with "volume mounts" that I know from experiments with the original integration test suite are such that they will work for domain-on-pv or logs-on-pv use cases.

The `K8S_NODEPORT_HOST` variable is set to the address of the worker node.  From the Kind documentation, this should consistently work on Linux, but I've not tested on MacOS.

Nodes in a Kind cluster do not have access to the Docker daemon on the host machine, but I've been able to find a pattern that works.  The script starts a Docker repository in another container and connects this to the network of the Kubernetes nodes.  This is an insecure registry (doesn't require credentials to push or pull).  Then I've adjusted the tests to use this repository name.  Since the tests are consistently using the TestConstants this was, thankfully, a small change to the test base.


rjeberhard:
Read about Kind: https://kind.sigs.k8s.io/

To install:
```
curl -Lo ./kind "https://kind.sigs.k8s.io/dl/v0.8.0/kind-$(uname)-amd64"
chmod +x ./kind
mv ./kind /some-dir-in-your-PATH/kind
```

kindtest.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
markxnelson, line 1: copyright notice

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+set -o errexit
+
+if [[ -z "$WORKSPACE" ]]; then
+  kinddir="/scratch/$USER/kindtest"
```
markxnelson, line 5: maybe we could print a usage message and stop, rather than arbitrarily chose this location?

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+else
+  kinddir="${WORKSPACE}/logdir/${BUILD_TAG}"
+fi
+mkdir -m777 -p "${kinddir}"
```
markxnelson, line 9: does it matter it is was already there?

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+
+export PV_ROOT="${kinddir}/k8s-pvroot"
+if [ -d "$PV_ROOT" ]; then
+  rm -Rf "$PV_ROOT/*"
+else
+  mkdir -m777 "$PV_ROOT"
+fi
+
+echo 'Remove old cluster (if any)...'
+kind delete cluster
```
markxnelson, line 25: should this have `--name ${KIND_CLUSTER_NAME}`?

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+
+reg_host="${reg_name}"
+if [ "${kind_network}" = "bridge" ]; then
+    reg_host="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "${reg_name}")"
+fi
+echo "Registry Host: ${reg_host}"
+
+echo 'Create a cluster with the local registry enabled in containerd'
+kind_version='v1.15.7@sha256:e2df133f80ef633c53c0200114fce2ed5e1f6947477dbc83261a6a921169488d'
```
markxnelson, line 55: make this a param? from a list of "approved" ones?

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+    endpoint = ["http://${reg_host}:${reg_port}"]
+nodes:
+  - role: control-plane
+    image: kindest/node:${kind_version}
+  - role: worker
+    image: kindest/node:${kind_version}
+    extraMounts:
```
markxnelson, line 68: do we need to expose 6443 or any node ports?

kindtest.sh
```diff
@@ -0,0 +1,102 @@
...
+      - hostPath: ${PV_ROOT}
+        containerPath: ${PV_ROOT}
+EOF
+
+kubectl cluster-info --context kind-kind
```
markxnelson, line 73: shouldn't this use your cluster name var?

kindtest.sh
rjeberhard: agreed... working on that now.  I needed this once I was testing against 1.16.

kindtest.sh
rjeberhard: no, not on Linux.  The kind documentation says that it's automatic for Linux networking.

kindtest.sh
rjeberhard: Yes, it should.  thanks.

kindtest.sh
rjeberhard: Ugh, okay... that's embarrassing. Thanks.

kindtest.sh
rjeberhard: I was trying to have a useful default when running on your own machine.  I think that is still valuable.  I can make another argument.

kindtest.sh
rjeberhard: No, I'm actually okay with that.  I'd like to be able to run repeatedly on my local box.

kindtest.sh
markxnelson: 2019?

kindtest.sh
markxnelson: ok cool

kindtest.sh
```diff
@@ -0,0 +1,143 @@
...
+    image: ${kind_image}
+  - role: worker
+    image: ${kind_image}
+    extraMounts:
+      - hostPath: ${PV_ROOT}
+        containerPath: ${PV_ROOT}
+EOF
+
+kubectl cluster-info --context "kind-${kind_name}"
+kubectl get node -o wide
```
sankarpn, line 115: This one is supposed to get the nodes from kind cluster or localhost kubernetes cluster ?

kindtest.sh
rjeberhard: More coffee needed... Thanks for having my back.  I'm not with it today.

kindtest.sh
rjeberhard: This gets the nodes from the kind cluster.  The call above (line 114) sets the current context to the kind cluster.  All kubectl calls after this will be to the kind cluster until a different context is set.

sankarpn: Got it. command cluster-info is kind of misleading :-(

markxnelson: oh yeah i did not notice that, a `kubectl config set-context ` might be clearer, but not a biggie, if it works, fair enough

rjeberhard: I'm happy to switch it... I used this command because after you create a kind cluster, the tool prints out this command as the suggested way to select the cluster's context.

kindtest.sh
```diff
@@ -0,0 +1,187 @@
...
+  echo "Unsupported Kubernetes version: ${k8s_version}"
+  exit 1
+fi
+
+echo "Using Kubernetes version: ${k8s_version}"
+
+mkdir -m777 -p "${outdir}"
+export RESULT_ROOT="${outdir}/wl_k8s_test_results"
+if [ -d "${RESULT_ROOT}" ]; then
+  rm -Rf "${RESULT_ROOT}/*"
```
tbarnes-us, line 93: RESULT_ROOT is set just above,  but never-the-less an rm -rf $ENV/* of env var looks a bit like playing with fire to me.  A "set -u" would make it a little safer maybe.  Putting the 'wl_k8s_test_results' string directly  in the rm expression makes it even safer IMO.  Ditto for PV_ROOT below.  This reduces chance that a future change  to the env var logic could leave a crater in the file system.

PR #1627, opened 2020-05-06
OWLS-80960 - Allow replicas count to drop below min dynamic cluster size

alai8:
Introduce a new Cluster CRD element `allowReplicasBelowMinDynClusterSize` that if set to true (the default), the operator would allow replicas count to be set below the minimum dynamic cluster size configured in the WebLogic domain home configuration. 

This allows users to shut down all servers in a dynamic cluster by setting the cluster replicas count to 0. This also prevents managed servers in a dynamic cluster from starting by operator even if the replicas is 0. This was the behavior before OWLS-79994.

To prevent WLDF from scaling below the min dynamic cluster size, set  `allowReplicasBelowMinDynClusterSize` to false.

### DIFF
--- a/docs/domains/Domain.json
+++ b/docs/domains/Domain.json
@@ -109,6 +109,10 @@
           "description": "The name of this cluster. Required",
           "type": "string"
         },
+        "allowReplicasBelowMinDynClusterSize": {
+          "description": "Whether to allow the number of replicas to drop below the minimum dynamic cluster size configured in the WebLogic domain home configuration.",
+          "type": "boolean"
+        },
         "serverPod": {
           "description": "Configuration affecting server pods.",
           "$ref": "#/definitions/ServerPod"
--- a/docs/domains/Domain.json
+++ b/docs/domains/Domain.json
@@ -110,7 +110,7 @@
           "type": "string"
         },
         "allowReplicasBelowMinDynClusterSize": {
-          "description": "If true (the default), the number of replicas is allowed to drop below the minimum dynamic cluster size configured in the WebLogic domain home configuration. Otherwise, the operator will ensure that the number of replicas is not below the minimum dynamic cluster. This setting applies to dynamic clusters only.",
+          "description": "If true (the default), then the number of replicas is allowed to drop below the minimum dynamic cluster size configured in the WebLogic domain home configuration. Otherwise, the operator will ensure that the number of replicas is not less than the minimum dynamic cluster setting. This setting applies to dynamic clusters only.",
           "type": "boolean"
         },
         "serverPod": {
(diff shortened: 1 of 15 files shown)


docs/domains/Domain.md
```diff
@@ -77,6 +77,7 @@ An element representing a cluster in the domain configuration.
 
 | Name | Type | Description |
 | --- | --- | --- |
+| `allowReplicasBelowMinDynClusterSize` | Boolean | If true (the default), the number of replicas is allowed to drop below the minimum dynamic cluster size configured in the WebLogic domain home configuration. Otherwise, the operator will ensure that the number of replicas is not below the minimum dynamic cluster. This setting applies to dynamic clusters only. |
```
rosemarymarano, line 80: If true (the default), the number -> If true (the default), then the number
that the number of replicas is not below the minimum dynamic cluster -> that the number of replicas is not less than the minimum dynamic cluster setting.

docs/domains/index.html
```diff
@@ -1029,6 +1029,10 @@
           "description": "The name of this cluster. Required",
           "type": "string"
         },
+        "allowReplicasBelowMinDynClusterSize": {
```
rosemarymarano, line 1032: If true (the default), the number -> If true (the default), then the number
that the number of replicas is not below the minimum dynamic cluster -> that the number of replicas is not less than the minimum dynamic cluster setting.

docs/domains/Domain.md
alai8: Thanks for the comment. I have updated the description to:

If true (the default), then the number of replicas is allowed to drop below the minimum dynamic cluster size configured in the WebLogic domain home configuration. Otherwise, the operator will ensure that the number of replicas is not less than the minimum dynamic cluster setting. This setting applies to dynamic clusters only.

docs/domains/index.html
alai8: ditto

docs/domains/Domain.json
```diff
@@ -109,6 +109,10 @@
           "description": "The name of this cluster. Required",
           "type": "string"
         },
+        "allowReplicasBelowMinDynClusterSize": {
```
tbarnes-us, line 112: Should this also be a global setting?  On the one hand, it's extra work and I'm not sure if the extra work is worth it at this time. On the other hand, there's precedent where 'replicas' itself is both global and per-cluster.  

One way to handle this would be check-in 'as-is' and file a JIRA to add a global setting...

tbarnes-us:
This LGTM overall. Made one minor comment above that I don't think is a show stopper. Delegating sign-off to Ryan.

docs/domains/Domain.json
alai8: Filed [OWLS-81769](https://jira.oraclecorp.com/jira/browse/OWLS-81769)

Issue #1457, opened 2020-03-03
SSL Enabled: false


dborysenko:
Setting "SSL enabled: false" causing introspectDomain.py script to fail with NullPointerException.

my wdt model:
```yaml
topology:
    AdminServerName: AdminServer
    Name: domain1
    ProductionModeEnabled: true
    Cluster:
        osb_cluster_1:
    Server:
        AdminServer:
            NetworkAccessPoint:
                T3Channel:
                    ListenPort: 30012
                    PublicAddress: "myexamplehost.us-east-2.eks.amazonaws.com"
                    PublicPort: 30012
            SSL:
                Enabled: false
                ListenPort: 7002
                HostnameVerificationIgnored: true
                ServerPrivateKeyPassPhraseEncrypted: 'passphrasechanged'
                JSSEEnabled: true
                ServerPrivateKeyAlias: DemoIdentity
        osb_server1:
            Cluster: osb_cluster_1
            ListenPort: 8001
            NumOfRetriesBeforeMsiMode: 0
            RetryIntervalBeforeMsiMode: 1
            JTAMigratableTarget:
                Cluster: osb_cluster_1
                UserPreferredServer: osb_server1
            CustomIdentityKeyStoreType: jks
            CustomIdentityKeyStoreFileName: wlsdeploy/servers/osb_server1/DemoIdentity.jks
            CustomIdentityKeyStorePassPhraseEncrypted: 'passphrasechanged'
            CustomTrustKeyStorePassPhraseEncrypted: 'passphrasechanged'
            CustomTrustKeyStoreType: jks
            CustomTrustKeyStoreFileName: wlsdeploy/servers/osb_server1/myTrust.jks
            ListenPort: 8011
            SSL:
                Enabled: false
                ListenPort: 7002
                HostnameVerificationIgnored: true
                ServerPrivateKeyPassPhraseEncrypted: 'passphrasechanged'
                JSSEEnabled: true
                ServerPrivateKeyAlias: DemoIdentity
    MigratableTarget:
        'osb_server1 (migratable)':
            Cluster: osb_cluster_1
            UserPreferredServer: osb_server1
```

exception during introspect phase:
```
@[2020-03-02T22:59:21.105 UTC][startNodeManager.sh:131][FINE] NODEMGR_HOME='/u01/nodemanager/domain1/introspector'
[...]

PR #1631, opened 2020-05-07
fix issue 1457 (and sundry)

tbarnes-us:
- issue https://github.com/oracle/weblogic-kubernetes-operator/issues/1457: 
  - update introspector job's jython script to work around an off-line WLST bug in SSL bean access (an attempt to access the bean when it's disabled throws an NPE)
  - update 'old WDT integration test' to reproduce/test

- fix stand-alone introspector test's MII path (depended on scripts that were removed)

- add option to old integration tests to bypass its desire for a docker login when running locally

### DIFF
--- a/integration-tests/src/test/resources/setupenv.sh
+++ b/integration-tests/src/test/resources/setupenv.sh
@@ -9,31 +9,41 @@ function cleanup {
 function create_image_pull_secret_wl {
 
   set +x 
+
+  if [ "$LOCAL_RUN" = "true" ]; then
+    if [ ! -z "$(docker images -q $IMAGE_NAME_WEBLOGIC:$IMAGE_TAG_WEBLOGIC)" ]; then
+      return 0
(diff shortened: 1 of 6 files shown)


tbarnes-us:
> i am ok for this an an interim step, but i really want to start getting away from all these scripts (that are difficult to test and to make portable) and have this stuff implemented in our Junit5 framework like the rest of the tests please

- if you're referring to the modified integration test, the modification will make its way to Junit5 once the old integration test is ported to Junit5. this is the long term plan for all of the old tests.

- if you're referring to the shell-script-operator-emulator + yaml templates 'instrospector' test, there's a JIRA tracking the need to get into the new framework. note that the new framework is kind of a set of square holes for that particular test's round pegs since this particular introspector test is closer to an internal unit test than an integration test. i'm sure that will change over time - but we should discuss if the process needs to be accelerated

PR #1635, opened 2020-05-08
Endless status updates

russgold:
This address two issues that caused the operator to send status updates repeatedly:

1. Our server status reader sometimes fails to get the status for a server. The code was defaulting to SHUTDOWN, meaning that we would intermittently change the state of a server to SHUTDOWN and then back to the real value. The code now defaults to the previous state of a server, if known.
2. The ServerStatus object has a volatile field called `isAdminServer` used for sorting to ensure that the admin server is listed first in the domain status; unfortunately, it was also included in the equals() method, meaning that it always looked as though the status had changed and needed to be updated. Since the field is volatile, that did nothing, so we updated the status about once a second. It has now been removed from equals() and hashCode().

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
@@ -322,8 +322,8 @@ public class DomainStatus {
   private ServerStatus getMatchingServer(ServerStatus server) {
     return getServers()
           .stream()
-          .filter(s -> s.getClusterName().equals(server.getClusterName()))
-          .filter(s -> s.getServerName().equals(server.getServerName()))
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -51,7 +52,7 @@
   @Description("Status of WebLogic Servers in this domain.")
   @Valid
   // sorted list of ServerStatus
-  List<ServerStatus> servers = new ArrayList<>();
+  private final List<ServerStatus> servers;
```
alai8, line 55: should we make clusters private too?

russgold: Shoot. I thought I had :)

Yeah, make it a TBD.

PR #1637, opened 2020-05-08
retry for app deployment and other fixes

vanajamukkara:
1. add checks to resolve the managed server service names from admin pod
2. print cluster status in py script before deploy
3. retry for app deployment
4. fixed managed coherence and multiple cluster tests for pods ready check and services created check.

With all these changes, I see good results on cluster-10. On cluster-1(quicktest) there are still ~15 failures on this branch vs ~30 failures on develop branch.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1464/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1460/

Only two failures on cluster-10 - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest-10/74/

### DIFF
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
@@ -1440,6 +1440,7 @@ public class TestUtils {
         }
       } else {
         LoggerHelper.getLocal().log(Level.INFO, "Application deployed successfully, stdout = " + result.stdout());
+        break;
       }
     }
 
--- a/integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
+++ b/integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
@@ -1401,7 +1401,27 @@ public class TestUtils {
    * @throws Exception exception
    */
   public static void callShellScriptByExecToPod(
-      String podName, String domainNS, String scriptsLocInPod, String shScriptName, String[] args)
+      String podName, String domainNS, String scriptsLocInPod, String shScriptName,
+      String[] args)
+      throws Exception {
+    callShellScriptByExecToPod(
+        podName, domainNS, scriptsLocInPod, shScriptName,args, 1);
+  }
+
+  /**
+   * exec into the pod and call the shell script with given arguments.
+   *
+   * @param podName         pod name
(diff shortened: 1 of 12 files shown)


integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -228,6 +228,51 @@ public static void checkServiceCreated(String serviceName, String domainNS, int
     }
   }
 
+  /**
+   * Resole a servive name inside a pod.
```
rjeberhard, line 232: Resole -> Resolve

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
```diff
@@ -228,6 +228,51 @@ public static void checkServiceCreated(String serviceName, String domainNS, int
...
+   */
+  public static void resolveServiceName(
+      String serviceName, String podName, String namespace, int numberOfRetries)
+      throws Exception {
+
+    String appLocationInPod = BaseTest.getAppLocationInPod();
+    String scriptLocationInPod = appLocationInPod + "/resolveServiceName.py";
+
+    // copy py script to check if managed service name can be resolved in admin pod
+    TestUtils.kubectlexec(podName, namespace," -- mkdir -p " + appLocationInPod);
```
rjeberhard, line 246: Surprisingly, this will break on k8s 1.16.  The `kubectl exec` command has always optionally taken a "--" as the separator between the kubectl arguments and the command that you want to run, but in 1.16 the parsing changes so that if you have this value it is taken as the beginning of the command.  We could fix this by always adding a "--" before the command on the `kubectl exec` command line.

vanajamukkara: @rjeberhard currently the command looks like this 
```
kubectl -n usab-domainns-4 exec -it usab-domain-4-admin-server  -- mkdir -p /u01/oracle/apps
```
what will be the command based on your suggestion?

integration-tests/src/test/java/oracle/kubernetes/operator/utils/TestUtils.java
vanajamukkara: done

PR #1641, opened 2020-05-12
update doc for K8s resource names

rosemarymarano:
OWLS-81892 - Update doc. for camel case names of Kubernetes resources. I found about 50 or so Kubernetes resource types, https://kubernetes.io/docs/reference/kubectl/overview/#resource-types. Our docs have hundreds of instances of them. I adopted the Kubernetes resource capitalization only when we referred to the resource itself, and used lower case in general discussion.


docs-source/content/samples/simple/domains/domain-home-in-image/_index.md
```diff
@@ -1,11 +1,11 @@
----
+PersistentVolumeClaim---
```
rjeberhard, line 1: Just curious... What does this mean at the top of the file?

rosemarymarano: I'd say "that was a boo-boo", an errant replacement that needs to be fixed.

PR #1643, opened 2020-05-12
Async http client change

ankedia:
Changes for async invocation of HTTP requests to read health of server instances using JDK11 java client (instead of JAX-RS HttpClient).


operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -34,4 +34,8 @@
   public static final String DOMAIN_INTROSPECT_VERSION = "weblogic.domainIntrospectVersion";
   public static final String DOMAIN_INPUTS_HASH = "weblogic.domainInputsHash";
   public static final String REMAINING_SERVERS_HEALTH_TO_READ = "serverHealthRead";
+  
+  public static final String KEY = "encodedCredentials";
+  public static final String RESULT = "httpResult";
+  public static final int HTTP_OK = 200;
```
rjeberhard, line 40: Very small thing ;)... This constant must be many places.  Do we need another? I bet it's available on the HTTP client classes.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientFactory.java
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
```
rjeberhard, line 1: New file... "2018, 2020," -> "2020,"

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
```diff
@@ -0,0 +1,118 @@
...
+      if (factory == null) {
+        factory = FACTORY;
+      }
+
+      httpClient = factory.get();
+    } catch (Throwable e) {
+      LOGGER.warning(MessageKeys.EXCEPTION, e);
+    }
+
+    if (isFirst.compareAndSet(true, false)) {
```
rjeberhard, line 80: This is copy/paste from K8s master logic... Not needed here.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
```diff
@@ -0,0 +1,118 @@
...
+    };
+  }
+
+  public static HttpClientPool getInstance() {
+    return SINGLETON;
+  }
+
+  @Override
+  protected HttpClient create() {
+    return instance.updateAndGet(
```
rjeberhard, line 55: Trying to figure out what you're doing here.  I thought you wanted to pool HttpClient instances, but you are creating a single instance and returning it every time.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -128,47 +242,62 @@ public NextAction apply(Packet packet) {
...
+                                    MessageKeys.HTTP_METHOD_FAILED,
+                                    "POST",
+                                    url,
+                                    input.getResponse());
+                              } else {
+                                packet.put(ProcessingConstants.RESULT, input);
+                              }
+                              fiber.resume(packet);
+                            });
+                  } catch (Exception e) {
```
rjeberhard, line 296: probably want Throwable here

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -128,47 +242,62 @@ public NextAction apply(Packet packet) {
...
+                            .build();
+                    httpClient
+                        .sendAsync(request, HttpResponse.BodyHandlers.ofString())
+                        .thenApply(
+                            response ->
+                                new Result(
+                                    response.body(),
+                                    response.statusCode(),
+                                    response.statusCode() == ProcessingConstants.HTTP_OK))
+                        .whenComplete(
```
rjeberhard, line 283: Is there any support for a timeout?  What if the response never comes back?

rjeberhard: Also, I don't yet understand the flow if the result is not HTTP_OK.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -183,15 +312,114 @@ public NextAction apply(Packet packet) {
       }
     }
 
+    /**
+     * Returns the URL to access the Service; using the Service clusterIP and port. If the service
```
rjeberhard, line 316: I know you probably copied this method, but this text doesn't describe what the method is doing. It looks like it's figuring out the URL of where the server can receive admin traffic.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -89,13 +103,110 @@ public NextAction apply(Packet packet) {
     V1Pod pod = info.getServerPod(serverName);
     if (service != null) {
       Step getClient =
-          HttpClient.createAuthenticatedClientForServer(
-              namespace, secretName, new ReadHealthWithHttpClientStep(service, pod, getNext()));
+          getEncodedCredentialsForServer(
```
alai8, line 106: The EncodeCredentialsForServerStep is just creating a readSecret step and then call doNext(). Can it be removed and instead we call SecretHelper.getSecretData() here?

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientFactory.java
ankedia: Fixed

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2017, 2020, Oracle Corporation and/or its affiliates.
```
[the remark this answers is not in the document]
ankedia: Fixed

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
ankedia: Fixed. Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
ankedia: I was modeling this based on the existing ClientPool class implementation in helper package.  JDK11 http client and request objects are thread-safe and can be safely shared. So I thought creating a single instance and sharing it every time should be fine. In my testing, it seem to work fine. Please let me know if you see any issues with this or if any changes are needed. Thanks.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
ankedia: yes, changed to throwable. Thanks.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
ankedia: > Is there any support for a timeout? What if the response never comes back?
Timeout is set on request object and it's currently set to 1 min. I'll test the negative case to make sure request times out after specified duration and doesn't hang forever. Please let me know if current timeout value needs to be changed.

                        HttpRequest.newBuilder()
                            .uri(URI.create(url))
                            .timeout(Duration.ofSeconds(HTTP_TIMEOUT_SECONDS))

ankedia: Tested negative case where request times out due to HttpConnectTimeoutException. It logs below message in operator logs every 3 seconds.

{"timestamp":"05-13-2020T17:57:50.600+0000","thread":189,"fiber":"","domainUID":"","level":"SEVERE","class":"oracle.kubernetes.operator.steps.ReadHealthStep$ReadHealthWithHttpClientStep","method":"lambda$apply$1","timeInMillis":1589392670600,"message":"HTTP POST method on URL http://192.168.3.200:7001/management/weblogic/latest/serverRuntime/search failed with return code java.util.concurrent.CompletionException: java.net.http.HttpConnectTimeoutException: HTTP connect timed out.","exception":"","code":"","headers":{},"body":""}

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
ankedia: Fixed. Thanks.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
ankedia: Fixed, Using HTTP_OK constant from HttpURLConnection class. Thanks.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
ankedia: This method was originally in HttpClient class and I just moved it to ReadHealthStep class (since HttpClient class was removed). I discussed the Java doc text with Anthony and have changed it based on that discussion. Please let me know if this is not clear enough.

markxnelson:
Note: I am not going to review this, I don't think I have the necessary domain knowledge.  If you feel you need me to review it, let me know.

docs/charts/index.yaml
```diff
@@ -2,68 +2,68 @@ apiVersion: v1
 entries:
   weblogic-operator:
   - apiVersion: v1
-    created: 2020-03-21T22:03:53.042035124Z
+    created: 2020-05-14T14:50:06.509204738-04:00
```
rjeberhard, line 5: If you merge in latest from develop, do changes to this file and the tgz file go away? These shouldn't be part of your change.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -34,4 +34,6 @@
   public static final String DOMAIN_INTROSPECT_VERSION = "weblogic.domainIntrospectVersion";
   public static final String DOMAIN_INPUTS_HASH = "weblogic.domainInputsHash";
   public static final String REMAINING_SERVERS_HEALTH_TO_READ = "serverHealthRead";
+
+  public static final String KEY = "encodedCredentials";
```
rjeberhard, line 38: You need a better constant name.  Why not ENCODED_CREDENTIALS?

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
```diff
@@ -123,6 +126,90 @@ public void listenTimeoutDoubled() {
...
 
+  class AsyncRequestStepProcessing {
+
+    final Packet packet;
+    final RetryStrategy retryStrategy;
+    final String cont;
+    final AtomicBoolean didResume = new AtomicBoolean(false);
+    final ApiClient client;
+
+    public AsyncRequestStepProcessing(Packet packet, RetryStrategy retry, String cont) {
```
rjeberhard, line 137: JavaDoc.  What is the "cont" parameter?

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2020, Oracle Corporation and/or its affiliates.
```
rjeberhard, line 1: I think this class is unnecessary because you don't actually pool HttpClient.  Instead, you can share a single instance.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
rjeberhard: As above, let's rid of this class.

operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
```diff
@@ -0,0 +1,122 @@
...
+
+    void process(AsyncFiber fiber) {
+      future = factory.createFuture(request);
+      future.whenComplete((response, throwable) -> resume(fiber, response, throwable));
+      fiber.scheduleOnce(timeoutSeconds, TimeUnit.SECONDS, () -> checkTimeout(fiber));
+    }
+
+    private void checkTimeout(AsyncFiber fiber) {
+      if (!future.isDone()) {
+        fiber.terminate(new RuntimeException("timeout"), packet);
```
rjeberhard, line 101: What is the full set of tasks in the steps for this fiber? This says that if the HTTP request times-out, then the full fiber stops.  If this fiber is only reading the status for one server than this is fine. If the fiber is running through a set of servers then none of them will have status updated. If might make more sense to resume the fiber and have the next step (that processes the result) handle that there is no data.

operator/src/main/java/oracle/kubernetes/operator/http/HttpResponseStep.java
```diff
@@ -0,0 +1,65 @@
...
+public abstract class HttpResponseStep extends Step {
+  static final String RESPONSE = "httpResponse";
+
+  public HttpResponseStep(Step next) {
+    super(next);
+  }
+
+  @Override
+  public NextAction apply(Packet packet) {
+    return Optional.ofNullable(getResponse(packet)).map(r -> doApply(packet, r)).orElse(null);
```
rjeberhard, line 24: Looks like this code already handles no response.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
russgold: It's the same one the K8s APIs use, apparently having to do with continuing requests. Do you have a recommendation on a better name? Should I put the underscore before it, as some k8s code seems to?

operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
russgold: Agreed. I will add a commit that changes this behavior.

operator/src/main/java/oracle/kubernetes/operator/http/HttpResponseStep.java
russgold: Agreed, although it returns null, which doesn't work. Fixing that as well.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
ankedia: > You need a better constant name. Why not ENCODED_CREDENTIALS?

Yes, I'll change it to ENCODED_CREDENTIALS. The constant name was previously used to store httpClient instance in packet and I was lazy to use same name for encoded credentials

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
rjeberhard: I'm okay with "cont" if I know what it is.  This is a cursor so that you can iterate over a long list.  Say, you had a namespace with 10,000 domains.  You could do a list and say give me the first 200, and then use the cont parameter to ask for the next 200 and so on. I can't understand why it's on this method though.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
ankedia: ok, I'll remove HttpClientPool class and HttpClientFactory classes. I will introduce a static

docs/charts/index.yaml
russgold: I merged in the starting version.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
russgold: It's needed for this call:

`
    // Create a call to Kubernetes that we can cancel if it doesn't succeed in time.
    private CancellableCall createCall(AsyncFiber fiber) throws ApiException {
      return factory.generate(requestParams, client, cont, new ApiCallbackImpl(this, fiber));
    }
`

That was in the original version of the class, and I don't see an easy way to get rid of it.

rjeberhard: Okay, resolving this issue.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
rjeberhard: Yes, looks good.

The pool you copied this from is also no longer needed, but removing that is technical debt. When the underlying HTTP client implementation was OkHttp 2.x, the pooling was needed, but OkHttp 3.x has its own internal pooling.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
ankedia: Removed in commit ba303f0. Thanks.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
ankedia: Fixed in commit ba303f0.

operator/src/main/java/oracle/kubernetes/operator/helpers/HttpClientPool.java
ankedia: ok, got it. I merged change based on above comments in ba303f0.

PR #1651, opened 2020-05-15
Implement assertion that only maxUnavailable pods are replaced at a time

sankarpn:
* Implemented assertion that only maxUnavailable pods are replaced in a rolling restart fashion
* Moved some of the pod assertion methods directly calling Kubernertes primitive to a intermediary Pod class.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+   * @param namespace name of the namespace in which the pod is running
+   * @return true if given pod is terminating otherwise false
+   * @throws Exception when more than one pod is terminating or cluster query fails
+   */
+  private static Callable<Boolean> onlyGivenPodTerminating(String podName, String domainUid, String namespace)
+      throws Exception {
+    return () -> {
+      String labelSelectors = String.format("weblogic.serverName", domainUid);
+      V1PodList listPods = Kubernetes.listPods(namespace, labelSelectors);
+      if (listPods.getItems().isEmpty()) {
```
tbarnes-us, line 122: I could be wrong, but I suspect there may be multiple issues with this assertion, some of it functional, some of it due to incompleteness.  All of the following should be double checked with Ryan, Lenny, and Anthony:

- Seems to assume a single cluster and no non-clustered servers. Consequently, this test seems to assume that only one server can go down a time.  This is incorrect when there are multiple clusters, unmanaged servers, a replica count change, or a deleted cluster.

- Seems to dynamically obtain a list of servers when checking if only one server is 'down'.  IMO, it'd be better to get the list of servers at the beginning of a test before any changes are made to initiate the roll (otherwise, it won't catch if two servers in the cluster 'disappear' at the same time).  This 'original list' should be then compared to what's expected by the domain resource in terms of replicas, restartVersions, and image in order to determine the expected behavior for the roll.

- Note that this may get a little tricky when 'replicas' changes at the same time as everything else.  When replicas increases, the number of 'not ready' can be higher than 1 since the replicas are allowed to start in parallel.   When replicas decreases I assume (but am not sure) the roll is expected to defer terminating the servers until _after_ it's completed the roll of the previous servers, and then only terminate them one at a time.   Similar issues may be seen when a cluster is deleted (the roll doesn't apply in that case I assume) or a new cluster is added (multiple servers can start at once).

- Seems to use 'Terminating' as a criteria when checking if a single server is down. For any particular domain it should check the admin server is ready - if not, then the admin server should be the only cycling server.  Then it should check that all other servers are 'ready' except one for their particular cluster as compared to the list of pods obtained at the beginning of the test (except in the case of varying replicas).
- Does not seem to verify that admin server always is 'ahead' of other servers (first to roll and reach target state, and only one that should be down until this occurs)?
- Does not verify that rolls are expected to occur 'in order' within a given cluster.
- At the conclusion of the test, does not check if all servers have (a) reached their target domain/cluster/server restartVersion and (b) reached their target image.  (This should be done in addition to checking for  'ready' state.)

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+      return false;
+    } else {
+      logger.info("WebLogic pods found in namespace {0}", namespace);
+      for (V1Pod item : listPods.getItems()) {
+        logger.info(item.getMetadata().getName());
+        podNames.add(item.getMetadata().getName());
+      }
+    }
+
+    // check the pods termination status in concurrent thread
```
markxnelson, line 63: check that all of the pods are in termination status in a set of concurrent threads

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+    ExecutorService executorService = Executors.newFixedThreadPool(podNames.size());
```
markxnelson, line 64: again, not sure this is a good idea when you are running under maven/failsafe plugin

markxnelson: you could just use the kubernetes get pods api and set watch = true .....

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+      for (var future : threads) {
+        if (!future.get(10, MINUTES)) {
+          return false;
+        }
+      }
+    }
+    executorService.shutdownNow();
+
+    // wait for pods to become ready
+    for (var podName : podNames) {
```
markxnelson, line 90: do you need to wait for them to be ready, or just running?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+      logger.info("Wait for pod {0} to be ready in namespace {1}", podName, namespace);
+      retry
```
markxnelson, line 92: this will wait another ten minutes right?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,150 @@
...
+    for (var pod : pods) {
+      retry
+          .conditionEvaluationListener(condition -> logger.info("Waiting for pod {0} to be "
+          + "terminating in namespace {1} "
+          + "(elapsed time {2}ms, remaining time {3}ms)",
+          pod,
+          namespace,
+          condition.getElapsedTimeInMS(),
+          condition.getRemainingTimeInMS()))
+          .until(assertDoesNotThrow(() -> onlyGivenPodTerminating(pods, pod, namespace),
```
xiancao, line 47: You assume the pods will be terminated in a specific order. This needs to match how operator does the termination.

sankarpn:
I have refactored the code and changed the method signature.
Now the assert method will take list of pod names and check if the pods are terminated the in the sequence it is in the list. Also while checking whether a particular pod is terminating , it checks that no other pods in the list are terminating at that time.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,150 @@
...
+   * @throws Exception when more than one pod is terminating or cluster query fails
+   */
+  public static Callable<Boolean> onlyGivenPodTerminating(ArrayList<String> pods, String podName, String namespace)
+      throws Exception {
+    return () -> {
+      int terminatingPods = 0;
+      boolean givenPodTerminating = false;
+      for (var pod : pods) {
+        if (Kubernetes.isPodTerminating(namespace, null, pod)) {
+          terminatingPods++;
```
vanajamukkara, line 80: print the pod names that are terminating

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiConfigMapOverride.java
```diff
@@ -372,6 +374,16 @@ public void testCreateMiiConfigMapDomain() {
         "patchDomainCustomResource(restartVersion)  failed ");
     assertTrue(rvPatched, "patchDomainCustomResource(restartVersion) failed");
 
+    ArrayList<String> pods = new ArrayList<>();
```
vanajamukkara, line 377: get the pods list before patching domain

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,150 @@
...
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static oracle.weblogic.kubernetes.extensions.LoggedTest.logger;
+import static org.awaitility.Awaitility.with;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+public class Pod {
+
+  /**
+   * Check pods are restarted in the same order as in the pods list.
```
vanajamukkara, line 20: Check only one pod is restarted at a time in the same order as in the pods list ...

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: Its printed inside the Kubernetes.isPodTerminating method

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiConfigMapOverride.java
sankarpn: moved it before patching

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: Removed getting pods dynamically and instead the pod names are passed to the assertion method as a list. The assertion method will simply go through the list and verify its restarted in the sequence it is added in the list, at the same verifies no other pods are restarted at the same time. verifying the restart version after rolling should be done in the test IMO.

sankarpn:
> 
> 
> i think there are some fundamental design issues to discuss here

Reworked the method implementation and removed most of the old code. Please take a look.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+
+public class Pod {
+
+  // reusable condition factory
+  private static ConditionFactory retry
+      = with().pollInterval(5, SECONDS).atMost(10, MINUTES).await();
+
+  /**
+   * Check the pods in the given namespace are restarted in a rolling fashion.
+   * Waits until all pods are restarted, for upto 10 minutes.
```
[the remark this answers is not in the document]
sankarpn: changed javadoc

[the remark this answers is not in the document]
sankarpn: > 
> 
> but your logic below seems to wait longer than ten minutes

removed old wait code

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+   *
+   * @param domainUid UID of the WebLogic domain
+   * @param namespace name of the namespace in which to check for the pods restart sequence
+   * @return true if pods in the namespace are restarted in a rolling fashion otherwise false
+   * @throws ApiException when Kubernetes cluster query fails
+   * @throws InterruptedException when pod status check threads are interrupted
+   * @throws ExecutionException when pod status checks times out
+   * @throws TimeoutException when waiting for the threads times out
+   */
+  public static boolean isARollingRestart(String domainUid, String namespace)
```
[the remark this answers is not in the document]
sankarpn: changed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+      throws ApiException, InterruptedException, ExecutionException, TimeoutException {
+
+    // query cluster and get pods from the namespace
```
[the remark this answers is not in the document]
sankarpn: now the assertion method takes the list of pods to check for rolling restart from test and there is a requirement for using this assertion , it should be called right after the patching was done.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,200 @@
...
+    String labelSelectors = "weblogic.serverName";
+    V1PodList listPods = Kubernetes.listPods(namespace, labelSelectors);
+    ArrayList<String> podNames = new ArrayList<>();
+
+    //return if no pods are found
+    if (listPods.getItems().isEmpty()) {
+      logger.severe("No pods found in namespace {0}", namespace);
+      return false;
+    } else {
+      logger.info("WebLogic pods found in namespace {0}", namespace);
```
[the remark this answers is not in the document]
sankarpn: see above comment

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: checking the pod termination status is now sequential according to the order of pods in the list. Also while checking for termination status of a  given pod , the method will check no other pods are terminating along with it.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,152 @@
...
+import static oracle.weblogic.kubernetes.extensions.LoggedTest.logger;
+import static org.awaitility.Awaitility.with;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+public class Pod {
+
+  /**
+   * Check only one pod is restarted at a time in the same order as in the pods list.
+   * This assertion method needs to be called right after the domain is patched to ensure
+   * it doesn't miss any of the pods restart.
```
anpanigr, line 22: I have a use case where I want to verify there I want to see the rolling restart should shutdown 2 servers simultaneously where the maxUnavailable attribute is set to 2.  In this case what will be my pod list?  Say I have 1 admin and 3 managed servers ( ms1,ms2, and ms3 ) Here  I should see ms1 and ms2 can both go to terminating state simultaneously while ms3 will be re-stated after both ms1 and ms2 are up

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: no concurrent checking now, its sequential

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: The refactored code will check the termination status and then the ready status

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: now its a standard  wait time of 5 minutes/pod

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: It is up to the test provide the list of pods in a order for rolling restart

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiConfigMapOverride.java
sankarpn: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: According to the JIRA only one server can restart at a time.
https://jira.oraclecorp.com/jira/browse/OWLS-81478
If 2 servers can go down at the same time then how can it be a rolling restart?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -157,9 +160,20 @@
...
+    return Pod.podTerminating(namespace, domainUid, podName);
   }
 
+  /**
+   * Check pods are restarted in the same order as in the pods list.
+   * @param pods names of the pods in a list
+   * @param namespace name of the namespace in which to check for pods rolling restart
+   * @return true if pods are restarted in rolling fashion
+   */
+  public static boolean verifyRollingRestartOccurred(ArrayList<String> pods, String namespace) {
```
rjeberhard, line 172: Type of argument can just be List<String>

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
```diff
@@ -154,27 +153,30 @@ public static boolean isPodReady(String namespace, String domainUid, String podN
...
-   * Checks if a pod exists in a given namespace and in Terminating state.
+   * Check if a pod exists in a given namespace and is terminating.
    * @param namespace in which to check for the pod
    * @param domainUid the label the pod is decorated with
    * @param podName name of the pod to check for
-   * @return true if pod is in Terminating state otherwise false
+   * @return true if pod is terminating otherwise false
    * @throws ApiException when there is error in querying the cluster
    */
-  public static boolean isPodTerminating(String namespace, String domainUid, String podName) throws ApiException {
```
rjeberhard, line 164: As an alternative, you can check that the pod's metadata.deletionTimestamp is set. If this field has a value, then the pod has been marked for deletion. It will either then be in it's terminating phase or soon will be.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
rjeberhard: We have a field that controls how many servers can be down at the same time -- it's maxUnavailable. The default value is 1.

rjeberhard: This was added for customers with larger clusters who want to improve the time to roll the entire cluster.

anpanigr: Yes, still, it is a rolling re-start as long as the cluster service is available. The maxUnavailable attribute is designed for this purpose. It can be any non-zero value less than the replica.  See the reference page
https://github.com/oracle/weblogic-kubernetes-operator/blob/master/docs/domains/Domain.md

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
rjeberhard: The operator does roll the servers in sorted order.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
rjeberhard, line 22: On this statement: "This assertion method needs to be called right after the domain is patched to ensure it doesn't miss any of the pods restart."...  This seems likely to intermittently fail. I think it would be better to collect the current V1PodList in advance of starting the roll and using that as the input.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
```diff
@@ -0,0 +1,152 @@
...
+    if (pods.isEmpty()) {
+      logger.severe("The pods list is empty");
+      return false;
+    }
+
+    // reusable condition factory
+    ConditionFactory retry
+        = with().pollInterval(5, SECONDS).atMost(5, MINUTES).await();
+
+    // check pods are terminated and started.
```
rjeberhard, line 39: Not sure if I've read this correctly, but even though WebLogic terminates and restarts slowly, I'd like the code to be defensive to the condition that you missed detecting the termination. It would be enough to find that the "new" Pod has a creation timestamp later than when the roll started.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
tbarnes-us: +1 to that

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
sankarpn: Yes the implementation checks for the existence of pod deletion time stamp.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: now checking the max number of pods that can restart concurrently to be not more than maxUnavailable.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
sankarpn: Now the argument is a Map with podnames and its creation time stamps.
The test will create a LinkedHashMap and add all the pods with its creation time stamps.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
sankarpn: Yes, the new implementation uses pod deletion timestamp.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: The test supplies the list of pods with its creation timestamps(created before the patching).

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
sankarpn: The new implementation will compare creation time stamps to determine the pods replacement status.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Pod.java
tbarnes-us: > verifying the restart version after rolling should be done in the test IMO.

tbarnes-us: Why? A "proper" restart version and image are required for every single roll.  They're essential for corrects behavior.  So why not centrally ensure they match the domain resource's values for every roll?

sankarpn: The requirement is to assert pods replacement doesn't exceed maxUnavailable at a time. We can create another assertion to ensure the integrity of the restartVersion and image.

tbarnes-us: the 'spirit' of the requirement IMO, is to check that roll is working as designed from a customer perspective, this includes restart-version and image

PR #1659, opened 2020-05-18
Add check for unzip

jshum2479:
This add the unzip command check to the MII image,  the container-registry image already have unzip but there may be a version at one point where the unzip is missing.  unzip was not used until we added wdt version check.

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.sh
+++ b/operator/src/main/resources/scripts/introspectDomain.sh
@@ -119,23 +119,23 @@ if [ ${DOMAIN_SOURCE_TYPE} == "FromModel" ]; then
     trace "Beginning Model In Image"
     command -v gzip
     if [ $? -ne 0 ] ; then
-      trace SEVERE "DomainSourceType is FromModel, 'gzip' is missing in the image. Please use an image with 'gzip' installed" && exit 1
+      trace SEVERE "DomainSourceType is 'FromModel', 'gzip' is missing in the image. Please use an image with 'gzip' installed" && exit 1
     fi
     command -v tar
     if [ $? -ne 0 ] ; then
-      trace SEVERE "DomainSourceType is FromModel, 'tar' is missing in the image. Please use an image with 'tar' installed" && exit 1
+      trace SEVERE "DomainSourceType is 'FromModel', 'tar' is missing in the image. Please use an image with 'tar' installed" && exit 1
     fi
     command -v unzip
     if [ $? -ne 0 ] ; then
-      trace SEVERE "DomainSourceType is FromModel, 'unzip' is missing in the image. Please use an image with 'unzip' installed" && exit 1
+      trace SEVERE "DomainSourceType is 'FromModel', 'unzip' is missing in the image. Please use an image with 'unzip' installed" && exit 1
     fi
     mkdir -p ${DOMAIN_HOME}
     if [ $? -ne 0 ] ; then
-      trace SEVERE "DomainSourceType is FromModel, cannot create domain home directory '${DOMAIN_HOME}'" && exit 1
+      trace SEVERE "DomainSourceType is 'FromModel', cannot create domain home directory '${DOMAIN_HOME}'" && exit 1
     fi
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -119,19 +119,23 @@ if [ ${DOMAIN_SOURCE_TYPE} == "FromModel" ]; then
...
+      trace SEVERE "DomainSourceType is FromModel, tar is missing in the image. Please use an image with tar installed" && exit 1
+    fi
+    command -v unzip
+    if [ $? -ne 0 ] ; then
+      trace SEVERE "DomainSourceType is FromModel, unzip is missing in the image. Please use an image with unzip installed" && exit 1
     fi
     mkdir -p ${DOMAIN_HOME}
     if [ $? -ne 0 ] ; then
-      trace SEVERE "cannot create domain home directory '${DOMAIN_HOME}'" && exit 1
+      trace SEVERE "DomainSourceType is FromModel, cannot create domain home directory '${DOMAIN_HOME}'" && exit 1
```
tbarnes-us, line 134: LGTM. 

Minor nits for the SEVEREs:  I think it helps to put '' around each keywords like 'tar', 'gzip', 'unzip', and 'FromModel', and end sentences with a period '.'.

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -119,19 +119,23 @@ if [ ${DOMAIN_SOURCE_TYPE} == "FromModel" ]; then
...
     fi
```
tbarnes-us, line 135: Minor nit:  It'd be nice to put a test file in the domain home directory just to verify that it's  accessible - so we can  'fail fast' when hitting a directory access issue.

touch ${DOMAIN_HOME}/testaccess.tmp
if [ $? -ne 0 ]; then
   yadayada Cannot write to domain home directory '${DOMAIN_HOME}' yadayayada
fi
rm -f ${DOMAIN_HOME}/testaccess.tmp

operator/src/main/resources/scripts/introspectDomain.sh
rjeberhard: Let's make these edits and then I'll commit.

operator/src/main/resources/scripts/introspectDomain.sh
jshum2479: done

operator/src/main/resources/scripts/introspectDomain.sh
jshum2479: done

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -119,20 +119,25 @@ if [ ${DOMAIN_SOURCE_TYPE} == "FromModel" ]; then
...
-      trace SEVERE "DomainSourceType is FromModel, unzip is missing in the image. Please use an image with unzip installed" && exit 1
+      trace SEVERE "DomainSourceType is FromModel, 'unzip' is missing in the image. Please use an image with 'unzip' installed" && exit 1
     fi
     mkdir -p ${DOMAIN_HOME}
     if [ $? -ne 0 ] ; then
       trace SEVERE "DomainSourceType is FromModel, cannot create domain home directory '${DOMAIN_HOME}'" && exit 1
     fi
+    touch ${DOMAIN_HOME}/testaccess.tmp
+    if [ $? -ne 0 ]; then
+      trace SEVERE "DomainSourceType is FromModel, cannot write to domain home directory '${DOMAIN_HOME}'" && exit 1
```
tbarnes-us, line 138: Missing period.

operator/src/main/resources/scripts/introspectDomain.sh
tbarnes-us: Minor nits still buzzing around:  Still missing periods at end of each SEVERE sentence.  Would be nice to have quotes around FromModel too.

PR #1663, opened 2020-05-18
Fix Rolling Upgrade

lennyphan:
Changes for fixing rolling upgrade:

1. For Kubernetes.exec API command, handle IllegalStateException thrown when reading from stderr when stream already closed.

2. Normalize the DomainSpec to resolve a mismatch in the image pull policy field for the updated and current Domain spec (the value for imagePullPolicy is IfNotPresent in the updated Domain whereas the value in current.getSpec() is <null>) which was causing the numerous rolling upgrade code path (canceling and running) by different fibers.

NOTE: Will address checking of all default values for Pod.spec fields (those fields on the pod that Kubernetes appears to fill in) in another JIRA.  Pod spec fields not updated in the DomainSpec.

3. Ryan's changes to remove RestartOneClusteredServerStep and just do the servers.poll() one level higher.

4. While there are server's to be rolled, the same fiber will periodically verify (using doDelay()) when pods are in ready state to initiate rolling of the next set of servers while honoring the configured minimum available servers.


operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -513,6 +513,7 @@ public void makeRightDomainPresence(
           // change just by looking at metadata.generation.
 
           // Has the spec actually changed? We will get watch events for status updates
+          DomainPresenceControl.normalizeDomainSpec(current.getSpec());
```
russgold, line 516: The problem with the call to 'normalize' is that it puts control of the internals of domain matching outside the domain classes. We should really fix this by correcting the equals methods to use the effective values of things rather than the raw fields.

operator/src/main/java/oracle/kubernetes/operator/helpers/RollingHelper.java
```diff
@@ -208,31 +208,18 @@ public NextAction apply(Packet packet) {
 
       LOGGER.info(MessageKeys.ROLLING_SERVERS, dom.getDomainUid(), servers, readyServers);
 
-      int countToRestartNow = Math.max(1, countReady - dom.getMinAvailable(clusterName));
+      int countToRestartNow = countReady - dom.getMinAvailable(clusterName);
       Collection<StepAndPacket> restarts = new ArrayList<>();
       for (int i = 0; i < countToRestartNow; i++) {
-        restarts.add(new StepAndPacket(new RestartOneClusteredServerStep(servers, null), packet));
+        StepAndPacket serverToRestart = servers.poll();
```
russgold, line 214: Functional programming structures can be much easier to read. In this case, 

```
   Optional.ofNullable(servers.poll()).ifPresent(serverToRestart -> restarts.add(serverToRestart));
```

makes it clearer that you are adding the result of the call to servers.poll() to the collection when it exists. It takes some getting used to, but the result is easier-to-read code.

operator/src/main/java/oracle/kubernetes/operator/helpers/RollingHelper.java
```diff
@@ -208,31 +208,18 @@ public NextAction apply(Packet packet) {
...
 
-    @Override
-    public NextAction apply(Packet packet) {
-      StepAndPacket serverToRestart = servers.poll();
-      if (serverToRestart != null) {
-        Collection<StepAndPacket> col = Collections.singleton(serverToRestart);
-        return doForkJoin(this, packet, col);
-      }
-      return doNext(packet);
+      return restarts.isEmpty()
```
russgold, line 220: Please avoid magic numbers in the code - constants are clearer. In addition, nested ternaries can be hard to read. This would be easier to follow as a set of if's, choosing the simplest case first, such as:

```
    if (!restarts.isEmpty()) {
        doForkJoin(this, packet, restarts);
    } else if (!servers.isEmpty()) {
        doDelay(this, packet, DELAY_IN_SECONDS, TimeUnit.SECONDS);
    } else {
        doNext(packet);
    }
```

rjeberhard:
@lennyphan, Russ has some good comments. Let's please address these. The only one that seems functionally different is to check for default values in the equals() method rather than an external normalization.  How hard would that be?

lennyphan:
I've addressed the last two review suggestions by Russ, easy cut-paste.  :-)  Thank you Russ!

Currently looking at how easy to change the check for default values to equals() instead of external normalization.  I'm first looking at seeing how it works with 'image' and 'imagePullPolicy' fields.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
lennyphan: Removed DomainPresenceControl class and used effective default values for 'image' and 'imagePullPolicy' directly in DomainSpec class.  DomainSpec.equals() method uses those effective values rather than raw fields by calling getImage() and getImagePullPolicy().

operator/src/main/java/oracle/kubernetes/operator/helpers/RollingHelper.java
lennyphan: Updated with your functional programming structure.

operator/src/main/java/oracle/kubernetes/operator/helpers/RollingHelper.java
lennyphan: Updated with your suggestion to make it easier to read.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -490,7 +490,6 @@ public void makeRightDomainPresence(
     DomainSpec spec = null;
     if (domain != null) {
```
rjeberhard, line 491: You can remove the entire if

lennyphan: Is it ever possible for 'domain == null' ?

russgold: It could be, since I think we may be handling delete cases, but you can delete the if because we're not doing anything in it.

lennyphan: Removed the entire if and moved it closer to where we compare DomainSpec equality.

PR #1669, opened 2020-05-21
Null check the KIND_REPO

sankarpn:
Doing a null check instead of calling isEmpty since its not a Optional instance.


sankarpn:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/717/

sankarpn:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/146/

PR #1671, opened 2020-05-22
Dynamically update operator for topology changes

russgold:
- changing the introspectVersion on the domain will now force the introspector to run, and will not automatically roll the pods
- a new field, `distributionStrategy` has been added to the `configuration` element of the domain spec. If it is set to ROLLING, a change to the `overridesConfigMap` or `secrets` elements of `configuration` (or their corresponding deprecated elements on the spec) will force all pods to roll, to uptake the changes produced by the introspector.


operator/src/main/java/oracle/kubernetes/operator/ConfigOverrideDistributionStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+package oracle.kubernetes.operator;
+
+public enum ConfigOverrideDistributionStrategy {
+  DYNAMIC, ROLLING;
+
+  public static final ConfigOverrideDistributionStrategy DEFAULT = DYNAMIC;
+}
+
+
+/*
```
alai8, line 13: This looks like implementation notes. Does it belong here?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -157,34 +160,33 @@ private static Step readExistingPods(DomainPresenceInfo info) {
   // pre-conditions: DomainPresenceInfo SPI
   // "principal"
   static Step bringAdminServerUp(
-        DomainPresenceInfo info, PodAwaiterStepFactory podAwaiterStepFactory, Step next) {
-    return Step.chain(bringAdminServerUpSteps(info, podAwaiterStepFactory, next));
+        DomainPresenceInfo info, PodAwaiterStepFactory podAwaiterStepFactory) {
```
alai8, line 163: is it ok to remove `next`?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -475,91 +481,107 @@ private void logThrowable(Throwable throwable) {
   }
 
   /**
-   * Begin activity to align domain status with domain resource.
-   * @param info domain presence info
-   * @param explicitRecheck if explicit recheck
+   * Begin activity to align the cached domain status with the value reaa from Kubernetes.
```
alai8, line 484: reaa -> read

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -475,91 +481,107 @@ private void logThrowable(Throwable throwable) {
...
+   * @param liveInfo domain presence info read from Kubernetes
+   * @param explicitRecheck true if we are to skip comparing the reported domain to the cached value
    * @param isDeleting if is deleting domain
    * @param isWillInterrupt if will interrupt already running activities
    */
   public void makeRightDomainPresence(
-      DomainPresenceInfo info,
+      DomainPresenceInfo liveInfo,
```
alai8, line 491: should we mark `liveInfo` `@Nonnull` too?

docs/domains/Domain.json
```diff
@@ -183,6 +183,14 @@
           "description": "Model in image model files and properties.",
           "$ref": "#/definitions/Model"
         },
+        "distributionStrategy": {
+          "description": "The strategy for handling changes to changes to the secrets or overridesConfigMap fields. Legal values are DYNAMIC and ROLLING.",
```
jshum2479, line 187: duplicate "changes to"

docs/domains/Domain.md
```diff
@@ -92,6 +92,7 @@ An element representing a cluster in the domain configuration.
 
 | Name | Type | Description |
 | --- | --- | --- |
+| `distributionStrategy` | string | The strategy for handling changes to changes to the secrets or overridesConfigMap fields. Legal values are DYNAMIC and ROLLING. |
```
jshum2479, line 95: duplicate "changes to"

operator/src/main/java/oracle/kubernetes/operator/DomainSourceType.java
```diff
@@ -4,7 +4,33 @@
 package oracle.kubernetes.operator;
 
 public enum DomainSourceType {
-  Image,
-  PersistentVolume,
-  FromModel
+  Image {
+    @Override
+    public String getDefaultDomainHome(String uid) {
+      return "/u01/oracle/user_projects/domains";
```
jshum2479, line 10: not appending uid ?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -475,91 +481,107 @@ private void logThrowable(Throwable throwable) {
...
-      if (existing != null) {
-        Domain current = existing.getDomain();
-        if (current != null) {
-          // Is this an outdated watch event?
-          if (domain != null
-              && KubernetesUtils.isFirstNewer(current.getMetadata(), domain.getMetadata())) {
-            LOGGER.fine(MessageKeys.NOT_STARTING_DOMAINUID_THREAD, domainUid);
-            return;
-          }
+    if (!delegate.isNamespaceRunning(liveInfo.getNamespace())) {
```
alai8, line 495: should we also check for isDeleting?

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainTopology.java
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) 2019, 2020, Oracle Corporation and/or its affiliates.
```
jshum2479, line 1: New file ? - copyrights should be 2020 only.

operator/src/main/java/oracle/kubernetes/operator/ConfigOverrideDistributionStrategy.java
russgold: Absolutely not. Oops. Deleting.

russgold: Nope.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: When it's in a call to Step.chain(), yes. That links the specified steps together, allowing them to be listed sequentially and not nested.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: Oops. Fixing.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: Good idea, thanks.

docs/domains/Domain.json
russgold: Thanks - actually generated from a comment on Configuration.java

docs/domains/Domain.md
russgold: Also generated from Configuration.java

operator/src/main/java/oracle/kubernetes/operator/DomainSourceType.java
russgold: We have never have in the past for Domain-in-image. I presume that this is because the image can only have one domain, but it's a question for @rjeberhard

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: No. `isDeleting` is passed into `internalMakeRightDomainPresence` and handled there. The purpose of the checks in this method is to determine whether to call that method.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainTopology.java
russgold: It's actually extracted from its original location in ConfigMapHelper, so the code is older than the file, which is why I did this, but I think you are correct. The dates should go with the file.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
```diff
@@ -536,6 +536,10 @@ public ServerStartupInfo(
       this.isServiceOnly = isServiceOnly;
     }
 
+    public String getName() {
+      return this.serverConfig.getName();
```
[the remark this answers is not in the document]
russgold: Actually, it cannot, which means that we don't need the null checks for getServerName - and that this method is a duplicate. Nice catch.

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -36,4 +65,10 @@ if [ -f ${STATEFILE} ] && [ `grep -c "FAILED_NOT_RESTARTABLE" ${STATEFILE}` -eq
   trace SEVERE "WebLogic Server state is FAILED_NOT_RESTARTABLE."
   exit $RETVAL
 fi
+
+newCopySitCfg /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
```
alai8, line 69: It seems odd to me to do processing other than checking whether the container is running or not in the liveness probe, but I am by no means expert in this.

russgold: This is a key part of the new functionality - the sit config files can be changed in the domain spec, which means that they need to be copied into their proper locations. If we roll the pods, that happens automatically, but the customer wants to be able to apply non-dynamic changes without rolling. The only way we could figure out to do that was by the liveness script, which is already running at intervals.

I agree that it seems clumsy and non-obvious. My original thought was that the operator could run a script, but it doesn't know when the config map with the files changed has actually happened. 

One possibility that we might try is having the operator run a script (via Kubernetes exec()) that does this testing and reports on whether the files have been changed. It could then decide whether it needs to do it again. That seems as though it might be cleaner, if it would work.

alai8: I am just worried if there might be some undesirable side-effect of using the liveness probe script this way. 

Another possibility could be starting a script from startServer.sh that wakes up periodically to do this.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -287,42 +288,55 @@ protected V1PodSpec createPodSpec(TuningParameters tuningParameters) {
...
+  }
+
+  private void addConfigOverrideVolume(V1PodSpec podSpec, String configOverrides) {
+    podSpec.addVolumesItem(
+          new V1Volume()
+                .name(configOverrides + "-volume")
+                .configMap(getOverridesVolumeSource(configOverrides)));
+  }
+
+  private boolean isSourceWdt() {
```
jshum2479, line 315: It probaby ok, any reason not name it as isSourceFromModel to more accurately reflect the logic>

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -7,6 +7,35 @@
 # the pod should be restarted. The script checks a WebLogic Server state file which
 # is updated by the node manager.
 
+function newCopySitCfg() {
+  (
+  shopt -s nullglob  # force file matching 'glob' for loops below to run 0 times if 0 matches
```
jshum2479, line 12: no unset of the nullglob may have side effect

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
russgold: No - when I refactored, I kept the names. Everything else there was calling in Wdt, so I didn't change it. But you're correct. We're using the name "fromModel" in a number of other places, and consistency would be good. I'll see where else we use that nomenclature.

operator/src/main/resources/scripts/livenessProbe.sh
russgold: @tbarnes-us any comments on this?

tbarnes-us: The nullglob is scoped to the surrounding "(" and ")".   @jshum2479

tbarnes-us: example output from

tbarnes-us:
```
fn ()  {
  (
  shopt -s nullglob
  echo "--start fn--"
  for file in doesnotexist*; do
    echo fileinfn: $file
  done
  echo "--end fn--"
  )
}

for file in doesnotexist*; do
  echo fileinmain1: $file
done
fn

for file in doesnotexist*; do
  echo fileinmain2: $file
done

shopt -s nullglob

for file in doesnotexist*; do
  echo fileinmain3: $file
done
```

tbarnes-us:
```
fileinmain1: doesnotexist*
--start fn--
--end fn--
fileinmain2: doesnotexist*
```

tbarnes-us:
Why do we have a 'distribution strategy' setting when customers can already cause a roll by changing the 'restart version'.

What's the expected behavior if there's an 'instrospect version' change in 'FromModel' and the introspector detects a configuration change?

operator/src/main/java/oracle/kubernetes/operator/KubernetesConstants.java
```diff
@@ -29,9 +29,9 @@
 
   String CONTAINER_NAME = "weblogic-server";
 
-  String DOMAIN_CONFIG_MAP_NAME = "weblogic-domain-cm";
+  String SCRIPT_CONFIG_MAP_NAME = "weblogic-domain-cm";
   String DOMAIN_DEBUG_CONFIG_MAP_SUFFIX = "-weblogic-domain-debug-cm";
-  String INTROSPECTOR_CONFIG_MAP_NAME_SUFFIX = "-weblogic-domain-introspect-cm";
+  String DOMAIN_CONFIG_MAP_NAME_SUFFIX = "-weblogic-domain-introspect-cm";
```
tbarnes-us, line 34: Just an observation:  extensive refactoring/renaming makes it harder to code review.  Would be easier if such changes were in a separate pull.

russgold: That would be difficult. In general, I refactor in order to make it easier to make changes. I don't know what refactorings I am going to need until I need to make those changes, and making them often leads to new refactorings.

The best way to address this, IMO, is for people to refactor aggressively as they go - then big after-the-fact refactorings would be rare.

russgold:
> Do we need a 'distribution strategy' setting in the first place? Customers can already 
> cause a roll by changing the domain restart version. 

It's part of the requirement.  It may be that the requirement is not needed. That's a question for @rjeberhard - I suspect our requirements process needs to be reconsidered.

> Also, if a customer is changing multiple secrets and the config map, and the distribution
> strategy is rolling, it looks like there be a 'premature roll' if the customer changes one secret,
> waits a few seconds then changes the next secret, etc, etc.

Only if they do a `kubectl apply` between changes. 

> What's the expected behavior if there's an 'instrospect version' change in 'FromModel' and
> the introspector detects a configuration change?

It's always possible for a configuration change to cause pods to roll.

docs/domains/Domain.md
```diff
@@ -92,6 +92,7 @@ An element representing a cluster in the domain configuration.
 
 | Name | Type | Description |
 | --- | --- | --- |
+| `distributionStrategy` | string | The strategy for handling changes to the secrets or overridesConfigMap fields. Legal values are DYNAMIC and ROLLING. |
```
tbarnes-us, line 95: The thing is, a change to a secret in FromModel is not dynamic at present - custom config overrides are not used in FromModel.

tbarnes-us:
>> What's the expected behavior if there's an 'instrospect version' change in 'FromModel' and
the introspector detects a configuration change?

> It's always possible for a configuration change to cause pods to roll.

I don't understand. This isn't true before your change.  Before your change, a 're-instrospect' in FromModel only occurred when restartVersion changed.  A 'configuration change' therefore does not always cause a roll.  Now, does it occur when when introspectVersion changes?  It seems necessary.   (It would have been nice to brain-storm the requirements even just a little bit more before coding started - I have a feeling they were set before the FromModel requirements were finalized.  Maybe they there are no conflicts - but I'm a little lost not having a solid feel for them.)

tbarnes-us:
> a new field, distributionStrategy has been added to the configuration element of the domain spec. If it is set to ROLLING, a change to the overridesConfigMap or secrets elements of configuration (or their corresponding deprecated elements on the spec) will force all pods to roll, to uptake the changes produced by the introspector.

This implies that when the contents of a secret or cm changes, a roll can occur prematurely before all secrets/cm change.  You responded the roll only occurs `Only if they do a kubectl apply between changes.`.  I don't understand.  Are you saying a roll _wont_ occur when the contents of a secret or cm changes?

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -7,6 +7,35 @@
 # the pod should be restarted. The script checks a WebLogic Server state file which
 # is updated by the node manager.
 
+function newCopySitCfg() {
```
tbarnes-us, line 10: code comment that describes function?
name of function leaves something to be desired (this was an ad hoc name).

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -7,6 +7,35 @@
...
+  (
+  shopt -s nullglob  # force file matching 'glob' for loops below to run 0 times if 0 matches
+  local src_dir=${1?}
+  local tgt_dir=${2?}
+  local fil_prefix=${3?}
+  local local_fname
+  local tgt_file
+  trace "Copying files starting with '$src_dir/$fil_prefix' to '$tgt_dir' without the prefix."
```
tbarnes-us, line 18: do these traces have a side effect on the liveness probe event messages?

russgold:
> > It's always possible for a configuration change to cause pods to roll.

> I don't understand. This isn't true before your change. Before your change, a 're-instrospect'
> in FromModel only occurred when restartVersion changed. A 'configuration change' therefore
> does not always cause a roll.

I didn't says that a configuration change always causes a roll; I said that it is always _possible_ for one to change a roll. That is, the change can affect the definition of a pod.

>  Now, does it occur when when introspectVersion changes? It seems necessary. 

No, all that does is trigger an introspect. If the introspect changes the configuration of a pod, that pod will roll.

operator/src/main/resources/scripts/livenessProbe.sh
russgold: I have not changed any of what you wrote, not being comfortable with shell scripting.

russgold:
> Are there JIRAs to track the corresponding required doc updates?
> JIRAs for functional tests?

None have been written, according to @rjeberhard

operator/src/main/resources/scripts/livenessProbe.sh
tbarnes-us: > > > It's always possible for a configuration change to cause pods to roll.
> 
> > I don't understand. This isn't true before your change. Before your change, a 're-instrospect'
> > in FromModel only occurred when restartVersion changed. A 'configuration change' therefore
> > does not always cause a roll.
> 
> I didn't says that a configuration change always causes a roll; I said that it is always _possible_ for one to change a roll. That is, the change can affect the definition of a pod.
> 
> > Now, does it occur when when introspectVersion changes? It seems necessary.
> 
> No, all that does is trigger an introspect. If the introspect changes the configuration of a pod, that pod will roll.

The question is about whether FromModel + introspectVersion + WebLogic configuration changes will lead to a roll.  Please do not assume pod configuration changes in the context of this question.

tbarnes-us: For example, name this "copySitCfgLive" and rename original function "copySitCfgStart".

For inspiration on comment contents, see comment from the original function in "startServer.sh".  Also add to both Live and Start something like "Note: this function is similar to copySitCfgXXXX in xxxx.sh, but with different error handling."

russgold: Can you add a commit that does that?

russgold:
Closing to split and revise the behavior after discussion with @tbarnes-us and @rjeberhard

PR #1679, opened 2020-05-26
Fix introspectVersion to force introspection and not roll pods

russgold:
This is a revised PR based on https://github.com/oracle/weblogic-kubernetes-operator/pull/1671, splitting the issues into two separate changes.

Changing the introspectVersion on the domain will now force the introspector to run, and will not automatically roll the pods


tbarnes-us:
I think I already asked about some of the following in the original pull, but forgot the answers if there were any (sorry):  

Is `introspectVersion` supported for the FromModel case?  Mii does not support custom sit config, and anything that'd change the introspector output in MII would likely also require a roll in order to propagate the new domain home with the new config. (I'm trying to gauge whether MII support for introspectVersion is harmless - or 'too much rope' - not sure at all.)

Setting aside MII, if a 're-introspect` fails for DII or DIPV, what are the consequences?  Specifically, does the operator act as if the old 'topology' is still valid and so allow scaling operations and what-not?

russgold:
> I think I already asked about some of the following in the original pull, but forgot the answers
> if there were any (sorry):
> 
> Is `introspectVersion` supported for the FromModel case? Mii does not support custom sit
> config, and anything that'd change the introspector output in MII would likely also require a
> roll in order to propagate the new domain home with the new config. (I'm trying to gauge
> whether MII support for introspectVersion is harmless - or 'too much rope' - not sure at all.)

My understanding is that this is primarily intended for the domain-on-PV case. I don't know what value it has in others. 

> Setting aside MII, if a 're-introspect` fails for DII or DIPV, what are the consequences?
> Specifically, does the operator act as if the old 'topology' is still valid and so allow scaling
> operations and what-not?

It has the same effect as if the original introspect fails, including the same reporting.

tbarnes-us:
>> Setting aside MII, if a 're-introspect` fails for DII or DIPV, what are the consequences?
Specifically, does the operator act as if the old 'topology' is still valid and so allow scaling
operations and what-not?

> It has the same effect as if the original introspect fails, including the same reporting.

I don't think it's equivalent.  When the original introspect fails, there's are WL pods running, no topology known by the operator, etc.   Thus in the original case, an attempted scale of an existing domain, etc, wouldn't do anything - but in this case ???

tbarnes-us:
>> Is introspectVersion supported for the FromModel case? Mii does not support custom sit
config, and anything that'd change the introspector output in MII would likely also require a
roll in order to propagate the new domain home with the new config. (I'm trying to gauge
whether MII support for introspectVersion is harmless - or 'too much rope' - not sure at all.)

> My understanding is that this is primarily intended for the domain-on-PV case. I don't know what value it has in others.

I assume there's value in DII - for example, updating the custom sit-cfg to dynamically turn on a debug flag.  But I don't know about MII.  If there isn't value, then we should disallow it IMO - if only to reduce the surface area that needs QA.

rjeberhard:
@russgold @tbarnes-us, based on a conversation with Russ and looking at your comments, Tom, I'm seeing two different kinds of issues:

1) Questions about whether we should handle introspection failure differently if servers are running. This is a good question and does indicate that we missed something during the analysis done when the JIRA was first written. That said, it's not obvious yet that we should do something different. So, I'd lean toward going with what we have and making a second JIRA to research and think about if there are failure cases that should be handled differently when there are servers running. I think we all agree that we have to be very careful about taking existing servers down, but we may need to more explicitly warn the customer for instance.

2) Questions about related use cases, such as the debug flags. Here, I don't want to hold up a solution for the customer's use case, which is the topology case. For domain-on-pv and debug flags, the customer would just make an online change, so this new discussion would be for an important but distinct use case of how to handle dynamically enabling debug flags for domain-in-image. Model in image would be handled in yet again a different way. So, I think this discussion also needs to be separated from this review.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -475,91 +479,107 @@ private void logThrowable(Throwable throwable) {
   }
 
   /**
-   * Begin activity to align domain status with domain resource.
-   * @param info domain presence info
-   * @param explicitRecheck if explicit recheck
+   * Begin activity to align the cached domain status with the value reaa from Kubernetes.
```
rjeberhard, line 482: reaa -> real

russgold: darn it. I fixed that in the last PR in a separate commit and missed it again.

PR #1682, opened 2020-05-27
Implement distribution strategy for config override changes

russgold:
This is a revised PR based on #1671, splitting the issues into two separate changes.

The essence of this change is simply to set an environment variable based on the selected distribution strategy, and to modify the scripts to watch for it.

There is a fair bit of refactoring in this change as well, initial done in order to enable more extensive processing, some of which turned out not to be needed. I'm leaving them here as they add to the test coverage and simplify the code.


docs/domains/Domain.json
```diff
@@ -183,6 +183,14 @@
           "description": "Model in image model files and properties.",
           "$ref": "#/definitions/Model"
         },
+        "distributionStrategy": {
+          "description": "The strategy for handling changes to the secrets or overridesConfigMap fields. Legal values are DYNAMIC and ON_RESTART.",
```
tbarnes-us, line 187: or overridesConfigMaps fields. --> or overridesConfigMaps fields once the introspectVersion is changed.  

- Or something similar. 
- The idea is to make it clear that the strategy is related to handling a change to the introspectVersion field.  
- Do the same edit for all related discussion of distributionStrategy since they're all tied to introspectVersion.
- Also, consider changing the name to 'introspectVersionDistributionStrategy' or similar...

docs/domains/Domain.json
```diff
@@ -183,6 +183,14 @@
...
+          "type": "string",
+          "enum": [
```
tbarnes-us, line 189: All descriptions for 'introspectVersion' should also have something like 'See also 'distributionStrategy'.'.

docs/domains/Domain.json
russgold: As far as I know, this has nothing to do with introspection. It is triggered by a change to the specified fields in the domain resource only.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -675,7 +675,7 @@ Step createDomainUpPlan(DomainPresenceInfo info) {
 
     return Step.chain(
           createDomainUpInitialStep(info),
-          ConfigMapHelper.readExistingSituConfigMap(info.getNamespace(), info.getDomainUid()),
+          ConfigMapHelper.readExistingDomainConfigMap(info.getNamespace(), info.getDomainUid()),
```
tbarnes-us, line 678: This is a confusing rename IMO.  

We have at least four config maps:   'Scripts' configmap deployed per namespace, Introspector Output per domain, Optional Custom Situational Config Input per domain, and Optional Model Input per domain.   

This change renames situational to 'Domain' which is a bit confusing because (a) both Situational and Model are both domain scoped, and (b) the namespaced cm is called 'weblogic-domain-cm'.

tbarnes-us: Suggested nomenclature:   Model, Situational, Introspector, Scripts ???

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -7,6 +7,49 @@
...
+
+if [ ! "${DYNAMIC_CONFIG_OVERRIDE:-notset}" = notset ]; then
+    (
+    shopt -s nullglob  # force file matching 'glob' for loops below to run 0 times if 0 matches
+    local src_dir=${1?}
+    local tgt_dir=${2?}
+    local fil_prefix=${3?}
+    local local_fname
+    local tgt_file
+    trace "Copying files starting with '$src_dir/$fil_prefix' to '$tgt_dir' without the prefix."
```
tbarnes-us, line 31: Does this trace have a side effect on eventing/logging?  This will be called every 10 seconds, which seems distracting if so.    Maybe only trace at line 38 (just before the cp) to only trace when copying a change over?

docs/domains/Domain.json
tbarnes-us: Wait, what?  We need to discuss with @rjeberhard

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: @rjeberhard noted that the config map in question started with just situation config, but now can includes other elements. It's worth bringing up again, though, if it's confusing.

operator/src/main/resources/scripts/livenessProbe.sh
russgold: That's in a loop, isn't it? Wouldn't that result in multiple log messages?

tbarnes-us: The liveness probe is called every  10 seconds.  So this trace will occur every 10 seconds in theory somewhere.   I'm suggesting moving the trace down to the copy - which only occurs when the file in question changed - so yes, multiple log messages, but only when a file changes instead of every 10 seconds...

docs/domains/Domain.json
rjeberhard: Here is my understanding of this field. In operator 2.5.0, introspection could only happen on initial domain creation or following a cold shutdown, so no servers presently running. Therefore, we had no issues about how changes to any of the inputs that affect the final configuration overrides would affect running servers. Now, we have a specific need to run introspection when servers are running and even to distribute updated configuration overrides to those running servers. However, since we can't test every combination of changes to these various inputs we introduced two guards: 1) The introspectVersion, which lets the customer explicitly trigger introspection and thereby confirming that all of the inputs are ready, and 2) the distribution strategy, which lets the customer select not distributing the updated strategy dynamically to already running servers.

tbarnes-us: I guess I don't know what Russ means by 'It' and 'specified fields' in 'It is triggered by a change to the specified fields in the domain resource only.'  

I've been assuming a re-introspect for a running in-pv or in-image domain only occurs if the introspectVersion changes.  Is that wrong?

rjeberhard: Or following a cold restart, as before. We had to increase the cases where we ran introspection for model-in-image, but that was not supposed to affect in-pv or image.

rjeberhard: I know what he might mean. There was a question about how and when changes to the distribution strategy field itself should be activated. Since we proposed that the distribution strategy would be realized as an env. variable in the server container then the active distribution strategy would be the value set when a given pod was started.

tbarnes-us: So, back to my original comment in this thread, is it valid?

rjeberhard: We tell customers about introspection, so we can reference that.  How about... "The strategy for how configuration overrides are distributed to WebLogic servers following introspection. Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology. Legal values are DYNAMIC and ON_RESTART."

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
rjeberhard: There are input config maps, but this is the one generated per-domain config map, correct? The changed name here seems okay to me.

docs/domains/Domain.json
russgold: Ah, so simply changing the domain spec to reference a different override config map will not  have any effect until the next introspection?

operator/src/main/resources/scripts/livenessProbe.sh
rjeberhard: Agreed, although without an opinion on exactly where the trace belongs, as a trace every 10 seconds will overwhelm other logging.

docs/domains/Domain.json
rjeberhard: That's correct.  It's also intentional here since the customer will likely make coordinating changes to secrets, the input overrides config map, and even domain topology (using online WLST) and so we only want to do introspection when the customer is ready.

russgold: What about changing the distribution strategy? Should that force the pods to roll? Currently it does, which seems troubling. 

I've tried to be specific about the effects of this feature by writing up a testing JIRA https://jira.oraclecorp.com/jira/browse/OWLS-82325. I think that may be the best way to communicate about it, and based on what you've just said, it's wrong.

rjeberhard: Yes, I agree that changing the distribution strategy should not cause the pods to roll. I'd prefer a behavior like, "the active distribution strategy is the distribution strategy configured when the server was started."  We have the same behavior for the shutdown-related options.

rjeberhard:
I've reviewed all of the files in this PR... Other than the issues that we are already discussing, everything looks good. Thanks for bringing up the issue about the changing distribution strategy rolling pods.

docs/domains/Domain.json
russgold: Can you point to code for those shutdown-related options?

tbarnes-us: "So it sounds to me as though what we're now calling the domain config map is the one which contains the generated overrides, as opposed to the overrideConfigMap, which contains information for describing the overrides. "   Again, calling it 'domain config map' is quite confusing name IMO, for the reasons I stated in the other thread.    How about 'introspector config map, which contains output from the introspector, as opposed to the override and model config maps, which are inputs to the introspector'.

tbarnes-us: Ryan:  How about  "The strategy for how configuration overrides are distributed to WebLogic servers following introspection. Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology. Legal values are DYNAMIC and ON_RESTART."

Sounds good to me.  Please also add:   "The default is WHATEVER."   And  please also use the same wording in the other locations that describe the new field.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: It's not OK by me - i find it quite confusing.  The 'script' config map is literally named 'weblogic-domain-cm' but the domain cm is _not_ the 'script' config map.   The domain config map is the introspector config map.  I find this all very confusing. Why not simply call it the introspector config map?  It's contents mirror the contents of the introspector output?  And the name of the config map in k8s has the keyword 'introspectsor' right in it.

docs/domains/Domain.json
tbarnes-us: The strategy for configuration override distribution to WebLogic servers following introspection. Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology. Legal values are DYNAMIC and ON_RESTART. The default value is XXXXX. Changes to the strategy require a full domain restart to take effect.

docs/domains/Domain.json
```diff
@@ -183,6 +183,14 @@
           "description": "Model in image model files and properties.",
           "$ref": "#/definitions/Model"
         },
+        "distributionStrategy": {
```
tbarnes-us, line 186: Can we consider changing the name to 'introspectVersionDistributionStrategy' or 'situationalConfigurationDistributionStrategy'?   The distribution strategy term is very generic - it doesn't let you know what the field is meant for.  Also the generic nature introduces potential conflicts - for example, we'll likely have a 'modelDistributionStrategy' at some point.

russgold: It's under the `configuration` element. That should give it the relevant context. If it were at the spec level, I'd agree with you.

russgold:
I do wish generated files were not included in PR reviews.

docs/domains/Domain.json
tbarnes-us: I disagree. The configuration element is getting crowded - there are attributes that have nothing to do with the new setting.  And there are attribute(s) that may have there own 'strategy' setting soon enough.

tbarnes-us: After some discussion:  'overridesDistributionStrategy'.

docs/domains/Domain.md
```diff
@@ -95,6 +95,7 @@ An element representing a cluster in the domain configuration.
 | `introspectorJobActiveDeadlineSeconds` | number | The introspector job timeout value in seconds. If this field is specified it overrides the Operator's config map data.introspectorJobActiveDeadlineSeconds value. |
 | `model` | [Model](#model) | Model in image model files and properties. |
 | `opss` | [Opss](#opss) | Configuration for OPSS security. |
+| `overrideDistributionStrategy` | string | Determines how updated configuration overrides are distributed to already running WebLogic servers following introspection when the domainHomeSourceType is PersistentVolume or Image.  Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology.Legal values are DYNAMIC (the default) and ON_RESTART. |
```
mriccell, line 98: Typo missing space after the period  -> ".  Legal values are DYNAMIC (the default) and ON_RESTART."

docs/domains/index.html
```diff
@@ -1095,6 +1095,14 @@
     "Configuration": {
       "type": "object",
       "properties": {
+        "overrideDistributionStrategy": {
+          "description": "Determines how updated configuration overrides are distributed to already running WebLogic servers following introspection when the domainHomeSourceType is PersistentVolume or Image.  Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology.Legal values are DYNAMIC (the default) and ON_RESTART.",
```
mriccell, line 1099: Typo missing space after the period -> ". Legal values are DYNAMIC (the default) and ON_RESTART."

docs/domains/index.html
```diff
@@ -1279,7 +1287,7 @@
           "type": "string"
         },
         "introspectVersion": {
-          "description": "If present, every time this value is updated, the operator will start introspect domain job",
+          "description": "If the domain is running, the operator will restart its introspector job when this value is changed.This field is ignored when the domainHomeSourceType source type is FromModel.",
```
mriccell, line 1290: Typo missing space after the period -> ".  This field is ignored when the domainHomeSourceType source type is FromModel."

kubernetes/crd/domain-crd.yaml
```diff
@@ -29,6 +29,17 @@ spec:
                 type: object
                 description: Models and overrides affecting the WebLogic domain configuration.
                 properties:
+                  overrideDistributionStrategy:
+                    type: string
+                    description: Determines how updated configuration overrides are
+                      distributed to already running WebLogic servers following introspection
+                      when the domainHomeSourceType is PersistentVolume or Image.  Configuration
+                      overrides are generated during introspection from secrets, the
+                      overrideConfigMap field, and WebLogic domain topology.Legal
```
mriccell, line 38: Typo missing space after the period -> ". Legal values are DYNAMIC (the default) and ON_RESTART."

kubernetes/crd/domain-crd.yaml
```diff
@@ -10861,8 +10872,9 @@ spec:
                   will restart the required servers.
               introspectVersion:
                 type: string
-                description: If present, every time this value is updated, the operator
-                  will start introspect domain job
+                description: If the domain is running, the operator will restart its
+                  introspector job when this value is changed.This field is ignored
```
mriccell, line 10876: Typo missing space after the period -> ". This field is ignored when the domainHomeSourceType source type is FromModel."

kubernetes/crd/domain-v1beta1-crd.yaml
```diff
@@ -32,6 +32,17 @@ spec:
               type: object
               description: Models and overrides affecting the WebLogic domain configuration.
               properties:
+                overrideDistributionStrategy:
+                  type: string
+                  description: Determines how updated configuration overrides are
+                    distributed to already running WebLogic servers following introspection
+                    when the domainHomeSourceType is PersistentVolume or Image.  Configuration
+                    overrides are generated during introspection from secrets, the
+                    overrideConfigMap field, and WebLogic domain topology.Legal values
```
mriccell, line 41: Typo missing space after the period -> ". Legal values are DYNAMIC (the default) and ON_RESTART."

kubernetes/crd/domain-v1beta1-crd.yaml
```diff
@@ -10643,8 +10654,9 @@ spec:
                 will restart the required servers.
             introspectVersion:
               type: string
-              description: If present, every time this value is updated, the operator
-                will start introspect domain job
+              description: If the domain is running, the operator will restart its
+                introspector job when this value is changed.This field is ignored
```
mriccell, line 10658: Typo missing space after the period -> ". This field is ignored when the domainHomeSourceType source type is FromModel."

russgold: This file is generate. It looks correct in Git. Why doesn't it look right in the review?!

russgold: Could you please check out the branch locally and see if these errors are there?

rjeberhard:
@mriccell, are your concerns resolved? @tbarnes-us and @russgold, you probably said in the scrum today, but I don't remember -- sorry... what is the status of this PR?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: Looks like this was fixed.

docs/domains/Domain.json
```diff
@@ -175,6 +175,14 @@
     "Configuration": {
       "type": "object",
       "properties": {
+        "overrideDistributionStrategy": {
+          "description": "Determines how updated configuration overrides are distributed to already running WebLogic servers following introspection when the domainHomeSourceType is PersistentVolume or Image.  Configuration overrides are generated during introspection from secrets, the overrideConfigMap field, and WebLogic domain topology. Legal values are DYNAMIC (the default) and ON_RESTART.",
```
tbarnes-us, line 179: Add "See also introspectVersion."

operator/src/main/resources/scripts/livenessProbe.sh
tbarnes-us: Looks fixed.

tbarnes-us:
> from @rjeberhard:  @tbarnes-us and @russgold, you probably said in the scrum today, but I don't remember -- sorry... what is the status of this PR?

Looks OK to me, apart from recommending 'overridesDistributionStrategy' to the 'overridesDistributionStrategy' description.  Delegating final sign off to @mriccell

PR #1683, opened 2020-05-27
Improvements found while diagnosing problems with fibers under load

rjeberhard:
Anil,

Here are the changes I have so far:
- Use 8.0.2
- Naming and numbering of fibers was wrong
- Added reporting of suspend fibers with where they are suspended


operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -219,6 +220,28 @@ public void stopNamespace(String ns) {
...
+   * Report on currently suspended fibers. This is the first step toward diagnosing if we need special handling
+   * to kill or kick these fibers.
+   */
+  public void reportSuspendedFibers() {
+    if (LOGGER.isFineEnabled()) {
+      BiConsumer<String, FiberGate> consumer =
+          (namespace, gate) -> {
+            gate.getCurrentFibers().forEach(
+                (key, fiber) -> {
+                  Step lastIfSuspended = fiber.getLastIfSuspended();
```
russgold, line 233: Consider using Optional.ofNullable(fiber.getLastIfSuspended()) rather than the separate get and if statement.

operator/src/main/java/oracle/kubernetes/operator/work/Fiber.java
```diff
@@ -318,6 +319,22 @@ public boolean isDone() {
     return status.get() == DONE;
   }
 
+  /**
+   * The most recently invoked step if the fiber is currently suspended.
+   * @return Last invoked step for suspended fiber.
+   */
+  public Step getLastIfSuspended() {
```
russgold, line 326: The name is a bit unclear. Maybe `getSuspendedStep` would be clearer. Its semantics would be the same, as a fiber can have at most one suspended step.

PR #1684, opened 2020-05-27
second pass edit

rosemarymarano:
Did a careful second pass literary edit on the entire sample doc. Aligned new sections to the original ones (initial use case and update1). Also, fixed/cleaned up some of the original wording.


tbarnes-us:
Changes look good to me.  Please hold off on any more changes to MII doc in order minimize the chance of merge difficulties.

rosemarymarano:
Will do.

PR #1685, opened 2020-05-28
Changing the podCreationTimestamp data type to DateTime from String

sankarpn:
Changing the podCreationTimestamp data type to DateTime from String and using isAfter method for comparing the timestamps.

All tests passed.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/204/

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/780/


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
```diff
@@ -571,16 +571,16 @@ public static boolean adminServerReady(String domainUid, String namespace) {
    * @param domainUid the label the pod is decorated with
    * @param namespace in which the pod is running
    * @param timestamp the initial podCreationTimestamp
-   * @return true if the pod's new timestamp is later than the initial PodCreationTimestamp
+   * @return true if the pod's creation timestamp is later than the initial PodCreationTimestamp
    * @throws ApiException when query fails
    */
   public static boolean isPodRestarted(
```
doxiao, line 577: Is this method still used any where?  Now we have verifyRollingRestartOccurred() method that performs more thorough checks for rolling-restart, and all tests have been updated to use it.  And verifyRollingRestartOccurred does not seem to use this method. So my guess is that this method is not longer needed.

sankarpn: Loos like it is still used in the following 
CommonTestUtils.scaleAndVerifyCluster(....)
ItUsabilityOperatorHelmChart.java

doxiao: Hmm, it looks like it is only called in CommonTestUtils.checkPodRestarted(), which is not used anywhere. It you prefer, we could clean that up in a different PR.

xiancao: It is still used in Domain.podStateNotChanged.

doxiao: I see. somehow my search missed that one.  thanks.

PR #1688, opened 2020-05-28
Mii sample jrf fix

tbarnes-us:
Maggie - Please test this fix.


maggiehe00:
With this PR's change now initial JRF domain is up with both admin server and 2 managed servers
kubectl get pods -n sample-domain1-ns --watch
NAME                                READY   STATUS    RESTARTS   AGE
sample-domain1-admin-server         1/1     Running   0          2m40s
sample-domain1-managed-server1-c1   1/1     Running   0          64s
sample-domain1-managed-server2-c1   1/1     Running   0          64s

PR #1695, opened 2020-06-01
Fix javadoc

doxiao:
Fix javadoc errors and warnings in the new test code.


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -137,8 +139,10 @@ public static boolean podRestartVersionUpdated(
    * @param domainUid ID of the domain resource
    * @param namespace Kubernetes namespace in which the domain custom resource object exists
    * @param podName name of the WebLogic server pod
+   * @param containerName name of the container inside the pod where the image is used
    * @param image name of the image that was used to patch the domain resource
    * @return true if the pod is patched correctly
+   * @throws ApiException if Kubernetes client API call fails
```
sankarpn, line 145: Does this throw exception?

doxiao: It makes Kubernetes client API calls, and therefore, may throw an ApiException as the method signature indicates.

sankarpn: Yes but this method returns a Callable object

doxiao: You are right. I was just fixing the errors and warnings that the generate javadoc tool reported.
I looked around and found a couple of other places where the method signature has throws while the method returns a Callable. I'll fix those too.

PR #1698, opened 2020-06-02
fix error in WatchBuilderTest.tearDown() caused by NamespaceTest

alai8:
Fixes test failure in WatchBuilderTest on Linux that is caused by NamespaceTest having a value of domainPresenceRecheckIntervalSeconds in MainTuning that is too low. 

Example of test failure (could happen in a different test in WatchBuilderTest):             

[ERROR] whenServiceWatchSpecifiesParameters_verifyAndReturnResponse(oracle.kubernetes.operator.builders.WatchBuilderTest)  Time elapsed: 0.005 s  <<< ERROR!
io.kubernetes.client.openapi.ApiException: java.net.ConnectException: Failed to connect to localhost/127.0.0.1:45387
        at oracle.kubernetes.operator.builders.WatchBuilderTest.whenServiceWatchSpecifiesParameters_verifyAndReturnResponse(WatchBuilderTest.java:203)
Caused by: java.net.ConnectException: Failed to connect to localhost/127.0.0.1:45387
        at oracle.kubernetes.operator.builders.WatchBuilderTest.whenServiceWatchSpecifiesParameters_verifyAndReturnResponse(WatchBuilderTest.java:203)
Caused by: java.net.ConnectException: Connection refused (Connection refused)
        at oracle.kubernetes.operator.builders.WatchBuilderTest.whenServiceWatchSpecifiesParameters_verifyAndReturnResponse(WatchBuilderTest.java:203)

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/DomainUpPlanTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/DomainUpPlanTest.java
@@ -156,7 +156,6 @@ public class DomainUpPlanTest {
 
   @Test
   public void whenAdminPodCreated_hasListenPort() throws NoSuchFieldException {
-    mementos.add(TuningParametersStub.install());
     mementos.add(UnitTestHash.install());
 
     WlsDomainConfigSupport configSupport = new WlsDomainConfigSupport("domain");
--- a/operator/src/test/java/oracle/kubernetes/operator/NamespaceTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/NamespaceTest.java
@@ -21,6 +21,7 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import static com.meterware.simplestub.Stub.createStrictStub;
 import static java.util.function.Function.identity;
 import static oracle.kubernetes.operator.DomainProcessorTestSetup.NS;
 import static org.hamcrest.Matchers.not;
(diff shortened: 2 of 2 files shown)


russgold:
Can you explain what you did? I see that you removed the install of a stub in one test, and created a new stub to install for another test. The shared TuningParametersStub doesn't work there?

I'm curious as to why this change fixes things and why any connections were attempted.

alai8:
In`Main.recheckDomains()`:

 

      int recheckInterval  = tuningAndConfig.getMainTuning().domainPresenceRecheckIntervalSeconds;
      DateTime now = DateTime.now();
      boolean isFullRecheck = false;
      if (lastFullRecheck.get().plusSeconds(recheckInterval).isBefore(now)) {
        processor.reportSuspendedFibers();
        isFullRecheck = true;
        lastFullRecheck.set(now);
      } else {
        // check for namespaces that need to be started
        namespacesToStart = new TreeSet<>(targetNamespaces);
        namespacesToStart.removeAll(namespaceStatuses.keySet());
        for (String ns : targetNamespaces) {
          if (namespacesToStop.contains(ns)) {
            namespacesToStart.remove(ns);
          }
        }
      }
      if (!namespacesToStart.isEmpty()) {
        runSteps(new StartNamespacesStep(namespacesToStart, isFullRecheck));
      }
    };


When `NamespaceTest` is run with the whole test suite, I found that the value of `domainPresenceRecheckIntervalSeconds` is 2. This results in the `if` block being run and `namespacesToStart` not being emptied, resulting in `StartNamespacesStep` being invoked. This causes the exception in the the `teardown()` method in the next test, `WatchBuilderTest`.

There was an extra call to `TuniningParametersStub.install()` in `DomainUpPlanTest.whenAdminPodCreated_hasListenPort()` as it was already done in `setup()`. After removing the extra call, the value of `domainPresenceRecheckIntervalSeconds` when `NamespaceTest` is run would become the default value of 120. Thus I have removed that extra call from `DomainUpPlanTest`, and also adding a stub in `NamespaceTest`.

PR #1699, opened 2020-06-02
MII sample docs

rosemarymarano:
MII sample doc preview. Do not merge.


docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -17,12 +13,12 @@ This feature is supported only in 3.0.0-RC1.
      - [Sample directory structure](#sample-directory-structure)
    - [Prerequisites for all domain types](#prerequisites-for-all-domain-types)
    - [Additional prerequisites for JRF domains](#additional-prerequisites-for-jrf-domains)
-   - [Initial use case](#initial-use-case): An initial WebLogic domain
-   - [Update1 use case](#update1-use-case): Dynamically adding a data source using a model ConfigMap
-   - [Update2 use case](#update2-use-case): Deploying an additional domain
-   - [Update3 use case](#update3-use-case): Updating an application in an image
-   - [Cleanup](#cleanup)
-   - [References](#references)
+   - [Initial](#initial-use-case) use case: An initial WebLogic domain
```
tbarnes-us, line 16: The 'Inital' and 'Update' use cases are indented differently.
How about doing this instead:
- Use cases:
   - Initial
   - Update1
   - Update2
   - Update3

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -17,12 +13,12 @@ This feature is supported only in 3.0.0-RC1.
...
+     - [Update 1]({{< relref "/samples/simple/domains/model-in-image/update1.md" >}}): Dynamically adding a data source using a model ConfigMap
+     - [Update 2]({{< relref "/samples/simple/domains/model-in-image/update2.md" >}}): Deploying an additional domain
+     - [Update 3]({{< relref "/samples/simple/domains/model-in-image/update3.md" >}}): Updating an application in an image
```
tbarnes-us, line 19: I prefer 'Update1' to 'Update 1'.  It's just as understandable/readable IMOO, it conveniently makes the doc easier to search/edit for a specific use case - no need to use two different search terms, plus it lines up with the internal documentation/scripts for same.

docs-source/content/samples/simple/domains/model-in-image/_index.md
rosemarymarano: Good idea.

docs-source/content/samples/simple/domains/model-in-image/_index.md
```diff
@@ -75,26 +71,26 @@ This sample demonstrates four Model in Image use cases:
...
     - An updated web application `v2` at the `myapp-v2` directory path instead of `myapp-v1`
     - An updated model that points to the new web application path
   - Kubernetes Secrets and ConfigMap:
-    - Same as the Update1 use case
+    - Same as the Update 1 use case
   - A domain resource:
-    - Same as the Update1 use case, except `spec.image` is `model-in-image:WLS-v2`
+    - Same as the Update 1 use case, except `spec.image` is `model-in-image:WLS-v2`
 
 #### Sample directory structure
```
tbarnes-us, line 95: Might as well split out pre-req-wls pre-req-jrf and initial into different files too?

docs-source/content/samples/simple/domains/model-in-image/_index.md
rosemarymarano: Personally, I don't like the titles Update 1, Update 2 and such. Perhaps, we can think of a more descriptive heading for each updated use case. That might solve our dilemma.

docs-source/content/samples/simple/domains/model-in-image/_index.md
rosemarymarano: We need to strike a balance between navigating to separate files versus not. I kept all the prereqs together with the initial use case for this reason: if a user wants to run through just the first scenario, it's all contained in one file.

rjeberhard:
This looks very good to me. It's quite hard though to compare the options (what we have vs. this change) in a diff tool.

docs-source/content/samples/simple/domains/model-in-image/_index.md
tbarnes-us: Hmm, I still lean towards separating them. IMO, bundling overview + pre-reqs + initial together makes for an overly long file in of itself. Also, I find it a little confusing that some of the use cases are included in the introduction document, but not all of them.

rosemarymarano:
True. It looks better in the Hugo UI.

docs-source/content/samples/simple/domains/model-in-image/_index.md
rosemarymarano: Perhaps, I should break out all the use cases into separate files and keep the intro and prereqs together.

tbarnes-us: I personally would prefer an "overview" top level doc that points to "intro" "pre-reqs" "initial" "update1" "update2" "update3" and "cleanup" - similar to the organization of the MII user docs   In for a penny, in for a pound.   Maybe wouldn't look too overwhelming if if there's a way to suppress all but "overview" in hugo's left side panel menu.

rosemarymarano: I think I will try something along these lines because the doc is so long. I don't think that there is a way to temporarily suppress entries in Hugo's left-side panel menu. However, it might be possible to make it expandable/collapsible (but I don't currently know how to do that).

PR #1700, opened 2020-06-02
Remove workarounds in the integration test code

doxiao:
1) Remove the property that enables the check for application availability in the duration of patching a domain with newer version of the application. The property was put in because the check would intermittently fail due to a product issue in rolling restart. With the property the test could be used to debug issue and verify a fix. The product issue has been fixed recently so the check can be in place all the time now.
2) Remove workarounds that were put in place to avoid an missing stream close issue in the Kubernetes Java client.  This includes removal of the timeout on thread join when using Kubernetes java client exec api, and removal the code that uses kubectl exec. Now the operator uses the latest version where the issue has been fixed so all assertions can use the Kubernetes Java client exec api.


doxiao:
Here are the results of Jenkins runs, and the test cases in ItMiiDomain that are related to this PR are all clean. The runs marked with (*) had a failure in ITSessionMigraiton that is not related to this PR.

https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/246
(*) https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/247
(*) https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/248
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/252

(*) https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/820
(*) https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/825

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java
```diff
@@ -2063,11 +2063,9 @@ public static ExecResult exec(V1Pod pod, String containerName, boolean redirectT
       // wait for the process, which represents the executing command, to terminate
       proc.waitFor();
 
-      // wait for reading thread to finish any last remaining output
+      // wait for reading thread to finish any remaining output
       if (out != null) {
```
alai8, line 2067: probably no need for null check here

doxiao: yes, I agree.

PR #1702, opened 2020-06-03
Correct affinity merging logic

rjeberhard:
Without this change, I was seeing concurrent modification exceptions and duplicated elements in the merged data structures. It looks like we already had similar problems because I was able to use logic similar to what we have for other types in `serverPod`.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
@@ -270,25 +270,15 @@ class ServerPod extends KubernetesResource {
 
   private void copyValues(V1Capabilities to, V1Capabilities from) {
     if (from.getAdd() != null) {
-      List<String> allAddCapabilities = new ArrayList<>();
-      if (to.getAdd() != null) {
-        allAddCapabilities =
-            Stream.concat(to.getAdd().stream(), from.getAdd().stream())
-                .distinct()
-                .collect(Collectors.toList());
-      }
-      to.setAdd(allAddCapabilities);
+      Stream<String> stream = (to.getAdd() != null)
+          ? Stream.concat(to.getAdd().stream(), from.getAdd().stream()) : from.getAdd().stream();
+      to.setAdd(stream.distinct().collect(Collectors.toList()));
     }
 
     if (from.getDrop() != null) {
-      List<String> allDropCapabilities = new ArrayList<>();
-      if (to.getDrop() != null) {
-        allDropCapabilities =
-            Stream.concat(to.getDrop().stream(), from.getDrop().stream())
-                .distinct()
-                .collect(Collectors.toList());
-      }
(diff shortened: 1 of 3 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -307,12 +311,18 @@ private void copyValues(V1Affinity to, V1Affinity from) {
...
+    if (from.getPreferredDuringSchedulingIgnoredDuringExecution() != null) {
+      List<V1PreferredSchedulingTerm> preferredDuringSchedulingIgnoredDuringExecution = new ArrayList<>();
+      if (to.getPreferredDuringSchedulingIgnoredDuringExecution() != null) {
+        preferredDuringSchedulingIgnoredDuringExecution =
+            Stream.concat(to.getPreferredDuringSchedulingIgnoredDuringExecution().stream(),
+                from.getPreferredDuringSchedulingIgnoredDuringExecution().stream())
+                .distinct()
+                .collect(Collectors.toList());
+      }
+      to.setPreferredDuringSchedulingIgnoredDuringExecution(preferredDuringSchedulingIgnoredDuringExecution);
```
alai8, line 323: If `to.getPreferred..()` was `null`, does the array `preferredDuringSchedulingIgnoredDuringExecution` contains element from `from.getPreferred...()`?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -307,12 +311,18 @@ private void copyValues(V1Affinity to, V1Affinity from) {
...
     }
+
     if (to.getRequiredDuringSchedulingIgnoredDuringExecution() == null) {
       to.setRequiredDuringSchedulingIgnoredDuringExecution(from.getRequiredDuringSchedulingIgnoredDuringExecution());
```
alai8, line 327: do we need deep copy here?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
rjeberhard: Good catch... This bug is also in the code I copied this from (look at copyValues(V1Capabilities, V1Capabilities)).

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
rjeberhard: No, I don't think we do. You will get lists that contain all of collected terms, which is the intent.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
rjeberhard: @alai8, this is fixed... can you please recheck?

PR #1703, opened 2020-06-03
Node heating FAQ

rjeberhard:
I'm targeting this PR to `develop` now because I've only tested this with the bug fix from #1702. 

* Created an FAQ on the Node heating problem
* Updated some existing documentation about Pod generation
* [Significant] I've updated existing samples to include content for cluster-level anti-affinity


docs-source/content/faq/node-heating.md
```diff
@@ -0,0 +1,56 @@
+---
+title: "Node heating problem"
+date: 2020-06-03T08:08:19-04:00
+draft: false
+weight: 22
+---
+
+The WebLogic Operator creates a Pod for each WebLogic Server instance that is started. The [Kubernetes Scheduler](https://kubernetes.io/docs/concepts/scheduling-eviction/kube-scheduler/) then selects a Node for each Pod. Because the default scheduling algorithm gives substantial weight to selecting a Node where the necessary Docker images have already been pulled, this often results in Kubernetes running many of the Pods for WebLogic Server instances on the same Node while other Nodes are not fairly utilized.
```
rosemarymarano, line 8: WebLogic Operator -> WebLogic Server Kubernetes Operator (when we use operator capitalized, I prefer that we use the product name, either WebLogic Server Kubernetes Operator or Oracle WebLogic Server Kubernetes Operator. Please make this change globally, if you want to use the full product name, or use lowercase operator.)

docs-source/content/faq/node-heating.md
```diff
@@ -0,0 +1,56 @@
...
+
+The WebLogic Operator Team recommends a different solution that is based on [inter-pod affinity and anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity). This solution has the advantage of both resolving the Node heating problem and of explicitly directing the Kubernetes Scheduler to spread the Pods for WebLogic Server instances from a given cluster or domain more widely across the available Nodes. Inter-pod affinity and anti-affinity are features of the Kubernetes Scheduler that allow the scheduler to choose a Node for a new Pod based on details of the Pods that are already running. For WebLogic Server use cases, the intent will often be for anti-affinity with the Pods for other WebLogic Server instances so that server instances spread over the available Nodes.
```
rosemarymarano, line 12: WebLogic Operator Team -> WebLogic Server Kubernetes Operator team (team does not need to be capitalized, like a product name)

docs-source/content/faq/node-heating.md
```diff
@@ -0,0 +1,56 @@
...
+              labelSelector:
+                matchExpressions:
+                  - key: "weblogic.clusterName"
+                    operator: In
+                    values:
+                    - $(CLUSTER_NAME)
+              topologyKey: "kubernetes.io/hostname"
+```
+
+Because the `serverPod` element here is scoped to a cluster, the content of the `affinity` element will be added to the Pod generated for each WebLogic Server instance that is a member of this WebLogic cluster. This inter-pod anti-affinity statement expresses a preference that the scheduler select a Node for the new Pod avoiding as much as possible Nodes that already have Pods with the label "weblogic.clusterName" and the name of this cluster. Note that the `weight` is set to `100`, which is the maximum weight, so that this term will outweigh any possible preference for a Node based on availability of Docker images.
```
rosemarymarano, line 35: the new Pod avoiding as much as possible Nodes -> the new Pod avoiding, as much as possible, Nodes (the commas add clarity to a long sentence)

docs-source/content/userguide/managing-domains/domain-resource.md
```diff
@@ -6,9 +6,9 @@ pre = "<b> </b>"
...
+Use this document to set up and configure your own [domain resource](https://github.com/oracle/weblogic-kubernetes-operator/blob/master/docs/domains/Domain.md) which can be used to configure the operation of your WebLogic Server domain. The domain resource does not replace the traditional domain configuration files, but instead cooperates with those files to describe the Kubernetes artifacts of the corresponding domain.  For instance, the domain configuration will still specify deployed applications, data sources, and most other details about the domain while the domain resource will specify the number of cluster members currently running or the persistent volumes that will be mounted into the containers running WebLogic Server instances.
```
rosemarymarano, line 9: your own [domain resource](https://github.com/oracle/weblogic-kubernetes-operator/blob/master/docs/domains/Domain.md) which can be used -> your own [domain resource](https://github.com/oracle/weblogic-kubernetes-operator/blob/master/docs/domains/Domain.md), which can be used (you need a comma (before which) to set off a non-restrictive clause)

docs-source/content/userguide/managing-domains/domain-resource.md
```diff
@@ -164,18 +164,27 @@ spec:
...
+* `DOMAIN_NAME`: The WebLogic Server domain name.
 * `DOMAIN_UID`: The domain unique identifier.
 * `DOMAIN_HOME`: The domain home location as a file system path within the container.
-* `SERVER_NAME`: The WebLogic Server name.
+* `SERVER_NAME`: The WebLogic Server instance name.
 * `CLUSTER_NAME`: The WebLogic cluster name, if this is a cluster member.
 * `LOG_HOME`: The WebLogic log location as a file system path within the container.
 
-This example domain YAML file specifies that pods for WebLogic Server instances in the `cluster-1` cluster will have a per-Managed Server volume and volume mount (similar to a Kubernetes `StatefulSet`), an `init` container to initialize some files in that volume, and anti-affinity scheduling so that the server instances are scheduled as much as possible on different nodes:
+This example domain YAML file specifies that Pods for WebLogic Server instances in the `cluster-1` cluster will have a per-Managed Server volume and volume mount (similar to a Kubernetes `StatefulSet`), an `init` container to initialize some files in that volume, and anti-affinity scheduling so that the server instances are scheduled as much as possible on different Nodes:
```
rosemarymarano, line 187: are scheduled as much as possible on different Nodes -> are scheduled, as much as possible, on different Nodes

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/domain-resources/JRF/mii-initial-d1-JRF-v1.yaml
```diff
@@ -92,6 +92,21 @@ spec:
   clusters:
```
tbarnes-us, line 92: This file and the other domain resource files in the MII sample are generated from a template and checked-in.  
So the corresponding file 'mii-domain.yaml.template-JRF' and also the WLS template file 'mii-domain.yaml.template-WLS' also need to be modified.   

If you want to test that templates and sample files match, run 'cd src/integration-tests/model-in-image/mii-sample-wrapper  ; ./generate-sample-doc.sh '.  This will generate YAML from the templates and compare it to the files checked into source.  It should pass.   (The same script is called as part of the MII sample testing - which isn't merged into develop yet.)

kubernetes/samples/scripts/create-weblogic-domain/manually-create-domain/domain.yaml
```diff
@@ -99,6 +99,21 @@ spec:
   clusters:
   - clusterName: cluster-1
     serverStartState: "RUNNING"
+    serverPod:
+      # Instructs Kubernetes scheduler to prefer nodes for new cluster members where there are not
+      # already members of the same cluster.
+      affinity:
```
tbarnes-us, line 105: Is podAntiAffinity necessary for the sample? It complicates the MII sample's domain resource YAML.

If it _is_ necessary, can there instead be a single spec.configuration.podAntiAffinityEnabled or somesuch option to add the podAntiAffinity expression automagically?  With default true?  (If it is necessary it seems like it'd be a common enough issue we'd even want it be easy to configure and on by default.)

kubernetes/samples/scripts/create-weblogic-domain/manually-create-domain/domain.yaml
```diff
@@ -99,6 +99,21 @@ spec:
...
+        podAntiAffinity:
```
tbarnes-us, line 106: Do the MII sample tests pass with this change?  See the WLS instructions for running the stand-alone version in 'src/integration-tests/model-in-image/README'.  (I've only tested on OCI boxes - might not work directly on a mac.)    This will run the sample _and_ also run the 'check generated source' test I describe below in my other comment.

kubernetes/samples/scripts/create-weblogic-domain/manually-create-domain/domain.yaml
rjeberhard: Necessary is, of course, debatable. The reasons I went with adding the content are that a lot of customers have already hit this issue, customers blindly use samples to create their own YAML, and there aren't any functional downsides to including the content.

I don't really like the automatic insertion idea because that means this content is either hidden in a script or the template is even more complicated.  If you look at the other places where we conditionally include content in a script, the content is there in the script always but is, by default, commented out until we use `sed` to remove the commenting.

docs-source/content/faq/node-heating.md
```diff
@@ -0,0 +1,56 @@
...
+            matchExpressions:
+            - key: "weblogic.domainUID"
+              operator: In
+              values:
+              - $(DOMAIN_UID)
+          topologyKey: "kubernetes.io/hostname"
+```
+
+Details about how the WebLogic Operator generates Pods for WebLogic Server instances, including details about labels and variable substitution, are available [here]({{< relref "/userguide/managing-domains/domain-resource#pod-generation" >}}).
+
```
tbarnes-us, line 56: It could be useful to to have an example that covers all WL pods - not just for a particular domain-uid or cluster.   Eg. Any pod with a weblogic.domainUID - not just a particular domainUID.

Also, what is the precedence hierarchy for anti-affinity? In other words, what happens when multiple anti-affinity expressions are in effect?  E.g. any && domain && cluster

rjeberhard: Affinity content from the various scopes (domain, admin, cluster, etc.) are merged, so you get content with all of the terms from any of the scopes that apply.  So, yes, you can do use cases where cluster members have anti-affinity with each other and where all servers in the domain have some other kind of affinity or anti-affinity with other workloads.

rjeberhard: I'm not sure that I want more examples here unless we have a use case customers have described.

tbarnes-us: So why not have the samples and FAQ emphasize 'all'?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/domain-resources/JRF/mii-initial-d1-JRF-v1.yaml
rjeberhard: I ran this on my Linux box and it succeeded; however, it's not clear that it compared what is currently in the tree -- where would I see that?  It ends like this:
```
@@
@@ Info: Finished 'generate-sample-doc.sh'! See target directory '/tmp/mii-sample'.
@@
```

tbarnes-us:
Once the pending 'mii-doc-updates' branch is merged, and the affinity yaml in this branch for the mii sample domain resource is finalized, then the mii sample documentation should be updated to include the changes to the domain resource yaml.  (The mii sample doc includes a cut-and-paste of the sample's domain resource yaml.)

PR #1706, opened 2020-06-04
Adding introspectVersion tests

sankarpn:
- Adding two new tests in ItDomainInPV.java for newly introduced introspectVersion field under domain.spec

-   testDomainIntrospectVersionNotRolling
    Test domain status updated when introspector run triggered by introSpectVersion.
    Test Creates a domain in pv using WLST
    Updates the cluster configuration; cluster size using online WLST.
    Patches the domain custom resource with introSpectVersion.
    Verifies the introspector runs and cluster maximum and minimum replicas are updated
    under domain status.

- testDomainIntrospectVersionRolling
   Test server pods are rolled and updated when introspector run triggered by introSpectVersion.
   Updates the admin server listen port using online WLST.
   Patches the domain custom resource with introSpectVersion.
   Verifies the introspector runs and pods are rolled in rolling fashion.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/860/

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/282/


tbarnes-us:
Increasing the cluster size in this manner should also include tests that verify that the managed servers within a cluster are able to find each-other.  Specifically:

new-integration-tests/src/test/resources/python-scripts/introspect_version_script.py
```diff
@@ -0,0 +1,77 @@
...
+    cmo.setMinDynamicClusterSize(2)
+    save()
+    activate()
+    disconnect()
+  except NameError, e:
+    print('Apparently properties not set.')
+    print('Please check the property: ', sys.exc_info()[0], sys.exc_info()[1])    
+    exit(exitcode=1)
+  except:
+    print 'Deployment failed'
```
anpanigr, line 23: There is no Deployment is happening.  May be cut-n-paste error

new-integration-tests/src/test/resources/python-scripts/introspect_version_script.py
```diff
@@ -0,0 +1,77 @@
...
+  try:
+    if connected == 'false':    
+      print 'connecting to admin server'
+      connect(admin_username, admin_password, 't3://' + admin_host + ':' + admin_port)
+  except NameError, e:
+    print('Apparently properties not set.')
+    print('Please check the property: ', sys.exc_info()[0], sys.exc_info()[1])
+    exit(exitcode=1)
+  except:
+    print 'Deployment failed'
```
anpanigr, line 59: Modify the Deployment Message

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/WLSTUtils.java
```diff
@@ -0,0 +1,207 @@
...
+        .conditionEvaluationListener(
+            condition -> logger.info("Waiting for job {0} to be completed in namespace {1} "
+                + "(elapsed time {2} ms, remaining time {3} ms)",
+                jobName,
+                namespace,
+                condition.getElapsedTimeInMS(),
+                condition.getRemainingTimeInMS()))
+        .until(jobCompleted(jobName, null, namespace));
+
+    // check job status and fail test if the job failed to execute WLST
```
anpanigr, line 154: Will the Job fail if the corresponding WLST script fails ?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/WLSTUtils.java
```diff
@@ -0,0 +1,207 @@
...
+          .atMost(5, MINUTES).await();
+
+  /**
+   * Execute WLST script.
+   *
+   * @param wlstScript WLST script path
+   * @param domainProperties domain property file path
+   * @param namespace namespace in which to run the job
+   */
+  public static void executeWLSTScript(Path wlstScript, Path domainProperties, String namespace) {
```
anpanigr, line 69: Can I get the standout of the WLST script after the Job is complected ?  May be thru job log or some other mechanism

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
```diff
@@ -573,20 +596,291 @@ public void testDomainInPvUsingWdt() {
         .isTrue();
   }
 
+
```
anpanigr, line 604: For a dynamic config.xml change (e.g. changing  cluster size) updating the introSpectVersion will initiate a simultaneous re-start.

For a non-dynamic config.xml change (e.g. changing  admin server port ) updating the introSpectVersion will initiate a rolling re-start.

Is this assumption is true ?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
```diff
@@ -573,20 +596,291 @@ public void testDomainInPvUsingWdt() {
...
+    p1.setProperty("test_name", "change_server_count");
+    assertDoesNotThrow(() -> p1.store(new FileOutputStream(wlstPropertiesFile), "wlst properties file"),
+        "Failed to write the WLST properties to file");
+
+    // change the server count of the cluster by running online WLST
+    Path configScript = Paths.get(RESOURCE_DIR, "python-scripts", "introspect_version_script.py");
+    executeWLSTScript(configScript, wlstPropertiesFile.toPath(), introDomainNamespace);
+
+    // construct a patch string to update introspectVersion
+    StringBuffer patchStr = new StringBuffer("[{")
```
anpanigr, line 771: Here we can use the logic used in TestActions.patchDomainResourceWithNewRestartVersion to dynamically generate the introspectVersion string, instead of hard-coding.

This utility can be moved to TestActions class as patchDomainResourceWithNewIntrospectVersion

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
tbarnes-us: I don't think so. As far as I know, the re-introspect behavior is the same with either type of change, and an introspectVersion change is not expected to trigger a roll.  @russgold ^^^

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
```diff
@@ -573,20 +596,291 @@ public void testDomainInPvUsingWdt() {
...
+              && (res.getStatus().getClusters().get(0).getMinimumReplicas() == 2);
+        }
+        );
+  }
+
+  /**
+   * Test server pods are rolled and updated when introspector run triggered by introSpectVersion.
+   * Updates the admin server listen port using online WLST.
+   * Patches the domain custom resource with introSpectVersion.
+   * Verifies the introspector runs and pods are rolled in rolling fashion.
```
anpanigr, line 856: Description should be read as  "Verifies the introspector runs and pods are restated in a rolling  fashion"

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
```diff
@@ -573,20 +596,291 @@ public void testDomainInPvUsingWdt() {
         .isTrue();
   }
 
+
   /**
-   * Uninstall Nginx.
-   * The cleanup framework does not uninstall Nginx release.
-   * Do it here for now.
+   * Test domain status updated when introspector run triggered by introSpectVersion.
```
anpanigr, line 606: Should we have a test to just update the introSpectVersion with out any config change.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/WLSTUtils.java
sankarpn: I  am printing it when job is done executing, so you don't need to look for it under diagnostic_logs. It will be available in console log itself.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/WLSTUtils.java
sankarpn: Yes if there are any exceptions or cannot connect to the admin server the job will fail.
Of course whoever calling this utility by supplying py script will have to make sure that they exit the WLST script with exit code 1 when something goes wrong.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
```diff
@@ -0,0 +1,846 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.with;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
```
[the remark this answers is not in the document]
sankarpn: Fixed

sankarpn:
> 
> 
> Increasing the cluster size in this manner ("Updates the cluster configuration; cluster size using online WLST") should also include tests that verify that the managed servers within a cluster are able to find each-other. This is because the old servers will rely on lazily duplicated config overrides to locate the new servers, which means, unlike our previous use of overrides, the address of new servers that the old servers cache is incorrect when they first learn about the new servers (they learn about them via mbean notifications due to the on-line updates which occur far before the overrides lazily replicate).
> 
> Specifically:
> 
>     * An RMI stub to an EJB obtained from an old server should be able to load balance to the new servers (obtain several).
> 
>     * JNDI lookups from an old servers should be able to find entries published by new servers (have the EJB lookup something that's only published by the new server - such as a JMS destination member of a distributed destination targeted to the cluster).

Created a new JIRA to create such a test
https://jira.oraclecorp.com/jira/browse/OWLS-82544

new-integration-tests/src/test/resources/python-scripts/introspect_version_script.py
sankarpn: Fixed

new-integration-tests/src/test/resources/python-scripts/introspect_version_script.py
sankarpn: Fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
sankarpn: > For a dynamic config.xml change (e.g. changing cluster size) updating the introSpectVersion will initiate

It will only start the new pod if replica is also changed, won't restart existing pods

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
sankarpn: Fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
sankarpn: fixed

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
sankarpn: we can but what is the purpose of it?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDomainInPV.java
```diff
@@ -375,6 +380,8 @@ public void testDomainInPvUsingWlst() {
         .withFailMessage("NGINX can not access the test web app from one or more of the managed servers")
         .isTrue();
 
+    assertTrue(shutdownDomain(domainUid, wlstDomainNamespace), "Failed to shutdown domain");
```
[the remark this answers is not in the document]
sankarpn: Initially I added the tests in ItDominInPV.java and these tests were running after the first 2 which created 4 server pods. In order not to strain the cluster I thought of shutting down the domain before the introspectVersion test runs.

Since the introspectVersion tests are moved to a different class I removed these shutdown calls.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
```diff
@@ -0,0 +1,854 @@
...
+
+    logger.info("Updating replicas in cluster {0} using patch string: {1}", clusterName, patchStr);
+    V1Patch patch = new V1Patch(patchStr);
+    assertTrue(patchDomainCustomResource(domainUid, introDomainNamespace, patch, V1Patch.PATCH_FORMAT_JSON_PATCH),
+        "Failed to patch domain");
+
+    //verify the introspector pod is created and runs
+    logger.info("Verifying introspector pod is created, runs and deleted");
+    String introspectPodName = domainUid + "-" + "introspect-domain-job";
+    checkPodExists(introspectPodName, domainUid, introDomainNamespace);
```
rjeberhard, line 394: I think this will be fine, but I'm wondering if we can eventually do better... What I mean is that this could fail in two different ways: 1) unlikely -- but the introspector could run very fast and the test would miss that the pod ever existed, and 2) the introspector could run but fail quickly and similarly be missed. Let's see if this test starts to fail intermittently.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
```diff
@@ -0,0 +1,854 @@
...
+   * with introSpectVersion when non dynamic changes are made.
+   * Updates the admin server listen port using online WLST.
+   * Patches the domain custom resource with introSpectVersion.
+   * Verifies the introspector runs and pods are restated in a rolling fashion.
+   * Verifies the new admin port of the admin server in services.
+   * Verifies accessing sample application in admin server works.
+   */
+  @Order(2)
+  @Test
+  @DisplayName("Test introSpectVersion rolling server pods when admin server port is changed")
```
rjeberhard, line 495: no capital 'S'

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
sankarpn: Yes I agree when I was manually testing it I missed the introspector lifecycle and noticed that within 30 seconds it finished running.
The withStandardRetryPolicy has a poll delay of 2 seconds so hopefully it shouldn't miss it. But lets see if fails intermittently.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
sankarpn: Fixed it.

PR #1710, opened 2020-06-04
Fix image name

doxiao:
Remove the REPO_NAME from the image name in ITMIIMultiModel test's BeforeAll method because the common utility method it uses, createMiiImageAndVerify(), already adds REPO_NAME.


doxiao:
Jenkins job is still running, but the ItMiiMultiModel tests were clean.
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/278.

PR #1714, opened 2020-06-05
Fix a customer issue when SSL is enabled

doxiao:
Fix a bug in introspector that caused the SSL NodePort not being exposed by the admin server external service.

Jenkins job is still running https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/287/

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.py
+++ b/operator/src/main/resources/scripts/introspectDomain.py
@@ -399,11 +399,12 @@ class TopologyGenerator(Generator):
   def getSSLOrNone(self,server):
     try:
       # this can throw if SSL mbean not there
-      ret = server.getSSL()
+      ssl = server.getSSL()
       # this can throw if SSL mbean is there but enabled is false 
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -469,7 +470,7 @@ def getServerClusterPortPropertyValue(self, server, clusterListenPortProperty):
       sslListenPort = ssl.getListenPort()
     sslListenPortEnabled = None
     if ssl is not None:
-      sslListenPortEnabled = ssl.isListenPortEnabled()
+      sslListenPortEnabled = ssl.isEnabled()
```
alai8, line 473: The only reference to `getServerClusterPortPropertyValue` is from `validateClusterServersListenPortProperty`. But I cannot find `validateClusterServersListenPortProperty` being called from anywhere. If so this change in line 472 would not make any difference. I may be mistaken though.

PR #1722, opened 2020-06-10
doc operator upgrade

rosemarymarano:
As per OWLS-82389 DOCUMENTATION FOR WEBLOGIC-OPERATOR UPGRADE PROCEDURE, added procedure for upgrading the operator.

### DIFF
--- a/docs-source/content/userguide/managing-operators/installation/_index.md
+++ b/docs-source/content/userguide/managing-operators/installation/_index.md
@@ -92,7 +92,7 @@ the `helm upgrade` command requires that you supply a new Helm chart and image. 
 ```
 $ helm upgrade \
   --reuse-values \
-  --set image=oracle/weblogic-kubernetes-operator:x.x.x \
+  --set image=oracle/weblogic-kubernetes-operator:2.5.0 \
   --namespace weblogic-operator-namespace \
   --wait \
   weblogic-operator \
--- a/docs-source/content/userguide/managing-operators/installation/_index.md
+++ b/docs-source/content/userguide/managing-operators/installation/_index.md
@@ -86,12 +86,13 @@ $ helm install weblogic-operator weblogic-operator/weblogic-operator
 
 #### Upgrade the operator
 
-To upgrade the operator to a newer version, use the `helm upgrade` command:
+To upgrade the operator, use the `helm upgrade` command. When upgrading the operator,
+the `helm upgrade` command requires that you supply a new Helm chart and image. For example:
 
 ```
 $ helm upgrade \
   --reuse-values \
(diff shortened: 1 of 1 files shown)


rosemarymarano:
@mriccell @ankedia I can certainly add this text, if you want: "helm upgrade" needs to be performed with new chart and image.

docs-source/content/userguide/managing-operators/installation/_index.md
```diff
@@ -76,7 +84,22 @@ Install the operator from the repository:
...
-#### Removing the operator
+#### Upgrade the operator
+
+To upgrade the operator to a newer version, use the `helm upgrade` command:
+
+```
+$ helm upgrade \
+  --reuse-values \
+  --set image=oracle/weblogic-kubernetes-operator:3.0.0-rc1 \
+  --namespace weblogic-operator-namespace \
```
mriccell, line 95: The namespace is just consistent with the namespace used in the other helm commands in this doc.  
The name of the operator image could be more generic, not sure what is the format used to document.  The problem  with giving a tag is that every time we have a new version of the operator we need to change documentation.

rosemarymarano: @mriccell I can change the tag to weblogic-kubernetes-operator:x.x.x Would that work?
Also, should I change the text to, "To upgrade the operator, use the `helm upgrade` command. When upgrading the operator, the `helm upgrade` command requires that you supply a new chart and image."

mriccell: Yes to both

rosemarymarano: Super, done.

rjeberhard: I'm okay with having to update this doc. on each release as I always grep the full codebase. I probably should take the time to somehow "mark" or document all of the places we need to change on a release, but it's been a fairly small set so far.

rjeberhard: Looks like you already changed it :).  That's fine too.

mriccell: After Ryan's comment about version change I agree that we should use a version. The other Helm examples do have a version like 2.5.0.  Rosemary do you mind changing the x.x.x to 2.5.0 to be consistent with other helm examples?

rosemarymarano: No problem. I added the release number 2.5.0 to the image. BTW, I do have a list of the places in the docs that need to change when we release a new version.

PR #1723, opened 2020-06-10
changes for OWLS-82320 - Recieve API error and operator restarts when namespaces is removed

ankedia:
This PR addresses the periodic restart of operator pod when operator's target namespace contains a non-existent namespace during initial startup. The second issue of API exception being received and logged is because of permission error or missing namespace (403 or 404 status code). This might be expected behavior.


operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -191,11 +192,9 @@ private static void begin() {
 
       Step strategy = Step.chain(
           new InitializeNamespacesSecurityStep(targetNamespaces),
-          new NamespaceRulesReviewStep(),
-          CrdHelper.createDomainCrdStep(version,
-              new StartNamespacesStep(targetNamespaces, false)));
+          new NamespaceRulesReviewStep());
       if (!isDedicated()) {
-        strategy = Step.chain(strategy, readExistingNamespaces());
+        strategy = Step.chain(strategy, readExistingNamespaces(targetNamespaces));
```
doxiao, line 197: If I understand the change correctly, domain CRD creation and  start namespaces logic are associated with readExistingNamespaces(), which is skipped in the dedicated mode. Does that mean in dedicated mode, CRD and domains will not be created? I may have missed something. Have you tested dedicated mode use cases?

ankedia: Thanks Dongbo. I'm trying to run integration tests in external jenkins with my changes and seeing some failures. I'm not sure if failures are related to my change or intermittent env issue. I'm also trying to run tests in my machine but seeing a different issue regarding JAVA_HOME.  Do you know if there are any integration test for dedicated mode?

ankedia: Hi Dongbo, I have adjusted the change for dedicated mode as we discussed. Please let me know if you see something missing. I'll also try running tests for dedicated mode use-case. Thanks again for your comment and catching this.

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -858,24 +866,63 @@ public NextAction onFailure(Packet packet, CallResponse<V1NamespaceList> callRes
...
-      if (namespaceWatcher == null) {
-        namespaceWatcher = createNamespaceWatcher(getInitialResourceVersion(result));
+      String intialResourceVersion = getInitialResourceVersion(result);
+      List<String> nsList = getExistingNamespaces(result);
+      
+      Set<String> namespacesToStart = new TreeSet<>(targetNamespaces);
+      for (String ns : targetNamespaces) {
+        if (!nsList.contains(ns)) {
+          namespacesToStart.remove(ns);
+        }
```
doxiao, line 884: recheckDomains method also has similar filtering logic. Can yon double check and see if the logic there takes care of deleted namespaces correctly as well?

operator/src/main/java/oracle/kubernetes/operator/Main.java
doxiao: The modified code here looks good.

operator/src/main/java/oracle/kubernetes/operator/Main.java
ankedia: My understanding is that recheckDomains runs periodically and it takes care of non-existing namespaces except in full recheck case. It also checks for namespaces that are removed from operator's targetNamespaces list or that are deleted from k8s cluster and stops those namespaces. In case of full recheck (which happens every 2 min), it'll check for all namespaces in targetNamespaces (including non-existing ns). However, I believe that doesn't cause liveness probe failure since it's executed in a thread separate from main thread.

doxiao: Shall we add a log (warning?) message when we detect that a namespace that is in the target namespace but does not exist?
We may also want to add test cases for different lifecycle sequences of namespaces w.r.t targetNamesapces. For example, a namespace that is in the targetNamespaces but does not exist initially, but is added while the operator is running or not running (and later starts to run). There are multiple permutations to be tested.  If we agree that those test cases are needed, we need to create a task to track it.

ankedia: Yes, I have added a warning message when we detect that a namespace that is in target namespace but doesn't exist. Adding test cases for different lifecycle sequences of namespaces is a good idea. I'll check on that. Thanks.

rjeberhard:
@doxiao and @ankedia, what's the status of this change. Anil, were you able to test the case Dongbo described?

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -858,24 +866,70 @@ public NextAction onFailure(Packet packet, CallResponse<V1NamespaceList> callRes
     @Override
     protected NextAction onFailureNoRetry(Packet packet, CallResponse<V1NamespaceList> callResponse) {
       return isNotAuthorizedOrForbidden(callResponse)
-          ? doNext(packet) : super.onFailureNoRetry(packet, callResponse);
+          ? doNext(createDomainCrdAndStartNamespaces(targetNamespaces), packet) :
```
doxiao, line 869: Sorry, I did not notice this earlier. 
I am not sure what this change is for. It tries to create domain CRD and start the namespaces if isNotAuthorizedOrForbidden() returns true. Not sure this is correct. in addition, this change does not seem to relate to non-existing namespaces.

ankedia: Previously create domain CRD and start namespaces steps were called before the async call to read existing namespaces (unconditionally). I made this change to cover the case where  async call fails and comes to "onFailureNoRetry" after multiple retries. We still execute create domain CRD and start namespaces steps before execute next step. If failure is due to some reason other than 401 or 403, it calls "super.onFailureNoRetry". I'm not sure if calling create domain CRD and start namespaces steps will be useful here. The async call is just to read  existing namespaces and if it fails and goes onFailureNoRetry after multiple retries, there's some fundamental problem with cluster.

@rjeberhard and @doxiao  - Please let me know if this looks incorrect or if you have any suggestions. Thanks.

ankedia:
> @doxiao and @ankedia, what's the status of this change. Anil, were you able to test the case Dongbo described?

ankedia:
> @doxiao and @ankedia, what's the status of this change. Anil, were you able to test the case Dongbo described?
Yes, I have completed running the tests. Dongbo is suggesting adding new test cases to automated test suite for various namespace lifecycle sequences. I'm not sure what tests already exists and what needs to be added. I'm planning to discuss in 3:30pm meeting today. Thanks.

operator/src/main/java/oracle/kubernetes/operator/Main.java
ankedia: Just had a discussion with Ryan about this change and it looks fine. Namespace is global k8s resource and some customers don't want operator to have permission on global resources. So we still want to go ahead and create CRD and start namespaces when API call to list namespaces fails due to authorization error. In other failure cases, operator pod liveness probe will fail and pod will be restarted by kubernetes. So create domain CRD and start namespace will be executed when pod restarts.

doxiao: Okay. It sounds fine. thanks.

PR #1725, opened 2020-06-11
Add test to verify interserver communication

sankarpn:
Added test to verify the members of a cluster can see each other.

- The clusterview application binds each server instance name in the local JNDI tree for other cluster members to discover through a local JNDI lookup. 

- The sample application uses a Servlet to bind and unbind in the lifecycle methods.

- Also fixed the build application to use only the persistent volume for all build scripts and application source.

- Fixed Cleanup util to delete pv and pvc created by the build application util.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/323/

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
@@ -93,7 +93,7 @@ public class TestUtils {
    * Call the curl command and check the managed servers can see each other.
    *
    * @param curlCmd curl command to call the clusterview app
-   * @param managedServerNames managed server names that should be see other
+   * @param managedServerNames managed server names part of the cluster
    * @param maxIterations max iterations to call the curl command
    * @return true if the managed servers can see each other, false otherwise
    */
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
@@ -10,6 +10,7 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import static oracle.weblogic.kubernetes.TestConstants.K8S_NODEPORT_HOST;
 import static oracle.weblogic.kubernetes.extensions.LoggedTest.logger;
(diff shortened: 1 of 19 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
```diff
@@ -88,6 +89,66 @@ public static boolean callWebAppAndCheckForServerNameInResponse(
...
+      } else {
+        return true;
+      }
+    }
+    // after the max iterations, if hit here, one or more servers cannot see other
+    managedServers.forEach((key, value) -> {
+      if (value) {
+        logger.info("The server {0} can see other cluster members", key);
+      } else {
+        logger.info("The server {0} cannot see other cluster members", key);
```
tbarnes-us, line 145: Another possible reason is that the response contained unexpected content (did not contain ServerName/Bound).  It'd be helpful to accumulate an 'error' string that contains 'unexpected' responses (e.g. does not contain at least ServerName of one of the servers), and dump the String here on a failure.

tbarnes-us: And yet another is that all managedServers didn't respond.   Or an 'extra' managed server responded.

new-integration-tests/src/test/resources/apps/clusterview/src/java/oracle/weblogic/kubernetes/applications/clusterview/ClusterViewServlet.java
```diff
@@ -0,0 +1,150 @@
...
+  protected void processRequest(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    response.setContentType("text/html;charset=UTF-8");
+    try (PrintWriter out = response.getWriter()) {
+      out.println("<!DOCTYPE html>");
+      out.println("<html>");
+      out.println("<head>");
+      out.println("<title>ClusterViewServlet</title>");
+      out.println("</head>");
+      out.println("<body>");
```
tbarnes-us, line 84: I'm not sure about this, but if you were to view this output in a browser, I suspect the newlines would be ignored and it'd all end up munged into one line. If so, you can put a <pre> after <body> and a </pre> before </body> to correct this.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
sankarpn: > 
> 
> Another possible reason is that the response contained unexpected content (did not contain ServerName/Bound). It'd be helpful to accumulate an 'error' string that contains 'unexpected' responses (e.g. does not contain at least ServerName of one of the servers), and dump the String here on a failure.

Its already there, when http request is made in a loop it will print this until all expected servers are bound.

<!DOCTYPE html>
<html>
<head>
<title>ClusterViewServlet</title>
</head>
<body>
ServerName:ms-2
Alive:3
Health:0
Members:ms-3,ms-1,ms-2
Unable to resolve 'ms-3'. Resolved ''
Unable to resolve 'ms-1'. Resolved ''
Bound:ms-2
</body>
</html>

new-integration-tests/src/test/resources/apps/clusterview/src/java/oracle/weblogic/kubernetes/applications/clusterview/ClusterViewServlet.java
sankarpn: Added pre tag

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
tbarnes-us: - I don't see how it's already there? I see line 120 `String response = result.stdout().trim();` but no corresponding print for the response if the response doesn't contain ServerName:.
- Regardless, the error message at line 145 itself is misleading if it hits the error cases I mentioned above (e.g. an error case that isn't due to 'unbound').

sankarpn: The response from the server is always printed, that happens when the curl command is executed. line # 119
So it unnecessary to accumulate the unexpected response and print it again.

I can change the error message in line # 145 to include the case for unbound.

Please see the https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/323/consoleFull

search for `Unable to resolve 'ms-1'. Resolved ''`

PR #1727, opened 2020-06-11
give the lowest named servers in a configured cluster precedence when…

tbarnes-us:
… choosing which ones to start (fixes OWLS-82374)


alai8:
LGTM, but what are the changes for pom.xml for? I can't find any file name `SampleTestToExclude.java`

tbarnes-us:
> LGTM, but what are the changes for pom.xml for? I can't find any file name `SampleTestToExclude.java`

The main reason to change the pom.xml was to add the comment about the danger of 'M4'.
The SampleTestToExclude.java is just a random sample file name - I added the 'excludes' stanza to the unit testing as a convenience for those that want to exclude a particular unit test during development...

PR #1729, opened 2020-06-12
fix cpu request comparison issue in ItPodRestart

xiancao:
In some cases, the server pod compute resources requests cpu request returns a value with different precision. For example, the error msg below:

ItPodsRestart.testServerPodsRestartByChangingResource:216 server pod compute resources requests was not updated correctly, set cpu request to 0.5, got 0.500 ==> expected: <0.500> but was: <0.5>

Change the test to use BigDecimal comparison instead of String comparison.

Jenkins test result: https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-model-in-image-tests-10/334/


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItPodsRestart.java
```diff
@@ -201,9 +201,9 @@ public void testServerPodsRestartByChangingResource() {
     // verify the server pod resources limits got updated
     logger.info("Checking that the server pod resources cpu limit was updated correctly");
     assertNotNull(limits.get("cpu"), domain1 + "/spec/serverPod/resources/limits/cpu is null");
-    assertEquals(limits.get("cpu").getNumber().toString(), cpuLimit,
+    assertEquals(limits.get("cpu").getNumber().intValue(), cpuLimit,
```
rjeberhard, line 204: Why do you keep switching between intValue() and doubleValue()? Alternatively, the test constants could also be BigDecimal and then you could just compare them without intValue() or doubleValue().

xiancao: From the BigDecimal java doc: https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#equals-java.lang.Object-

public boolean equals(Object x)
Compares this BigDecimal with the specified Object for equality. Unlike compareTo, this method considers two BigDecimal objects equal only if they are equal in value and scale (thus 2.0 is not equal to 2.00 when compared by this method).

We will get the same issue if compare two BigDecimals.

xiancao: I can use this instead if you prefer:
assertNotEquals(limits.get("cpu").getNumber().compareTo(new BigDecimal(cpuLimit)), 0,
          String.format("server pod compute resources cpu limit is already set to %s, set cpu limit to "
              + "a different value", cpuLimit));

rjeberhard: Sure, that's fine. The cpuLimit could also just be a BigDecimal earlier as the above gets more complicated than it needs to be.

PR #1730, opened 2020-06-12
Correct introspect version

russgold:
1. The value of introspectVersion used to trigger an introspect is now recorded as a label on the introspector CM to preserve the desire for a new introspection across operator restarts.

2. Added to the REST API:
A POST to /operator/v1/domains/{uid} may have one of the following values:
{ "action" : "INTROSPECT"} to trigger an introspection on domain {uid}
{ "action" : "RESTART"} to trigger a full restart of the domain

It does so by updating the introspectVersion or restartVersion fields of the domain, which the operator can then detect.

3. Changed the order of the checkstyle and spotbugs executions, as errors in the former seem to be more common causes of build failure.


operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
```diff
@@ -68,7 +82,6 @@
    * @param accessToken is the access token of the Kubernetes service account of the client calling
    *     the WebLogic operator REST api.
    * @param targetNamespaces a list of Kubernetes namepaces that contain domains that the WebLogic
-   *     operator manages.
```
rjeberhard, line 71: This is an odd edit to this Javadoc. A mistake?

russgold: Yes.

russgold:
I'm going to revisit my REST examples over the weekend to see if there is a better approach.

operator/src/main/resources/Operator.properties
```diff
@@ -54,9 +54,10 @@ WLSKO-0052=Existing managed server Service is correct for WebLogic domain with U
 WLSKO-0053=Creating cluster Service for WebLogic domain with UID: {0}. Cluster name: {1}.
 WLSKO-0054=Replacing cluster Service for WebLogic domain with UID: {0}. Cluster name: {1}.
 WLSKO-0055=Existing cluster Service is correct for WebLogic domain with UID: {0}. Cluster name: {1}.
-WLSKO-0056=Creating domain config map, {0}, for namespace: {1}.
-WLSKO-0057=Replacing domain config map, {0}, for namespace: {1}.
-WLSKO-0058=Existing domain config map, {0}, is correct for namespace: {1}.
+WLSKO-0056=Creating config map, {0}, for namespace: {1}.
+WLSKO-0057=Replacing config map, {0}, for namespace: {1}.
+WLSKO-0171=Patching config map, {0}, for namespace: {1}.
```
lennyphan, line 59: Not a big deal and we're grouping config map messages together but should we move this down to the end for the sake of ordering?

russgold: I wasn't really sure how to handle cases like this. It would have been better, perhaps to have left gaps in the numbering.

OTOH, maybe we don't even need to preserve the numbers from one commit to the next. I just don't know the intent.

PR #1731, opened 2020-06-12
Add a function to delete namespaces starting with 'ns-'

sankarpn:
- Added a function to delete namespace artifact with names starting with 'ns-'

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/907/consoleFull


rjeberhard:
@bhavaniravichandran, click approve please in addition to the "lgtm" comment.

src/integration-tests/bash/cleanup.sh
```diff
@@ -685,6 +689,8 @@ genericDelete "$g_arg1" "$g_arg2" "$g_arg3" -friendlyDelete
 genericDelete "$g_arg1" "$g_arg2" "$g_arg3" -forceDelete
 SUCCESS="$?"
 
+deleteNamespacesIT
+
```
tbarnes-us, line 693: Unlike every other delete, this approach is done in foreground and has no retries.

Let me see if there's an easy alternative.

tbarnes-us: Also, it is not specifying the same timeout/force parameters we use everywhere else.

tbarnes-us: Finally, it is not affecting the return code for the script.

I'll see if I can come up with something more in-line with the spirit of the rest of the script.

sankarpn:
This changes in this PR is covered by 
https://github.com/oracle/weblogic-kubernetes-operator/pull/1733 
hence closing it.

PR #1735, opened 2020-06-12
Add namespace to operator log messages

doxiao:
Add domain namespace to operator's log messages.
1) When the current fiber and DomainPresenceInfo are both available, LoggingFormatter uses the existing DomainPresenceInfo to obtain the namespace information.
2) When the current fiber is available but DomainPresenceInfo is not available, add a new component to the packet to pass the namespace to the LoggingFormatter.
3) When the current fiber is not available, add the logging context to the thread local to pass the namespace to the LoggingFormatter. This is not ideal; still looking into possible alternatives.
4) Use ThreadLocal for log messages in the Watch events.

Jenkins job had one known failure in ItSessionMigration.testSessionMigration https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/350


operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -231,8 +232,23 @@ public void reportSuspendedFibers() {
...
+                    // find out the domainUID 
+                    Packet packet = fiber == null ? null : fiber.getPacket();
+                    String domainUid = null;
+                    if (packet != null) {
+                      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+                      if (info != null) {
+                        domainUid = info.getDomainUid();
+                      }
+                    }
+                    LoggingContext.context(new LoggingContext().namespace(namespace).domainUid(domainUid));
```
alai8, line 245: consider adding a utility method in LoggingContext that handles setting the context, invoke the code, and removal of context, something like `LoggingContext.runUnderContext(LoggingContext, Runnable)`.

alai8:
> When the current fiber is not available, add the logging context to the thread local

we need to be careful that while under a logging context, the fiber won't be suspended and the thread is used for running step from another fiber.

alai8:
Not sure if this PR is work in progress, but it would be nice to add some unit tests before merging. Thanks

operator/src/main/java/oracle/kubernetes/operator/ServerStatusReader.java
```diff
@@ -167,6 +168,8 @@ public NextAction apply(Packet packet) {
             String state = null;
             ClientPool helper = ClientPool.getInstance();
             ApiClient client = helper.take();
+            String namespace = pod.getMetadata().getNamespace();
+            LoggingContext.context(new LoggingContext().namespace(namespace));
```
tbarnes-us, line 172: Does this need a corresponding LoggingContext().remove() in a finally?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: Or maybe have a Closable wrapper to allow for a 'try (closable) {}' pattern?

doxiao: Changed to use AutoCloseable.

doxiao:
> Not sure if this PR is work in progress, but it would be nice to add some unit tests before merging. Thanks

Russell added the unit tests for LoggignFormatter to cover what is there without the changes in this PR. I then added test cases for the new code logic.

operator/src/main/java/oracle/kubernetes/operator/ServerStatusReader.java
doxiao: fixed

doxiao:
The latest code is being tested in the Jenkins job https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/408.

PR #1739, opened 2020-06-16
Fix for OWLS-82511 - OPERATOR NOT RECOVERING GRACEFULLY FROM LACK OF …

ankedia:
Fix for OWLS-82511 - OPERATOR NOT RECOVERING GRACEFULLY FROM LACK OF CRD.
In replace CRD Step, in case of failure due to missing permissions, continue to onSuccess case. okhttp3 client throws StreamResetException when operator doesn't have update/patch permissions for CRD. It checks for StreamResetException in onFailure and continues to onSuccess.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
@@ -37,6 +37,7 @@ import io.kubernetes.client.openapi.models.V1beta1CustomResourceSubresources;
 import io.kubernetes.client.openapi.models.V1beta1CustomResourceValidation;
 import io.kubernetes.client.openapi.models.V1beta1JSONSchemaProps;
 import io.kubernetes.client.util.Yaml;
+import okhttp3.internal.http2.StreamResetException;
 import oracle.kubernetes.json.SchemaGenerator;
 import oracle.kubernetes.operator.KubernetesConstants;
 import oracle.kubernetes.operator.calls.CallResponse;
@@ -54,6 +55,7 @@ import oracle.kubernetes.weblogic.domain.model.DomainStatus;
 public class CrdHelper {
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
   private static final String SCHEMA_LOCATION = "/schema";
+  public static final String NO_ERROR = "NO_ERROR";
   private static final CrdComparator COMPARATOR = new CrdComparatorImpl();
 
   private static final FileGroupReader schemaReader = new FileGroupReader(SCHEMA_LOCATION);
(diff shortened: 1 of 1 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -607,7 +614,12 @@ public NextAction onSuccess(
       @Override
       public NextAction onFailure(
           Packet packet, CallResponse<V1beta1CustomResourceDefinition> callResponse) {
-        return super.onFailure(conflictStep, packet, callResponse);
+        // Continue to onSuccess if failure is due to missing permissions
+        // okhttp3 client throws StreamResetException when permissions are missing
+        return ((callResponse.getE().getCause() instanceof StreamResetException)
```
doxiao, line 619: Is this the only way to identifying a permission issue? In other failure cases, operator checks the callResponse's status code to determine if it encountered a permission issue (see isNotAuthorizedOrForbidden() method). Are we checking different permissions issues here? Shall we check the status code here too?

ankedia: Yes. Initially I had made change to check for permissions in onFailureNoRetry (with isNotAuthorizedOrForbidden() call) similar to other failure cases. Unfortunately API call (in case of permission issue) fails with status code 0 and ApiException "okhttp3.internal.http2.StreamResetException: stream was reset: NO_ERROR". This is likely a bug in okhttp3 client . Also, in my testing, I noticed that retry mechanism in AsyncRequestStep is not working as expected so I could not get the steps to go if I added check for StreamResetException in onFailureNoRetry. I'll be creating a JIRA to review and fix the retry behavior in AsyncRequestStep. For now, above is the only way to fix and test the fix for permissions issue.

doxiao: I see. Yes, Ideally, the logic should be in onFailureNoRetry(), instead of onFailure(), I think. 
Shall we check status code zero too so that we know for sure we are hitting the permission issue that we want to handle? And shall we log a message to let customer know that we cannot update the CRD?

doxiao: I am okay with the current solution as long as we will try to move the logic to onFailureNoRetry once we resolve the issue in retry logic in AsyncRequestStep.

ankedia: Yes, I have created JIRA OWLS-82761 to track issue in retry logic AsyncRequestStep and am currently investigating it. I'll move logic to onFailureNoRetry once that JIRA is resolved.

ankedia: I created PR https://github.com/oracle/weblogic-kubernetes-operator/pull/1745 to fix retry logic issue in AsyncRequestStep and have also moved logic in onFailureNoRetry. If that PR is reviewed and approved before this PR, this PR will no longer be needed and can be closed.

rjeberhard:
This PR is superseded.

PR #1743, opened 2020-06-17
Add owner references to resources associated with a specific domain

rjeberhard:
Customers noticed that we could not support safely draining a node prior to shutting down that node because `kubectl drain <node-name>` would return an error indicating that some of the pods were not governed by controllers. 

I reviewed the Kubernetes doc. on draining nodes and looked at the Kubernetes code. It turns out that we needed to mark the metadata of our resources with an owner reference so that `kubectl drain` knows it can safely delete pods on the target node because a controller will recreate and reschedule pods on an available node.

With this change, Kubernetes will delete our pods on a node that is being drained and the operator will respond by creating new pods that will then be scheduled on available nodes. 

Again, after reviewing the Kubernetes doc on ownership, it made sense to mark all of the resources created for a specific domain resource with an owner reference, so I've marked the introspector job, introspector config map, and the services as well.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
@@ -3,6 +3,7 @@
 
 package oracle.kubernetes.operator.helpers;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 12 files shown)


rjeberhard:
Oh, implementation note... I moved a bunch of methods from StepContextBase to BasePodStepContext so that StepContextBase could be a common base class for the service and config map-related step context classes.

PR #1745, opened 2020-06-18
Changes to fix retry logic and recover gracefully from lack of RBAC for CRD

ankedia:
Fixed retry logic in AsyncRequestStep (OWLS-82761) and changed CrdHelper so operator can recover gracefully from lack of RBAC for CRD.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
@@ -377,14 +377,18 @@ public class AsyncRequestStep<T> extends Step implements RetryStrategyListener {
         }
 
         NextAction na = new NextAction();
-        if (statusCode == 0 && retryCount <= maxRetryCount) {
-          na.invoke(Optional.ofNullable(conflictStep).orElse(retryStep), packet);
-        } else {
-          LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
-          na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
+        if (retryCount <= maxRetryCount) {
+          if (statusCode == 0) {
+            na.invoke(retryStep, packet);
+          } else {
+            LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
+            na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
+          }
+          return na;
         }
-        return na;
-      } else if (statusCode == 409 /* Conflict */ && conflictStep != null) {
+      } else if (statusCode == 409 /* Conflict */
+          && conflictStep != null
+          && retryCount <= maxRetryCount) {
         // Conflict is an optimistic locking failure.  Therefore, we can't
         // simply retry the request.  Instead, application code needs to rebuild
         // the request based on latest contents.  If provided, a conflict step will do that.
--- a/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
@@ -377,21 +377,16 @@ public class AsyncRequestStep<T> extends Step implements RetryStrategyListener {
         }
 
         NextAction na = new NextAction();
-        if (retryCount <= maxRetryCount) {
-          if (statusCode == 0) {
-            na.invoke(retryStep, packet);
-          } else {
-            LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
-            na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
-          }
-          return na;
+        if (!retriesLeft()) {
+          return null;
+        } else if (statusCode == 0) {
+          na.invoke(retryStep, packet);
+        } else {
+          LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
+          na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
         }
-      } else if (statusCode == 409 /* Conflict */
-          && conflictStep != null
-          && retryCount <= maxRetryCount) {
-        // Conflict is an optimistic locking failure.  Therefore, we can't
-        // simply retry the request.  Instead, application code needs to rebuild
-        // the request based on latest contents.  If provided, a conflict step will do that.
+        return na;
+      } else if (retriesLeft() && isRestartableConflict(conflictStep, statusCode)) {
 
         // exponential back-off
         long waitTime = Math.min((2 << ++retryCount) * SCALE, MAX) + (R.nextInt(HIGH - LOW) + LOW);
@@ -404,6 +399,17 @@ public class AsyncRequestStep<T> extends Step implements RetryStrategyListener {
 
       // otherwise, we will not retry
       return null;
+    }
+
+    // Conflict is an optimistic locking failure.  Therefore, we can't
+    // simply retry the request.  Instead, application code needs to rebuild
+    // the request based on latest contents.  If provided, a conflict step will do that.
+    private boolean isRestartableConflict(Step conflictStep, int statusCode) {
+      return statusCode == 409 /* Conflict */ && conflictStep != null;
+    }
+
+    private boolean retriesLeft() {
+      return retryCount <= maxRetryCount;
     }
 
     @Override
--- a/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
@@ -386,7 +386,7 @@ public class AsyncRequestStep<T> extends Step implements RetryStrategyListener {
           na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
         }
         return na;
-      } else if (retriesLeft() && isRestartableConflict(conflictStep, statusCode)) {
+      } else if (isRestartableConflict(conflictStep, statusCode)) {
 
         // exponential back-off
         long waitTime = Math.min((2 << ++retryCount) * SCALE, MAX) + (R.nextInt(HIGH - LOW) + LOW);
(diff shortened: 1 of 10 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/CrdHelperTest.java
```diff
@@ -179,9 +180,7 @@ public void whenNoCrd_retryOnFailure() {
 
     Step scriptCrdStep = CrdHelper.createDomainCrdStep(KUBERNETES_VERSION_15, null);
     testSupport.runSteps(scriptCrdStep);
-
```
russgold, line 182: So the functionality change is that we _never_ retry on a create CRD failure? If so, you need to change the names of the tests you modified, since they assert that the retry will happen.

operator/src/test/java/oracle/kubernetes/operator/helpers/CrdHelperTest.java
```diff
@@ -242,8 +241,7 @@ public void whenReplaceFails_scheduleRetry() {
     Step scriptCrdStep = CrdHelper.createDomainCrdStep(KUBERNETES_VERSION_16, null);
     testSupport.runSteps(scriptCrdStep);
 
-    testSupport.verifyCompletionThrowable(FailureStatusSourceException.class);
```
russgold, line 245: As above, the test no longer does what its name says it does.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
```diff
@@ -377,14 +377,18 @@ public NextAction doPotentialRetry(Step conflictStep, Packet packet, int statusC
         }
 
         NextAction na = new NextAction();
```
russgold, line 379: This is changed behavior and there is no corresponding unit test.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
```diff
@@ -377,14 +377,18 @@ public NextAction doPotentialRetry(Step conflictStep, Packet packet, int statusC
...
+            na.invoke(retryStep, packet);
+          } else {
+            LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
+            na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
+          }
+          return na;
         }
-        return na;
-      } else if (statusCode == 409 /* Conflict */ && conflictStep != null) {
+      } else if (statusCode == 409 /* Conflict */
```
russgold, line 389: This conditional has now become complex enough that it should probably be broken out into another method. I would recommend changing it to```
    } else if (retriesLeft() && isRestartableConflict(conflictStep, status)) {```

where you also have:```

    private isRestartableConflict(Step conflictStep, int status) {
      return conflictStep != null && status == 409;
    }```

This would also allow you to move the comment explain optimistic locking failures to just above the new method

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
```diff
@@ -377,14 +377,18 @@ public NextAction doPotentialRetry(Step conflictStep, Packet packet, int statusC
         }
 
         NextAction na = new NextAction();
-        if (statusCode == 0 && retryCount <= maxRetryCount) {
-          na.invoke(Optional.ofNullable(conflictStep).orElse(retryStep), packet);
-        } else {
-          LOGGER.finer(MessageKeys.ASYNC_RETRY, identityHash(), String.valueOf(waitTime));
-          na.delay(retryStep, packet, waitTime, TimeUnit.MILLISECONDS);
+        if (retryCount <= maxRetryCount) {
```
russgold, line 380: We're now testing this condition tries, so it is worth breaking it out into its own method:```

    private boolean retriesLeft() {
      return retryCount <= maxRetryCount;
    };```

In addition, rather than using a nested if, negate the above condition and make it the first branch if an if-else if-else sequence, that simply returns na.

operator/src/test/java/oracle/kubernetes/operator/helpers/CrdHelperTest.java
ankedia: We still retry on create CRD failure. The functionality change is to log a failure message in onFailureNoRetry method when operator doesn't have permission to create CRD (status code 401) and continue to next step . Some customers want to manage their own CRD and don't want operator to fail if it doesn't have permission to create/update/patch CRD.
I have changed the name of method to reflect this and also added back one of the previous assertion. The only change is to assert that CRD creation failure message is logged instead of verifying throwable.

operator/src/test/java/oracle/kubernetes/operator/helpers/CrdHelperTest.java
ankedia: I have changed the name of test method to reflect current purpose and also added back one of the previous assertion. The only change is to assert that CRD replace failure message is logged instead of verifying throwable.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
ankedia: I have made the suggested changes, please review, Thanks.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
ankedia: I have made the suggested changes, please review, Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -559,6 +561,13 @@ public NextAction onSuccess(
         LOGGER.info(MessageKeys.CREATING_CRD, callResponse);
         return doNext(packet);
       }
+
```
russgold, line 564: This is new functionality without a unit test

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -597,6 +613,15 @@ public NextAction onSuccess(
...
+      @Override
+      protected NextAction onFailureNoRetry(Packet packet, CallResponse<V1CustomResourceDefinition> callResponse) {
```
russgold, line 618: New functionality without a unit test

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -616,6 +641,15 @@ public NextAction onSuccess(
         LOGGER.info(MessageKeys.CREATING_CRD, callResponse);
         return doNext(packet);
       }
+
+      @Override
+      protected NextAction onFailureNoRetry(Packet packet, CallResponse<V1beta1CustomResourceDefinition> callResponse) {
```
russgold, line 646: New functionality without a unit test

operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
```diff
@@ -151,8 +151,11 @@ public NextAction onFailure(Packet packet, CallResponse<T> callResponse) {
    * @return Next action for fiber processing, which may be a retry
    */
   public NextAction onFailure(Step conflictStep, Packet packet, CallResponse<T> callResponse) {
-    return Optional.ofNullable(doPotentialRetry(conflictStep, packet, callResponse))
-        .orElse(onFailureNoRetry(packet, callResponse));
+    Optional<NextAction> optionalNextAction =
```
russgold, line 154: New functionality without a unit test

ankedia: This is a bug fix and not a new functionality. Previously, "onFailureNoRetry()" method in orElse clause was getting called even when Optional.ofNullable(doPotentialRetry()) returned a non-empty Optional. The change is to make sure that onFailureNoRetry() is called when Optional.ofNullable(doPotentialRetry()) returns empty optional.

operator/src/main/java/oracle/kubernetes/operator/calls/AsyncRequestStep.java
ankedia: Just to clarify - changed behavior is to fix a bug as retry logic was not working as intended. I have added unit tests multipleFailedCallbackNoRetriesLeft_verifyCompletionThrowable() and multipleFailedCallbackRetriesLeft_nextStepAppliedWithValue() in AsyncRequestStepTest to cover this changed behavior.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
ankedia: Just to clarify - this is not a new functionality but resiliency related bug fix. This changed behavior allows operator to recover gracefully from lack of RBAC on CRD. Previously operator was not able to startup in case of missing permissions. In any case, I have added "whenNoCrd_retryOnFailureAndLogFailedMessageInOnFailureNoRetry()" test to cover this scenario. Please let me know if more unit tests needs to be added. Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
ankedia: Same as above, it's resiliency bug fix to allow operator to start in case of missing permissions. I have added whenReplaceFails_scheduleRetryAndLogFailedMessageInOnFailureNoRetry() unit test to cover this.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
ankedia: I think this is covered by whenBetaCrdReplaceFails_scheduleRetryAndLogFailedMessageInOnFailureNoRetry() unit test. Please let me know if more tests needs to be added.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
ankedia: I have added a new unit test "whenReplaceFailsThrowsStreamException_scheduleRetryAndLogFailedMessageInOnFailureNoRetry"  for StreamResetException check .

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
ankedia: I have added a new unit test whenBetaCrdReplaceThrowsStreamResetException_scheduleRetryAndLogFailedMessageInOnFailureNoRetry() for StreamResetException check.

PR #1746, opened 2020-06-18
Standardize behavior of introspectVersion and restartVersion

russgold:
As discussed, the intended behavior is now:

1. On `restartVersion` change, all affected pods are restarted in all modes
2. On `introspectVersion` change, introspection is redone in all modes. For model-in-image, if the introspection generates a new domain home then any running servers will be rolled to the new domain home.
3. Introspection will also occur if there is no current introspection-results config map or if the domain presently has zero server instances running and is now starting a server.
4. For model-in-image only, introspection will also run for every ‘make-right’ flow in which it is determined that a server pod must be updated. In particular, that entails rerunning from the start of the step chain, and this time running the introspector.
5. If introspection fails, the domain resource status must reflect the failure; however, the ‘make-right’ flow otherwise stops and any currently running pods are left intact.


tbarnes-us:
General comment:  Now that MII honors introspectVersion just like any other domain source type, the corresponding documentation needs to change throughput.

operator/src/main/java/oracle/kubernetes/operator/MakeRightDomainOperation.java
```diff
@@ -0,0 +1,70 @@
...
+
+  static void recordInspection(Packet packet) {
+    fromPacket(packet).ifPresent(MakeRightDomainOperation::setInspectionRun);
+  }
+
+  static boolean isInspectionRequired(Packet packet) {
+    return domainRequiresIntrospectionInCurrentMakeRight(packet) && !wasInspectionRun(packet);
+  }
+
+  /**
```
russgold, line 51: checkstyle demanded that I add javadoc to this method, even though it is private. That doesn't seem right to me.

PR #1747, opened 2020-06-18
Convert stickysess tests use Junit5 

hzhao-github:
use case:
--------------
Test that two HTTP requests sent to WebLogic are directed to same WebLogic server.
Server-affinity is achieved by Voyager ingress controller based on HTTP session information.


hzhao-github:
will create a new PR

PR #1748, opened 2020-06-18
OWLS-80884: Removing a Domain namespace from Operator TargetNamespace should not shutdown the domain

lennyphan:
When a namespace is removed from 'domainNamespaces' list (Targetnamespaces), the operator will not shutdown domain resources if the namespace is active (running).  This is to allow customer's to upgrade the Operator without incurring downtime to the currently running domains in the active namespace.


operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -228,14 +228,23 @@ private static void completeBegin() {
...
-    AtomicBoolean stopping =
-        remove ? isNamespaceStopping.remove(ns) : isNamespaceStopping.get(ns);
+  private static void stopNamespace(String ns, boolean inTargetNamespaceList) {
+    AtomicBoolean isNamespaceStopping = isNamespaceStopping(ns);
 
-    if (stopping != null) {
-      stopping.set(true);
+    // Remove if namespace not in targetNamespace list
+    if (!inTargetNamespaceList) {
+      Main.isNamespaceStopping.remove(ns);
```
alai8, line 236: Should we rename `isNamespaceStopping` field in Main to something else, such as `isNamespaceStoppingMap` to avoid having same name?

russgold: The name is really confusing. For example, we pass the keys from that map to the rest server:```

    startRestServer(principal, isNamespaceStopping.keySet())
```

when what we clearly mean is the list of all namespaces currently known. The problem is that it is a map of known namespace names to an atomic boolean which tells us if we are stopping that namespace, and there really isn't a good name for that. What that really suggests is that we ought to rethink the map itself, but that's a bit further off, I think.

For now, I would suggest dropping the "is." The usual convention is to use the "is" prefix for methods returning booleans, but not for the corresponding fields.

lennyphan: Yes, changed field in Main to `isNamespaceStoppingMap`.

lennyphan: Dropped the 'is.' from the 'isNamespaceStoppingMap' field in Main and renamed to 'namespaceStoppingMap' as suggested by Russ.

PR #1752, opened 2020-06-19
Fix build application utility

sankarpn:
The fix involves   

- Removing the persistent volumes for mounting application source directory  to the pod , since PV_ROOT is shared only between the Kubernetes nodes. Jenkins client doesn't have access to the PV_ROOT  so its not working in Jenkins runs.

-   Instead of a Kubernetes job the utility will create a WebLogic pod and copies the source application folder to the pod and builds the application.    

- Once built the archives directory is copied back to the local file system for use by the test code.

Results

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-model-in-image-tests/1032

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/493/testReport/oracle.weblogic.kubernetes/

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
@@ -60,10 +60,9 @@ public class BuildApplication {
    * Build application.
    *
    * <p>The appSrcPath, your application source directory is zipped up and copied to a WebLogic server pod for building.
-   * The application is built under /u01/application inside the pod. If your archives are placed under
-   * /u01/application/build after building, use <b>build</b> as the archiveDistDir param value. This method copies the
-   * folder /u01/application/<b>archiveDistDir</b> to local file system under
-   * WORK_DIR/<b>your_application_name</b>/u01/application/<b>archiveDistDir</b>
+   * If your archives are placed under &lt application_source &gt /build after building, use <b>build</b> as the
+   * archiveDistDir param value. This method copies the folder <b>archiveDistDir</b> to local file system and absolute
+   * path of the <b>archiveDistDir</b> directory is returned by this method.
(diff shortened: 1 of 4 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
```diff
@@ -75,233 +52,176 @@
...
    *
-   * @param application path of the application source folder
-   * @param parameters system properties for ant
-   * @param targets ant targets to call
-   * @param namespace name of the namespace to use for pvc
+   * <p>The appSrcPath, your application source directory is zipped up and copied to a WebLogic server pod for building.
+   * The application is built under /u01/application inside the pod. If your archives are placed under
+   * /u01/application/build after building, use <b>build</b> as the archiveDistDir param value. This method copies the
+   * folder /u01/application/<b>archiveDistDir</b> to local file system under
+   * WORK_DIR/<b>your_application_name</b>/u01/application/<b>archiveDistDir</b>
```
bhavaniravichandran, line 66: Can we just copy the ear/war file to WORK_DIR/your_application_name? /u01 is generally in the pod for our weblogic servers. So, it is a bit confusing.

sankarpn: The Kubernetes java api Copy.copyDirectoryFromPod(...) which actually returns the entire absolute path of the directory inside the pod and hence the /u01/application/...

I can get rid of the /u01/application and can form a path like WORK_DIR/applicationName/archiveDistDir ,  since we cannot put a restriction on what output directories the applications build system will use, the archiveDistDir  will be required parameter.

bhavaniravichandran: lgtm

PR #1762, opened 2020-06-24
Version CRD's by apiVersion and then product version and remove use of weblogic.resourceVersion label

rjeberhard:
The initial intention of this change was to support versioning the CRD both based on apiVersion and then by operator product version. Before this change, the CRD was versioned based on the apiVersion (e.g. "v7") and on a not-well-defined "weblogic.resourceVersion" that never changed. The problem with this pattern is that while the "v1beta1" version of CRD's just have one schema even as you support multiple apiVersions, the newer "v1" CRD's must provide a schema for every supported version and this quickly exceeds the maximum size of a CRD (because of etcd limitations).

Therefore, I want a pattern that lets us stay at apiVersion "v8" for as many releases as possible. It is reasonable to stay at the same apiVersion if from operator release to release we only add to the schema new, optional fields. 

I thought about using this "weblogic.resourceVersion" label, but that would require defining an ordering on the values of this field and it seemed clearer just to use the semantic version ordering of our product release numbers.

Since we are also going to 3.0.0 and many of these "weblogic.resourceVersion" label values contained "v2", I decided to just remove these.


rjeberhard:
> I noticed below 2 doc files with label "weblogic.resourceVersion: domain-v2" and was wondering if label reference should be removed from these?
weblogic-kubernetes-operator/site/v2.0.1/config-overrides.md
weblogic-kubernetes-operator/site/v2.0/config-overrides.md

@ankedia, I left these files alone because they are doc from the 2.0.1 and 2.0 releases of the operator. At some point, we should probably just delete all of the saved doc. from these very own releases. In general though, when fixing doc., we don't go back and fix the doc from older versions unless it's critical.

PR #1764, opened 2020-06-24
Owls 82147

tbarnes-us:
Include env vars in MII re-introspect MD5 checks so that the introspector job recreates the domain home if/when an environment variable changes between runs.


operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -294,6 +294,16 @@ private String getAsServiceName() {
         addEnvVar(vars, ServerEnvVars.DATA_HOME, dataHome);
       }
 
+      // Populate env var list used by the MII introspector job's 'short circuit' MD5
+      // check. To prevent a false trip of the circuit breaker, the list must be the
+      // same regardless of whether domainTopology == null.
+      StringBuffer sb = new StringBuffer(vars.size() * 32);
+      for (V1EnvVar var : vars) {
+        sb.append(var.getName()).append(',');
```
jshum2479, line 302: Is the name of the environment variable alone sufficient, no need to have the value?

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -245,6 +245,7 @@ function buildWDTParams_MD5() {
          " but the secret does not have this key."
       exit 1
     else
+      # TBD code review comment: pass credentials via filename instead of risking putting them in an env var
```
jshum2479, line 248: We can probably enhance WDT to take file.

tbarnes-us: OK - I'll leave the comment here as a reminder.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: Yes, the name alone is sufficient.  The 'modelInImage.sh' MD5 check iterates over the names and gets the values directly from the env vars.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -305,10 +315,6 @@ private String getAsServiceName() {
           addEnvVar(vars, "ADMIN_PORT_SECURE", "true");
         }
         addEnvVar(vars, "AS_SERVICE_NAME", getAsServiceName());
-
-        // TBD Tom Barnes, Johnny Shum
```
jshum2479, line 309: I think Tom added that for a reason. At one point in early stage of MII,  I use it for dynamic update - i.e. the job pod will use wlst online to update configuration.  That's one reason I can think of.  util.sh also has a function using it.

operator/src/test/java/oracle/kubernetes/operator/helpers/Matchers.java
```diff
@@ -165,29 +170,63 @@ public void describeTo(Description description) {
...
+    private EnvVarMatcher(String expectedName, String expectedValueRegEx) {
+      this.expectedName = expectedName;
+      this.expectedValueRegEx = expectedValueRegEx;
     }
 
     static EnvVarMatcher envVarWithName(@Nonnull String name) {
       return new EnvVarMatcher(name);
     }
 
+    static EnvVarMatcher envVarWithName(@Nonnull String name, String value) {
```
russgold, line 191: The method name implies that it is only matching the name, but it is actually matching the name and value. It should be named in a way that makes it clear what is happening. Simple "envVar" would probably be a more readable choice.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -78,6 +79,7 @@
   private static final String DOMAIN_UID = "JobHelperTestDomain";
   private static final String RAW_VALUE_1 = "find uid1 at $(DOMAIN_HOME)";
   private static final String END_VALUE_1 = "find uid1 at /u01/oracle/user_projects/domains";
+  private static final String OSEV = "OPERATOR_SUPPLIED_ENV_VARS";
```
russgold, line 82: So this is the name of an environment variable whose value is a comma-separated list of environment variable names? That's far from obvious. At the very least, please include a Javadoc comment explaining that.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -351,6 +392,13 @@ public void whenAdminServerHasValueFromEnvironmentItems_introspectorPodStartupWi
             hasItem(configMapKeyRefEnvVar),
             hasItem(secretKeyRefEnvVar),
             hasItem(fieldRefEnvVar)));
+
+    assertThat(
```
russgold, line 396: "hasEnvVarContains" is a confusing name. It is not checking to see if it has a "envVarContains," but rather whether the special env var OSEV includes the specified name in a list, correct? I had to do a fair amount of searching through the code even to come up with that suggestion.

I would suggest then, that the actual test needs to be explicit. That is, fetch the envVar and check its contents. Among other things, you are testing that this envVar does not contain an entry for the admin user and password, but the matcher `not(hasEnvVarContains(OSEV, "ADMIN_PASSWORD"))` could easily be satisfied if there is no envVar whose name matches OSEV.

To make it easier to read, I would suggest an assertion such as:

assertThat(getOSEVValues(), allOf(hasItem(configMapKeyRefEnvVar.getName()), etc.

where getOSEVValues() returns a (possibly empty) list of the names in the OSEV variable.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: I have the same recollection - and agree that the comment seems outdated now.  
Please resolve the comment if you're OK with it's removal.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
tbarnes-us: I'm not quite grokking, can you make this change?

tbarnes-us: Or suggest the exact code you're thinking about?

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
tbarnes-us: Done.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
tbarnes-us: I will fix the 'not'.
Can you suggest a better name?
As for changing to 'getOSEVValues' ... pattern instead of leveraging the Matcher, can we leave the matcher in place as is, and instead, can you suggest a better name for 'hasEnvVarContains'?  I agree the matcher is confusing - but, on the other hand (1) the matcher was _already_ confusing - that's its nature - i just expanded its purpose slightly, (2) the matcher lets us have more clear error messages when/if the assertion fails.

tbarnes-us: I made changes - not same as you asked - but hopefully clear enough. Please re-review.

operator/src/test/java/oracle/kubernetes/operator/helpers/Matchers.java
tbarnes-us: Changed to hasEnvVarRegEx(str,str), this corresponds with hasEnvVar(str,str) above...

PR #1774, opened 2020-06-29
Parse all resource versions using BigInteger

rjeberhard:
Based on an internal team's testing, we now know that the resourceVersion can be larger than Java's max. integer.


operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -226,12 +227,16 @@ private long computeNextResourceVersionFromMessage(V1Status status) {
         if (index2 > 0) {
           String val = message.substring(index1 + 1, index2);
           if (!isNullOrEmpty(val)) {
-            return Long.parseLong(val);
+            try {
+              return new BigInteger(val);
+            } catch (NumberFormatException nfe) {
```
dldnh, line 232: Long.parseLong can throw NumberFormatException - why would you swallow it for BigInteger?

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -128,14 +129,22 @@ public static boolean isFirstNewer(V1ObjectMeta first, V1ObjectMeta second) {
...
 
-  private static int getResourceVersion(V1ObjectMeta metadata) {
-    return Integer.parseInt(metadata.getResourceVersion());
+  private static BigInteger getResourceVersion(V1ObjectMeta metadata) {
+    String resVersion = metadata.getResourceVersion();
+    if (resVersion != null) {
+      try {
+        return new BigInteger(resVersion);
+      } catch (NumberFormatException nfe) {
+        // no-op
```
dldnh, line 144: why swallow the NumberFormatException?

operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -226,12 +227,16 @@ private long computeNextResourceVersionFromMessage(V1Status status) {
...
+              // no-op
```
markxnelson, line 233: no op or return BI.ZERO?

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
markxnelson, line 144: no op or zero?

rjeberhard:
@dldnh @markxnelson, thanks for the comments... Yes, I do intend to no-op on unparsable resource version. I'll update the code to make what I want more clear.

russgold:
According to the [documentation](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions), Kubernetes does not even guarantee that resource versions are integers:

``
Resource versions are strings that identify the server's internal version of an object. Resource versions can be used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server. For example, clients _must not assume resource versions are numeric_, and may only compare two resource version for equality (i.e. must not compare resource versions for greater-than or less-than relationships).
``

Does that cause problems for the Watcher logic?

rjeberhard:
@russgold, I'm about to push a change with some more description and that answers this point. As you note, the doc says that resource versions are opaque, but also requires that watch calls be able to determine if a change is "after" a resource version.

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -135,13 +136,41 @@ public static boolean isFirstNewer(V1ObjectMeta first, V1ObjectMeta second) {
...
+   *  api-conventions.md#concurrency-control-and-consistency, the resource version is technically opaque; however,
+   * the Kubernetes design also requires that clients be able to list changes to resources "after" the last
+   * change to the same or different resource. Therefore, all Kubernetes implementations use a increasing positive
+   * integer value for the resource version. This can be useful to detect out-of-order watch events. This method
+   * parses the metadata's resource version into a big integer or to 0, if the value is not parsable.
+   * @param metadata Meta data containing resource version
+   * @return The integer value of the resource version or 0, if the value is not parsable
+   */
+  public static BigInteger getResourceVersion(V1ObjectMeta metadata) {
+    if (metadata != null) {
```
russgold, line 151: Let's get in the habit of using functional programming consistently:

```
return Optional.ofNullable(metadata).map(V1ObjectMethod::getResourceVersion).orElse(BigInteger.ZERO);
```

rjeberhard:
Per conversation with Russ, I'm going to update this PR to have the watcher handle the case where the resource version can't be parsed as an integer. In such a case, the watcher won't attempt to compare the resource version to detect out-of-order watch events, but it will continue to use the literal value in the next watch. This is, admittedly, a bit academic as we've never seen a case where the resource version was not able to be parsed as an integer; only that the value became larger than max int. Separately, I could have switched to using longs -- and max. long is very big; however, I think I'll stay with BigInteger.

operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -128,14 +130,47 @@ public static boolean isFirstNewer(V1ObjectMeta first, V1ObjectMeta second) {
...
+  /**
+   * Parse the resource version from the metadata. According to the Kubernetes design documentation,
+   * https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/
+   *  api-conventions.md#concurrency-control-and-consistency, the resource version is technically opaque; however,
+   * the Kubernetes design also requires that clients be able to list changes to resources "after" the last
+   * change to the same or different resource. Therefore, all Kubernetes implementations use a increasing positive
+   * integer value for the resource version. This can be useful to detect out-of-order watch events. This method
+   * parses the metadata's resource version into a big integer or to 0, if the value is not parsable.
+   * @param metadata Meta data containing resource version
+   * @return The integer value of the resource version or 0, if the value is not parsable
```
markxnelson, line 148: you say you will return the value or zero, but you seem to return the value or null

rjeberhard: That's not correct. This method may pass null as the parameter to getResourceVersion(String), which will return 0.

markxnelson: ah ok, my bad. thanks

PR #1777, opened 2020-06-30
Improve test verification

doxiao:
1) In ItMiiDomain testPatchAppV2 test case, stop the application availability thread when it detects a failure, instead of waiting for the patched application is accessible on all managed servers.
2) In ItOperatorRestartWhenPodRoll test case, check and wait for the operator pod to be restarted, instead of,  ready to close a window between async deletion of the operator and check for podIsReady.

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
@@ -410,7 +410,7 @@ class ItMiiDomain {
         } catch (InterruptedException ie) {
           // do nothing
         }
- 
+
         // check the application availability data that we have collected, and see if
         // the application has been available all the time since the beginning of this test method
         logger.info("Verify that V2 application was available when domain {0} was being patched with image {1}",
@@ -1045,10 +1045,11 @@ class ItMiiDomain {
       String appPath
   ) {
     boolean v2AppAvailable = false;
- 
+    boolean failed = false;
+
     // Access the pod periodically to check application's availability across the duration
     // of patching the domain with newer version of the application.
-    while (!v2AppAvailable)  {
+    while (!v2AppAvailable && !failed)  {
       v2AppAvailable = true;
       for (int i = 1; i <= replicaCount; i++) {
         v2AppAvailable = v2AppAvailable && appAccessibleInPod(
@@ -1074,6 +1075,7 @@ class ItMiiDomain {
       
       if (count == 0) {
         logger.info("XXXXXXXXXXX: application not available XXXXXXXX");
+        failed = true;
       } else {
         logger.fine("YYYYYYYYYYY: application available YYYYYYYY count = " + count);   
       }
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
@@ -1045,11 +1045,10 @@ class ItMiiDomain {
       String appPath
   ) {
     boolean v2AppAvailable = false;
-    boolean failed = false;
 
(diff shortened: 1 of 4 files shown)


doxiao:
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/566/

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
```diff
@@ -1074,6 +1075,7 @@ private static void collectAppAvaiability(
       
       if (count == 0) {
         logger.info("XXXXXXXXXXX: application not available XXXXXXXX");
+        failed = true;
```
rjeberhard, line 1078: This is an odd programming style. Why not use break?

rjeberhard: Related question... why aren't you using our ordinary retry pattern?

doxiao: > This is an odd programming style. Why not use break?

Yes, actually I can just exit the loop with break.

doxiao: > Related question... why aren't you using our ordinary retry pattern?

Did you mean using ConditionFactory?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItOperatorRestartWhenPodRoll.java
```diff
@@ -235,18 +234,14 @@ private void restartOperatorAndVerify() {
             opNamespace,
             condition.getElapsedTimeInMS(),
             condition.getRemainingTimeInMS()))
-        .until(assertDoesNotThrow(() -> operatorIsReady(opNamespace),
-          "operatorIsReady failed with ApiException"));
+        .until(assertDoesNotThrow(() -> isOperatorPodRestarted(opNamespace, opPodCreationTime),
```
vanajamukkara, line 237: you already have operator pod name in this method, you can directly call isPodRestarted(podName, namespace, timestamp)?

doxiao: I only have the old name, and a restarted pod has a different name.

doxiao: isOperatorPodRestarted() does not require the operator pod name because we'll only have one operator in a namespace.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
doxiao: The pattern that we use in the Unit test cases are easier for a simple condition retries. Here we have additional operation to perform. So I did not think that was a better pattern for this.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDomain.java
```diff
@@ -1074,6 +1074,7 @@ private static void collectAppAvaiability(
       
       if (count == 0) {
         logger.info("XXXXXXXXXXX: application not available XXXXXXXX");
```
rjeberhard, line 1076: What are the "XXX" and "YYY"? What application is it?

doxiao: Those "XXX" and "YYY" marks just make it easier for people to notice the failure/healthy conditions. The failure case "XXX" is INFO, and "YYY" is "FINE".

PR #1780, opened 2020-07-01
Remove old situation configuration entries from the introspector map

russgold:
When an introspection result lacks Sit-Cfg entries that are in the old introspector config map, remove the old entries.

Also cleaned up some dead code.


operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -284,10 +301,10 @@ boolean isCompatibleMap(V1ConfigMap existingMap) {
       public NextAction onSuccess(Packet packet, CallResponse<V1ConfigMap> callResponse) {
         DomainPresenceInfo.fromPacket(packet).map(DomainPresenceInfo::getDomain).map(Domain::getIntrospectVersion)
               .ifPresent(value -> addLabel(INTROSPECTION_STATE_LABEL, value));
-        V1ConfigMap existingMap = callResponse.getResult();
+        V1ConfigMap existingMap = withPreservedData(callResponse.getResult());
```
tbarnes-us, line 304: If we have  from the introspect job result that short circuited, I assume we can still get here?  If so, the existingMap will be stripped of its Sit-Cfg keys, and the update at line 308 looks it will consequently (improperly) remove all Cfg from the introspect cm.

sankarpn: I am not familiar with this code. But can it handle the situation where the existing override cm is recreated with different config.xml? Is the code only looks for the files in the map or does it looks for changes in the data of those files as well?

russgold: I added a unit test, and it was indeed removing the entries because of the way the test code works, but would not have done so in production.

This code, in ConfigMapHelper.IntrospectionConfigMapStep```
      IntrospectionLoader loader = new IntrospectionLoader(packet, this);
      if (loader.isTopologyNotValid()) {
        return doNext(reportTopologyErrorsAndStop(), packet);
      } else if (loader.getDomainConfig() == null)  {
        return doNext(loader.createIntrospectionVersionUpdateStep(), packet);
      } else {
        LOGGER.fine(MessageKeys.WLS_CONFIGURATION_READ, timeSinceJobStart(packet), loader.getDomainConfig());
        loader.updatePacket();
        return doNext(loader.createValidationStep(), packet);
      }
    }
```
detects the short-circuit case, and creates a step that only patches a label. Thus, it wouldn't change any of the entries; however, since the local copy in the configmap steps was actually the same instance as the simulated configmap, it was causing the test to fail. I have fixed that by not updating the local copy in the patch-only case, but I should really fix the simulated Kubernetes. This isn't the first time I've had trouble testing a complex case.

tbarnes-us: Thanks Russ - I hope that's the only possible path where the candidate map is empty of Sit-Cfg files.  If we can't be sure, then the code could guard against the possibility by only calling 'withPreservedData' when the candidate replacement map actually contains Sit-Cfg files (the introspector currently always generates at least one Sit-Cfg file when it doesn't short-circuit).

BTW I too am curious about Sankar's question above, and also think that this code path only appears to check for key existence when making decisions - it doesn't seem to check for differences.

russgold: It checks for differences in the pre-existing code.

PR #1785, opened 2020-07-02
Dependency and plugin version updates

rjeberhard:


rjeberhard:
Yes, this was the only reference to sldf.

PR #1791, opened 2020-07-09
Use fixed number of threads to run the tests in parallel and other fixes

vanajamukkara:
Number of threads used to run the tests in parallel for kind cluster is set 2 by default.
Fix for some of the tests which are failing when ran in parallel.

There are still some intermittent failures in parallel runs which needs to be fixed -
Voyager resource conflicts - Huiling is working on a fix
JRF DB port conflicts - Maggie is looking into this
MII Samples Traefik port conflict - https://jira.oraclecorp.com/jira/browse/OWLS-83143
if any other failures - individual test owners will have to look into them to make the test work in parallel runs

Parallel run with two threads - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/661/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/657/

Sequential run on the branch, no regression - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/656/


kindtest.sh
```diff
@@ -36,7 +36,7 @@ script="${BASH_SOURCE[0]}"
 scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
 
 function usage {
-  echo "usage: ${script} [-v <version>] [-n <name>] [-o <directory>] [-t <tests>] [-c <name>] [-p true|false] [-h]"
+  echo "usage: ${script} [-v <version>] [-n <name>] [-o <directory>] [-t <tests>] [-c <name>] [-p true|false] [-th <number_of_threads>] [-h]"
```
anpanigr, line 39: Generally the command line arguments are single letter. we can make add -p <parallel thread count>

vanajamukkara: -p is already there, -t and -n are also there, do you have any suggestions?

anpanigr: I am also out of letters use some letter like -x

PR #1797, opened 2020-07-13
Using different ports instead of starting at default 30000

sankarpn:
Using a different starting node port range for t3channel nodeport. The tests tries to get a unused port and then creates a domain with that t3channel nodeport. But by the time the domain starts up the port is gone and domain startup fails.

So using a different starting port ranges in hope of reducing the port clash.


vanajamukkara:
Looks good. Please update the PR with Jenkins results, both parallel and sequential runs.

sankarpn:
closing this PR since it is covered by PR # 1800

PR #1799, opened 2020-07-13
Owls 83019 stop JRF create primordial retry if WDT failed

jshum2479:
I tested this for failure situation and also works after fixing the failure with new restart version,  let me know if you see any holes in it.


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -679,7 +679,11 @@ function wdtCreatePrimordialDomain() {
     > ${WDT_OUTPUT}
   ret=$?
   if [ $ret -ne 0 ]; then
-    trace SEVERE "WDT Create Domain Failed ${ret}"
+    if [ "JRF" == "$WDT_DOMAIN_TYPE" ] ; then
```
doxiao, line 682: Do we need to do the same for update code path in wdtUpdateModelDomain method?

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -33,9 +33,8 @@
 
   String ENCODED_CREDENTIALS = "encodedCredentials";
   String DOMAIN_VALIDATION_ERRORS = "domainValidationErrors";
-
+  //  String UNRECOVERABLE_INTROSPECT_ERRORS = "unrecoverableIntrospectErrors";
```
doxiao, line 36: shall we remove the added code that is commented out?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -625,10 +626,20 @@ public boolean wasInspectionRun() {
...
+          .map(DomainPresenceInfo::getDomain)
+          .map(Domain::getStatus)
+          .map(DomainStatus::getMessage)
+          .orElse(null);
       if (cachedInfo == null || cachedInfo.getDomain() == null) {
         return true;
       } else if (isCachedInfoNewer(liveInfo, cachedInfo)) {
         return false;  // we have already cached this
+      } else if (existingError != null && existingError.startsWith("MII Fatal Error")
+          && !isSpecChanged(liveInfo, cachedInfo)) {
```
doxiao, line 639: what if the spec is changed and the change is causing a fatal error?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -625,10 +626,20 @@ public boolean wasInspectionRun() {
 
     private boolean isShouldContinue() {
       DomainPresenceInfo cachedInfo = getExistingDomainPresenceInfo(getNamespace(), getDomainUid());
+      String existingError = Optional.ofNullable(liveInfo)
```
doxiao, line 629: I don't think all introspector errors are reported in the domain status. We need to make sure that all SEVERE JRF ones are.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -679,7 +679,11 @@ function wdtCreatePrimordialDomain() {
...
+      trace SEVERE "MII Fatal Error: WDT Create Domain Failed ${ret}"
```
tbarnes-us, line 683: Add comment that "MII Fatal Error" is a keyword that the operator detects, and if detected, the operator will suppress introspection retries.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -679,7 +679,11 @@ function wdtCreatePrimordialDomain() {
...
+    else
```
tbarnes-us, line 684: It looks like MII Fatal is used for _any_ JRF failure path, instead of just the JRF schema create path.  

I strongly suspects we still want retries for failures other than JRF schema create.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
```diff
@@ -521,6 +520,25 @@ public void afterChangeTriggersIntrospection_doesNotRunIntrospectionOnNextExplic
     assertThat(job, nullValue());
   }
 
+  @Test
+  public void whenDomainTypeIsFromModelDomainAndAdminServerModified_runIntrospectionJobFirst2() throws Exception {
```
russgold, line 524: There are two things being checked here; please split it into two tests. Use blank lines to separate setup, execution, and validation.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -64,6 +64,7 @@
 import oracle.kubernetes.weblogic.domain.model.AdminService;
 import oracle.kubernetes.weblogic.domain.model.Channel;
 import oracle.kubernetes.weblogic.domain.model.Domain;
```
tbarnes-us, line 66: Where is the Info level message we discussed in slack that informs users that no wallet file is specified, and therefore a schema creation attempt is likely about to be made?

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: The JRF schema data is only created and inserted during domain creation, the update usually deal with resources and app deployment, that's why it only in creates

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
jshum2479: Yes, it should be and shouldn't be there

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: This is one area that I depends on, I believe Anil has fixed this in an earlier PR that makes the SEVER error shows up the domain status.

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: There is no schema create path in WDT create domain,  this is a single entry point and unless there is a way WDT can differentiate it and return differently

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
jshum2479: The test tries to simulate the first introspection that returns an error in the domain status and then in the next retry will stop the introspection retries.   How to make it into two tests and the second one runs after the first one without affecting all other tests?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: It's not there yet.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: Good point, I think I can remove the check

jshum2479: Well, if I don't put it there, then after fixing the underlying error and changing the spec will not run the introspector ever again and essentially have to delete the domain and retry.

doxiao: If the spec change is causing a fatal error, and the operator continues retry, will the isSpecChanged() returns false in the next retry?  In other word, does the operator update the current info on each retry? If so, I think you are good.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: I noticed that the task that tracks Anil's work is still open (https://jira.oraclecorp.com/jira/browse/OWLS-82011).

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: added

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: Yes, that's the idea.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: We can wait till 82011 is merged and recheck

operator/src/main/resources/scripts/modelInImage.sh
tbarnes-us: If there's no wallet file, then you know, right?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: Can you add it?  It can go in the java or the shell script.

operator/src/main/resources/scripts/modelInImage.sh
tbarnes-us: How about:
    # MII Fatal Error is handled by DomainProcessorImpl.isShouldContinue
-->
    # The string "MII Fatal Error" is detected by DomainProcessorImpl.isShouldContinue

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: If the wallet is wrong or passphrase is wrong, should we auto retry?  I am just on the safe side but we can try to open it up

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -674,12 +674,25 @@ function wdtCreatePrimordialDomain() {
...
   ${WDT_BINDIR}/createDomain.sh -oracle_home ${ORACLE_HOME} -domain_home ${DOMAIN_HOME} $model_list \
   ${archive_list} ${variable_list}  -domain_type ${WDT_DOMAIN_TYPE} ${OPSS_FLAGS}  ${UPDATE_RCUPWD_FLAG}  \
     > ${WDT_OUTPUT}
   ret=$?
   if [ $ret -ne 0 ]; then
-    trace SEVERE "WDT Create Domain Failed ${ret}"
+    # MII Fatal Error is handled by DomainProcessorImpl.isShouldContinue
+    # If it is detected then it will stop the periodic retry
+    if [ "JRF" == "$WDT_DOMAIN_TYPE" ] && [ -z "${OPSS_FLAGS}" ] ; then
+      trace SEVERE "MII Fatal Error: WDT Create Primordial Domain Failed ${ret}"
```
tbarnes-us, line 692: Please add a comment explaining why we're disabling retries, something like:
# We need to prevent retries with a "MII Fatal Error" because JRF without the OPSS_FLAGS indicates
# a likely attempt to initialize the RCU DB schema for this domain, and we don't want to retry when this fails
# without admin intervention (retrying can compound the problem and obscure the original issue).

operator/src/main/resources/scripts/modelInImage.sh
tbarnes-us: The update looks good to me, thanks.  Added a separate comment asking for more comments.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: I am okay with getting this change in and doing more testing once owls-82011 is complete.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: I'm OK with this pull going in if the Info message isn't in yet - it can go into another pull.

jshum2479: It is there  line 677 in modelInImage.sh

  if [ "JRF" == "$WDT_DOMAIN_TYPE" ] ; then
    if [ -z "${OPSS_FLAGS}" ] ; then
      trace "Creating JRF Primordial Domain"
    else
      trace "Creating JRF Primordial Domain with OPSS Wallet and Password"
    fi
  fi

tbarnes-us: That doesn't cover it. (1) Primordial is an internal term. (2) Default trace level is FINE, so it won't show up in the Operator log by default.   How about:
if [ -z "${OPSS_FLAGS}" ] ; then
trace "Creating JRF Primordial Domain"
else
trace "Creating JRF Primordial Domain with OPSS Wallet and Password"
fi

tbarnes-us: How about:

if [ -z "${OPSS_FLAGS}" ] ; then
trace "Creating JRF Primordial Domain"
else
trace "Creating JRF Primordial Domain with OPSS Wallet and Password"
trace INFO "An OPSS wallet was not supplied for Model in the Image JRF domain 'spec.configuration.opss.walletFileSecret' attribute; therefore, it's assumed that this is the first time the OPSS RCU database is being accessed by the domain, so a schema and a wallet file will be created. Consult the Model in Image documentation for preserving the OPSS wallet file."
fi

tbarnes-us: Let's continue the discussion on Slack if needed.

tbarnes-us: Oops:  The INFO goes above the else!  Editing above to fix/refine.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -676,9 +676,12 @@ function wdtCreatePrimordialDomain() {
 
   if [ "JRF" == "$WDT_DOMAIN_TYPE" ] ; then
     if [ -z "${OPSS_FLAGS}" ] ; then
-      trace "Creating JRF Primordial Domain"
+      trace INFO "An OPSS wallet was not supplied for Model in the Image JRF domain " \
+        "'spec.configuration.opss.walletFileSecret' attribute; therefore, it's assumed that this is the first time " \
+        "the OPSS RCU database is being accessed by the domain, so a schema and a wallet file will be created. " \
+        "Consult the Model in Image documentation for preserving the OPSS wallet file."
```
tbarnes-us, line 682: This is the uncorrected suggested contents for the INFO.  Refresh the browser to get the corrected contents.

tbarnes-us: Here are the corrected contents again in case you can't find them easily:

An OPSS wallet was not supplied for the Model in Image JRF domain in its 'spec.configuration.opss.walletFileSecret' attribute; therefore, it's assumed that this is the first time the OPSS RCU database is being accessed by the domain, so a schema and a wallet file will be created. Consult the Model in Image documentation for instructions about preserving the OPSS wallet file.

Issue #1769, opened 2020-06-26
Deployment's api version in 'oracle.db.yaml' should be changed to 'apps/v1' for K8S version 1.18.x


sbattagi:
Hi, 

Creating a Database deployment using the file "weblogic-kubernetes-operator/kubernetes/samples/scripts/create-oracle-db-service/common/oracle.db.yaml" throws below error on K8S version 1.18.x.
ERROR:
error: unable to recognize "/scratch/sbattagi/workdir/weblogic-kubernetes-operator-2.6.0/kubernetes/samples/scripts/create-oracle-db-service/common/oracle.db.yaml": no matches for kind "Deployment" in version "extensions/v1beta1"

As per the K8S 1.18 release notes shown below, we have to use "apps/v1" instead of "extensions/v1beta1".
"daemonsets, deployments, replicasets resources under extensions/v1beta1 - use apps/v1 instead" 
https://kubernetes.io/docs/setup/release/notes/#kube-apiserver-1

Please review and fix the issue.

Thanks
Samba

sbattagi:
This issue is seen in operator 2.6.0 release branch.

rjeberhard:
Fixed in 3.0.

PR #1805, opened 2020-07-14
Update Deployment and Ingress apiVersions

rjeberhard:
This PR resolves issue #1769, which we've been asked to expedite. The issue only mentions the Deployment apiVersion, but while reviewing recent changes to Kubernetes, I noticed that both Ingress and Deployment have been long-since moved to new apiVersions and the updated versions are now supported by our minimum Kubernetes version (1.14).

Therefore, I've updated references to both Deployment and Ingress in this PR.

### DIFF
--- a/docs-source/content/samples/simple/domains/model-in-image/prerequisites.md
+++ b/docs-source/content/samples/simple/domains/model-in-image/prerequisites.md
@@ -52,8 +52,8 @@ weight: 1
 
     - Option 1: To create the ingresses, use the following YAML file to create a file called `/tmp/mii-sample/ingresses/myingresses.yaml` and then call `kubectl apply -f /tmp/mii-sample/ingresses/myingresses.yaml`:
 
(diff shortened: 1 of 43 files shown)


rjeberhard:
Converting this to draft because I realized there are more files that need to change.

rjeberhard:
Okay, fixed... Tests running now. I needed to update the Java classes too.

alai8:
Do we also need to update apps/v1beta to apps/v1 in these files? 
./src/integration-tests/kubernetes/logstash.yaml
./src/integration-tests/kubernetes/elasticsearch.yaml
./src/integration-tests/kubernetes/kibana.yaml
./kubernetes/samples/scripts/elasticsearch-and-kibana/elasticsearch_and_kibana.yaml

rjeberhard:
Yes, thanks, @alai8. Those files are updated now.

PR #1807, opened 2020-07-14
correct host-name in health check REST url when Pod IP is null

ankedia:
Remove the default "cluster.local" search path when creating host-name for health check REST URL. This search path can be different for different cloud providers for e.g. it's ".svc.occloud" in GBU CNE team's environment. If the services are in same cluster, just adding namespace to service-name should be sufficient.


alai8:
Looks good, but can you look into adding new or modifying existing unit test to test this? Thanks

PR #1815, opened 2020-07-16
Fixes necessary for 2.6.0 to 3.0.0 upgrade

rjeberhard:
Thank goodness @vanajamukkara tested the upgrade path again!

I found and fixed a few problems:
- When the CRD replacement failed, you could only see the error message with FINE logging
- The new schema version was added to the CRD in the wrong place
- For the v1beta1 CRD version, we weren't updating the top-level version field
- We were failing to mark the older schema versions as no longer being storage versions
- We were failing to mark the new schema version as being the storage version

We were also seeing a problem with etcd complained that the updated CRD was "too big". The only solution to this was to change schema generation to not include descriptions from the built-in Kubernetes types that we reference. I compensated for this by referring to the `kubectl explain` documentation for those types.


integration-tests/src/test/java/oracle/kubernetes/operator/ItOperatorUpgrade.java
```diff
@@ -157,6 +189,31 @@ private void upgradeOperator() throws Exception {
...
+    operatorMap.put("serviceAccount", OP_SA);
+    operatorMap.put("externalRestEnabled", true);
+    operatorMap.put("javaLoggingLevel", "FINE");
+    List<String> domNs = new ArrayList<String>();
+    domNs.add(DOM_NS);
+    operatorMap.put("domainNamespaces", domNs);
+    operator = TestUtils.createOperator(operatorMap, Operator.RestCertType.LEGACY);
+
+    checkCrdVersion();
+    checkDomainNotRestarted();
```
rjeberhard, line 212: For 2.6 to 3.0 upgrade, I do expect the domain to be restarted.

vanajamukkara: Yes, domain is getting restarted. Removed the check.

vanajamukkara:
@rjeberhard should the upgrade from Operator 2.5.0 to 3.0.0 work? I see the CRD version is updated to v8, but only admin server is restarted not the managed servers and NPE in operator logs on k8s 1.15.11, scaling is not working. Not sure if this should work.

rjeberhard:
@vanajamukkara 
> @rjeberhard should the upgrade from Operator 2.5.0 to 3.0.0 work? I see the CRD version is updated to v8, but only admin server is restarted not the managed servers and NPE in operator logs on k8s 1.15.11, scaling is not working. Not sure if this should work.

I don't know if this could be made to work, but I'm okay documenting that customers can only upgrade to 3.0.0 from 2.6.0.

rjeberhard:
@vanajamukkara, if you are happy :) then please go ahead and approve this PR.

PR #1817, opened 2020-07-16
update WDT version used by domain-in-image samples...

tbarnes-us:
…and make it configurable

Old version: Previously, the DII default was effectively 0.17, as DII depends on the docker image project's sample which has that default.  

Work around: Manually export WDT_VERSION first. The docker image sample accepts a WDT_VERSION env var with either a specific version or the keyword 'LATEST'.

Fix: Have the DII sample set the WDT_VERSION env var prior to calling the docker image sample build.sh script. Allow users to specify the value as an 'input' (input attr 'wdtVersion'). The DII default is now 1.9.1.

Note that the DIPV WDT samples already default to 1.9.1 (it's set to 1.9.1 in the DIPV 'create domain job' script).

### DIFF
--- a/kubernetes/samples/scripts/common/utility.sh
+++ b/kubernetes/samples/scripts/common/utility.sh
@@ -346,8 +346,6 @@ function createFiles {
     weblogicCredentialsSecretName="${domainUID}-weblogic-credentials"
   fi
 
-  wdtVersion="${WDT_VERSION}"
-
   if [ "${domainHomeInImage}" == "true" ]; then
     domainPropertiesOutput="${domainOutputDir}/domain.properties"
     domainHome="/u01/oracle/user_projects/domains/${domainName}"
@@ -389,6 +387,9 @@ function createFiles {
       sed -i -e "s|%IMAGE_NAME%|${image}|g" ${domainPropertiesOutput}
     fi
   else
+    # we're in the domain in PV case
+
+    wdtVersion="${WDT_VERSION:-${wdtVersion}}"
 
     createJobOutput="${domainOutputDir}/create-domain-job.yaml"
     deleteJobOutput="${domainOutputDir}/delete-domain-job.yaml"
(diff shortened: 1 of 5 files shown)


kubernetes/samples/scripts/common/utility.sh
```diff
@@ -389,6 +387,9 @@ function createFiles {
       sed -i -e "s|%IMAGE_NAME%|${image}|g" ${domainPropertiesOutput}
     fi
   else
+    # we're in the domain in PV case
+
+    wdtVersion="${WDT_VERSION:-${wdtVersion}}"
```
tbarnes-us, line 392: Implementation note:

The changes in this file only affect the PV path, which is not the main thrust of this pull, but I did this for clarity.  

Specifically, I moved the 'wdtVersion' initialization for PV down into the else clause to make it more clear that it's only used on the PV path, plus, instead of setting 'wdtVersion' to WDT_VERSION directly (WDT_VERSION is an undocumented setting), I set wdtVersion to come from the 'wdtVersion' env var if (a) it's already been set and (b) WDT_VERSION has not been set.  (If the wdtVersion env var has been set, then it probably came from the inputs file.)

Phew!

kubernetes/samples/scripts/create-fmw-infrastructure-domain/domain-home-in-image/create-domain-inputs.yaml
```diff
@@ -142,6 +142,11 @@ domainHomeImageBase: container-registry.oracle.com/middleware/fmw-infrastructure
 # which uses WDT, instead of WLST, to generate the domain configuration.
 domainHomeImageBuildPath: ./docker-images/OracleFMWInfrastructure/samples/12213-domain-home-in-image
 
+# Uncomment to customize the WebLogic Deploy Tool (WDT) version used by the WDT
+# "domain home in image" Docker image sample in the Docker images project. See
+# "domainHomeImageBuildPath". Set to "LATEST" to use the latest version.
+# wdtVersion: LATEST
```
mriccell, line 148: Why would we not want LATEST to be the default?

tbarnes-us: This change is already merged.

This change implicitly defaults to 1.9.1 since that's what we used in all of the other samples already - presumably because this helps make sure the samples are stable/reproducable.  (I don't document the default here because I assume that'll change over time.)

rjeberhard:
@mriccell, we have a bunch of other code points that are currently using a hardcoded 1.9.1. @ddsharpe mentioned this too. I'll create a JIRA to address this across the usages.

PR #1826, opened 2020-07-20
update the JAVA_URL in Dockerfiles

xiancao:
The latest openjdk build is 12. Change Dockefile to set the correct JAVA_URL:
ENV JAVA_URL https://download.java.net/java/GA/jdk14.0.2/205943a0976c4ed48cb16f1043c5c647/12/GPL/openjdk-14.0.2_linux-x64_bin.tar.gz


mriccell:
@markxnelson we just need this change right now so that the tests can build the Operator image and pass. Later we will create a PR to automate the lookup of the JAVA URL.

PR #1830, opened 2020-07-22
Test support: don't update status on replace if defined as subresource

russgold:
- Updated KubernetesTestSupport so that it replaceDomain preserves the old status.
- Added interface to synchronous domain status replace

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesTestSupport.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesTestSupport.java
@@ -110,11 +110,17 @@ public class KubernetesTestSupport extends FiberTestSupport {
   public static final String SELF_SUBJECT_RULES_REVIEW = "SelfSubjectRulesReview";
   public static final String TOKEN_REVIEW = "TokenReview";
 
-  private static RequestParams REQUEST_PARAMS
+  private static final String PATH_PATTERN = "\\w+(?:.\\w+)*";
+  private static final String OP_PATTERN = "=|==|!=";
+  private static final String VALUE_PATTERN = ".*";
+  private static final Pattern FIELD_PATTERN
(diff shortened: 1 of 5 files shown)


rjeberhard:
@russgold, just making sure we are on the same page... I think this PR is changing the test support layer to honor the behavior described here: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource:

When the status subresource is enabled, the /status subresource for the custom resource is exposed.

- The status and the spec stanzas are represented by the .status and .spec JSONPaths respectively inside of a custom resource.

- PUT requests to the /status subresource take a custom resource object and ignore changes to anything except the status stanza.

- PUT requests to the /status subresource only validate the status stanza of the custom resource.

- PUT/POST/PATCH requests to the custom resource ignore changes to the status stanza.

- The .metadata.generation value is incremented for all changes, except for changes to .metadata or .status.

russgold:
@rjeberhard Yes, that's the intent. In particular, Anil is working on an issue that he said he could not unit test, as the test support wasn't honoring those rules. I added these changes to support his work.

russgold:
Hold on. Anil has pointed out a case I missed.

operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesTestSupport.java
```diff
@@ -607,6 +633,10 @@ private T withOptionalCreationTimeStamp(T resource) {
       return resource;
     }
 
+    private void preserveStatusIfSeparate(String name, T resource) {
+
```
lennyphan, line 637: should this be removed?

rjeberhard: Lenny, I'm not understanding why this might need to be removed...

russgold: Yes; thanks for noticing that.

PR #1834, opened 2020-07-24
Use kubectl exec to fix a test hanging issue

doxiao:
Replace calls to Kubernetes java client exec API with kubectl exec calls to temporarily fix a test hanging issue in Jenkins that has started to happen recently.


doxiao:
Jenkins job before the last commit (no hanging in testPatchAppV2): https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/894/.
Jenkins job with the latest: https (still running): //build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/902/

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Application.java
```diff
@@ -78,6 +80,42 @@ public static boolean appAccessibleInPod(
...
+      String appPath,
+      String expectedResponse
+  ) {
+
+    // calling "kubectl exec" command to access the app inside a pod
+    String cmd = String.format(
+        "kubectl -n %s exec -it %s -- /bin/bash -c 'curl http://%s:%s/%s'",
+        namespace,
+        podName,
+        podName,
```
tbarnes-us, line 108: The podName can differ from the service name.  Should the second 'podName' be converted to DNS 1123?  Alternatively, it could  be changed to "$SERVICE_NAME" as I assume the bash command will have access to the pod's env vars...

tbarnes-us: (DNS 1123 == podName all lower case with "_" converted to "-")

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Application.java
```diff
@@ -78,6 +80,42 @@ public static boolean appAccessibleInPod(
...
+        port,
+        appPath);
+
+    CommandParams params = Command
+        .defaultCommandParams()
+        .command(cmd)
+        .saveResults(true)
+        .redirect(false)
+        .verbose(false);
+    return Command.withParams(params).executeAndVerify(expectedResponse);
```
tbarnes-us, line 118: For other tests, we've needed to have multiple retries for initial curl attempts as tests can sometimes try access an app before it or the service is fully deployed; in addition, it's been helpful to tune the curl command's connection and overall timeouts.

tbarnes-us: For MII sample testing, we use "--connect-timeout 5 --max-time 20" and loop the curl command up to 15 times on a failure (5 seconds between each try).

doxiao: The test that uses this assertion has a retry logic at the higher level; this is the pattern used in the Juni5 integration tests.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Application.java
doxiao: I totally agree that the code should convert the podName to a legal DNS1123 name. But this pattern is used everywhere in the integration test cases; the assumption is that the test code will not use any non-legal-1123 name as the domainUID, clusterName, managedServerPrefix and etc.  I know this assumption may not be true once more engineers start to write integration test cases. 
Do you want to have a task to do a global fix?  We need to get this change in to fix the hanging issue that impacts everyone that run Jenkins jobs to verify their changes.

tbarnes-us: Didn't know that.

PR #1840, opened 2020-07-30
Apache Webtier Support for non-priviledged port, Persistence Volume Claim and ImagePullSecrets

ashageetha:
Hi @mriccell, @rjeberhard, @doxiao,

This PR has fix for Apache Webtier to run with non-root user and non-privileged port.
Mainly this is fixed to be in sync with the latest changes done to the Apache Webtier Docker images as part of PR - https://github.com/oracle/docker-images/pull/1630.

Along with this below changes are also added so that it can be used in Cloud Setup where Kubernetes Cluster nodes are private and have restricted access:
1. Persistent Volume Claim support instead of Host path as volume
2. Image Pull Secrets that is needed to pull registry containing the apache webtier Docker image.
3. Readme changes to support helm v3

### DIFF
--- a/kubernetes/samples/charts/apache-samples/custom-sample/README.md
+++ b/kubernetes/samples/charts/apache-samples/custom-sample/README.md
@@ -8,7 +8,7 @@ $ kubectl create namespace apache-sample
 ```
 
 ## 2. Create WebLogic domains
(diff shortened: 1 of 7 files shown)


doxiao:
Note that the doc has been updated to use Helm V3 in https://github.com/oracle/weblogic-kubernetes-operator/pull/1847.

ashageetha:
Tracking with PR oracle/weblogic-kubernetes-operator#1916

PR #1846, opened 2020-08-04
Helm upgrade command removes private key from Kubernetes secret 'weblogic-operator-secrets'

lennyphan:
Customer OFSS discovered that the Operator's internal REST interface is not started after first executing a 'Helm upgrade' command and then recycling/restarting the Operator pod.

The Operator's helm chart manages the 'weblogic-operator-secrets' Kubernetes Secret and the 'Helm upgrade' command reverts the Secret to its original chart content, which is empty.  Subsequent restart of the Operator pod disables the start of the internal REST interface due to the missing private key. 

This change uses the 'lookup' function to check if the 'weblogic-operator-secrets' Kubernetes Secret exists and if so, stores the 'internalOperatorKey' to a template variable and sets the value when recreating the resource.  During install of the chart, the Secret resource creation behaves the same as before where the empty Secret is created and then the internal REST certificate and key are created and copied into the secret using the 'patch' command.

### DIFF
--- a/kubernetes/charts/weblogic-operator/templates/_operator-secret.tpl
+++ b/kubernetes/charts/weblogic-operator/templates/_operator-secret.tpl
@@ -9,6 +9,13 @@ data:
   {{- if (and .externalRestEnabled (hasKey . "externalOperatorKey")) }}
   externalOperatorKey: {{ .externalOperatorKey | quote }}
   {{- end }}
+  {{- $secret := (lookup "v1" "Secret" .Release.Namespace "weblogic-operator-secrets") }}
+  {{- if $secret }}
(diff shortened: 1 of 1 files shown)


lennyphan:
OFSS verified the change to helm template for REST secrets:

Hi @Lenny Phan this approach worked very well.....after doing helm upgrade the secret "weblogic-operator-secrets" didn't got emptied........it still had the same value as what was stored during helm install........also during restart of operator pod it uses the same secret value.
Now i am able to hit the operator svc even after helm upgrade/pod restart without any issue.Thanks!
If u feel this to be the solution to the raised problem then in which version of operator and when can we expect this fix to be available in an official release?

PR #1854, opened 2020-08-06
Retry failure fix

jshum2479:
This PR address the following issues

https://jira.oraclecorp.com/jira/browse/OWLS-83019
Disable introspect job retry when MII JRF path fails creating a new schema

https://jira.oraclecorp.com/jira/browse/OWLS-82989
Fix permanent failure logic for introspector job
(regression of 3.0/2.6 over older versions)

https://jira.oraclecorp.com/jira/browse/OWLS-82975
No errors reported in domain status when overrides file name is wrong, job retry indefinitely


docs/domains/Domain.json
```diff
@@ -421,6 +421,11 @@
             "$ref": "#/definitions/ServerStatus"
           }
         },
+        "retryCount": {
+          "description": "Number of retries.",
```
rjeberhard, line 425: This description isn't clear enough. Is this the number of attempts that have failed? What kind of failures does it include? When is it reset?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -630,10 +631,31 @@ public boolean wasInspectionRun() {
...
+
       if (cachedInfo == null || cachedInfo.getDomain() == null) {
         return true;
+      } else if (currentRetryCount > DomainPresence.getDomainPresenceFailureRetryMaxCount()) {
+        LOGGER.fine("Stop introspection retry - exceeded configured domainPresenceFailureRetryMaxCount: "
+            + DomainPresence.getDomainPresenceFailureRetryMaxCount());
+        return false;
       } else if (isCachedInfoNewer(liveInfo, cachedInfo)) {
         return false;  // we have already cached this
+      } else if (existingError != null && existingError.startsWith("MII Fatal Error")
```
rjeberhard, line 654: Is there a way we can return a fatal error more clearly? It seems brittle to rely on a specific string like this.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -630,10 +631,31 @@ public boolean wasInspectionRun() {
...
+
+      String existingError = Optional.ofNullable(liveInfo)
+          .map(DomainPresenceInfo::getDomain)
+          .map(Domain::getStatus)
+          .map(DomainStatus::getMessage)
+          .orElse(null);
+
       if (cachedInfo == null || cachedInfo.getDomain() == null) {
         return true;
+      } else if (currentRetryCount > DomainPresence.getDomainPresenceFailureRetryMaxCount()) {
```
rjeberhard, line 648: This PR is missing behavior that resets the retry count on introspectVersion or other changes.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
tbarnes-us: I don't see how to eliminate dependence on string parsing, but to make it less specific to MII and maybe a bit less brittle, perhaps use 'contains' instead of 'starts' and look for a string like 'FatalIntrospectorError'?  What do you think @rjeberhard ?

docs/domains/Domain.json
tbarnes-us: To save time on the back-and-forth, can you start a slack discussion for this?

More questions:

What is the interval between retries?  If greater than 0, is the interval configurable?

How does this interact with the operator 'domainPresenceFailureRetryMaxCount and domainPresenceFailureRetrySeconds'?  

What's the default?

How do you force the system to try again?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
rjeberhard: That would be okay, but my expectation is that the introspector is returning structured data, of which the error message is one value. The least brittle would be if there was a value for "isFatalError" or "isPermanentFailure". However, I'm okay with something like what you propose.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -630,10 +631,31 @@ public boolean wasInspectionRun() {
...
+        LOGGER.fine("Stop introspection retry - exceeded configured domainPresenceFailureRetryMaxCount: "
+            + DomainPresence.getDomainPresenceFailureRetryMaxCount());
```
tbarnes-us, line 650: Is it only introspection failures that cause a limited number of retries, or there other types of validation failures?  If so, then the wording is misleading.

Should the text say 'domain.spec.retryCount' instead of 'domainPresenceFailureRetryMaxCount'?

docs/domains/Domain.json
jshum2479: It only works with domainPresenceFailureRetryMaxCount (default is 5), I don't see domainPresenceFailureRetrySeconds used anymore.

jshum2479: We can force retry before the 2 minute interval by changing restart/introspect/image name in the spec (or after it exceeded the limit).   Other changes can happen in between the intervals (e.g. configmap, image with same name/tag)

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: There is a reset now

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: The max count actually is an operator tuning parameter not in domain spec

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
jshum2479: I changed it to contains "FatalIntrospectorError". Eventually If we can change the introspect job and operator java code has a more flexible way to communicate other than just a return code, then we can experiment the idea of always return 0 but have some text that gives more structured data.

operator/src/main/resources/Operator.properties
```diff
@@ -205,7 +205,8 @@ WLSDO-0012=Managed Server ''{0}'' specified in the domain resource does not exis
 WLSDO-0013=Istio is enabled and the domain resource specified to expose channel {0} in the adminServices section. The channel name default, \
 default-admin, default-secure are internal to the WebLogic Kubernetes Operator. Please create the network channel with a different name in the WebLogic Domain before \
 exposing it.
-WLSDO-0014="The introspect job had failed, retrying introspection job. Current failure count: {0}."
+WLSDO-0014="The domain {0} introspect job failed, it will automatically retry in {1} seconds."
+WLSDO-0015="Retrying failed domain {0} introspect job {1} of {2}."
```
tbarnes-us, line 209: (1) Suggested new text:  Retrying failed domain {0} introspect job, retry {1} of a maximum of {2}.

(2) Can you add a new 'Retrying' INFO for the 'infinite' retry case?

operator/src/main/resources/Operator.properties
```diff
@@ -205,7 +205,8 @@ WLSDO-0012=Managed Server ''{0}'' specified in the domain resource does not exis
 WLSDO-0013=Istio is enabled and the domain resource specified to expose channel {0} in the adminServices section. The channel name default, \
 default-admin, default-secure are internal to the WebLogic Kubernetes Operator. Please create the network channel with a different name in the WebLogic Domain before \
 exposing it.
-WLSDO-0014="The introspect job had failed, retrying introspection job. Current failure count: {0}."
+WLSDO-0014="The domain {0} introspect job failed, it will automatically retry in {1} seconds."
```
tbarnes-us, line 208: Can you add ", current number of retries = {2}" or ", total number of attempts = {2}" (where the latter == retries + 1)?

operator/src/main/resources/Operator.properties
jshum2479: 1 is done. 
2 has a severe message 

WLSKO-0157=Domain {0} is not valid: {1}

operator/src/main/resources/Operator.properties
jshum2479: done

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -276,6 +284,45 @@ public DomainStatus withReplicas(Integer replicas) {
     return this;
   }
 
+  /**
+   * The number of retries for an introspect job that fails.
+   *
+   * @return introspectJobFailureRetryCount
+   */
+  public Integer getIntrospectJobFailureRetryCount() {
```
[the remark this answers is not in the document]
jshum2479: done

jshum2479:
Is there anything that needs to be added to this PR, can we close this out first?

operator/src/main/resources/Operator.properties
```diff
@@ -205,6 +205,9 @@ WLSDO-0012=Managed Server ''{0}'' specified in the domain resource does not exis
 WLSDO-0013=Istio is enabled and the domain resource specified to expose channel {0} in the adminServices section. The channel name default, \
 default-admin, default-secure are internal to the WebLogic Kubernetes Operator. Please create the network channel with a different name in the WebLogic Domain before \
 exposing it.
+WLSDO-0014=The domain {0} introspect job failed, it will automatically retry in {1} seconds. Current number of  \
+  retries = {2}
+WLSDO-0015="Retrying failed domain {0} introspect job, retry {1} of a maximum {2}."
```
tbarnes-us, line 210: The maximum doesn't always apply - so maybe the easiest thing to do is remove it?  Sorry if I suggested including it.  Or maybe change maximum --> "potential maximum {2} (the retry limit does not apply to all types of failures).

jshum2479: I think it applies.  The domain validation fail case is not in the same code path, it spill out message in the log

WLSKO-0157=Domain {0} is not valid: {1}

tbarnes-us: Ok.

tbarnes-us (approved):

rjeberhard (approved):

PR #1855, opened 2020-08-06
Changes for OWLS-83431

ankedia:
Changes for OWLS-83431 - Create Pods one-at-a-time waiting on scheduling

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
@@ -34,6 +34,7 @@ import oracle.kubernetes.operator.work.Step;
 import oracle.kubernetes.weblogic.domain.model.ServerSpec;
 import oracle.kubernetes.weblogic.domain.model.Shutdown;
 
+import static oracle.kubernetes.operator.LabelConstants.CLUSTERNAME_LABEL;
 import static oracle.kubernetes.operator.ProcessingConstants.SERVERS_TO_ROLL;
 
 public class PodHelper {
@@ -90,13 +91,16 @@ public class PodHelper {
    * @param info Domain presence info
    * @return list containing scheduled pods
    */
-  public static List<String> getScheduledPods(DomainPresenceInfo info) {
+  public static List<String> getScheduledPods(DomainPresenceInfo info, String clusterName) {
     // These are presently scheduled servers
     List<String> scheduledServers = new ArrayList<>();
     for (Map.Entry<String, ServerKubernetesObjects> entry : info.getServers().entrySet()) {
       V1Pod pod = entry.getValue().getPod().get();
       if (pod != null && !PodHelper.isDeleting(pod) && PodHelper.getScheduledStatus(pod)) {
-        scheduledServers.add(entry.getKey());
+        String wlsClusterName = pod.getMetadata().getLabels().get(CLUSTERNAME_LABEL);
+        if ((wlsClusterName == null) || (wlsClusterName.contains(clusterName))) {
+          scheduledServers.add(entry.getKey());
+        }
       }
     }
     return scheduledServers;
@@ -107,75 +111,19 @@ public class PodHelper {
    * @param info Domain presence info
    * @return list containing ready pods
    */
-  public static List<String> getReadyPods(DomainPresenceInfo info) {
+  public static List<String> getReadyPods(DomainPresenceInfo info, String clusterName) {
     // These are presently Ready servers
     List<String> readyServers = new ArrayList<>();
     for (Map.Entry<String, ServerKubernetesObjects> entry : info.getServers().entrySet()) {
       V1Pod pod = entry.getValue().getPod().get();
       if (pod != null && !PodHelper.isDeleting(pod) && PodHelper.getReadyStatus(pod)) {
-        readyServers.add(entry.getKey());
+        String wlsClusterName = pod.getMetadata().getLabels().get(CLUSTERNAME_LABEL);
+        if ((wlsClusterName == null) || (wlsClusterName.contains(clusterName))) {
+          readyServers.add(entry.getKey());
+        }
       }
     }
     return readyServers;
-  }
-
-  /**
-   * Change pod status to scheduled.
-   * @param info Domain presence info
-   * @param serverName Name of server pod to be changed
-   */
-  public static void schedulePods(DomainPresenceInfo info, String serverName) {
-    for (Map.Entry<String, ServerKubernetesObjects> entry : info.getServers().entrySet()) {
-      V1Pod pod = entry.getValue().getPod().get();
-      if (pod.getMetadata().getName().contains(serverName)) {
-        PodHelper.setScheduledStatus(pod);
-      }
-    }
-  }
-
-  /**
-   * Change pod status to ready.
-   * @param info Domain presence info
-   * @param serverName Name of server pod to be changed
-   */
-  public static void makePodsReady(DomainPresenceInfo info, String serverName) {
-    for (Map.Entry<String, ServerKubernetesObjects> entry : info.getServers().entrySet()) {
-      V1Pod pod = entry.getValue().getPod().get();
-      if (pod.getMetadata().getName().contains(serverName)) {
-        PodHelper.setReadyStatus(pod);
-      }
-    }
-  }
-
-  /**
-   * get if pod is in scheduled state.
-   * @param pod pod
-   * @return true, if pod is scheduled
-   */
-  public static boolean setScheduledStatus(V1Pod pod) {
-    V1PodSpec status = pod.getSpec();
-    if (status != null) {
-      status.setNodeName("Node1");
-    } else {
-      pod.spec(new V1PodSpec().nodeName("Node1"));
-    }
-    return true;
-  }
-
-  /**
-   * get if pod is in scheduled state.
-   * @param pod pod
-   * @return true, if pod is scheduled
-   */
-  public static boolean setReadyStatus(V1Pod pod) {
-    V1PodStatus status = pod.getStatus();
-    if (status != null) {
-      status.phase("RUNNING").addConditionsItem(new V1PodCondition().type("Ready").status("True"));
-    } else {
-      pod.status(new V1PodStatus().phase("Running")
-              .addConditionsItem(new V1PodCondition().type("Ready").status("True")));
-    }
-    return true;
   }
 
   /**
@@ -186,9 +134,7 @@ public class PodHelper {
   public static boolean getScheduledStatus(V1Pod pod) {
     V1PodSpec status = pod.getSpec();
     if (status != null) {
-      if (status.getNodeName() != null) {
-        return true;
-      }
+      return status.getNodeName() != null;
     }
     return false;
   }
@@ -552,7 +498,7 @@ public class PodHelper {
     protected V1ObjectMeta createMetadata() {
       V1ObjectMeta metadata = super.createMetadata();
       if (getClusterName() != null) {
-        metadata.putLabelsItem(LabelConstants.CLUSTERNAME_LABEL, getClusterName());
+        metadata.putLabelsItem(CLUSTERNAME_LABEL, getClusterName());
       }
       return metadata;
     }
@@ -610,7 +556,7 @@ public class PodHelper {
       if (oldPod != null) {
         Map<String, String> labels = oldPod.getMetadata().getLabels();
         if (labels != null) {
-          clusterName = labels.get(LabelConstants.CLUSTERNAME_LABEL);
+          clusterName = labels.get(CLUSTERNAME_LABEL);
         }
 
         ServerSpec serverSpec = info.getDomain().getServer(serverName, clusterName);
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
@@ -86,6 +86,114 @@ public class PodHelper {
   }
 
   /**
+   * Get list of scheduled pods.
+   * @param info Domain presence info
+   * @return list containing scheduled pods
+   */
+  public static List<String> getScheduledPods(DomainPresenceInfo info) {
+    // These are presently scheduled servers
+    List<String> scheduledServers = new ArrayList<>();
+    for (Map.Entry<String, ServerKubernetesObjects> entry : info.getServers().entrySet()) {
+      V1Pod pod = entry.getValue().getPod().get();
+      if (pod != null && !PodHelper.isDeleting(pod) && PodHelper.getScheduledStatus(pod)) {
+        scheduledServers.add(entry.getKey());
+      }
+    }
+    return scheduledServers;
+  }
+
+  /**
+   * Get list of ready pods.
+   * @param info Domain presence info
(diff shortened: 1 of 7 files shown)


russgold:
Can you explain the logic in ManagedServerUpIteratorStep.apply() ?

ankedia:
> Can you explain the logic in ManagedServerUpIteratorStep.apply() ?

ankedia:
> Can you explain the logic in ManagedServerUpIteratorStep.apply() ?

The apply method runs StartManagedServersStep (which starts managed servers) for 2 different types of servers -
1. Non-clustered (standalone) servers
2. Clustered servers from one or more clusters

It creates a child fiber for each cluster and a fiber for standalone (non-clustered) servers and it executes StartManagedServersStep. Since each cluster can have a different value for "maxConcurrentStartups", I'm creating different fiber for each cluster (and for non-clustered servers). The idea for maxConcurrent startup is that only n number of pods can be created simultaneously until all previous pods are ready.

In StartManagedServersStep, I check to see how many pods are scheduled and how many pods are ready. This is used to decide if a new managed server pod can be created. Each managed server pod is created in a separate child fiber. Please let me know if this is not clear or if you have any questions. Thanks.

russgold:
It's clear but incomplete. I'm trying to understand the logic. Can you describe it in pseudocode? What is the logic in the method?

ankedia:
> It's clear but incomplete. I'm trying to understand the logic. Can you describe it in pseudocode? What is the logic in the method?



> It's clear but incomplete. I'm trying to understand the logic. Can you describe it in pseudocode? What is the logic in the method?
ok, I will work on that. Do you or @alai8 know if there's any psuedocode or document for original ManagedServerUpIteratorStep or other related steps? It'll help me with the terminology and structure. Thanks.

ankedia:
> It's clear but incomplete. I'm trying to understand the logic. Can you describe it in pseudocode? What is the logic in the method?
ok, I will work on that. Do you or @alai8 know if there's any psuedocode or document for original ManagedServerUpIteratorStep or other related steps? It'll help me with the terminology and structure. Thanks.

russgold:
There is no formalism for pseudocode. I'm just asking you to describe what is going on in english, not at the code level.

Let me get specific. The body of the method is very roughly
1. The first condition is whether the following step for the head of the queue is a ServerDownStep. Why? What does that mean? It seems a very brittle thing to test
2. If the queue is empty you advance to the after-step
3. The next condition (after verifying that the queue is not empty) uses some combination of the number of scheduled pods and the ready pods to decide to fork join. Why? What is the intent?
4. The third condition decides to delay if the queue is non-empty. Why delay? How do you know that there is nothing left to process? Is there any way for it to be empty at this point, given that you already checked that (see 2)
5. Once again, if the queue is empty you advance to the after-step, which seems redundant

What have I missed?

ankedia:
Step StartManagedServerStep (Argument clusterName, Argument Collection<StepAndPacket> startDetails, Argument nextStep){ 

    Initialize current scheduled count for cluster "clusterName" to 0 
	Initialize FIFO Queue - it'll hold StepsAndPackets for starting managed servers for this cluster
    
    Constructor {
	  Populate FIFO queue containing StepsAndPackets from startDetails to satrt servers in cluster "clusterName".
	  Get maxConcurrency for this cluster
    }    
	
	apply method {
	  Check first StepAndPacket in queue
	  if (step in StepsAndPackets is for bringing server down i.e. for scale down scenario) {
	    start a fiber to run ServerDown step and then invoke this step again
	  } else if (all steps in startDetails completed) {
	    proceed to nextStep
      } 	  
	  
	  // This makes assumption that admin server is already scheduled 
	  if (current scheduled count is less than total scheduled servers for cluster "clusterName") { //initial value of total scheduled servers is 1
	    if (current scheduled count is less than (maxConcurrency + ready managed servers for cluster "clusterName")) {
		   Dequeue StepAndPacket from FIFO queue
		   Increment current scheduled count
		   Start a child fiber to run step for managed server pod and service creation and then invoke this step again
		}
	  }
	  if (More items in queue) {
	     Wait 100 mill-seconds to give a chance for previous server to get scheduled and then invoke this step again
      } else { // All managed server start steps executed
	     proceed to nextstep
	  }
	} 
}

ankedia:
1. The `bringManagedServerUp()` has below existing code. My understanding of this code is that it checks ServerStartupInfo to see if only service needs be created for managed server (i.e. when pre-creating server service) then it creates a ServerDownStep in startDetails. This is also executed during scale down scenario to bring managed server down based on ServerStartupInfo. Previously it was executing all steps in startDetails array list in parallel. I added check make sure it executes ServerDownStep before we check for scheduled and ready servers. 

```
  private static Step bringManagedServerUp(ServerStartupInfo ssi) {
    return ssi.isServiceOnly()
        ? ServiceHelper.createForServerStep(
            true, new ServerDownStep(ssi.getServerName(), true, null))
        : ServiceHelper.createForServerStep(PodHelper.createManagedPodStep(null));
  }
```
The queue empty check in "else" condition is to handle cases where startDetails only contains ServerDownSteps and no steps to create managed server pods/service. In this case, we don't need to rely on scheduled count etc. I can look to change/refactor this code if it's brittle to test.

3. Please let me know if intent is not clear from above pseudocode.

4. This is executed after checking the condition of previous pods to see if they're scheduled and/or ready. If previous pod is not yet scheduled, it'll wait 100ms and invoke this step again

5. Again, this check is after checking the condition of previous pods. I had added it before I added the check for ServerDownStep and previous check to see if queue is empty. I forgot to revisit this after adding check for ServerDownStep, I'll check and remove it.

russgold:
The first simplification I see is that you should check if the queue is empty first. If it is not, there is no need to check if peek returns a null, as it won't.

I would then extract the check on that first item into a method with a name such as isShuttingDown. It will make it clear as to what is going on much better than reading the computation logic. And since that logic does a lot of null checks, Optional.ofNullable would probably make it easier to read the implementation.

The next pair of nested ifs seems to be the heart of the functionality. My first observation is that it should really be a single if with a complex condition extracted into a method for clarity. I suppose it means something like "there are more servers to start in the cluster"? It is important to make it as easy to understand as possible, and right now it isn't obvious to me why this logic is the one to use. What does the scheduled count mean? Isn't it actually the number of pods in the cluster that have been started, not scheduled? Also, I suspect this could be another else if.

ankedia:
Thanks Russ for the comments and suggestions. I have made changes to incorporate these suggestions. Please let me know if I missed something.

russgold:
It's definitely improved. I would further suggest:
1. Use isEmpty() rather than size() = 0
2. Combine the second group to the first using if else
3. You don't need the local variable servers.

So, for example:

`    @Override
      if (startDetailsQueue.isEmpty()) {
        return doNext(new ManagedServerUpAfterStep(getNext()), packet);
      } else if (isServiceOnlyOrShuttingDown()) {
        return doForkJoin(this, packet, Collections.singletonList(startDetailsQueue.poll()));
      } else if (serverAvailableToStart(packet.getSpi(DomainPresenceInfo.class))) {
        this.countStarted++;
        return doForkJoin(this, packet, Collections.singletonList(startDetailsQueue.poll()));
      } else {
        return doDelay(this, packet, 100, TimeUnit.MILLISECONDS);
      }
    }
`

Why is `maxConcurrency` a variable? What values can it take?

russgold:
A suggest about naming variables: Use a consistent pattern such as "numXXX." Right now you have `countStarted` and `readyManagedPods`

ankedia:
I have made changes as per above suggestions. `maxConcurrency ` is the value of `MaxConcurrentStartup` cluster configuration in domain. It's an instance variable and I have changed canStartConcurrently() to refer it by this. Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -85,6 +86,59 @@ public static boolean isReady(V1Pod pod) {
     return ready;
   }
 
+  /**
+   * Get list of scheduled pods.
+   * @param info Domain presence info
+   * @return list containing scheduled pods
```
rjeberhard, line 92: Missing Javadoc @param for clusterName

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -85,6 +86,59 @@ public static boolean isReady(V1Pod pod) {
...
+   */
+  public static List<String> getScheduledPods(DomainPresenceInfo info, String clusterName) {
```
rjeberhard, line 94: I'll verify below, but it's surprising that you have a method to list scheduled pods. I would think that you would be interested in unscheduled pods.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -85,6 +86,59 @@ public static boolean isReady(V1Pod pod) {
...
+    return readyServers;
+  }
+
+  /**
+   * get if pod is in scheduled state.
+   * @param pod pod
+   * @return true, if pod is scheduled
+   */
+  public static boolean getScheduledStatus(V1Pod pod) {
+    V1PodSpec status = pod.getSpec();
```
rjeberhard, line 135: Fix variable name

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -142,19 +154,53 @@ private boolean isServerInCluster(ServerStartupInfo ssi) {
...
+      }
+    }
+
+    private boolean isServiceOnlyOrShuttingDown() {
+      return Optional.ofNullable(startDetailsQueue.peek().step)
+              .map(step -> step.getNext() instanceof ServerDownStep).orElse(false);
+    }
+
+    private boolean serverAvailableToStart(DomainPresenceInfo info) {
+      return ((this.numStarted < PodHelper.getScheduledPods(info, clusterName).size())
```
rjeberhard, line 197: What do you think about switching this to test PodHelper.unscheduledPodsExist() rather than comparing the count of scheduled pods to an expected number?

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -142,19 +154,53 @@ private boolean isServerInCluster(ServerStartupInfo ssi) {
...
     public NextAction apply(Packet packet) {
-      return doForkJoin(new ManagedServerUpAfterStep(getNext()), packet, startDetails);
+
+      if (startDetailsQueue.isEmpty()) {
+        return doNext(new ManagedServerUpAfterStep(getNext()), packet);
+      } else if (isServiceOnlyOrShuttingDown()) {
+        Collection<StepAndPacket> servers = Collections.singletonList(startDetailsQueue.poll());
+        return doForkJoin(this, packet, servers);
+      } else if (serverAvailableToStart(packet.getSpi(DomainPresenceInfo.class))) {
+        this.numStarted++;
```
rjeberhard, line 184: What is the concurrency of this code? Will multiple threads ever be running here?

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -142,19 +154,53 @@ private boolean isServerInCluster(ServerStartupInfo ssi) {
...
+              .map(step -> step.getNext() instanceof ServerDownStep).orElse(false);
+    }
+
+    private boolean serverAvailableToStart(DomainPresenceInfo info) {
+      return ((this.numStarted < PodHelper.getScheduledPods(info, clusterName).size())
+              && (canStartConcurrently(PodHelper.getReadyPods(info, clusterName).size())));
+    }
+
+    private boolean canStartConcurrently(int numReady) {
+      return ((this.maxConcurrency > 0) && (this.numStarted < (this.maxConcurrency + numReady - 1)))
```
rjeberhard, line 202: Similarly, it seems like you ought to be checking for the number of unready pods.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
ankedia: It's one thread per cluster and one thread for non-clustered servers. It'll be single threaded for one cluster and 2 threads for 2 cluster case. If domain also have non-clustered (standalone) servers, it'll create a separate fiber for those servers.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
ankedia: I believe the reason I went with current `count` based approach is because we can only schedule/start `n` number pods in parallel based on value of maxConcurrency. Previously the code was creating `maxConcurrency` number of fibers and fibers used to wait until pod becomes ready. With current approach, I don't make the fibers wait until pod is ready but start servers (create fibers) only if all previous pods are scheduled and it matches the criteria for `maxConcurrency`. I think it's possible to test  PodHelper.unscheduledPodsExist() but we need to check for number of ready pods before deciding to start a new server (start a new fiber). I can try the approach you suggested and test it out for different use-case scenarios.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
ankedia: Please see above comment about creating fibers based on value of `maxConcurrency`. I think we can't wait for all previous pods to be ready and need to check for number of ready pods in order to decide whether to create a new fiber/thread or not.

ankedia:
Thanks @rjeberhard for your comments and suggestions. I have addressed some of your comments and have replied to other comments. I have thoroughly tested current approach by running functional, integration and unit tests. If we decide to change the approach, these tests will need to be rerun. Please let me know your thoughts.

rjeberhard:
Okay, I'm happy enough with the current approach.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
rjeberhard: Let me be clearer in my question. Do I need to worry about the concurrency of updating the numStarted variable? I can't really tell from your answer.

ankedia: Sorry for not being clear in my previous answer. In 2 cluster scenario, 2 threads could reading/updating in parallel but overall concurrency will be low. I have `numStarted` to AtomicInteger as we discussed and made `numConcurrency ` final by setting it in constructor. Please let me know if I missed anything. Thanks.

PR #1856, opened 2020-08-07
Add domainUID to operator log messages in async call request/response code path

doxiao:
1) Add domainUID to CallBuilder's RequestParams when applicable. 
2) Pass in domainUID when build a call request when applicable.
3) Refine some of the existing logging context code to not accidentally associate domainUID with a fiber/pool related log message.


operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -630,7 +633,7 @@ private Call readDomainAsync(
    */
   public Step readDomainAsync(String name, String namespace, ResponseStep<Domain> responseStep) {
     return createRequestAsync(
-        responseStep, new RequestParams("readDomain", namespace, name, null), readDomain);
+        responseStep, new RequestParams("readDomain", namespace, name, null, name), readDomain);
```
alai8, line 636: is domain name always the same as uid?

doxiao: It is the domain resource's metadata.getName(), which is the domainUID.

rjeberhard: Just clarifying this: .metadata.name is the default value for domainUID, but there is also a .spec.domainUid field that, if set, overrides .metadata.name.

The domain name from config.xml can be a different value.

PR #1867, opened 2020-08-13
duplicate -srcstorepass parameter used in script

jshum2479:
initialize-internal-operator-identity.sh  keytool command specify the parameter -srcstorepass twice.


rjeberhard:
@jshum2479, you need to create this PR from a branch created off of master.

PR #1871, opened 2020-08-14
Kubernetes Exec API intermittently hangs when executing curl command for ItMiiDomain test

lennyphan:
When using the Kubernetes java client exec api, to invoke curl commands, intermittent hangs occur with no response.  This was seen in the JUnit 5 ItMiiDomain test.  It seems that by reading the both stdout and stderr streams while the command is executing resolves the hang on the Kind cluster.

Here is the full test suite result for kind cluster run: 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1292/

There is one failure which is different than the develop run Pani referred me to compare known failures: https://build.weblogick8s.org:8443/job/wko-nightly-parallel/3/


lennyphan:
Here is the full test suite result for kind cluster run:  https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1292/

There is one failure which is different than the develop run Pani referred me to compare known failures:  https://build.weblogick8s.org:8443/job/wko-nightly-parallel/3/

The single failure in my run 1292 (above) doesn't look related to my change.

PR #1876, opened 2020-08-17
Owls 83538

russgold:
Rework the unit tests to reduce dependency on threads and I/O. In particular, this PR:

- changes the test point for watchers from the watch wrapper in some tests and PseudoServer in others, to the creation of the Watch object itself, taking advantage of the addition of a Watchable interface in the client library,
- disables starting threads during invocations of functions in Main,
- isolates creating the steps to check domains from the runnable that runs them in Main,
- some miscellaneous cleanup in tests that had gotten a bit confused.

This is intended to improve reliability and repeatability of the unit tests, removing environmental dependencies (there are still some left to deal with in future PRs).

This was motivated by a bug filed by @xiancao and I don't want it merged until it's been shown definitely to fix the unit test failures she'd been seeing.


russgold:
I didn't change any docs/charts. How did they get into this PR?

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -158,10 +158,14 @@ boolean isOutdatedBetaCrd(
...
-        return doNext(context.verifyCrd(getNext()), packet);
-      } else {
-        return doNext(context.verifyBetaCrd(getNext()), packet);
+      try {
+        if (context.version.isCrdV1Supported()) {
+          return doNext(context.verifyCrd(getNext()), packet);
+        } else {
+          return doNext(context.verifyBetaCrd(getNext()), packet);
+        }
+      } catch (NullPointerException e) {
```
doxiao, line 167: Can we avoid catching and throwing NullPointerException here?

rjeberhard:
I don't know how the docs/charts got it, but I've backed out the change to that file.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
russgold: Oops - thanks for catching that.  It was something I added to debug one of the test failures I found in the develop branch.

doxiao (approved):
LGTM.

russgold:
It appears that @xiancao is no longer seeing these failures, but is now seeing different ones. So this is ready to merge and I will look at what else is going on with her machine.

PR #1879, opened 2020-08-19
owls-83918 an idle domain's resource version should stay unchanged

doxiao:
The fix contains two parts:
1) Instead of adding Progressing steps unconditionally, the operator only add Progressing steps when there are changes to make to the domain.
2) In the domain make-right code path, populates the server state and health info in the packet based on what is in the domain object, so that the domain's server state/health is not cleared accidentally.

Whole IT test suite run (had one failure): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1551.
Rerun of the failed test class alone (passed): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1555.


operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -840,6 +840,7 @@ public NextAction onSuccess(Packet packet, CallResponse<DomainList> callResponse
         for (Domain dom : callResponse.getResult().getItems()) {
           String domainUid = dom.getDomainUid();
           domainUids.add(domainUid);
+          boolean absent = dpis.get(domainUid) == null;
```
russgold, line 843: Why compute absent up here, if it is used below? Unless there is a very good reason, it should be computed when you need it - you can even remove the local variable that way.

doxiao: The line after the current assignment will change the content of the map, therefore we need to catch the old value before it.

doxiao:
If we do need to re-bring the domain or any of it servers up, we would want to have the corresponding Progressing condition recorded. Let me see if we can add the Progressing condition step only when the pod needs to be restarted/replaced.

doxiao:
Let me see if we can add the Progressing condition step only when the pod needs to be restarted/replaced.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -330,9 +332,11 @@ private Step patchCurrentPod(V1Pod currentPod, Step next) {
     KubernetesUtils.addPatches(
         patchBuilder, "/metadata/annotations/", getAnnotations(currentPod), getPodAnnotations());
 
-    return new CallBuilder()
-        .patchPodAsync(getPodName(), getNamespace(), getDomainUid(),
-            new V1Patch(patchBuilder.build().toString()), patchResponse(next));
+    return Step.chain(
```
alai8, line 335: would the domain status be updated when we stop a server, or scale down a cluster?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -669,6 +671,23 @@ private void internalMakeRightDomainPresence() {
             willInterrupt);
     }
 
+    private void populateServerStateHealthFromDomain(Packet packet) {
+      Map<String, ServerHealth> serverHealth = new HashMap<String, ServerHealth>();
+      Map<String, String> serverState = new HashMap<String, String>();
```
ankedia, line 676: Should these be ConcurrentHashMap instead of regular HashMap?

doxiao: yes, good catch! Changed to ConcurrentHashMap.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
doxiao: Once the domain is started, the operator periodically checks the server's state and health condition, and update the domain status/condition accordingly. Right now, we don't have a progressing REASON for cluster scaling down; the condition we are adding here is for managed server starting only. 
I am looking into the scaling down code path to see if we need to add additional Progressing step or not.

russgold:
How are you handling intermittent server status and where are the unit tests for it?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: The scheduleDomainStatusUpdating creates a new packet for periodically domain status update, which is not the same packet that operator creates here in internalMakeRightDomainPresence.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
doxiao: I added ProgressingStep to scaling down code path to match what we had without the changes in this PR. I thought about adding a new reason (for example "managedServersScalingDown") for this, but I decided to use what we used before this change. Ryan, your thoughts on this?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -611,6 +611,16 @@ void modifyStatus(DomainStatus status) {
     }
   }
 
+  /**
+   * A utility method for unit testing.
+   */
+  public static Step excludeProgressingStep(Step step) {
```
alai8, line 617: I understand that the ProgressingStep is a private static class, but I think it is better to have this method to unit test code.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -669,6 +671,27 @@ private void internalMakeRightDomainPresence() {
             willInterrupt);
     }
 
+    private void populateServerStateHealthFromDomain(Packet packet) {
```
russgold, line 674: Aren't we already updating these maps in ServerStatusReader? Why copy the logic to here?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
doxiao: I initially had this method in the test code, and made the ProcessingStep public. I changed to the current implementation because I slightly preferred keeping the class private. But I am okay with either approach.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: The domain status contains DomainConditions and ServerStatus. The ServerStatus contains the WebLogic servers' state and health information, which is updated based on what is in the two maps (serverState and serverHealth) in the packet. The domain conditions are only handled by the domain status update steps in DomainStatusUpdater. 

Prior to the code change in this PR, the server status update is already triggered from two code paths: make-right code path and ServerStatusReader/update  code path (see the places where the DomainStatusUpdater.createStatusUpdateStep is called). If I understand the logic correctly, the update in the make-right is to record possible cluster scale-up and scale-down changes. The update in the ServerStatusReader is to record state/health changes on the WebLogic server processes. The two code paths do NOT share the same packet instances and the contents. The ServerStatusReader populates the maps based on the responses that the reader gets, and add them to the packet. The make-right code does not populate the maps,  and therefore no maps are added to the packet. As a result, when the operator does a full recheck of an idle domain (running but no activities), the make-right code path will update the domain status to remove the server state and health, and the next update from the ServerStatusReader will update the states and health again, and put the actual state and health back. This behavior will repeat due to the periodical ServerStatusReader server status update logic and make-right full recheck logic, thus the domain version number keeps to be bumped up.

The solution in this PR is to populate the packet in the make-right code path with the current known state/health that the operator gets from the domain object, so that the logic starts with the correct initial server status. 

I agree that it is hard to unit test the combined behavior and end results from the code paths. I have been trying to find a way to test only the make-right code path instead, for example checking if the maps in the packet are populated correctly.

rjeberhard: Could the make-right code path simply leave the per-server status alone by copying the content from the current status, if any?

russgold: A better solution, then, would probably be to encapsulate the logic, possibly in the ServerStatus itself. Both paths could then use that logic, and it could be unit-testable.

doxiao: @rjeberhard That is the change in the PR does; get the current status from the domain object. We need to populate the maps in the packet with those status contents because the remaining logic around this uses that information to determine the behavior.

doxiao: @russgold are you suggesting that both code paths read the status from the servers, and populate the maps with the responses?  Readers reaches out to each servers to get its state and health information, and I am not sure if we would want to do that in the make-right code path.

russgold: I'm suggesting that both paths do what they need to do, but that the logic be encapsulated in ServerStatus, where it can be unit tested.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -659,12 +703,11 @@ private void internalMakeRightDomainPresence() {
               Component.createFor(liveInfo, delegate.getVersion(),
                   PodAwaiterStepFactory.class, delegate.getPodAwaiterStepFactory(getNamespace()),
                   V1SubjectRulesReviewStatus.class, delegate.getSubjectRulesReviewStatus(getNamespace())));
-
       runDomainPlan(
             getDomain(),
             getDomainUid(),
             getNamespace(),
-            new StepAndPacket(createSteps(), packet),
+            new StepAndPacket(createPopulatePacketServerMapsStep(createSteps()), packet),
```
russgold, line 710: Wouldn't it make sense to add the new step into `createSteps`, in the branch where we aren't shutting down the domain?

doxiao:
Jenkins test result with the latest code: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1756/console.

doxiao:
Jenkins test result with the latest code: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1756/console.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: I thought about that, but noticed that createSteps is called in other places outside of the make-right code path.

doxiao: Only in the case where createSteps is called from the make-right code path (internalMakeRightDomainPresence method), the packet does not already have the server maps;

russgold: Then pull that whole line, 
```
new StepAndPacket(createPopulatePacketServerMapsStep(createSteps()), packet)
```

into its own method, such as `createDomainPlanSteps`. There is otherwise too much going on in that code block.

rjeberhard (approved):

PR #1881, opened 2020-08-19
 List continuation and watch bookmark support

rjeberhard:
This PR resolves two problems:

1. Anytime the operator lists resources (Pods, Services, ConfigMaps, Namespaces, etc.), Kubernetes can limit the set of resources returned. Kubernetes then includes in the result a "continuation" token that may be used to repeat the search and retrieve the next batch of results. The operator had only a partial implementation of this continuation support that was not yet functional.

With this proposed change, the operator will handle continuations for all list operations. Since we didn't previously support this pattern, our tuning parameters defaulted to only limiting resources at 500 results so that it was very unlikely that a customer would hit this issue. Therefore, I've now tuned the default limit to 50.

2. With Kubernetes 1.15, the watch pattern supports a BOOKMARK event type that is used to tell watch listeners to no longer use resource versions older than the value given in the bookmark. While researching the listing support, I discovered that we could easily support bookmarks since the functionality was complete. I've added unit-tests and enabled the necessary flag.


docs/charts/index.yaml
```diff
@@ -3,108 +3,108 @@ entries:
   weblogic-operator:
```
russgold, line 3: Should this really be part of the PR? I don't see a change in the CRD.

operator/src/test/java/oracle/kubernetes/operator/DomainWatcherTest.java
```diff
@@ -43,6 +45,13 @@ public void initialRequest_specifiesStartingResourceVersion() {
         hasEntry("resourceVersion", INITIAL_RESOURCE_VERSION.toString()));
   }
 
+  @Test
```
russgold, line 48: I don't understand the name of this test. It seems to be testing that if a watcher receives  a bookmark callback, it updates its resource version for future queries. Is that correct? If so, I'd suggest that the test be named something like, `whenWatcherReceivesBookmarkEvent_updateResourceVersion`. The current name seems to be talking about what Kubernetes does, not what the operator does.

operator/src/test/java/oracle/kubernetes/operator/WatcherTestBase.java
```diff
@@ -101,10 +105,20 @@ void sendInitialRequest(BigInteger initialResourceVersion) {
     createAndRunWatcher(NAMESPACE, stopping, initialResourceVersion);
   }
 
+  DomainWatcher sendBookmarkRequest(BigInteger initialResourceVersion, String bookmarkResourceVersion) {
```
russgold, line 108: Generally, we've kept the type of watcher generic in the test base. This could simply return Watcher<?> and the unit test could either cast the result or just accept the generic.

russgold:
So this PR has what appears to be two completely separate changes, and it's a big change, so that made it a bit harder for me to follow. The changes to the watcher functionality seem fine. I'm still trying to figure out the list changes.

operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
```diff
@@ -92,17 +94,34 @@ private NextAction getPotentialRetryAction(Packet packet) {
...
    * @return Next action for list continue
    */
-  protected final NextAction doContinueList(Packet packet) {
-    RetryStrategy retryStrategy = packet.getSpi(RetryStrategy.class);
-    if (retryStrategy != null) {
-      retryStrategy.reset();
+  protected final NextAction doContinueListOrNext(CallResponse<T> callResponse, Packet packet) {
+    return doContinueListOrNext(callResponse, packet, getNext());
+  }
+
```
russgold, line 106: Here you have the retry logic and list continue logic mixed together. It seems to me that they really need to be separate, as a retry and and a continue are orthogonal. You seem to be conditioning the processing of a retry on the presence of a continue token. At the very least, the code is unclear.

russgold:
I see no unit tests that demonstrate the handling of chunked lists. At the very least, I'd want to see unit tests that actually try to retrieve chunked lists (not just tests on how the AsyncRequestStep behaves), and can handle cases when the resource version has incremented partway through, because a new item was added asynchronously.

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -369,14 +370,33 @@ private static Step readExistingPods(String ns) {
...
-                                             boolean isFullRecheck) {
+                                     Collection<String> domainNamespaces,
+                                     boolean isFullRecheck) {
     CallBuilder builder = new CallBuilder();
     String selector = selectionStrategy.getLabelSelector();
     if (selector != null) {
       builder.withLabelSelectors(selector);
     }
-    return builder.listNamespaceAsync(new NamespaceListStep(selectionStrategy, domainNamespaces, isFullRecheck));
+    return builder.listNamespaceAsync(
```
russgold, line 380: This code is becoming increasingly complex. The coupling between the listNamespace calls and the CRD processing should not be done at this level. One of the great things about the Step/Fiber architecture is that you have those separate steps, and let the code that sets them up define the chain. Putting the dependency directly in the response to the namespace list makes future maintenance ever more difficult.

rjeberhard: That's fair, but the complexity here really isn't new -- I just moved this code here. I'll see if I can separate them.

rjeberhard: Okay, I pulled the CRD steps out separately and earlier in the overall flow.

operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
rjeberhard: I've changed this to try and clarify... I believe that the disconnect is that I see that you implemented a solution in your PR that would iterate until there was no continue token and then present the final list to the response step, but I wanted to present each partial list to the response step. Therefore, the response step needs a mechanism to request the next request. When a new request happens, I do want to reset any failure count that may have been accumulated from the previous request.

operator/src/test/java/oracle/kubernetes/operator/DomainWatcherTest.java
rjeberhard: Changed

rjeberhard:
@russgold, I looked at your PR. As I said in one of the responses to your review comments, there is a difference between our proposed solutions. You iterate over the requests until the continue token is cleared and then present the total list to the response step. I wanted to present each partial list to the response step and then let the response step decide if it needed to continue to receive more data.

I do see that there are lots of improvements in code clarity that I would like to preserve, such as the Helm variable access. How should we merge what we've done?

russgold:
I can think of a few possibilities, including:

1. cherry-pick any commits that will work with few or no merges conflicts
2. checkout one branch and shelve whatever parts you want from it in IntelliJ and then checkout the other and unshelve
3. Brute-force copy.
4. Tell me which parts you want to keep from mine and have me split them into independent commits, which you can cherry-pick.

My goal was to hide the details from the higher-level code (including the response step), since it could then be independent from whatever follow-on logic is needed.

operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
```diff
@@ -126,6 +147,20 @@ private NextAction doPotentialRetry(Step conflictStep, Packet packet, CallRespon
     return null;
   }
 
+  /**
+   * Resets any retry strategy, such as a failed retry count and invokes the request again. This
+   * will be useful for patterns such as list requests that include a "continue" value.
+   * @param packet Packet
+   * @return Next action for the original request
+   */
+  private NextAction resetResetAndReinvokeRequest(Packet packet) {
```
ankedia, line 156: The method name is bit unclear. Should it be something like `resetRetryStrategyAndReinvokeRequest`?

rjeberhard: Yup... it's a typo. Thanks.

operator/src/main/java/oracle/kubernetes/operator/builders/WatchBuilder.java
```diff
@@ -29,7 +29,7 @@
   /** Ignored for watches. */
   private static final String START_LIST = null;
 
-  private static final Boolean ALLOW_BOOKMARKS = false;
+  private static final Boolean ALLOW_BOOKMARKS = true;
```
ankedia, line 32: The ALLOW_BOOKMARKS is set to `true` in WatchBuilder but I don't see BOOKMARK watch event handled in Watcher.java. My understanding of Watch bookmark feature is that it will send an event of type BOOKMARK to mark that all changes up to a given resourceVersion the client is requesting have already been sent. I was imagining that we'll update the resource version in Watcher.java with resource version set in BOOKMARK event . https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks

operator/src/main/java/oracle/kubernetes/operator/Watcher.java
```diff
@@ -93,6 +93,11 @@ void waitForExit() {
     }
   }
 
+  // for test
```
ankedia, line 96: Sorry, I hit enter by mistake in my previous comment. I was thinking that getNewResourceVersion() method of `Watcher.java` will check if event type is `BOOKMARK` and then update the resource version sent as part of bookmark event. Please let me know if my understanding is incorrect or if I missed something.

russgold:
Can you explain more about your approach? Do we currently have a requirement to get a partial list? If not, why have each ResponseStep decide whether to ask for more? And why add the new continue behavior to the generic Step class, when it is only needed in the specific case of responses to listXXXAsync calls?

operator/src/main/java/oracle/kubernetes/operator/builders/WatchBuilder.java
rjeberhard: I found that the behavior of the default branch of the switch was identical to what we would do for a specific BOOKMARK event, so there was no need to make further code change.

operator/src/main/java/oracle/kubernetes/operator/Watcher.java
rjeberhard: If you follow the logic through, you will see that this is what happens. I'm working on another branch, but tomorrow I'll find the existing code.

rjeberhard:
> Can you explain more about your approach? Do we currently have a requirement to get a partial list? If not, why have each ResponseStep decide whether to ask for more? And why add the new continue behavior to the generic Step class, when it is only needed in the specific case of responses to listXXXAsync calls?

I thought it would be a bad practice to take data that had been separated into slices and reassemble them unnecessarily into one giant list. Practically, it is unlikely that we would blow the heap because of this, but I noticed that almost every case where we process a list there was no need to have the whole list together as we just iterate through the results anyway.

Similarly, it makes sense to me that the code analyzing the response list would be best placed to decide if more data was needed.

Finally, we only have AsyncRequestStep and ResponseStep now. I didn't think it was work separating this pattern into a list version and a non-list version just for continue.

russgold:
I hear the memory concern; do we actually have cases now where we might choose not to process the entire list?

It seems to me that we now have code duplication, as each list response step needs to do the same kinds of processing to ask for a continue. That could be aggregated into a subclass of ResponseStep that does that, and possibly takes a Consumer<KubernetesObject> to decide what to do with each resource as it is retrieved. That separates the continuation logic from the processing logic. If at some time we do need to add partial list processing, we can change it to a Function<KubernetesObject,Boolean>

operator/src/main/java/oracle/kubernetes/operator/Watcher.java
ankedia: ok, got it. I wonder if we need to test it functionally (in integration test or performance test) to verify if bookmark events helps in mitigating the impact of short event history window. In any case, this looks fine to me.

russgold (approved):

ankedia (approved):

rjeberhard:
@russgold, @ankedia, thanks for the approvals. I'm going to merge this as-is so that I can work on getting 3.0.2 released and on story development. I'm very open to the improvements that Russ suggested here or that come from the other PR. Let's just do those as a separate commit, perhaps as part of the mob session.

PR #1890, opened 2020-08-27
Retrieve large numbers of namespaces in multiple calls

russgold:
This change takes advantage of the Kubernetes continue functionality to retrieve large result sets in list operations. Each list call returns only as many are available, up to the configured limit, and will perform multiple calls until the full set is retrieved. If new items are created or deleted, the list operation will restart. Currently, this is only used for namespaces.

The corresponding functionality has been added to KubernetesTestSupport in order to unit test it.


rjeberhard:
Hey, Russ. I already have a PR in-flight (PR #1881) that covers this functionality.

PR #1904, opened 2020-09-04
fix build application not to assert exec false exit value

bhavaniravichandran:
Occasionally, exec into the pod where an application is built fails with a non-zero exit code. The application builds successfully and the app seems to exist for the test to continue. This exit code is returned by kubernetes client api, so temporarily commenting out the assertion that checks for exitvalue of 0.  Now, any It test that calls buildApplication() to build an app, has to check that the packaged app exists before proceeding with the test. Seems like all the existing tests do that now since the run in kind cluster passed yesterday - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1808/

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
@@ -45,7 +45,7 @@ import static org.apache.commons.io.FileUtils.copyDirectory;
 import static org.apache.commons.io.FileUtils.deleteDirectory;
 import static org.awaitility.Awaitility.with;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertEquals;
+
 
 /**
  * Utility class to build application.
@@ -64,7 +64,8 @@ public class BuildApplication {
    * <p>The appSrcPath, your application source directory is zipped up and copied to a WebLogic server pod for building.
    * If your archives are placed under &lt application_source &gt /build after building, use <b>build</b> as the
    * archiveDistDir param value. This method copies the folder <b>archiveDistDir</b> to local file system and absolute
-   * path of the <b>archiveDistDir</b> directory is returned by this method.
+   * path of the <b>archiveDistDir</b> directory is returned by this method. In your It test, assert the
+   * application exists in the local <b>archiveDistDir</b> before proceeding with the test.
    * <p> Example Usage: </p>
 
    * <pre>{@literal
@@ -174,8 +175,14 @@ public class BuildApplication {
       if (exec.stderr() != null) {
         logger.info("Exec stderr {0}", exec.stderr());
       }
-      assertEquals(0, exec.exitValue(), "Exec into " + webLogicPod.getMetadata().getName()
-          + " to build an application failed with exit value " + exec.exitValue());
+
+      // Exec returns a non-zero return code intermittently even when the application builds
+      // successfully. This seems to be an issue with io.kubernetes.client.Exec.java. So, Commenting
+      // this assertion for now. it is now the responsibility of the It test class
+      // to assert the application exists before continuing with the test.
+
+      //assertEquals(0, exec.exitValue(), "Exec into " + webLogicPod.getMetadata().getName()
+      //    + " to build an application failed with exit value " + exec.exitValue());
 
       Kubernetes.copyDirectoryFromPod(webLogicPod,
           Paths.get(APPLICATIONS_PATH, archiveDistDir).toString(), destArchiveBaseDir);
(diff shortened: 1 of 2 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
```diff
@@ -174,8 +175,14 @@ public static Path buildApplication(Path appSrcPath, Map<String, String> antPara
       if (exec.stderr() != null) {
         logger.info("Exec stderr {0}", exec.stderr());
       }
-      assertEquals(0, exec.exitValue(), "Exec into " + webLogicPod.getMetadata().getName()
-          + " to build an application failed with exit value " + exec.exitValue());
+
+      // Exec returns a non-zero return code intermittently even when the application builds
+      // successfully. This seems to be an issue with io.kubernetes.client.Exec.java. So, Commenting
+      // this assertion for now. it is now the responsibility of the It test class
+      // to assert the application exists before continuing with the test.
```
vanajamukkara, line 182: why can't the check be done in this method instead of changing all the calling It test classes?

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
```diff
@@ -174,8 +175,14 @@ public static Path buildApplication(Path appSrcPath, Map<String, String> antPara
...
+
+      //assertEquals(0, exec.exitValue(), "Exec into " + webLogicPod.getMetadata().getName()
+      //    + " to build an application failed with exit value " + exec.exitValue());
 
       Kubernetes.copyDirectoryFromPod(webLogicPod,
```
vanajamukkara, line 187: doesn't the copy files/dir from pod fail if the build is not successful?

vanajamukkara (changes requested):

sankarpn (approved):
Looks good to me

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
sankarpn: The source application might build more than 1 archive and we don't know what file name it will be, so it is better to leave it to the It class to do that.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
sankarpn: Yes, but that is what we expect right?. Most of them the time what we  see is exit code issue even though build is successful.

vanajamukkara: Yes, we expect it to fail. So, if the build is not successful, will it fail at this point?

vanajamukkara: if it fails here, we don't need to add any additional checks in It class

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
vanajamukkara: I see the stderr is empty when it returns non-zero code on success, how about checking for 0 exit value or (non-zero and stderr is empty) for success? assuming on failure, stderr is not empty, this needs to be checked if not sure.
https://build.weblogick8s.org:8443/job/wko-kind-nightly/95/artifact/logdir/jenkins-wko-kind-nightly-95/wl_k8s_test_results/diagnosticlogs/ItCrossDomainTransaction/ItCrossDomainTransaction.out/*view*/

sankarpn: I don't know if we can rely on that logic. IMO a simple assertion in the It class would be sufficient. Like this
`assertTrue(clusterViewAppPath.toFile().exists(), "Application archive is not available");`

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
sankarpn: Depends on the source app. If the build failed but created a empty distDir this copy command may not fail. It will simply copy the empty directory to local file system.

new-integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/BuildApplication.java
vanajamukkara: Is that check done in all the existing it classes which calls this method? If not, that needs to be done in the It classes as part of this PR as this assertion is removed here, otherwise tests will proceed even if the application build fails.

sankarpn: Only 1 class was missing the assert. Added the assertion in ItManagedCoherence.java

vanajamukkara (approved):

PR #1911, opened 2020-09-11
OWLS-80038 and OWLS-80090: fix mountPath validation and token substitution

doxiao:
When the domain spec is validated before the operator starts a domain, the mountPath of a volumeMount is checked to see if it is valid. A mountPath may contain tokens, such as $(DOMAIN_HOME), or $(SERVER_NAME), which are substituted when the operator has the actual values. The problem is that the token substitution happens after the domain spec validation, so a mountPath that contains tokens are considered invalid. 

The PR contains the following changes/fixes with unit test cases:
1) Add code to detect if a mountPath value contains tokens of valid reserved/custom env var names, and if it does, the operator skips the mountPath part of the domain spec validation and defer it to later.
2) Add a validation step to check mountPath in WLS server pods and introspector job/pod spec after the token substitution happens. If the check fails, the operator quits with an failure message in the domain status. 
3) Fix a bug where volume mounts in the admin server's serverPod and cluster's serverPod are not checked when the domain spec is validated.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1901/


russgold (approved):
LGTM

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -441,6 +444,55 @@ public void introspectorPodStartupWithNullAdminUsernamePasswordEnvVarValues() {
...
+        .withAdditionalVolumeMount("volume1", VOLUME_MOUNT_PATH_1);
+
+    runCreateJob();
+
+    assertThat(job.getSpec().getTemplate().getSpec().getContainers().get(0).getVolumeMounts(),
+            hasVolumeMount("volume1", END_VOLUME_MOUNT_PATH_1));
+  }
+
+  @Test
+  public void whenDomainHasAdditionalVolumesWithCustomVariablesInvalidValue_reportValidationError() {
```
alai8, line 476: Should this test be named ..._jobNotCreated rather than ..._reportValidationError?

alai8 (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
doxiao: sure. thanks!

PR #1917, opened 2020-09-14
Testfixpostimageupdate

bhavaniravichandran:
New weblogic images have been uploaded to the registry and so we need to fix weblogic image tag for ItMiiDomain. The updated image also changed the uid/gid to oracle:root. WIT by default uses oracle:oracle and so changing the ItMiiSamole to use the chown option when running imagetool.
Will update this with the jenkins job once there is a successful run.

### DIFF
--- a/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
+++ b/new-integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
@@ -3,6 +3,7 @@
 
 package oracle.weblogic.kubernetes.extensions;
 
+
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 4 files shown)


new-integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
```diff
@@ -185,8 +204,21 @@ public void beforeAll(ExtensionContext context) {
           images.add(WLS_BASE_IMAGE_NAME + ":" + WLS_BASE_IMAGE_TAG);
           images.add(JRF_BASE_IMAGE_NAME + ":" + JRF_BASE_IMAGE_TAG);
           images.add(DB_IMAGE_NAME + ":" + DB_IMAGE_TAG);
+          /*
```
rjeberhard, line 207: Don't we still need the "PushToKind" side of this functionality?

rjeberhard: I think I've misread this -- you didn't comment out that code, instead, you've added the entire comment... why?

bhavaniravichandran: I think I need more than what I changed here. This PR is not ready for merge yet. I created a PR so you could get the changes to the 2 files I mentioned. I will have to still fix ImageBuilders.java. Running it locally once before pushing the changes.

PR #1919, opened 2020-09-14
MII jrf - improve wallet password handling

tbarnes-us:
Don't include a JRF OPSS wallet passphrase in the WDT 'createDomain.sh' command line since createDomain.sh echos its command line to stdout.


jshum2479 (approved):

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -674,16 +674,46 @@ function wdtCreatePrimordialDomain() {
...
+
+  if [ -z "${OPSS_FLAGS}" ]; then
+
+    # We get here for WLS domains, and for the JRF 'first time' case
+
+    # JRF wallet generation note:
+    #  If this is JRF, the unset OPSS_FLAGS indicates no wallet file was specified
+    #  via spec.configuration.opss.walletFileSecret and so we assume that
+    #  this is the first time this domain started for this RCU database.
+    #  We also assume that 'createDomain.sh' will setup new RCU tables for the domain
```
jshum2479, line 687: It should be create_rcu_schema.sh

jshum2479 (approved):

rjeberhard (approved):

rjeberhard:
@tbarnes-us and @jshum2479, I'm not sure how to resolve this conflict.

tbarnes-us:
> @tbarnes-us and @jshum2479, I'm not sure how to resolve this conflict.

What's the conflict?

rjeberhard:
Click the button that says "Resolve conflicts" and see :). I may have communicated the wrong tone with my statement that I "didn't know how." It looks like two changes that are just next to each other, so the tool complains, but I'm not sure so I wanted you two to weigh in.

tbarnes-us:
> Click the button that says "Resolve conflicts" and see :). I may have communicated the wrong tone with my statement that I "didn't know how." It looks like two changes that are just next to each other, so the tool complains, but I'm not sure so I wanted you two to weigh in.

I think the only difference is that Johnny's pull adds detailed log messages.   You can resolve the problem by including the entirety of my change within the conflict, and then adding the following before my change:

```
  if [ "JRF" == "$WDT_DOMAIN_TYPE" ] ; then
    if [ -z "${OPSS_FLAGS}" ] ; then
      trace INFO "An OPSS wallet was not supplied for the Model in Image JRF domain in its " \
        "'spec.configuration.opss.walletFileSecret' attribute; therefore, it's assumed that this is the first time " \
        "the OPSS RCU database is being accessed by the domain, so a schema and a wallet file will be created. " \
        "Consult the Model in Image documentation for instructions about preserving the OPSS wallet file."
    else
      trace "Creating JRF Primordial Domain"
    fi
  fi
```

PR #1926, opened 2020-09-15
replace StepTest with FiberTest, remove unused Fiber code

russgold:
1. Replace StepTest with FiberTest that uses the FiberTestSupport to validate Fiber by simulating threads and timers
2. Removed Fiber functionality that we actually haven't ever needed
3. Found and fixed a bug in FiberTest that was causing incorrect results when advancing time
4. Found and fixed an error in ServerDownIteratorStepTest caused by the aforementioned FiberTest bug

This cuts the time taken when running the full unit tests in the IDE by about a third.


lennyphan (approved):
LGTM

rjeberhard (approved):

PR #1930, opened 2020-09-17
OWLS-84141 fix operator upgrade issues related to domainNamespaceSelectionStrategy

doxiao:
Modified the operator helm chart template with the following:
1) set the default value to take care of the use cases where the new attribute domainNamespaceSelectionStrategy is missing;
2) if the new attribute is set to a no-default value, ignore the deprecated "dedicated" setting;
3) make the new attribute optional so that the helm validation does not fail if the attribute is missing.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/1985/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2023/  ( Upgrade Test )


rjeberhard (approved):

alai8 (approved):

vanajamukkara (approved):
LGTM

anpanigr (approved):
Both Upgrade and Full Test are Clean

PR #1932, opened 2020-09-18
Initial version of audit tool scripts for OWLS-83656

ankedia:
Initial version of audit tool scripts for OWLS-83656 based on JIRA specs and meeting discussion. Additional notes are captured in http://aseng-wiki.us.oracle.com/asengwiki/display/ASDevWLS/Image+Audit+Tool+Notes wiki,

### DIFF
--- a/src/scripts/audit/tool.sh
+++ b/src/scripts/audit/tool.sh
@@ -14,8 +14,8 @@ function usage {
   echo "  -k Node information file containing mapping of nodes to images. For Kubernetes use-case, this is the output of 'kubectl get nodes -o json' command."
   echo "  -m Container CLI command e.g. 'docker' or 'podman'. Defaults to 'docker'."
   echo "  -w Workspace directory, generated files will be stored in this directory. Defaults to './work' dir."
+  echo "  -d Disable validation check of images that are present in the Node information file but have not been pulled on the current machine."
   echo "  -s Silent mode."
-  echo "  -d Disable validation check of images that are present in node information file but have not been pulled on current machine."
   echo "  -h Help"
   exit 1
 }
--- a/src/scripts/audit/tool.sh
+++ b/src/scripts/audit/tool.sh
@@ -11,7 +11,7 @@ function usage {
   echo "usage: ${script} [-i] [-s] [-w] [-h]"
   echo "  -i File containing list of images to scan in [Repository]:[Tag]@[Hash] format."
   echo "  -k File containing Kubernetes Node information in json format (output of 'kubectl get nodes -o json')."
-  echo "  -w Working directory for the generated files. Defaults to '$scriptDir/work' dir."
+  echo "  -w Working directory for the generated files. Defaults to './work' dir."
   echo "  -s Silent mode."
   echo "  -h Help"
   exit 1
@@ -23,6 +23,7 @@ function usage {
 silentMode=false
 imageFile=""
 kubernetesFile=""
+nodeImages=""
 artifactsFile="util/artifacts/weblogic.txt"
 workDir=work
 while getopts "s:i:k:a:w:h:" opt; do
@@ -81,14 +82,27 @@ function initialize {
   if [ ! -f "$artifactsFile" ]; then
     validationError "Unable to locate the artifacts file ${artifactsFile}"
   fi
-
+   
   failIfValidationErrors
+
+  # Initialize workspace dir and generate report header
+  workDir=$scriptDir/work
+  initWorkDir
+  generateReportHeader
+
+  # validate availale images
+  if [[ -n "${kubernetesFile}" && -z "$imageFile" ]]; then
+    echo "Scanning all images on current machine, use '-i' option if you want to scan specific set of images."
+    imageFile=$(docker images --digests --format "{{.Repository}}:{{.Tag}}@{{.Digest}}@{{.ID}}" | grep -v "<none>:<none>")
+    imageNoTagFile=$(docker images --digests --format "{{.Repository}}@{{.Digest}}" | grep -v "<none>:<none>")
+    imageNoDigestFile=$(docker images --digests --format "{{.Repository}}:{{.Tag}}" | grep -v "<none>:<none>")
+
+    #echo "validating available images"
+    validateAvailableImages
+  fi
 
   # Read or create image file
   if [ -z "$imageFile" ]; then
-    echo "Scanning all images on current machine, use '-i' option if you want to scan specific set of images."
-    imageFile=$(docker images --format "{{.Repository}}:{{.Tag}}@{{.Digest}}" | grep -v "<none>:<none>")
-  else 
     IFS=$'\n' read -d '' -r -a images < "${imageFile}"
     imageFile=${images[@]}
   fi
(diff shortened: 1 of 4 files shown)


src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
+#!/bin/bash
+# Copyright 2020, Oracle Corporation and/or its affiliates. All rights reserved.
```
rjeberhard, line 2: We no longer use the phrase "All rights reserved" and some other corrections needed:

```
# Copyright (c) 2020, Oracle Corporation and/or its affiliates.
# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
```

rjeberhard: I want to know where you copied your line from because that place needs to be fixed too.

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+
+#set -x
+script="${BASH_SOURCE[0]}"
+scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
+source ${scriptDir}/utility.sh
```
rjeberhard, line 8: For the benefit of Joe's customers, we probably need to keep this script from referencing others so that customers have less to review. Remember that some of the customers using this script will not be using the operator and so additional dependencies will be suspicious.

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+
+function usage {
+  echo "usage: ${script} [-i] [-s] [-w] [-h]"
+  echo "  -i File containing list of images to scan in [Repository]:[Tag]@[Hash] format."
```
rjeberhard, line 12: Will the list be one-per-line or comma-separated, etc.?

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+  echo "  -k File containing Kubernetes Node information in json format (output of 'kubectl get nodes -o json')."
+  echo "  -w Working directory for the generated files. Defaults to './work' dir."
+  echo "  -s Silent mode."
+  echo "  -d Disable Image Validation"
```
rjeberhard, line 16: What kind of image validation is this? Does this mean "Disable image hash validation" or something else?

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+# Copyright 2020, Oracle Corporation and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+
+#set -x
+script="${BASH_SOURCE[0]}"
+scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
+source ${scriptDir}/utility.sh
+
+function usage {
+  echo "usage: ${script} [-i] [-s] [-w] [-h]"
```
rjeberhard, line 11: Wouldn't the usage have something like `[-i <filename>]`?

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+    imageHash=$(echo $image | cut -d'@' -f2)
+    imageId=$(echo $image | cut -d'@' -f3)
+    repository=$(echo $imageRepoTag | cut -d':' -f1)
+    tag=$(echo $imageRepoTag | cut -d':' -f2)
+
+    if [ -z ${imageId} ]; then
+      imageId=$(getImageId "${imageRepoTag}")
+    fi
+
+    imageSize=`docker inspect ${imageId} | jq .[].Size`
```
rjeberhard, line 174: I should have mentioned this earlier, but we will need to support other command-line tools, such as podman.

src/scripts/audit/tool.sh
```diff
@@ -0,0 +1,305 @@
...
+  local imageRepo=$1
+  local imageHash=$2
+  local imageSize=$3
+  local __result=$4
+
+  #sameSize=false
+  hashMatchFailed=false
+  imageNode=()
+  if [ -z "${kubernetesFile}" ]; then
+    eval $__result="FailedNotAvailable"
```
rjeberhard, line 218: Okay, what does this mean? (This is just the limit of my Bash knowledge)

src/scripts/audit/tool.sh
ankedia: Fixed. I don't remember exactly but I may have taken it from some old script which no longer exist in github.

src/scripts/audit/tool.sh
ankedia: ok, yes. I had considered this but later thought that extracting common library type functions and utilities to a separate script might make the main script (`tool.sh`) more readable and easier to review. This might also be helpful if we need to add more library type functions in future. Please note that utility.sh doesn't have a dependency on operator, it'll resides in same directory as `tool.sh` and will be shipped with the zip file containing audit tool script and artifacts file. 

I can consolidate it into a single script as you recommended. I just wanted to discuss before merging two scripts into one.

src/scripts/audit/tool.sh
ankedia: It'll be one-per-line, I'll mention this in the usage to clarify this.

src/scripts/audit/tool.sh
ankedia: It's a validation check for images that are present in the node information file but have not been pulled on current machine. In other words, it validates if any images are missing on current machine when option to "scan all images" is selected. I have clarified the usage and changed variable name. Please let me know if you have any suggestions for the wordings. Thanks.

src/scripts/audit/tool.sh
ankedia: I have fixed this.

src/scripts/audit/tool.sh
ankedia: I tried running `podman images` command in Openshift cluster to understand it's usage. It required me to run as `sudo` and it doesn't return any images when run as a regular user. I'm not sure about other environment. Do you know if there's any other environment where I can test `podman`?

src/scripts/audit/tool.sh
ankedia: This function gets the information about k8s nodes where a particular image has been pulled. It also verifies that hash value of image on current machine matches with hash value of image in node information file. If the Kubernetes node information file (file containing `kubectl get node` output) is not provided then I skip the hash verification and logic to get node information. Above line (line 218) returns a value indicating that either node information file is not provided or hash verification has failed.  In case of "hash verification failure", script will also print a warning message that "hash verification failed for a particular image" and will continue scanning other images. I'm also not an expert on bash scripting and looked up different ways to return value from function. Above approach of returning values is from this article (3rd approach) https://www.linuxjournal.com/content/return-values-bash-functions . I preferred this approach over "echoing" return value  as it allows to echo warning message in addition to retuning failure. Please let me know if this is not clear or need to be refactored. Thanks.

ankedia: I realize that it's hard to read bash function, below is the pseudo code for this function. I have also refactored the code a little bit to remove above confusing check. Please let me know if you have any suggestions to improve or if something is missing.

```
function verifyHashAndGetNodeNames (Argument imageRepository, Argument imageHash, Argument imageSize, Argument result ){ 

    PreConditions:
    Array "nodeToImageMapping" is populated in initialize method by calling "getImagesInNodeFile" function.
	This array contains Node to Image mapping (i.e. what images are pulled on which k8s nodes)
	It is populated by reading 'k8s node info' file.
	
	for each NodeAndImage pair in "nodeToImageMapping" array:
	  1. Parse NodeName, Image string and Image Size
	  2. Parse Image repo, tag and hash value from Image string obtained in step 1.
	  
	  If (Image repo in step 2 matches with input imageRepository) 
	    a) Check if image size from step 1 matches with imageSize input parameter
		   If match fails and size verification is enabled then provide a warning message and break from loop
		b) Add node name to the array of node-names where image is pulled.
		c) Check if hash value from step 2 matches with input imageHash value
		   If match fails then provide a warning message, set a flag and break from loop
	
    If (lengh of node array is 0 or hash match failed)
        return error value "NotAvailableOrFailed"
    else 
        return array containing node names where image is pulled.

}		
```

src/scripts/audit/tool.sh
ankedia: As discussed in meeting, utility.sh is part of this PR and will be shipped with tool.sh. I can rename it to helper.sh or something else if that helps avoiding the confusion.

src/scripts/audit/tool.sh
ankedia: I have incorporated changes to support other command-line tools such as podman. The tool can be run with "-m podman" option to replace docker with podman. I have tested tool with 'podman' on Openshift helper node.

rjeberhard (approved):

ankedia:
Thanks @rjeberhard and @lennyphan for approving this PR. As discussed, I have checked in the scripts in gitlab project at https://gitlab-odx.oracledx.com/weblogic/image-audit-tool . I'll be deleting this branch shortly.

PR #1933, opened 2020-09-21
Owls83813 fix a scaling issue when upgrade from 2.5.0

doxiao:
Add a check to void NullPointerException; minDynamicClusterSize does not exist in 2.5.0 release. 
Remove the workaround in integration test case because the test passes now without the workaround.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2010/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2028/


vanajamukkara:
@doxiao please update the PR with jenkins results

lennyphan (approved):
LGTM

alai8:
should we add a unit test even though the change is trivial? Thanks

doxiao:
> should we add a unit test even though the change is trivial? Thanks
This is an upgrade use case, and we have an integration test case that covers this already. I think we are fine.

alai8 (approved):

vanajamukkara (approved):
LGTM

PR #1937, opened 2020-09-22
Kubernetes Java Client 10.0.0

rjeberhard:
It's that time again... The Kubernetes Java Client team released a new version with some more important bug fixes and support for k8s 1.18.

I've received LT/BA approval in PLS.


docs/charts/index.yaml
```diff
@@ -3,116 +3,116 @@ entries:
   weblogic-operator:
   - apiVersion: v1
     appVersion: 3.1.0
-    created: "2020-09-16T21:57:54.167065-04:00"
```
russgold, line 6: We seem to have to update this file an awful lot. If it were published as a result of generating a website rather than updating the git repo, that wouldn't be necessary. Is it really so advantageous to avoid web site generation?

rjeberhard: I'll open a JIRA to investigate how we can do this differently. I don't care when we do this. There is some concern that not doing it during the build means that we don't test the chart generation until we push to master.

rjeberhard: https://jira.oraclecorp.com/jira/browse/OWLS-84737

ddsharpe (approved):

alai8 (approved):

PR #1938, opened 2020-09-22
Nightly intermittent failure in operator restart test

vanajamukkara:
operator restart during pod rolling restart test fails intermittently on nightly, currently after the domain is patched, the test waits for the admin pod to be deleted to make sure rolling restart has started, that is now modified to check if admin pod has restarted instead of checking for admin pod deletion as its possible that admin pod deletion may have happened too soon. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2025/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2026


anpanigr (approved):

xiancao (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItOperatorRestart.java
```diff
@@ -289,10 +290,11 @@ public void testOperatorRestartWhenPodRoll() {
         replicaCount,
         adminSecretName);
 
-    // wait till rolling restart has started by checking admin server pod is deleted
-    logger.info("Waiting for rolling restart to start by checking {0} pod is deleted in namespace {0}",
+    // wait till rolling restart has started by checking admin server pod has restarted
```
doxiao, line 293: The test case, as its name indicated, is intended to test operator restart DURING the managed servers are rolling-restarted. This change made it testing operator restart AFTER the managed servers' rolling-restart completes.

vanajamukkara: For the rolling restart, first admin pod is restarted and then managed server pods, so here the test waits for the admin pod to be restarted and then restarts operator while the managed servers are getting rolled.

doxiao: I think that makes sense. thanks.

PR #1950, opened 2020-09-24
Fix missing ODL configuration that may presents in the model

jshum2479:
ODL configuration is written to a logging.xml file under $domain/config/fmwconfig/servers/<server name> and not in config.xml file, archive the directories for any customization may have been supplied by user's customization.


jshum2479:
This depends on the latest to be released WDT.  Where will be the good place to put documentation, essentially if user wants to use WDT to configure ODL logging then it needs to use a WDT version and this fix.

tbarnes-us (approved):

tbarnes-us:
> This depends on the latest to be released WDT. Where will be the good place to put documentation, essentially if user wants to use WDT to configure ODL logging then it needs to use a WDT version and this fix.

The only place I know that publicly mentions a WDT version in the context of MII is the MII sample's `./docs-source/content/samples/simple/domains/model-in-image/prerequisites.md` (search for "1.9").

The general location where we document all pre-reqs is `/docs-source/content/userguide/introduction/introduction.md`, but we don't have any information there.  Maybe we should.

Regardless, it seems to me like updating either would need to wait until the WDT version is actually available...

rjeberhard (approved):

PR #1953, opened 2020-09-25
Unit test and fix detection of stranded namespaces

russgold:
Reworked the namespace list logic to eliminate the ActionStep and move the post-processing to after all of the domain namespaces have been identified. Simplified the domain selection logic to move it into the enum, and varied the namespaces selected by the three relevant tests to ensure that they were using different rules.

Note that this starts a refactoring which will end the superfluous passing of several intermediate values from method to method; however, it cannot proceed until a dependency between the NamespaceListAfterStep and some variables inside Main has been fixed - and that will require some additional unit testing to ensure coverage. I hope to work that in the future, but it was not needed for this bug fix.

I am running the change on Jenkins at https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2155/


russgold:
The method Main.createDomainRecheckSteps appears to check _namespaces_ rather than domains. Is that a mistake?

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -447,94 +421,237 @@ ServiceWatcher getServiceWatcher(String namespace) {
...
+        nextSteps.add(getNext());
+        return Step.chain(nextSteps.toArray(new Step[0]));
       }
-    };
 
-    public boolean isRequireList() {
-      return false;
-    }
+      private Set<String> getNamespacesToStart(List<String> namespaceNames) {
+        return namespaceNames.stream().filter(selectionStrategy::isSelected).collect(Collectors.toSet());
```
rjeberhard, line 611: This is substantially less efficient than the old code as the number of namespaces increases. Why the change? Here's what was done before: 
```
            namespacesToStartNow = namespacesToStartNow.stream().filter(
                    Pattern.compile(regexp).asPredicate()).collect(Collectors.toSet());
```
The new version compiles the regular expression for every namespace tested.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
```diff
@@ -196,6 +206,62 @@ private V1ObjectMeta createMetadata(DateTime creationTimestamp) {
...
+  }
+
+  // todo add unit tests for namespace shutdown. Must manipulate the namespaceStopping map. Can that be abstracted?
+
+  private void defineSelectionStrategy(SelectionStrategy selectionStrategy) {
+    TuningParameters.getInstance().put(Main.Namespaces.SELECTION_STRATEGY_KEY, selectionStrategy.toString());
+  }
+
+  @Test
+  public void whenNamespacesListedInMultipleChunks_dontDeclarePresentNamespacesAsMissing() {
```
alai8, line 242: what is the difference between this test and the whenNamespacesListedInOneChunk...()?

alai8 (approved):

operator/src/main/java/oracle/kubernetes/operator/Main.java
russgold: It was done for code simplification. I agree that it is less efficient; the real question should be, I think, whether there is a noticeable performance degradation. Do you have a way to test that?

russgold: I can cache the result, which should achieve roughly the same performance while adding only a small bit of extra code.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
russgold: The difference is the number of namespaces. A chunk, by default, is 50 items (see the following code in TuningParametersImpl.java:```
    CallBuilderTuning callBuilder =
        new CallBuilderTuning(
            *(int) readTuningParameter("callRequestLimit", 50)*,
            (int) readTuningParameter("callMaxRetryCount", 5),
            (int) readTuningParameter("callTimeoutSeconds", 10));
```

In the old code, the detection of whether a namespace was stranded was being done on each chunk, meaning that any namespace not in the current chunk was incorrectly reported as stranded. The new code does the check after all namespaces have been located.

operator/src/main/java/oracle/kubernetes/operator/Main.java
russgold: With the caching, the performance should actually be a bit better, as it should not need to recompile the pattern unless it is changed. I'm still not persuaded that it matters.

rjeberhard: You could create a test that compiles a slightly more complicated regular expression and measure the time.  The difference in code complexity seems fairly small though between having a method that checks if a namespace matches or returning a predicate function that does the same.

russgold: The approach I have taken avoids the need for other code to test the strategy; it makes the same call in all cases, and the logic is in the enum. At any rate, the new implementation should be even faster than the original.

rjeberhard: Let's add a test to demonstrate that the new implementation is still faster as the number of namespaces increases.

rjeberhard (approved):

PR #1972, opened 2020-10-06
OWLS-84786 - Use Kubernetes Java Client 10.0.0

ankedia:
Integrate the Kubernetes Java Client 10.0.0 into the operator and changes for resolving the cluster rolling issue.

Changed PodStepContext to wait for pod to be deleted before proceeding to replace/create pod.
Changed KubernetesTestSupport to return V1Pod instead of V1Status when deleting pod
Removed NullPodAwaiterStepFactory in PodHelperTestBase as it seemed redundant after above PodStepContext change and use PassthroughPodAwaiterStepFactory instead.

Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2266/


docs/charts/index.yaml
```diff
@@ -3,178 +3,116 @@ entries:
   weblogic-operator:
   - apiVersion: v1
     appVersion: 3.1.0
-<<<<<<< HEAD
```
rjeberhard, line 6: This is a bad merge :(. The easiest fix I've found is to checkout develop, copy the contents of docs/charts to a tmp directory, checkout your branch, copy docs/charts back, and commit. We have a separate JIRA to move the chart generation out of the build step, but we obviously haven't done this yet.

rjeberhard (approved):
Approving, except that you need to backout the change to docs/charts.

docs/charts/index.yaml
ankedia: I have backed out the changes to docs/charts as you suggested. Let me know if I missed something. Thanks.

alai8 (approved):

PR #1989, opened 2020-10-14
copy out scalingAction.log from admin pod when scaling cluster with WLDF

xiancao:
When scaling cluster with WLDF, sometime we need to check scalingAction.log for the root cause of the failure. Copy it out to the ittestsresults directory. 

Jenkins result:
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/2407/
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/2403/
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/2402/
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/2401/


anpanigr (commented):
Can u move the method  copyFileFromPod to utils/FileUtils.java to complement the existing method  copyFileToFPod

xiancao:
@anpanigr I added the method copyFileFromPod to utils/FileUtils.java which calls Kubernetes.copyFileFromPod following the pattern for the method copyFileToFPod.

anpanigr (approved):
LGTM

integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Domain.java
```diff
@@ -684,4 +698,33 @@ private static boolean executeCommandOnPod(V1Pod pod,
...
+   * @return true if no exception thrown, false otherwise
+   */
+  private static boolean copyFileFromPod(String namespace,
+                                         String pod,
+                                         String container,
+                                         String srcPath,
+                                         Path destPath) {
+    try {
+      getLogger().info("Copy file {0} from pod {1} in namespace {2} to {3}", srcPath, pod, namespace, destPath);
+      Kubernetes.copyFileFromPod(namespace, pod, container, srcPath, destPath);
```
lennyphan, line 718: Even though copyFileFromPod() was moved to FileUtils, you just going to call it directly in Kubernetes or shouldn't this be changed to use FileUtils?

xiancao: in Domain.java, all other methods are directly called from Kubernetes.java. That's why I did not change to use FileUtils.java

lennyphan: Still doesn't make sense to me to even have it in FileUtils if we're not going to call through it.  copyFileToPod() is only called through FileUtils when I do a search so makes sense to me to do the same.  My .02

lennyphan (approved):
LGTM

PR #1990, opened 2020-10-14
OWLS 84741: Scaling failed on Jenkins when setting Dedicated to true & io.kubernetes.client.openapi.ApiException: Not Found

lennyphan:
Currently, before any Operator REST Api invocation is allowed, the client is authenticated by the Operator if a TokenReview (based on the client's Bearer Token) is successfully created.  Subsequently, the Operator then does authorization checks by using a SubectAccessReview request to verify the client has the correct permissions to do the operation on a Resource.  If the client is successfully authorized, the WebLogic Operator REST implementation makes all K8s Api Server operations (on behalf of the client) using the Operator's service account credentials.  

When the Operator is configured to run with 'domainNamespaceSelectionStrategy=Dedicated' mode, all Resource modification operations are namespaced scope, except for TokenReview creation (which is a cluster scope operation).

The K8s Api Server already controls client's access (and hence any Resource modification operations) by authentication and authorization checks as outlined in https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/.

Since the K8s Api Server already does authentication and authorization checks, this change proposes to delegate these checks directly to the K8s Api Server by creating a K8s ApiClient object based on the client's 'Bearer Token'.  Thus all Resource operations through the Operator's REST api's are done using the client's service account and not the Operator's service account.  The only change in behavior is that permissions would have to be given to the client's service account to do 'patch' operations on the Domain resource.  Otherwise, the client's Operator REST Api invocations do not need to be changed.  Backward compatibility with the Operator's original authentication (TokenReview) and authorization (SubjectAccessReview) behavior can be enabled through an environment variable, "TOKEN_REVIEW_AUTHENTICATION=true".


operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
```diff
@@ -89,13 +97,38 @@
   RestBackendImpl(String principal, String accessToken, Collection<String> domainNamespaces) {
     LOGGER.entering(principal, domainNamespaces);
     this.principal = principal;
-    userInfo = authenticate(accessToken);
+    initializeCallBuilder(accessToken);
     this.domainNamespaces = domainNamespaces;
     LOGGER.exiting();
   }
 
+  private void initializeCallBuilder(String accessToken) {
```
russgold, line 105: It would probably be cleaner to add a fluent method to CallBuilder that would allow you to say something like:
```
  new CallBuilder().withAuthentication(new AccessTokenAuthentication(accessToken))
```

If the authentication is null, you'd get the current behavior. That would allow this logic to be moved into CallBuilder, where it really belongs. This isn't a REST thing.

But this whole bit of code seems needlessly complex. Find a way to simplify it, please.

operator/src/test/java/oracle/kubernetes/operator/rest/RestBackendImplTest.java
```diff
@@ -334,10 +411,86 @@ private void setupScanCache() {
...
+    /**
+     * Construct a RestBackendImpl that is used to handle one WebLogic operator REST request.
+     *
+     * @param principal        is the name of the Kubernetes user to use when calling the Kubernetes
+     *                         REST api.
+     * @param accessToken      is the access token of the Kubernetes service account of the client
+     *                         calling the WebLogic operator REST api.
+     * @param domainNamespaces a list of Kubernetes namepaces that contain domains that the WebLogic
+     */
+    RestBackEndStub(String principal, String accessToken,
```
russgold, line 457: Seems an odd place to break this method. Can it really not fit on one line?

russgold (changes requested):

russgold:
How do we validate the client's bearer token? This seems to deal with authorization. Is there a built-in module that checks the client certificate?

rjeberhard:
@russgold, 
> How do we validate the client's bearer token? This seems to deal with authorization. Is there a built-in module that checks the client certificate?

Previously, we were validating the client's bearer token by passing that token as part of an explicit token validation request. This required that the operator's service account have permission to create token authentication requests (which is, unfortunately, a cluster-level operation), which is not possible if the operator only has privileges in its own namespace.

The new design proposal is that the operator use the provided bearer token to create a new connection to the k8s master using that token (a delegation). In this way, k8s itself validates the token and the permissions of the service account that the token authenticates.

russgold:
> @rjeberhard 
> Previously, we were validating the client's bearer token by passing that token as part of an explicit token validation request.
> This required that the operator's service account have permission to create token authentication requests (which is,
> unfortunately, a cluster-level operation), which is not possible if the operator only has privileges in its own namespace.
> 
> The new design proposal is that the operator use the provided bearer token to create a new connection to the k8s master
> using that token (a delegation). In this way, k8s itself validates the token and the permissions of the service account that
> the token authenticates.

I think I may not have been clear. I have been assuming that the token was sent to the REST API by a client. Is that not the case? If so, how do we validate that the client has the right to that bearer token in the first place? It seems akin simply to allowing a client to send in any user name without a password.

rjeberhard:
> If so, how do we validate that the client has the right to that bearer token in the first place?

For lack of subtlety, yes, that's the model. If you have access to the token then it is yours to use. These tokens are protected by Secrets associated with a ServiceAccount or embedded in generated KUBECONFIG files. Once you have access to the bearer token though, then there is no further checking that it is yours to use.

integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Domain.java
```diff
@@ -534,6 +534,7 @@ private static boolean createRbacApiObjectsForWLDFScript(String domainNamespace,
               .addResourcesItem("domains")
               .addVerbsItem("get")
               .addVerbsItem("list")
+              .addVerbsItem("patch")
```
rjeberhard, line 537: We are adding a required role verb (this seems correct), but reminds me that we need to document this requirement in the scaling and/or RBAC documentation. Similarly, we need to document how to create the correct roles and role bindings so that the service account used with the scaling API has the necessary permissions.

operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
```diff
@@ -417,6 +450,15 @@ private WebApplicationException createWebApplicationException(int status, String
     return new WebApplicationException(rb.build());
   }
 
+  protected boolean authenticateWithTokenReview() {
+    return "true".equalsIgnoreCase(Optional.ofNullable(getEnvVariable.apply("TOKEN_REVIEW_AUTHENTICATION"))
```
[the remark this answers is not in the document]
rjeberhard: Please look at what Dongbo did in https://github.com/oracle/weblogic-kubernetes-operator/pull/1979. There the pattern is that there are values that can be passed to the Helm chart that preserve the pre-existing behavior. The chart puts the value into the operator's ConfigMap, which can then be accessed through the TuningParameters. Note: the Java code needs to handle the variable's not being present as it won't be present on a Helm upgrade.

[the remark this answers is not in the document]
lennyphan: Just to confirm Ryan, from looking at Dongbo's change, you want the property that controls the new vs old behavior to be defined the values.yaml of the Operator's helm chart and this will then be copied to the Operator's configmap which then can be accessed from TuningParameters? The property won't be defined as an environment variable nor system property.

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -191,3 +191,9 @@ externalServiceNameSuffix: "-ext"
 # cluster size.
 # The default value is true.
 clusterSizePaddingValidationEnabled: true
+
+# tokenReviewAuthentication specifies whether the the operator's REST API should use
+#   1. Kubernetes token review API for authenticating users, and
+#   2. Kubernetes subject access review API for authorizing a user's operation (get, list,
+#      patch, etc) on a resource.
+#tokenReviewAuthentication: false
```
russgold, line 199: "The default value is false." would be a good thing to note.

operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -1907,4 +1910,31 @@ private Call readPodLogAsync(
...
+  }
+
+  /**
+   * Create AccessTokenAuthentication component for authenticating user represented by
+   * the given token.
+   * @param accessToken - User's Bearer token
+   * @return - this CallBuilder instance
+   */
+  public CallBuilder withAuthentication(String accessToken) {
+    if (accessToken != null && !accessToken.isEmpty()) {
```
russgold, line 1925: We've been using the Guava method

```
Strings.isNullOrEmpty()
```

for cases like this.

operator/src/test/java/oracle/kubernetes/operator/rest/RestBackendImplTest.java
```diff
@@ -336,66 +335,52 @@ public void verify_getWlsDomainConfig_doesNotReturnNull_whenScanIsNull() {
   }
 
   @Test
-  public void verify_initializeCallBuilder_withAccessToken_userInfoIsNull() {
-    RestBackendImpl restBackendImpl = new RestBackendImpl("", "", Collections.singletonList(NS));
-    assertThat(restBackendImpl.getUserInfo(), nullValue());
+  public void vwhenUsingAccessToken_userInfoIsNull() {
```
russgold, line 338: typo: "when" rather than "vwhen"

russgold (approved):
With the few minor changes noted, this looks good to me.

operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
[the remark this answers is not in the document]
rjeberhard: That's correct.

kubernetes/charts/weblogic-operator/values.yaml
rjeberhard: This leaves ambiguous the meaning of "false." Would you please update this to describe that when the value is false that the caller's bearer token will be delegated so that the update to the Domain resource is done using the caller's privileges?

rjeberhard (changes requested):
Lenny, I'm very happy with the code changes. I believe we need some documentation changes though. This change needs to be mentioned in a few places:
1. Where we describe the REST API, we need to discuss that by default we will now use the caller's bearer token to perform the underlying update to the Domain resource: https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-operators/using-the-operator/the-rest-api/
2. Document the new Helm values: https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-operators/using-the-operator/using-helm/

operator/src/test/java/oracle/kubernetes/operator/rest/RestBackendImplTest.java
```diff
@@ -326,6 +335,74 @@ public void verify_getWlsDomainConfig_doesNotReturnNull_whenScanIsNull() {
...
+    RestBackEndStub restBackEndStub = new RestBackEndStub("", "", Collections.singletonList(NS));
+    assertThat(restBackEndStub.getUserInfo(), notNullValue());
+  }
+
+  @Test
+  public void verify_authorizationCheck_notCalled_whenAuthenticateWithTokenReviewIsFalse()
+      throws NoSuchFieldException, IllegalAccessException {
+    RestBackendImpl restBackendImpl = new RestBackendImpl("", "", Collections.singletonList(NS));
+    AuthorizationProxyStub authorizationProxyStub = new AuthorizationProxyStub();
+    Field nameField = restBackendImpl.getClass()
```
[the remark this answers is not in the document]
lennyphan: Changed to more fluent interface per recommendation.

operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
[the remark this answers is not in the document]
lennyphan: Change to use TuningParameters.

operator/src/main/java/oracle/kubernetes/operator/rest/RestBackendImpl.java
lennyphan: moved AccessTokenAuthentication logic CallBuilder as suggested.  Isolated to callbuilder as that's where it's more suited and relevant.

operator/src/test/java/oracle/kubernetes/operator/rest/RestBackendImplTest.java
lennyphan: Fixed.

docs-source/content/userguide/managing-domains/domain-lifecycle/scaling.md
```diff
@@ -87,8 +87,27 @@ For example, when using `curl`:
 curl -v -k -H X-Requested-By:MyClient -H Content-Type:application/json -H Accept:application/json -H "Authorization:Bearer ..." -d '{ "managedServerCount": 3 }' https://.../scaling
 ```
 
-If you omit the header, you'll get a `400 (bad request)` response. If you omit the Bearer Authentication header, then you'll get a `401 (Unauthorized)` response.
+If you omit the header, you'll get a `400 (bad request)` response. If you omit the Bearer Authentication header, then you'll get a `401 (Unauthorized)` response.  If the service account or user associated to the `Bearer` token does not have permission to `patch` the WebLogic domain resource, you'll get a `403 (Forbidden)` response.
```
rosemarymarano, line 90: associated to -> associated with
resource, you'll -> resource, then you'll

docs-source/content/userguide/managing-domains/domain-lifecycle/scaling.md
```diff
@@ -87,8 +87,27 @@ For example, when using `curl`:
...
 
+{{% notice note %}}
+To resolve a `403 (Forbidden)` response, when calling the operator's REST scaling API, you may need to add the `patch` request verb to the cluster role associated to the WebLogic `domains` resource.
```
rosemarymarano, line 93: associated to -> associated with

docs-source/content/userguide/managing-operators/using-the-operator/the-rest-api.md
```diff
@@ -14,7 +14,12 @@ You can access most of the REST services using `GET`, for example:
 * To obtain a list of domains, send a `GET` request to the URL `/operator/latest/domains`
 * To obtain a list of clusters in a domain, send a `GET` request to the URL `/operator/latest/domains/<domainUID>/clusters`
 
-All of the REST services require authentication.  Callers must pass in a valid token header and a CA certificate file.  Callers should pass in the `Accept:/application/json` header.
+All of the REST services require authentication.  Callers must pass in a valid token header and a CA certificate file.  In previous operator versions, the operator would perform authentication and authorization checks using the Kubernetes token review and subject access review API's and then update the Domain resource using the operator's privileges.  Now the operator will, by default, use the Caller's bearer token to perform the underlying update to the Domain resource using the Caller's privileges and thus delegating authentication and authorization checks directly to the Kubernetes API Server (see [REST interface configuration]({{< relref "/userguide/managing-operators/using-the-operator/using-helm.md#rest-interface-configuration" >}})).
```
rosemarymarano, line 17: operator would perform ->operator performed
API's -> APIs, (no apostrophe, add a comma)
and then update -> and then updated
Now the operator will, by default, use -> Now, by default, the operator will use
Caller's bearer token -> caller's bearer token 
Caller's privileges -> caller's privileges

docs-source/content/userguide/managing-operators/using-the-operator/the-rest-api.md
```diff
@@ -14,7 +14,12 @@ You can access most of the REST services using `GET`, for example:
...
+All of the REST services require authentication.  Callers must pass in a valid token header and a CA certificate file.  In previous operator versions, the operator would perform authentication and authorization checks using the Kubernetes token review and subject access review API's and then update the Domain resource using the operator's privileges.  Now the operator will, by default, use the Caller's bearer token to perform the underlying update to the Domain resource using the Caller's privileges and thus delegating authentication and authorization checks directly to the Kubernetes API Server (see [REST interface configuration]({{< relref "/userguide/managing-operators/using-the-operator/using-helm.md#rest-interface-configuration" >}})).  
+{{% notice note %}}
+When scaling up or scaling down a WebLogic cluster, through the operator's REST services, you may need to grant `patch` access to the user or service account associated with the Caller's bearer token. This can be done with an RBAC ClusterRoleBinding between the user or service account and the ClusterRole that defines the permissions for the WebLogic `domains` resource.
```
rosemarymarano, line 19: When scaling up or scaling down a WebLogic cluster, through the operator's REST services, -> When using the operator's REST services to scale up or down a WebLogic cluster, 
Caller's bearer token -> caller's bearer token (is there a reason that Caller is capitalized?)

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -468,7 +468,21 @@ Example:
 ```
 externalOperatorKey: QmFnIEF0dHJpYnV0ZXMKICAgIGZyaWVuZGx5TmFtZTogd2VibG9naWMtb3B ...
 ```
-
+##### `tokenReviewAuthentication`
+tokenReviewAuthentication, if set to `true`, specifies whether the the operator's REST API should use
```
rosemarymarano, line 472: tokenReviewAuthentication, if set to `true`, specifies -> If set to `true`, `tokenReviewAuthentication` specifies
should use -> should use:
@lennyphan Is the following meant to be a numbered list, in other words, does `tokenReviewAuthentication` do each step in order? If so, keep it numbered. If not, then change it to a bulleted list.

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -468,7 +468,21 @@ Example:
...
+   1. Kubernetes token review API for authenticating users, and
```
rosemarymarano, line 473: Remove this from the sentence: (comma), and

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -468,7 +468,21 @@ Example:
...
+   2. Kubernetes subject access review API for authorizing a user's operation (get, list,
+      patch, etc) on a resource.
```
rosemarymarano, line 475: get, list, patch, etc -> `get`, `list`, `patch`, and such

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -468,7 +468,21 @@ Example:
...
+   3. Update the Domain resource using the operator's privileges.
+ This parameter, if set to `false`, will use the caller's bearer token for any update
```
rosemarymarano, line 477: This parameter, if set to `false`, will use -> If set to `false`, this parameter will use

rosemarymarano (commented):
Some stuff. If there are questions to clear up, I'll try to look this over again later today.

integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/Domain.java
lennyphan: Updated documentation in /userguide/managing-operators/using-the-operator/using-helm.md, /userguide/managing-operators/using-the-operator/the-rest-api.md, /userguide/managing-domains/domain-lifecycle/scaling.md, and /kubernetes/charts/weblogic-operator/values.yaml

kubernetes/charts/weblogic-operator/values.yaml
lennyphan: updated with description.

operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
lennyphan: Changed to use Guava method.

operator/src/test/java/oracle/kubernetes/operator/rest/RestBackendImplTest.java
lennyphan: fixed typo

docs-source/content/userguide/managing-domains/domain-lifecycle/scaling.md
lennyphan: Fixed all review comments on documentation.

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -469,12 +469,13 @@ Example:
 externalOperatorKey: QmFnIEF0dHJpYnV0ZXMKICAgIGZyaWVuZGx5TmFtZTogd2VibG9naWMtb3B ...
 ```
 ##### `tokenReviewAuthentication`
-tokenReviewAuthentication, if set to `true`, specifies whether the the operator's REST API should use
-   1. Kubernetes token review API for authenticating users, and
-   2. Kubernetes subject access review API for authorizing a user's operation (get, list,
-      patch, etc) on a resource.
-   3. Update the Domain resource using the operator's privileges.
- This parameter, if set to `false`, will use the caller's bearer token for any update
+If set to `true`, tokenReviewAuthentication specifies whether the the operator's REST API should use:
```
rosemarymarano, line 472: tokenReviewAuthentication specifies -> `tokenReviewAuthentication` (code font) specifies

docs-source/content/userguide/managing-operators/using-the-operator/using-helm.md
```diff
@@ -469,12 +469,13 @@ Example:
...
+   * Kubernetes token review API for authenticating users and
+   * Kubernetes subject access review API for authorizing a user's operation (get, list,
+      patch, and such) on a resource.
```
rosemarymarano, line 475: get, list, patch, -> `get`, `list`, `patch`, (code font, surround with back ticks)

rosemarymarano (approved):
Few minor edits left, otherwise fine.

rosemarymarano (approved):

rjeberhard (approved):

PR #1992, opened 2020-10-14
October refactoring

russgold:
This refactoring eliminates a lot of duplicate code and centralizes much of the processing of domain namespaces, including management of watchers. It also addresses some misuse of Kubernetes resource versions: according to the documentation, the initial reference version for a watcher should be taken from the result of a list operation. In two cases, configmaps and jobs, we were using resource versions from other resources.

In the current K8s implementation, that doesn't matter, as it currently uses the etcd `modifiedIndex` which is consistent for a server; however, that is subject to change.

This cleanup has highlighted some additional issues I hope to address in the next PR, including the fact that we do different processing for a namespace that was in existence before the operator started than for one created while the operator is running. For example, the ADDED event handling of the Namespace watcher has special handling for the script map, while it should be handled by common namespace startup code.


russgold:
I'm seeing a lot of integration test failures, so I am closing this to figure out what is going on. I am going to add these changes on top of the latest commit that has clean IT results.

PR #2000, opened 2020-10-20
Update test dependencies and POM and change Dockerfile to use JDK 15

rjeberhard:
This PR updates the Docker file to use JDK 15 and updates the POM to move test dependencies to the most recent production versions. 

This also (unintentionally) cleans up an earlier bad merge to the generated Helm chart package.


json-schema-maven-plugin/pom.xml
```diff
@@ -44,13 +44,13 @@
...
             <artifactId>asm</artifactId>
-            <version>7.1</version>
+            <version>9.0</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
 
     <properties>
-        <junit-version>4.13</junit-version>
+        <junit-version>4.13.1</junit-version>
```
russgold, line 53: There's some duplication here, specifying the property in each POM. At the very least, this property should be moved to the top-level POM. Even better, the junit dependency (and in fact, any dependency used in more than one module) should be moved to a `dependencyManagement` section in the top-level POM if we don't want it automatically included in ever module. Those that use it would then list it in their dependencies without a version, inheriting the one from the project.

rjeberhard: Mark moved the JUnit 4 dependency down into the individual modules because he didn't want any chance that the new integration tests would use JUnit 4 symbols. I'm happy to change this however will work. I wonder if we could just universally switch to JUnit 5 now?

PR #2003, opened 2020-10-21
Add instructions for creating a custom Security Context Constraint

markxnelson:


docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+
+COPY --from=builder --chown=oracle:root /u01 /u01
+```
+
+However, if you build your own image, or obtain an image from another source, it
+may not have the necessary permissions.  You may need to configure similar file
+system permissions to allow your image to work in OpenShift.
+
+Alternatively, you may choose to configure OpenShift to allow use of UID 1000.  This
+can be done using a Security Context Constraint.  Oracle recommends that you define
```
rosemarymarano, line 41: Security Context Constraint -> Security Context Constraint (SCC)
I checked the capitalization of this term in the OpenShift docs and this is what I found: it does not need to be capitalized. It can be capitalized prior to providing its abbreviation, SCC, but otherwise, can be used lower case. (They are not 100% consistent in their docs, but that's my takeaway from them.)

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+volumes:
+- configMap
+- downwardAPI
+- emptyDir
+- persistentVolumeClaim
+- projected
+- secret
+```
+
+Assuming you called that file `uid1000.yaml`, you can create the Security Context Constraint
```
rosemarymarano, line 95: Security Context Constraint -> security context constraint

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+using the following command:
+
+```bash
+$ oc create -f uid1000.yaml
+```
+
+After you have created the security context constraint, you can install the WebLogic Operator.
```
rosemarymarano, line 102: the WebLogic Operator -> the operator (or WebLogic Server Kubernetes Operator)

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+Make sure you use the same service account that you granted permission to in the security
+context constraint (`weblogic-operator` in the example above).  The operator will then run
```
rosemarymarano, line 104: service account that you granted permission to -> service account to which you granted permission
the example above -> the preceding example

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+with UID 1000, and any WebLogic Domain it creates will also run with UID 1000.
```
rosemarymarano, line 105: Domain  -> domain

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
+
+#### Use the `anyuid` Security Context Constraint
+
+The built-in `anyuid` security context constraint
+will also ensure proper access to the file system within the
+Docker image, however this security context constraint actually grants more permissions
+than are actually needed. This means that the administrator must:
```
rosemarymarano, line 112: however this security context constraint actually grants more permissions than are actually needed. -> you can remove both instances of "actually", or at a minimum, remove the latter one.

rosemarymarano (commented):
Some style issues/comments

markxnelson:
@rosemarymarano i addressed all of your comments, please review again.

rosemarymarano (approved):
Consider changing Security Context Constraint -> security context constraint throughout the doc. Other than that, I approve this PR.

docs-source/content/security/openshift.md
```diff
@@ -5,14 +5,111 @@ weight: 7
...
-The Docker images that Oracle publishes default to the container user
-as `oracle`, which is UID `1000` and GID `1000`. When running the
-Oracle images or layered images that retain the default user as
-`oracle` with OpenShift, the `anyuid` security context constraint
-is required to ensure proper access to the file system within the
-Docker image. This means that the administrator must:
+#### Security requirements to run WebLogic in OpenShift
+
+WebLogic Server Docker images obtained from Oracle Container Registry have an `oracle` user with
+UID 1000 and it is in a group, which is also called `oracle` and has GID 1000.
```
rjeberhard, line 11: This is no longer true. Now the images have user `oracle` that is part of the `root` group. We did this precisely so that the images would work better with OpenShift. I'm going to keep reading, but my guess is that much of the rest of this content is affected by this change. Looping in @mriccell.

markxnelson: updated as discussed

rjeberhard (approved):

PR #2012, opened 2020-10-27
OWLS85461 add introspect version to server pod label

doxiao:
1. Add a new label to server pods with unit test cases.
2. Update the public doc.
3. some cleanup.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2691/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2693/


docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
```diff
@@ -44,6 +44,21 @@ Set `introspectVersion` to a new value.
...
+Namespace:      domain1-ns
+Priority:       0
+Labels:         weblogic.createdByOperator=true
+                weblogic.domainName=domain1
+                weblogic.domainUID=domain1
+                weblogic.introspectVersion=12345
+                weblogic.serverName=admin-server
+```
+
+When a domain's `spec.intrsopectVersion` is changed, the `weblogic.introspectVersion` label of each WebLogic Server pod is updated to the new `introspectVersion` value either when the operator restarts the pod, or when the operator determines that the pod does not need to be restarted.
```
rosemarymarano, line 60: value either -> value, either (comma here)
the pod, or when -> the pod or when (no comma here)

docs-source/content/userguide/managing-domains/domain-resource.md
```diff
@@ -161,7 +161,7 @@ Elements related to domain [startup and shutdown]({{< relref "/userguide/managin
...
+* `introspectVersion`: Changes to this field cause the operator to repeat its introspection of the WebLogic domain configuration (see [initiate introspection]({{< relref "/userguide/managing-domains/domain-lifecycle/introspection/_index.md#initiating-introspection" >}})). Repeating introspection is required for the operator to recognize changes to the domain configuration, such as adding a new WebLogic cluster or Managed Server instance, to regenerate configuration overrides, or to regenerate the WebLogic domain home when the `domainHomeSourceType` is FromModel. Introspection occurs automatically, without requiring change to this field, when servers are first started or restarted after a full domain shut down. For the FromModel `domainHomeSourceType`, introspection also occurs when a running server must be restarted because of changes to any of the fields [listed here]({{< relref "/userguide/managing-domains/domain-lifecycle/startup#fields-that-cause-servers-to-be-restarted" >}}). See also `overridesConfigurationStrategy`.
```
rosemarymarano, line 164: [initiate introspection] ->[Initiating introspection]
See also `overridesConfigurationStrategy`. -> I do not see `overridesConfigurationStrategy` on this page.

rosemarymarano (commented):
A few edits.

docs-source/content/userguide/managing-domains/domain-resource.md
doxiao: "See also overridesConfigurationStrategy" part is old  (was there before this PR). I'll double check it.

doxiao: This bullet is added in 3.0.0, and there has never been overridesConfigurationStrategy as far as I can tell. The closest is overrideDistributionStrategy. Tom? Ryan?

rjeberhard: Yes, it's a typo. It's referring to `overrideDistributionStrategy`.

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
```diff
@@ -57,7 +57,7 @@ Labels:         weblogic.createdByOperator=true
                 weblogic.serverName=admin-server
 ```
 
-When a domain's `spec.intrsopectVersion` is changed, the `weblogic.introspectVersion` label of each WebLogic Server pod is updated to the new `introspectVersion` value either when the operator restarts the pod, or when the operator determines that the pod does not need to be restarted.
+When a domain's `spec.intrsopectVersion` is changed, the `weblogic.introspectVersion` label of each WebLogic Server pod is updated to the new `introspectVersion` value, either when the operator restarts the pod or when the operator determines that the pod does not need to be restarted.
```
tbarnes-us, line 60: intrsopectVersion --> introspectVersion

operator/src/main/resources/Operator.properties
```diff
@@ -189,6 +189,7 @@ WLSKO-0176=Job {1} in namespace {0} failed, job details are {2}
 WLSKO-0177=Pod {0} in namespace {1} failed, the pod status is {2}
 WLSKO-0178=Patching administration server Pod labels with ''weblogic.introspectVersion'' {2} for WebLogic domain with UID: {0}. Administration server name: {1}.
 WLSKO-0179=Patching managed server Pod labels with ''weblogic.introspectVersion'' {2} for WebLogic domain with UID: {0}. Managed server name: {1}.
+WLSKO-0180=Operator cannot start. 'Dedicated' namespace strategy selected, but the CRD is not installed
```
tbarnes-us, line 192: Please reference the exact name/location of the namespace strategy setting and expand the CRD acronym to something like - Custom Resource for Domains (CRD)

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
```diff
@@ -44,6 +44,21 @@ Set `introspectVersion` to a new value.
...
+
+```
+Name:           domain1-admin-server
+Namespace:      domain1-ns
+Priority:       0
+Labels:         weblogic.createdByOperator=true
+                weblogic.domainName=domain1
+                weblogic.domainUID=domain1
+                weblogic.introspectVersion=12345
+                weblogic.serverName=admin-server
```
tbarnes-us, line 57: restartVersion?

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
```diff
@@ -44,6 +44,21 @@ Set `introspectVersion` to a new value.
 
 As with `restartVersion`, the `introspectVersion` field has no required format; however, we recommend using a value likely to be unique such as a continually increasing number or a timestamp.
 
+Beginning with operator 3.1.0, if a domain resource's `spec.introspectVersion` is set, each of the domain's WebLogic Server pods will have a label with the key `weblogic.IntrospectVersion` to indicate the `introspectVersion` at which the pod is running.
```
tbarnes-us, line 47: weblogic.IntrospectVersion --> weblogic.introspectVersion

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -364,6 +366,24 @@ protected Step patchPod(V1Pod currentPod, Step next) {
...
+    return createProgressingStep(updatePod(currentPod, next));
+  }
+
+  protected Step updatePod(V1Pod currentPod, Step next) {
+    JsonPatchBuilder patchBuilder = Json.createPatchBuilder();
+    String baseName = "/metadata/labels/";
+    if (!getPodLabels().containsKey(INTROSPECTION_STATE_LABEL)) {
+      patchBuilder.add(baseName + INTROSPECTION_STATE_LABEL, getDomain().getSpec().getIntrospectVersion());
+    } else {
+      patchBuilder.replace(baseName + INTROSPECTION_STATE_LABEL, getDomain().getSpec().getIntrospectVersion());
```
tbarnes-us, line 379: Minor minor nit: combine repeated patch string concatenation code into a single local variable...

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
doxiao: It is just an example to show the new label. startVersion is not always set.

operator/src/main/resources/Operator.properties
```diff
@@ -187,6 +187,8 @@ WLSKO-0175=Job {0} in namespace {1} failed with status {2}. Check log messages \
   copied from the introspector pod {3} log for additional information.
 WLSKO-0176=Job {1} in namespace {0} failed, job details are {2}
 WLSKO-0177=Pod {0} in namespace {1} failed, the pod status is {2}
+WLSKO-0178=Patching administration server Pod labels with ''weblogic.introspectVersion'' {2} for WebLogic domain with UID: {0}. Administration server name: {1}.
+WLSKO-0179=Patching managed server Pod labels with ''weblogic.introspectVersion'' {2} for WebLogic domain with UID: {0}. Managed server name: {1}.
```
tbarnes-us, line 191: There's a lot of overlap between the MANAGED/ADMIN _ POD_ PATCHED/CREATED/REPLACED/EXISTS log messages at KO-0120/0121/0041/0042/0043/0050/0051/0052.  

IMO, they should be aligned since that would simplify tracing pod transitions.  Something like:

```
Action on pod ''{0}'' of type ''managed server/adminstration server'', operation ''patched/created/replaced/exists'', domain UID ''{1}'', label ''weblogic.introspectVersion''=''{2}'', label ''weblogic.restartVersion''=''{3}''.
```
They could remain separate log messages, or combined into a single log message that parameterizes the operation and server type.

Also, I suspect 'domain UID' is no longer needed since it's probably always reported in the log message's domain UID field...

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
tbarnes-us: Up to you.  introspectVersion is not always set either.

operator/src/main/resources/Operator.properties
doxiao: I would suggest that we make this change in a different PR after everyone agrees.  Note that a lot of unit test cases check the log messages.

docs-source/content/userguide/managing-domains/domain-resource.md
doxiao: thanks. I made the change.

docs-source/content/userguide/managing-domains/domain-lifecycle/introspection.md
doxiao: I am sure there are many things that are in the pod but not in this example. We are talking about the use case where the  introspect version is set here.

operator/src/main/resources/Operator.properties
russgold: I don't think all that many tests are checking _each_ message. The change shouldn't be all that much work, but I agree it is not really part of this PR.

tbarnes-us: I've filed https://jira.oraclecorp.com/jira/browse/OWLS-85639 to track.  @rjeberhard FYI

docs-source/content/userguide/managing-domains/domain-resource.md
```diff
@@ -161,14 +161,14 @@ Elements related to domain [startup and shutdown]({{< relref "/userguide/managin
...
+* `introspectVersion`: Changes to this field cause the operator to repeat its introspection of the WebLogic domain configuration (see [initiating introspection]({{< relref "/userguide/managing-domains/domain-lifecycle/introspection/_index.md#initiating-introspection" >}})). Repeating introspection is required for the operator to recognize changes to the domain configuration, such as adding a new WebLogic cluster or Managed Server instance, to regenerate configuration overrides, or to regenerate the WebLogic domain home when the `domainHomeSourceType` is FromModel. Introspection occurs automatically, without requiring change to this field, when servers are first started or restarted after a full domain shut down. For the FromModel `domainHomeSourceType`, introspection also occurs when a running server must be restarted because of changes to any of the fields [listed here]({{< relref "/userguide/managing-domains/domain-lifecycle/startup#fields-that-cause-servers-to-be-restarted" >}}). See also `overridesConfigurationStrategy`.
```
rosemarymarano: [initiating introspection] -> [Initiating introspection] (capital "I" as in the section title.)

rosemarymarano (approved):
One minor nit, otherwise approve.

russgold (changes requested):
The Pod processing already has a way to change labels and uses a patch to do so. All that should be necessary to add the introspect version is to add it to the computation of the model. This seems a lot more code than necessary.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
russgold: Or better, into a method call. But I agree that this code is needless duplication.

doxiao:
> The Pod processing already has a way to change labels and uses a patch to do so. All that should be necessary to add the introspect version is to add it to the computation of the model. This seems a lot more code than necessary.

That is actually what I started to do initially. I then realized that it was much cleaner to separate the existing patching use case and this one The additional code is marginally more, especially after I consolidated the code in three different use cases. Having said that, let me go through the process of using the existing patch step again and see.

doxiao:
Kicked off a Jenkins run for the latest. https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2720/

docs-source/content/userguide/managing-domains/domain-resource.md
doxiao: fixed. thanks for your quick review!

tbarnes-us (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -359,6 +361,15 @@ protected Step patchPod(V1Pod currentPod, Step next) {
     KubernetesUtils.addPatches(
         patchBuilder, "/metadata/annotations/", getAnnotations(currentPod), getPodAnnotations());
 
+    if (introspectVersionChanged(currentPod)) {
```
russgold, line 364: This is still much more complex than is needed. Look at `PodStepContext.createMetadata()`. This is where we define the labels and annotations for server pods. To add a new label, all you need to do is define it there. The rest of the code will automatically handle detection of differences between what is supposed to be and what we find.

operator/src/main/resources/Operator.properties
```diff
@@ -187,7 +187,7 @@ WLSKO-0175=Job {0} in namespace {1} failed with status {2}. Check log messages \
   copied from the introspector pod {3} log for additional information.
 WLSKO-0176=Job {1} in namespace {0} failed, job details are {2}
 WLSKO-0177=Pod {0} in namespace {1} failed, the pod status is {2}
-WLSKO-0180=Operator cannot start. 'Dedicated' namespace strategy selected, but the CRD is not installed
+WLSKO-0178=Operator cannot start. Operator 'domainNamespaceSelectionStrategy' is set to 'Dedicated', but the Custom Resource Definition for ''domains.weblogic.oracle'' is not installed.
```
russgold, line 190: Is there a particular reason to renumber this? I deliberately skipped some numbers to allow 178 and 179 for messages that seemed logically connected to the previous ones. Do we have a rule banning skipped numbers?

russgold (changes requested):

operator/src/main/resources/Operator.properties
doxiao: Sorry, I had two log messages 0178 and 0179 in my previous version of the changes. Now they are gone and I thought yours could move up. I assume that we could skip numbers.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
doxiao: The label was added in the createPodModel method but in the non hashed part. If you look into getPodLabels you'll see that the labels that are to be patched here are the custom defined labels in the server specs.

russgold (approved):
LGTM

PR #2014, opened 2020-10-28
OWLS 85530: OPERATOR INTROSPECTOR THROWS VALIDATION ERRORS FOR STATIC CLUSTER

lennyphan:
With 14.1.1.0 WebLogic image, static/configured clusters are being mis-identified as dynamic clusters.  The workaround bug, where WLST throws an exception when there are no 'real' DynamicServers, may not occur as a DynamicServerMBean is returned although it is not configured with any dynamic servers:
```
<cluster>
   <name>c1</name>
   <dynamic-servers>
     <maximum-dynamic-server-count>0</maximum-dynamic-server-count>
   </dynamic-servers>
</cluster>
```
The above <cluster> configuration gets generated when running online WLST.  Validation errors similar to the following can be seen when a configured cluster is mis-identified as a dynamic cluster:

validationErrors:
- "The WebLogic dynamic cluster \"c1\"' is not referenced by any server template."
- "The WebLogic dynamic cluster \"c1\"'s dynamic servers use calculated listen ports."
- "The WebLogic dynamic cluster \"c1\" is referenced by configured server \"ms1\", the operator does not support 'mixed clusters' that host both dynamic (templated) servers and configured servers."
- "The WebLogic dynamic cluster \"c1\" is referenced by configured server \"ms2\", the operator does not support 'mixed clusters' that host both dynamic (templated) servers and configured servers."

### DIFF
--- a/src/integration-tests/introspector/createStaticCluster.py
+++ b/src/integration-tests/introspector/createStaticCluster.py
@@ -1,0 +1,50 @@
+# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+#
+# purpose:
+#
+#    invoke on-line WLST to create a configured static cluster and two managed
+#      servers that reference it.
+#
+#    assumes the admin user/pass was encoded to a userConfig/userKey by the
+#      introspector these files have been mounted to /weblogic-operator/introspector
+#
+# usage:
+#    wlst.sh createStaticCluster.py <url> <cluster_name>
+#
+# sample usage:
+#    wlst.sh createStaticCluster.py t3://domain1-admin-server:7001 c1
+#
+
+import sys
+
+url = sys.argv[1]
+cluster_name = sys.argv[2]
+
+connect(userConfigFile='/weblogic-operator/introspector/userConfigNodeManager.secure',userKeyFile='/tmp/userKeyNodeManager.secure.bin',url=url)
+
+edit()
+startEdit()
+
+cl=cmo.createCluster(cluster_name)
+
+# Create managed servers
+for index in range(0, 2):
+  cd('/')
+
+  msIndex = index+1
+  name = '%s%s' % ('ms', msIndex)
+
+  create(name, 'Server')
+  cd('/Servers/%s/' % name )
+  print('managed server name is %s' % name);
+  set('ListenPort', 8001)
+  set('ListenAddress', '')
+  set('Cluster', cl)
+
+save()
+activate()
+
+print 'ok'
+exit(exitcode=0)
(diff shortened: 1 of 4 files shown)


alai8 (approved):

src/integration-tests/introspector/introspectTest.sh
```diff
@@ -39,6 +39,12 @@
 #     To check for ISTIO
 #         export ISTIO_ENABELD=true - this only test for Non Model in Image
 #
+#     To check for introspection of configured clusters with WebLogic 14.1.1.0 image.
+#     NOTE: Test for introspection of configured clusters only for Non Model in Image.
+#
+#         export WEBLOGIC_IMAGE_TAG=14.1.1.0
+#         introspectTest.sh
```
tbarnes-us, line 46: Please update the "Usage:" and "Internal Test Flow" sections of the README file for this test to reflect this check-in's additions.  Also, if it's not too much trouble, please also update the README to discuss the ISTIO_ENABLED flag (I don't recall who added that - maybe Johnny).

src/integration-tests/introspector/createStaticCluster.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2018, 2020, Oracle Corporation and/or its affiliates.
```
tbarnes-us, line 1: Does the static cluster in this test need to be dynamically added?  it could be added via off-line WLST in the existing `wl-create-domain-pod.pyt` file.  If that's fine, then there'd be no need for introspectTest.sh to be updated to rerun the introspector, and no need to run this file -- the first run of the introspector would cover the issue under test.

lennyphan: The interesting thing is that creating a static cluster via off-line produces the following cluster configuration:
<cluster>
  <name>c1</name>
  <cluster-messaging-mode>unicast</cluster-messaging-mode>
</cluster>

whereas on-line WLST produces:

<cluster>
    <name>c1</name>
    <cluster-messaging-mode>unicast</cluster-messaging-mode>
    <dynamic-servers>
      <maximum-dynamic-server-count>0</maximum-dynamic-server-count>
    </dynamic-servers>
  </cluster>

The call to ClusterMBean.getDynamicServers() returns a DynamicServersMBean.  Hence, the reason I created the cluster using on-line WLST and rerun the introspector in order to test this issue.

tbarnes-us: Oh. That's ugly and helps explain why we haven't seen the problem before. Sigh. 

To prevent someone from mistakenly 'cleaning up' the code in the future, please add a comment in this script and also in one or two key places in the main script.  The comment should be something like:   

// NOTE: The static cluster must be configured using on-line WLST instead of off-line WLST in order to reproduce OWLS 85530.

sankarpn (approved):

src/integration-tests/introspector/createStaticCluster.py
tbarnes-us: Looks good! Thanks.

src/integration-tests/introspector/introspectTest.sh
tbarnes-us: Looks good! Thanks!

tbarnes-us (approved):

PR #2019, opened 2020-10-30
Fix dedicated mode test

russgold:
1. Fixes ItDedicatedMode that was failing as a result of my previous PR
2. Refactoring - DomainNamespaces is now a class with instance variables rather than the old static variables. This means that tests create instances of it rather than using reflection to access its internals. It also means that the code passes an instance of this class to code that needs it.

Jenkins links: 
[ItDedicatedMode](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2873/)
[Parallel tests](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2874/)


operator/src/test/java/oracle/kubernetes/operator/MainTest.java
```diff
@@ -153,14 +168,6 @@ public void tearDown() throws Exception {
...
 
-  @SuppressWarnings("unchecked")
-  private Map<String, NamespaceStatus> getNamespaceStatusMap()
-          throws NoSuchFieldException, IllegalAccessException {
-    Field field = DomainNamespaces.class.getDeclaredField(NAMESPACE_STATUS_MAP);
-    field.setAccessible(true);
-    return (Map<String, NamespaceStatus>) field.get(null);
-  }
-
   private Map<String, NamespaceStatus> createNamespaceStatuses() {
```
lennyphan, line 171: I think createNamespaceStatuses() can be removed.

lennyphan (approved):
LGTM other than my one nitpick comment.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
russgold: Yep, you're right - there's a few of these that can go away.

rjeberhard (approved):

PR #2020, opened 2020-10-30
Owls85582 take all ALWAYS servers before considering rest of the servers when meet cluster replicas requirement

doxiao:
1. Changed the operator code to take all severs with ALWAYS start policy before considering rest of the managed servers to meet the replicas requirement.
2. Add unit test cases for ALWAYS and NEVER start policy working with cluster replicas, including scale up and scale down.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2858/ (passed with the latest product changes)
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2869/ (in progress with additional unit test changes)


operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -118,23 +118,30 @@ public NextAction apply(Packet packet) {
...
+        .forEach(wlsClusterConfig -> addClusteredServersToFactory(
+            factory, clusteredServers, wlsClusterConfig, pendingServers));
 
     wlsDomainConfig.getServerConfigs().values().stream()
         .filter(wlsServerConfig -> !clusteredServers.contains(wlsServerConfig.getName()))
-        .forEach(wlsServerConfig -> factory.addServerIfNeeded(wlsServerConfig, null));
+        .forEach(wlsServerConfig -> factory.addServerIfAlways(wlsServerConfig, null, pendingServers));
+
+    for (ServerConfig serverConfig : pendingServers) {
+      factory.addServerIfNeeded(serverConfig.wlsServerConfig, serverConfig.wlsClusterConfig);
```
tbarnes-us, line 131: Would this approach affect the guaranteed 'lexi numeric' order in which we start or shutdown servers?  Or which servers are reported in status?   (For example, when we're shutting down a cluster's servers one-at-a-time, the goal is to shutdown only the 'highest' server first, then the second highest, and so on.)

doxiao: The approach considers this. The pending list is in the original order. We have unit test cases covering this too.

doxiao: But the final list does not maintain the original order. For example, if server3 is always and replicas is 2, server1 and server3 will be started, and server3 will be started before server1.

doxiao: In the sample example, if cluster later scales down, server1 will be taken down.

doxiao: The only behavior difference is the startup order among servers that need to be started up in the same round of make right check; servers with ALWAYS policy will be started before the servers with If-needed policy.

tbarnes-us: We want the overall startup and shutdown order to be very intuitive, predictable, and the exact reverse of each-other.

Question 1: I think it'd be better if server1 always first in a cluster, then server2, and so on, if they aren't configured to start concurrently -- regardless of whether any of the servers are marked 'always'.  So, based on your analysis of the cluster use case ^^^ where server3 (marked always) starts before server1 (marked if_needed), it sounds like this pull may need to be refined?

Question 2:  As for shutdown, if replica count is 3, and server1 & 3 are 'if needed' while server2 is always, then reducing replica count to 1 (or 0) should always shutdown server3 first and then shutdown server1 second.  (Side note: when setting the entire cluster to NEVER the entire cluster is expected to shutdown concurrently, so no worries there.)  Based on your analysis, it sounds like this will be honored?

rjeberhard: I understand why you started the ALWAYS servers first, but I agree that it's more important that servers be started (and stopped) in a predictable and consistent order. We've had a few customers ask about startup ordering, so I could see us doing something like that in a future release. I think that means we need to keep the order very, very simple now so that customers would understand what they are configuring later.

doxiao: "Yes" to both questions. Good catch on sequential startup order. I think we need to resort the final list.

doxiao: Fixed. 
Kicked off a new Jenkins job https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2893/.

rjeberhard (approved):

lennyphan (approved):
LGTM

tbarnes-us (approved):

PR #2027, opened 2020-11-04
Detect and shut down stuck server pods

russgold:
This code looks for stuck pods, which is detects by comparing the current time with any deleted time stamp, allowing for the specified grace period. If any pods managed by the operator is terminated, but are still present after the grace period, the operator will force-delete them and run a new make-right on their domain(s).

Note: we need to do some code cleanup on the async calls. Adding new parameters to be checked by unit tests is increasingly difficult. It is time to introduce an object rather than passing in a list (see Clean Code, chapter 3, regarding function parameters).

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
@@ -16,6 +16,8 @@ import io.kubernetes.client.openapi.models.V1PodList;
 import oracle.kubernetes.operator.calls.CallResponse;
 import oracle.kubernetes.operator.helpers.CallBuilder;
 import oracle.kubernetes.operator.helpers.PodHelper;
+import oracle.kubernetes.operator.logging.LoggingFacade;
+import oracle.kubernetes.operator.logging.LoggingFactory;
 import oracle.kubernetes.operator.steps.DefaultResponseStep;
 import oracle.kubernetes.operator.work.NextAction;
 import oracle.kubernetes.operator.work.Packet;
@@ -23,8 +25,14 @@ import oracle.kubernetes.operator.work.Step;
 import oracle.kubernetes.utils.SystemClock;
 import org.joda.time.DateTime;
 
-public class StuckPodProcessing {
+import static oracle.kubernetes.operator.logging.MessageKeys.POD_FORCE_DELETED;
 
+/**
+ * Under certain circumstances, when a Kubernetes node goes down, it may mark its pods as terminating, but never
+ * actually remove them. This code detects such cases, deletes the pods and triggers the necessary make-right flows.
+ */
+public class StuckPodProcessing {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
 
   private final MainDelegate mainDelegate;
 
@@ -114,7 +122,8 @@ public class StuckPodProcessing {
     private Step createForcedDeletePodStep(V1Pod pod) {
       return new CallBuilder()
             .withGracePeriodSeconds(0)
-            .deletePodAsync(getName(pod), getNamespace(pod), getDomainUid(pod), null, new DefaultResponseStep<>());
+            .deletePodAsync(getName(pod), getNamespace(pod), getDomainUid(pod), null,
+                  new ForcedDeleteResponseStep(getName(pod), getNamespace(pod)));
     }
 
     private String getName(V1Pod pod) {
@@ -129,4 +138,22 @@ public class StuckPodProcessing {
       return PodHelper.getPodDomainUid(pod);
     }
   }
+
+  static class ForcedDeleteResponseStep extends DefaultResponseStep<V1Pod> {
+
+    private final String name;
+    private final String namespace;
+
+    public ForcedDeleteResponseStep(String name, String namespace) {
+      this.name = name;
+      this.namespace = namespace;
+    }
+
+    @Override
+    public NextAction onSuccess(Packet packet, CallResponse<V1Pod> callResponse) {
+      LOGGER.info(POD_FORCE_DELETED, name, namespace);
+      return super.onSuccess(packet, callResponse);
+    }
+  }
+
 }
--- a/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
@@ -100,7 +100,7 @@ public class StuckPodProcessing {
         Collection<StepAndPacket> startDetails = new ArrayList<>();
 
         for (V1Pod pod : stuckPodList) {
-          startDetails.add(new StepAndPacket(createDeletePodStep(pod), packet.clone()));
+          startDetails.add(new StepAndPacket(createForcedDeletePodStep(pod), packet.clone()));
         }
         return doForkJoin(readExistingNamespaces(), packet, startDetails);
       }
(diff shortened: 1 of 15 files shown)


operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
```diff
@@ -0,0 +1,132 @@
...
+    private boolean isStuck(V1Pod pod, DateTime now)  {
+      return getExpectedDeleteTime(pod).isBefore(now);
+    }
+
+    private DateTime getExpectedDeleteTime(V1Pod pod) {
+      return getDeletionTimeStamp(pod).plusSeconds((int) getDeletionGracePeriodSeconds(pod));
+    }
+
+    private long getDeletionGracePeriodSeconds(V1Pod pod) {
+      return Optional.of(pod).map(V1Pod::getMetadata).map(V1ObjectMeta::getDeletionGracePeriodSeconds).orElse(1L);
```
rjeberhard, line 74: No issue... Reminder to myself to verify if 1 is the proper default for deletionGracePeriodSeconds if not set. I had thought that it defaulted to 30 seconds.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
```diff
@@ -0,0 +1,132 @@
...
+      final List<V1Pod> stuckPodList = getStuckPodList(packet);
+      if (stuckPodList.isEmpty()) {
+        return doNext(packet);
+      } else {
+        Collection<StepAndPacket> startDetails = new ArrayList<>();
+
+        for (V1Pod pod : stuckPodList) {
+          startDetails.add(new StepAndPacket(createForcedDeletePodStep(pod), packet.clone()));
+        }
+        return doForkJoin(readExistingNamespaces(), packet, startDetails);
```
rjeberhard, line 105: I'd like to make sure that we've thought through and verified the concurrency here. While we don't want to depend on watches, it's very likely that the deletion of a pod will trigger our pod watcher. The pod watcher would start a make-right for the associated domain. 

That seems right. This flow here (reading existing resources) will generate "recheck" make-right calls, which won't interrupt a make-right flow for the given domain that may have been started because of the watch event following the deletion.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Nonnull
+    private Step readExistingNamespaces() {
+      return mainDelegate.getDomainNamespaces().readExistingResources(namespace, mainDelegate.getDomainProcessor());
+    }
+
+    private Step createForcedDeletePodStep(V1Pod pod) {
+      return new CallBuilder()
+            .withGracePeriodSeconds(0)
+            .deletePodAsync(getName(pod), getNamespace(pod), getDomainUid(pod), null, new DefaultResponseStep<>());
```
rjeberhard, line 117: Does the DefaultResponseStep include typical retry behavior? We still have that issue where a delete will periodically fail because Kubernetes returns a V1Status rather than a V1Pod.

That's actually a little scary because this code might retry and accidentally interleave with make-right code triggered by the pod watcher that has already created a new pod.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
```diff
@@ -0,0 +1,132 @@
...
+    public NextAction onSuccess(Packet packet, CallResponse<V1PodList> callResponse) {
+      callResponse.getResult().getItems().stream()
+            .filter(pod -> isStuck(pod, now))
+            .forEach(pod -> addStuckPodToPacket(packet, pod));
+      
+      return doContinueListOrNext(callResponse, packet);
+    }
+
+    private boolean isStuck(V1Pod pod, DateTime now)  {
+      return getExpectedDeleteTime(pod).isBefore(now);
```
tbarnes-us, line 66: - The stuck check potentially involves two different machine clocks that might be out of sync - the operator's and the pod's. I just did some brief research, and it looks like k8s doesn't require strict clock synchronization although it's considered desirable (in addition, WL clusters have fairly tight needs).
- Should there be an additional grace after a detecting a potentially stuck delete in order to give K8S time to 'naturally' honor its timeout and perhaps to help account for a 'small' amount of potential clock drift?

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
russgold: This is not something we set, but what Kubernetes puts into the pod when it deletes it. My assumption is that it should always be set, and I need a default value here in case it isn't, just to help decide if the pod is stuck. I don't know what it would mean for Kubernetes to fail to set the value.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
russgold: I was not aware of that possibility - that seems a bit scary for it to be non-determinate in its behavior.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
russgold: That seems a reasonable precaution.

rjeberhard:
@russgold, I've tested a build from your branch and it does detect and delete a stuck pod. Can you please add an INFO level message indicating that the operator found and is deleting such a pod? Presently, I only see a message when the operator subsequently creates the replacement pod.

operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
tbarnes-us: Tough question. Off the top my head, maybe account for a few seconds of drift -- I've seen much worse on occasion, but, in general, WL clusters themselves dislike a drift any higher than that in my experience.   Plus perhaps account for a few extra seconds to give k8s time to handle the delete 'naturally'.  So that would total 10 seconds? And perhaps have an associated configurable?

@rjeberhard Thoughts?

rjeberhard: I keep any accounting for drift very small; perhaps just 1 or 2 seconds. It's subjective, but when I tested the solution it still "felt" like I waited a long time before the terminated pod was removed.

russgold: Actually, what's the harm? If we delete a pod that is going to be deleted anyway, it should still be fine, right?

tbarnes-us: It's my understanding this change initiates a forced delete - so in-memory-replicated-state data loss is possible if the forced delete occurs too early  for the operator to perform it's graceful shutdown

rjeberhard: The harm is that instead of catching only a pod that is "stuck" in the Terminating state that we could also detect pods that are in the process of shutting down normally and delete them before they've had a chance to complete normal shut down. This should only happen if the operator's pod is badly skewed from the master and the operator thinks the time is later than the master does.

rjeberhard: I wish that we could find a way to know that time at the master, but I've not been able to find it.

rjeberhard (approved):

rjeberhard:
Per discussion, merging this now.

PR #2033, opened 2020-11-05
Added anti affinity to the wls pods

marinakog:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/2994


anpanigr (approved):
LGTM  based on kind nightly result.

vanajamukkara:
have you verified that the pods are running on different nodes with these changes? Can you update with OKE jenkins results.

xiancao (approved):
LGTM

anpanigr (commented):
@vanajamukkara  we can write an explicit test to verify the distribution of pod across nodes.  We need to skip test while running locally where the node count is one. I have a backlog JIRA to address it.  Marina can you verify the Pod Distribution manually on OKE cluster.

PR #2035, opened 2020-11-06
owls-85910 - display correct minimumReplicas status for dynamic clusters

ankedia:
OWLS-85910 - Display correct value of `domain.status.clusters[].minimumReplicas`  for dynamic clusters when `domain.spec.allowReplicasBelowMinDynClusterSize` is at its default (true).

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -602,7 +602,8 @@ public class DomainStatusUpdater {
       }
 
       private Integer getClusterMinimumSize(String clusterName) {
-        return getDomainConfig()
+        return getDomain().isAllowReplicasBelowMinDynClusterSize(clusterName)
+              ? 0 : getDomainConfig()
               .map(config -> config.getClusterConfig(clusterName))
               .map(WlsClusterConfig::getMinClusterSize)
               .orElse(0);


rjeberhard (approved):

lennyphan (approved):

tbarnes-us:
Can/should the new sample leverage this information?

rjeberhard:
@tbarnes-us said:
> Can/should the new sample leverage this information?

I'm open to that, but I would want it to be a separate effort since we can more easily change the sample script.

What does having the script use that information get us? How hard would it be to change the script?

tbarnes-us:
I don't know off the top of my head how hard it would be to change the script.  If min-reps is active, the sample could end up incorrectly determining which servers are already up, incorrectly assuming that decreasing the replica count would shutdown a server that's actually forced up due to 'minimum replicas', incorrectly assuming that increasing the replica count would prevent a server from shutting down (keep=false and starting an arbitrary server), etc, etc

tbarnes-us:
It'd probably end up being  minimal change to the sample scripts - the hard part is going through each path through the code and making sure it'll work fine (by code inspection and ad hoc testing)

ankedia:
I'm wondering if it's feasible to change the operator to either update the cluster's replica count when min replicas is active or have an effective replica count field in domain status. I'm not sure if there is any downside to either approach or if it's better to keep this logic in script.

tbarnes-us:
Anil - With this pull's fix in place, shouldn't it already be pretty clear when min-replicas is active since it's reported in the status?  But you don't want the script to rely on the status, it could check the topology within the configmap (which I though it already did in one place?), as well as the `domain.spec.allowReplicasBelowMinDynClusterSize`

tbarnes-us:
Anil - With this pull's fix in place, shouldn't it already be pretty clear when min-replicas is active since it's reported in the status?

If you don't want the script to rely on the status, could it check the topology within the configmap in order to get the min-replicas, as well as the domain.spec.allowReplicasBelowMinDynClusterSize?

ankedia:
Yes. I was thinking to keep the script simple so we don't have to go through each code path and make sure it works in all cases. Also, I thought it might be more intuitive for customers to know why a server is running when min-replicas is active and number of running servers are more than replica count .

tbarnes-us:
It doesn't seem appropriate for the operator to change the configured 'replicas' value that the end-user has specified - as then the reported configured value wouldn't reflect the actual configured value. 

There is already 'status.cluster[].minimumReplicas' and 'status.cluster[].replicasGoal' in place:

```
  - clusterName: cluster-1
    maximumReplicas: 12
    minimumReplicas: 1
    readyReplicas: 9
    replicas: 9
    replicasGoal: 9
```

I assume 'replicasGoal' reflects the effective replicas count...

tbarnes-us:
If it doesn't, then that'd be another bug we need to fix.

ankedia:
ok, cool. In that case, I assume script can get replica count value from `status.clusters[].replicasGoal` instead of `domain.spec.clusters[].replicas` and there should not be any other changes needed. Is this correct assumption? Thanks.

ankedia:
I guess we also need a validation check in stop server script to fail if allowReplicasBelowMinDynClusterSize is false and minimum number of servers are running.

tbarnes-us:
> ok, cool. In that case, I assume script can get replica count value from status.clusters[].replicasGoal instead of domain.spec.clusters[].replicas and there should not be any other changes needed. Is this correct assumption? Thanks.

assuming that feature is correctly working...  :)

> I guess we also need a validation check in stop server script to fail if allowReplicasBelowMinDynClusterSize is false and minimum number of servers are running.

maybe for 'keep=false' but I assume not for 'keep=true'

PR #2036, opened 2020-11-06
Remove t3 channel in IT tests and add it to separate IT class.

sankarpn:
- This PR removes the t3 channel exposed in all IT classes due to intermittent failures caused by port conflict.
- Instead a separate test to expose t3 channel is added to the IT class ItT3Channel.java

Full test suite results
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3076/testReport/oracle.weblogic.kubernetes/


xiancao (commented):
Both Jenkins runs failed at oracle.weblogic.kubernetes.ItExternalRmiTunneling.testExternalRmiAccessThruHttpsTunneling. Is this expected?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
```diff
@@ -289,6 +326,197 @@ public void testDedicatedModeNoCrd() {
     }
   }
 
+  /**
```
vanajamukkara, line 329: Is it possible to add t3channel to the existing MII domain used in the dedicated mode tests and use that to deploy the application using WLST instead of adding new domain? It saves time by not bringing up a new domain.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
```diff
@@ -397,6 +625,45 @@ private void verifyDomainNotRunning(String domainNamespace) {
     }
   }
 
+  private static void verifyMemberHealth(String adminServerPodName, List<String> managedServerNames,
```
vanajamukkara, line 628: move this into an util class

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
```diff
@@ -289,6 +326,197 @@ public void testDedicatedModeNoCrd() {
...
+          managedServerPodNamePrefix + i, domain1Namespace);
+      checkPodReadyAndServiceExists(managedServerPodNamePrefix + i, domainUid, domain1Namespace);
+    }
+
+    // deploy application and verify all servers functions normally
+    logger.info("Getting node port for T3 channel");
+    int t3ChannelNodePort = assertDoesNotThrow(()
+        -> getServiceNodePort(domain1Namespace, getExternalServicePodName(adminServerPodName), "t3channel"),
+        "Getting admin server t3channel node port failed");
+    logger.info("techannel channel node port: {0}", t3ChannelNodePort);
```
vanajamukkara, line 501:
```suggestion
    logger.info("t3channel channel node port: {0}", t3ChannelNodePort);
```

vanajamukkara (changes requested):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
sankarpn: I don't want to modify the common model file used for this image MII_BASIC_IMAGE_NAME + ":" + MII_BASIC_IMAGE_TAG because it is shared by many tests. 
Since anyway this test is running alone, the time it take to run this test shouldn't matter IMO.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
sankarpn: This method uses a specific app deployed in the test method. Its not a member for util class.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDedicatedMode.java
sankarpn: fixed

sankarpn:
> 
> 
> Both Jenkins runs failed at oracle.weblogic.kubernetes.ItExternalRmiTunneling.testExternalRmiAccessThruHttpsTunneling. Is this expected?

Thanks for pointing out. Fixed it. The ItExternalRmiTunneling was using a wrong api to get the service port and the test was failing intermittently. Updated the test to use correct api.

vanajamukkara (approved):
LG.

anpanigr (commented):
Need to modify the kindtest.sh to exclude ItT3Channel test in Parallel run and include it in sequential run.

anpanigr (approved):
LGTM

PR #2037, opened 2020-11-06
Owls85476 attempt to fix intermittent integration test issues in nightlies

doxiao:
1. Use "kubectl exec" instead of Kubernetes Java client exec call to verify application accessibility.
2. Use Awaitility utilities instead of a while loop for retries.
3. Sync up with the completion of the main thread from the accounting thread.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3043/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3044/


anpanigr (approved):
LGTM

lennyphan (approved):
LGTM

PR #2038, opened 2020-11-06
Response from pod delete REST call can be Pod or Status

rjeberhard:
We've had a long-running issue where calls to delete a pod would fail even though the underlying deletion was successful. This failure was due to attempting to deserialize a Pod from the response even though the actual response from the REST call could either be a Pod or a Status depending on the timing of the underlying deletion operation.

The Java API handled this by only handling the Pod response and failing otherwise.

This is causing problems with the "stuck Terminating pod" deletion functionality because that activity is retried repeatedly interleaving with the operator's behavior to create a new pod -- therefore, the customer would see a pod deleted, immediately recreated, and then deleted again in a cycle that would go for five iterations.

This change replaces the pod deletion response code with one that only assumes that the response will be an Object. It turns out that we never really looked at the return value assuming that the call was successful.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/StuckPodProcessing.java
@@ -139,7 +139,7 @@ public class StuckPodProcessing {
     }
   }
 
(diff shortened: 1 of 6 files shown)


ddsharpe (approved):

lennyphan (approved):
LGTM

PR #2042, opened 2020-11-09
OWLS 85938 - Server lifecycle script changes to use domain.status.minimumReplicas when stopping/starting server

ankedia:
OWLS 85938 - Server lifecycle script changes to use `domain.status.minimumReplicas` when allowReplicasBelowMinDynClusterSize is set to `false`. It also fixes a minor bug when setting server start policy to ALWAYS.

This change adds a check in stop script to see if replica count value is same as minimum replica count value and prevents script from incorrectly assuming that decreasing the replica count would shut down a server. Additionally, when replica count value of cluster is lower than minimum replica count, it resets the replica count value to minimum replica count value. This helps server start script to know the value of effective replica count.

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/stopServer.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/stopServer.sh
@@ -82,6 +82,7 @@ serverStarted=""
 effectivePolicy=""
 managedServerPolicy=""
 stoppedWhenAlwaysPolicyReset=""
+replicasEqualsMinReplicas=""
 withReplicas="CONSTANT"
 withPolicy="CONSTANT"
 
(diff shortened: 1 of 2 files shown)


rjeberhard (approved):

kubernetes/samples/scripts/domain-lifecycle/stopServer.sh
```diff
@@ -153,6 +154,16 @@ else
   fi
 fi
 
+if [[ -n "${clusterName}" && "${keepReplicaConstant}" == 'false' ]]; then
+  # check if replica count can decrease below current value
+  isReplicaCountEqualToMinReplicas "${domainJson}" "${clusterName}" replicasEqualsMinReplicas
+  if [ "${replicasEqualsMinReplicas}" == 'true' ]; then
+    printInfo "Current replica count value is same as minimum number of replica count. \
```
tbarnes-us, line 161: Maybe provide a little more detail:

Not decreasing replica count: it is at its minimum. (See 'domain.spec.allowReplicasBelowMinDynClusterSize' and 'domain.status.clusters[].minimumReplicas' for details.)

ankedia: Changed the message in 6dae3f1. Please let me know if more details are needed.

ankedia:
Sorry, I had accidently deleted the branch which inadvertently closed the PR.

tbarnes-us (approved):

PR #2046, opened 2020-11-10
Update the startServer.sh copy functionality to only chmod owned files

rjeberhard:
This is part of making restrictive SCC work on OpenShift. In addition to the specific change to startServer.sh to only chmod when permitted, I noticed that many of the integration tests were initializing PV's with chown oracle:oracle rather than with oracle:root. I've updated those references too.

### DIFF
--- a/operator/src/main/resources/scripts/startServer.sh
+++ b/operator/src/main/resources/scripts/startServer.sh
@@ -34,8 +34,6 @@ function copyIfChanged() {
     trace "Copying '$1' to '$2'."
     cp $1 $2
     [ $? -ne 0 ] && trace SEVERE "failed cp $1 $2" && exitOrLoop
-    chmod 750 $2 
-    [ $? -ne 0 ] && trace SEVERE "failed chmod 750 $2" && exitOrLoop
   else
     trace "Skipping copy of '$1' to '$2' -- these files already match."
   fi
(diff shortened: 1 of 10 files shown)


bhavaniravichandran (commented):
LGTM

bhavaniravichandran (approved):

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -34,15 +34,17 @@ function copyIfChanged() {
     trace "Copying '$1' to '$2'."
     cp $1 $2
     [ $? -ne 0 ] && trace SEVERE "failed cp $1 $2" && exitOrLoop
-    chmod 750 $2 
-    [ $? -ne 0 ] && trace SEVERE "failed chmod 750 $2" && exitOrLoop
+    if [ -O "$2" ]; then
+      chmod 770 $2
+      [ $? -ne 0 ] && trace SEVERE "failed chmod 770 $2" && exitOrLoop
+    fi
```
tbarnes-us, line 40: What file isn't owned?  Note that there is similar logic in the liveness probe script.

rjeberhard: On OpenShift, the containers will not be run as the "oracle" user, but instead will be run as a new user (large UID), but that is part of the "root" group. In the case of the boot.properties, or any file that already exists and is being copied over, the file will still be owned by "oracle" and so the script cannot chmod the file.

rjeberhard: Yes, I'll change the liveness probe too

tbarnes-us: It's interesting that boot.properties is created and copied over by the operator at runtime (contains the WL credentials), but it still needs an unknown check and a chmod...

rjeberhard: It's not copied by the "operator". It's copied by the user principal inside the server instance running a script injected by the operator.

ankedia (approved):
LGTM

PR #2050, opened 2020-11-12
OWLS-86038: Helm upgrade error missing internalOperatorKey

lennyphan:
When using 'helm upgrade' command for the operator, the 'weblogic-operator-secrets' data field is nil (empty or null) if there are no key/value pairs defined in the 'data' dictionary.  In this specific scenario, the 'internalOperatorKey' key is not defined and so the 'data' field of the 'weblogic-operator-secrets' Secret resource has no defined key/value pairs hence it is defined as 'nil' (null or empty).  The 'index' function cannot operate on a 'nil' object.  The fix was to add a condition check for '$secret.data'.

When doing an upgrade, if either the key or certificate of the operator's internal SSL identity is missing then the operator will generate a new one when the operator is re-started.

### DIFF
--- a/kubernetes/charts/weblogic-operator/templates/_operator-cm.tpl
+++ b/kubernetes/charts/weblogic-operator/templates/_operator-cm.tpl
@@ -12,6 +12,13 @@ data:
   externalOperatorCert: {{ .externalOperatorCert | quote }}
     {{- end }}
   {{- end }}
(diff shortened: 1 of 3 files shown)


rjeberhard (approved):

alai8 (approved):

lennyphan:
Sorry about this, a little premature as I'm going to see if we can reuse the internal operator key if it already exists in the operator config map or to generate a new one.

PR #2062, opened 2020-11-18
Fix for mac os and validation error messages when no parameters are specified (OWLS-86060)

ankedia:
1. Fixed the python yaml module issue on Mac OS by using yq instead of python
2. Fix validation error messages for start/stop cluster and domain scripts when no parameters are specified as part of OWLS-86060.

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
@@ -78,9 +78,7 @@ initialize
 domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
 
 if [ -z "${domainJson}" ]; then
-  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
-    This script requires that the introspector job for the specified domain ran \
-    successfully and generated the domain resource. Exiting."
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. Exiting."
   exit 1
 fi
 
--- a/kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
@@ -67,13 +67,22 @@ function initialize {
 
   validateKubernetesCliAvailable
   validateJqAvailable
+  validateYqAvailable
+
   failIfValidationErrors
 }
 
 initialize
 
 # Get the domain in json format
-domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json)
+domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
+
+if [ -z "${domainJson}" ]; then
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
+    This script requires that the introspector job for the specified domain ran \
+    successfully and generated the domain resource. Exiting."
+  exit 1
+fi
 
 getDomainPolicy "${domainJson}" serverStartPolicy
 
--- a/kubernetes/samples/scripts/domain-lifecycle/stopDomain.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/stopDomain.sh
@@ -76,9 +76,7 @@ initialize
 domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
 
 if [ -z "${domainJson}" ]; then
-  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
-    This script requires that the introspector job for the specified domain ran \
-    successfully and generated the domain resource. Exiting."
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. Exiting."
   exit 1
 fi
 
--- a/kubernetes/samples/scripts/domain-lifecycle/stopDomain.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/stopDomain.sh
@@ -66,13 +66,21 @@ function initialize {
 
   validateKubernetesCliAvailable
   validateJqAvailable
+  validateYqAvailable
   failIfValidationErrors
 }
 
 initialize
 
 # Get the domain in json format
-domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json)
+domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
+
+if [ -z "${domainJson}" ]; then
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
+    This script requires that the introspector job for the specified domain ran \
+    successfully and generated the domain resource. Exiting."
+  exit 1
+fi
 
 getDomainPolicy "${domainJson}" serverStartPolicy
 
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -257,10 +257,7 @@ function getSortedListOfServers {
   local sortedServers=()
   local otherServers=()
 
-  getConfigMap "${domainUid}" "${domainNamespace}" configMap
-  topology=$(echo "${configMap}" | jq '.data["topology.yaml"]')
-  jsonTopology=$(python -c \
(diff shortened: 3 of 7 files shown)


kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
```diff
@@ -67,13 +67,22 @@ function initialize {
...
 initialize
 
 # Get the domain in json format
-domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json)
+domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
+
+if [ -z "${domainJson}" ]; then
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
+    This script requires that the introspector job for the specified domain ran \
+    successfully and generated the domain resource. Exiting."
```
tbarnes-us, line 83: the second sentence look like it should be removed ("This script..."), as the introspect job does not generate a domain resource...

tbarnes-us: (same problem  in 'stopDomain.sh' too)

kubernetes/samples/scripts/domain-lifecycle/startDomain.sh
```diff
@@ -67,13 +67,22 @@ function initialize {
...
 
 initialize
 
 # Get the domain in json format
-domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json)
+domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
+
+if [ -z "${domainJson}" ]; then
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
+    This script requires that the introspector job for the specified domain ran \
```
anpanigr, line 82: This information is not needed.  The first statement ( Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace} )  is sufficient convey the message.  Should we put internal detail about introspector job here ?

kubernetes/samples/scripts/domain-lifecycle/stopDomain.sh
```diff
@@ -66,13 +66,21 @@ function initialize {
...
 
 initialize
 
 # Get the domain in json format
-domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json)
+domainJson=$(${kubernetesCli} get domain ${domainUid} -n ${domainNamespace} -o json --ignore-not-found)
+
+if [ -z "${domainJson}" ]; then
+  printError "Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace}'. \
+    This script requires that the introspector job for the specified domain ran \
```
anpanigr, line 80: This information is not needed. The first statement ( Domain resource for domain '${domainUid}' not found in namespace '${domainNamespace} ) is sufficient convey the message. Should we put internal detail about introspector job here ?

ankedia: Done.

anpanigr (commented):
There are more places where we display wrong usage as follows ...
$  stopCluster.sh -c hello
[2020-11-19T03:24:55.685 UTC][ERROR] Domain config map 'sample-domain1-weblogic-domain-introspect-cm' not found. This script requires that the introspector job for the specified domain ran successfully and generated this config map. Exiting.
$  stopCluster.sh -c hello
[2020-11-19T03:41:57.871 UTC][ERROR] Domain config map 'sample-domain1-weblogic-domain-introspect-cm' not found. This script requires that the introspector job for the specified domain ran successfully and generated this config map. Exiting.

Instead we could display the message "There is no cluster with name 'hello' exist in domain resource 'sample-domain1' 
We should not display non complete sentence 'Exiting.'

anpanigr (commented):
$ startServer.sh -s myserver
Error from server (NotFound): namespaces "sample-domain1-ns" not found

Here the usage error does not help to diagnose the issue. We should display message "The server 'myserver' is not included in domain resource  'sample-domain1' "

ankedia:
> There are more places where we display wrong usage as follows ...
> $ stopCluster.sh -c hello
> [2020-11-19T03:41:57.871 UTC][ERROR] Domain config map 'sample-domain1-weblogic-domain-introspect-cm' not found. This script requires that the introspector job for the specified domain ran successfully and generated this config map. Exiting.
> 
> Instead we could display the message "There is no cluster with name 'hello' exist in domain resource 'sample-domain1'
> We should not display non complete sentence 'Exiting.'

This is expected behavior. The domain topology information is stored in the introspector config map. If we can't get the config map, we can't validate and know if ''hello' cluster exists in sample-domain1' or not.

ankedia:
> $ startServer.sh -s myserver
> Error from server (NotFound): namespaces "sample-domain1-ns" not found
> 
> Here the usage error does not help to diagnose the issue. We should display message "The server 'myserver' is not included in domain resource 'sample-domain1' " Moreover the error message format is not matching with other e.g. timestamp is missing.
> 
Pushed a change to display better error message with timestamp.

$ ./startServer.sh -s myserver
[2020-11-19T05:48:21.653 UTC][ERROR] Unable to get domain resource for domain 'sample-domain1' in namespace 'sample-domain1-ns'. Please make sure 'domain_uid' and 'namespace' provided with '-d' and '-n' arguments are correct.

As mentioned in previous comment, the domain topology information is stored in the introspector config map. If we can't get the config map, we can't validate and know if ''myserver" server exists in "sample-domain1" or not.

anpanigr:
> > There are more places where we display wrong usage as follows ...
> > $ stopCluster.sh -c hello
> > [2020-11-19T03:41:57.871 UTC][ERROR] Domain config map 'sample-domain1-weblogic-domain-introspect-cm' not found. This script requires that the introspector job for the specified domain ran successfully and generated this config map. Exiting.
> > Instead we could display the message "There is no cluster with name 'hello' exist in domain resource 'sample-domain1'
> > We should not display non complete sentence 'Exiting.'
> 
> This is expected behavior. The domain topology information is stored in the introspector config map. If we can't get the config map, we can't validate and know if ''hello' cluster exists in sample-domain1' or not.

[anpanigr] The process to find the corresponding config map is an internal to the script. The end user should able to resolve the issue simply by providing right cluster name. We do not need to display the internal detail.

ankedia:
> > This is expected behavior. The domain topology information is stored in the introspector config map. If we can't get the config map, we can't validate and know if ''hello' cluster exists in sample-domain1' or not.
> 
> [anpanigr] The process to find the corresponding config map is an internal to the script. The end user should able to resolve the issue simply by providing right cluster name. We do not need to display the internal detail.
[ANIL] We document the pre-requisite for the scripts that domain must be created and deployed before running the scripts. If domain is created and deployed then users will not see above message. Below is the note from documentation -

*Note**: Prior to running these scripts, you must have previously created and deployed the domain.

rjeberhard:
I agree that we don't want the scripts to discuss introspection except obliquely, such as something like "Domain topology information is only available after at least one server has started. If domain topology is not available then the caller must specify the cluster to which the server is a member, if appropriate."

ankedia:
> I agree that we don't want the scripts to discuss introspection except obliquely, such as something like "Domain topology information is only available after at least one server has started. If domain topology is not available then the caller must specify the cluster to which the server is a member, if appropriate."
[ANIL] I think in this case. the script is trying to start a server in the sample domain when the domain is not created and/or deployed. I can look into changing the error message as you suggested but I'm not sure if it's critical for this PR or not. Other suggestion was to *not* default the domainUid and namespace to the sample domains uid and namespace. Probably we can keep the JIRA OWLS-86060 open for now and make these changes in a separate PR. Please let me know your thoughts. Thanks.

rjeberhard (approved):

ankedia:
@anpanigr I have made a change to check for domain resource before trying to validate server/cluster name in domain. So if domain is not created/deployed, it will fail with "Unable to get domain resource" message instead of missing config map message. If you agree, please review and approve the PR. Thanks.

$ ./stopCluster.sh -c hello
[2020-11-19T18:43:35.500 UTC][ERROR] Unable to get domain resource for domain 'sample-domain1' in namespace 'sample-domain1-ns'. Please make sure 'domain_uid' and 'namespace' provided with '-d' and '-n' arguments are correct.

anpanigr (approved):
LGTM

Verified the usage for all scripts. Found there is no information about configmap or introspector in the error message.

ankedia:
> Please make sure the 'domain_uid' and 'namespace' values specified by the '-d' and '-n' arguments are correct.

Fixed in 3d36645

tbarnes-us (approved):

PR #2068, opened 2020-11-20
Add assertion to ItMiiDynamicUpdate test

alai8:
For  merging into dynamicupdate branch

Minor update to ItMiiDynamicUpdate test to also verify that servers did not roll during the dynamic update.


jshum2479 (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiDynamicUpdate.java
```diff
@@ -227,6 +233,9 @@ public void testMiiAddWorkManager() {
                         MANAGED_SERVER_NAME_BASE + "1",
                         workManagerName, "200"));
     logger.info("Found new work manager configuration");
+
```
tbarnes-us, line 236: Can the managed server timestamps be verified too?

Also, ideally this code should also verify that the 'weblogic.introspectVersion' label on each pod has also been updated.  Note that right now there's a wide race where the label may not be updated until several seconds after the online update has propagated - it waits until the next make-right. Johnny is working on a fix, but even when fixed there will still be a race - just less of a race (maybe less than a second?).

tbarnes-us (approved):

PR #2072, opened 2020-11-24
OWLS-77957 - Populate server and domain related env variables to init containers

ankedia:
Populate the same set of environment variables that are populated in WebLogic server container to init containters.


ankedia:
Getting some unit test failures, I will investigate and reopen PR.

alai8:
Please also add new unit tests for this change. Thanks

PR #2077, opened 2020-11-30
OWLS-77957 - Populate server and domain related env variables to init containers

ankedia:
Populate the same set of environment variables that are populated in WebLogic server container to init containters. Added unit test for new behavior. 
Integration test url - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3194/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -590,12 +590,18 @@ public abstract class PodStepContext extends BasePodStepContext {
     V1PodSpec podSpec = createPodSpec(tuningParameters)
         .readinessGates(getReadinessGates())
         .initContainers(getServerSpec().getInitContainers().stream()
-                .map(c -> c.env(getEnvironmentVariables(tuningParameters))).collect(Collectors.toList()));
+                .map(c -> c.env(createEnv(c, tuningParameters))).collect(Collectors.toList()));
 
     for (V1Volume additionalVolume : getVolumes(getDomainUid())) {
       podSpec.addVolumesItem(additionalVolume);
     }
     return podSpec;
+  }
+
+  private List<V1EnvVar> createEnv(V1Container c, TuningParameters tuningParameters) {
+    List<V1EnvVar> envVars = Optional.ofNullable(c.getEnv()).orElse(new ArrayList<>());
+    envVars.addAll(getEnvironmentVariables(tuningParameters));
+    return envVars;
   }
 
   // ---------------------- model methods ------------------------------
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.json.Json;
 import javax.json.JsonPatchBuilder;
 
@@ -588,7 +589,8 @@ public abstract class PodStepContext extends BasePodStepContext {
   protected V1PodSpec createSpec(TuningParameters tuningParameters) {
     V1PodSpec podSpec = createPodSpec(tuningParameters)
         .readinessGates(getReadinessGates())
-        .initContainers(getServerSpec().getInitContainers());
+        .initContainers(getServerSpec().getInitContainers().stream()
+                .map(c -> c.env(getEnvironmentVariables(tuningParameters))).collect(Collectors.toList()));
 
     for (V1Volume additionalVolume : getVolumes(getDomainUid())) {
       podSpec.addVolumesItem(additionalVolume);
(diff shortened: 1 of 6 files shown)


alai8 (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -588,14 +589,22 @@ private static String formatHashLabel(String hash) {
...
 
     for (V1Volume additionalVolume : getVolumes(getDomainUid())) {
       podSpec.addVolumesItem(additionalVolume);
     }
     return podSpec;
   }
 
+  private List<V1EnvVar> createEnv(V1Container c, TuningParameters tuningParameters) {
+    List<V1EnvVar> envVars = new ArrayList<>();
+    Optional.ofNullable(c.getEnv()).ifPresent(envVars::addAll);
```
alai8, line 603: should env var configued in initContainer take precedence over those configured in server or cluster?

ankedia: That's a good point. I had thought that the env var list will be mutually exclusive or env variables configured in server or cluster will take precedence. I'm not sure if customers would want to override  server/domain env variables in the init containers. @rjeberhard thoughts? Thanks.

rjeberhard: I guess that I would lean the other direction... That is, the highest precedence would be the most local configuration, which in this case, I think would be the environment variables specifically on the initContainer.

ankedia: I have made the change to set higher precedence for environment variables set in init containers in cad0c30. Please let me know if I missed something. Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -599,10 +599,11 @@ protected V1PodSpec createSpec(TuningParameters tuningParameters) {
...
 
   private List<V1EnvVar> createEnv(V1Container c, TuningParameters tuningParameters) {
-    List<V1EnvVar> envVars = new ArrayList<>();
-    Optional.ofNullable(c.getEnv()).ifPresent(envVars::addAll);
-    envVars.addAll(getEnvironmentVariables(tuningParameters));
-    return envVars;
+    List<V1EnvVar> initContainerEnvVars = new ArrayList<>();
+    Optional.ofNullable(c.getEnv()).ifPresent(initContainerEnvVars::addAll);
+    getEnvironmentVariables(tuningParameters).forEach(envVar ->
+            addDefaultEnvVarIfMissing(initContainerEnvVars, envVar.getName(), envVar.getValue()));
```
alai8, line 605: addDefaultEnvVarIfMissing does not handle V1EnvVar with ValueFrom.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
alai8: please add a test case for testing the precedence. Thanks

ankedia: Added `whenInitContainersHaveEnvVar_verifyInitContainersEnvVarTakesPrecedenceOverPreConfiguredEnvVar` test in 66f56c5

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
ankedia: As discussed, V1EnvVar with ValueFrom is already handled in getEnvironmentVariables(tuningParameters) method.  I have created a new method `addIfMissing` to add the env variable to list only if it's missing from the list. I also tested setting env var with ValueFrom in init containers and it's handled correctly.

alai8: can you also verify env var with ValueFrom in server spec are also set correctly in initContainer? Thanks

ankedia: I have pushed a change in 4646578 to handle ValueFrom in `addIfMissing` which will add the env variable with ValueFrom if env value is empty or null. I have also functionally tested that env var is set in initContainer. Thanks.

alai8 (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/Matchers.java
```diff
@@ -66,6 +78,39 @@ private static V1Container createContainer(String name, String image, String...
...
+    }
+    return false;
+  }
+
+  protected static void addIfMissing(List<V1EnvVar> vars, String name, String value) {
+    if (!listHasEnvVar(vars, name)) {
+      addEnvVar(vars, name, value);
+    }
+  }
+
```
doxiao, line 113: Those new methods (createXXX, listXXX and addXXX) do not seem to belong to this class.

doxiao (commented):
Looks good; just have one comment.

operator/src/test/java/oracle/kubernetes/operator/helpers/Matchers.java
ankedia: Thanks. These methods are used by the helper methods `hasInitContainer` and `hasInitContainerWithEnvVar` to verify if the init containers in the pod has appropriate pre-populated environment variables. Any suggestions on what would be a good place to move them?

doxiao: Now I see other helper methods in this class too.

doxiao (approved):

PR #2078, opened 2020-11-30
Fix domain lib

jshum2479:
This fix is for wdt archive and model containing domainLibraries which is not correctly handled in MII.

### DIFF
--- a/operator/src/main/resources/scripts/startServer.sh
+++ b/operator/src/main/resources/scripts/startServer.sh
@@ -232,8 +232,8 @@ function prepareMIIServer() {
 
         if [ -d ${WLSDEPLOY_DOMAINLIB} ] && [ "$(ls -A ${WLSDEPLOY_DOMAINLIB})" ] ; then
           mv ${WLSDEPLOY_DOMAINLIB}/* .
-          rm -fr wlsdeploy/
         fi
+        rm -fr wlsdeploy/
 
         # expand the archive apps and shared lib to the wlsdeploy/* directories
         # the config.xml is referencing them from that path
--- a/operator/src/main/resources/scripts/startServer.sh
+++ b/operator/src/main/resources/scripts/startServer.sh
@@ -215,15 +215,18 @@ function prepareMIIServer() {
     trace  SEVERE "Domain Source Type is FromModel, cannot create ${DOMAIN_HOME}/lib "
     return 1
   fi
+  local WLSDEPLOY_DOMAINLIB="wlsdeploy/domainLibraries"
 
   for file in $(sort_files ${IMG_ARCHIVES_ROOTDIR} "*.zip")
     do
         # expand the archive domain libraries to the domain lib
         cd ${DOMAIN_HOME}/lib || return 1
-        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/
+        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} ${WLSDEPLOY_DOMAINLIB}
         # Flatten the jars to the domain lib root
-        mv wlsdeploy/domainLibraries/* .
-        rm -fr wlsdeploy/
+        if [ -d ${WLSDEPLOY_DOMAINLIB} ] && [ "$(ls -A ${WLSDEPLOY_DOMAINLIB})" ] ; then
+          mv ${WLSDEPLOY_DOMAINLIB}/* .
+          rm -fr wlsdeploy/
+        fi
         if [ $? -ne 0 ] ; then
           trace SEVERE  "Domain Source Type is FromModel, error in extracting domain libs ${IMG_ARCHIVES_ROOTDIR}/${file}"
           return 1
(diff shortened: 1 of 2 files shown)


tbarnes-us:
(1) There are many changes in this review.  Did you intend to merge to 'master'?  Maybe it should be 'develop'?
(2) I looked at your commits and have two comment so far:
```
 # Flatten the jars to the domain lib root
        mv wlsdeploy/domainLibraries/* .
        rm -fr wlsdeploy/
```
What if 'wlsdeploy/domainLibraries' doesn't exist to star with?Do you want to

jshum2479:
Oops, did it again.  It should be develop instead

jshum2479:
> (1) There are many changes in this review. Did you intend to merge to 'master'? Maybe it should be 'develop'?
> (2) I looked at your commits and have two comment so far:
> 
> ```
>  # Flatten the jars to the domain lib root
>         mv wlsdeploy/domainLibraries/* .
>         rm -fr wlsdeploy/
> ```
> 
> Comment A) What if 'wlsdeploy/domainLibraries' doesn't exist to start with?
> Comment B) Should this code check to see if there are any unexpected files in the wlsdeploy dir once the mv is complete?

I am not sure if this is necessary,  the archive contains jar files in that location.  If they put something wrong over there and if it causes problem starting the server then they should see it in the server log.  Unless WDT has a command just to extract the libraries according to the model - it doesn't check much other than matching the path in the archive.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -221,7 +221,9 @@ function prepareMIIServer() {
         # expand the archive domain libraries to the domain lib
         cd ${DOMAIN_HOME}/lib || return 1
         ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/
-
+        # Flatten the jars to the domain lib root
+        mv wlsdeploy/domainLibraries/* .
+        rm -fr wlsdeploy/
```
tbarnes-us, line 226: >>  Should this code check to see if there are any unexpected files in the wlsdeploy dir once the mv is complete?

> I am not sure if this is necessary, the archive contains jar files in that location. If they put something wrong over there and if it causes problem starting the server then they should see it in the server log. Unless WDT has a command just to extract the libraries according to the model - it doesn't check much other than matching the path in the archive.

If the domain lib files aren't normally in an image archive, this will always generate an error message to stderr - which is confusing to see in a WL server pod log file.

If it's an error not to have domain lib files in an image archive, and the files are missing, then (a) the error will be a relatively hidden and cryptic 'wlsdeploy/domainLibraries/* not found' error on stderr in a WL pod log - not an ERROR trace message, and (b) since 'set -e' isn't on, the error will be ignored.  

So if  problem is possible but it's not always expected to have these files, then it'd be helpful to skip the flatten if there are no files in the directory.  Or if the the files must always be there, then the introspector job should verify that it exists and emit a helpful error before the WL pods start. or  to have the introspector check for this - it's harder to diagnose issues that are generated at the WL Server level...

jshum2479: I agreed is not very common to use domain/lib,  at first I thought you are referring to checking what type of files are in the directory and whether they are valid jar files.  I have added check for directory existence before moving the files.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -215,13 +215,18 @@ function prepareMIIServer() {
...
         cd ${DOMAIN_HOME}/lib || return 1
-        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/
-
+        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} ${WLSDEPLOY_DOMAINLIB}
+        # Flatten the jars to the domain lib root
+        if [ -d ${WLSDEPLOY_DOMAINLIB} ] && [ "$(ls -A ${WLSDEPLOY_DOMAINLIB})" ] ; then
+          mv ${WLSDEPLOY_DOMAINLIB}/* .
+          rm -fr wlsdeploy/
+        fi
         if [ $? -ne 0 ] ; then
```
tbarnes-us, line 230: It looks like the if check at line 230 belongs directly after line 224.

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -215,13 +215,18 @@ function prepareMIIServer() {
...
     do
         # expand the archive domain libraries to the domain lib
         cd ${DOMAIN_HOME}/lib || return 1
-        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/
-
+        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} ${WLSDEPLOY_DOMAINLIB}
+        # Flatten the jars to the domain lib root
+        if [ -d ${WLSDEPLOY_DOMAINLIB} ] && [ "$(ls -A ${WLSDEPLOY_DOMAINLIB})" ] ; then
+          mv ${WLSDEPLOY_DOMAINLIB}/* .
+          rm -fr wlsdeploy/
```
tbarnes-us, line 228: should this call rm -fr regardless of whether the domain lib was found in the wlsdeploy directory?

operator/src/main/resources/scripts/startServer.sh
tbarnes-us: Looks good.

tbarnes-us (approved):

rjeberhard (approved):

PR #2084, opened 2020-12-03
doc change - add note on shutting down adminServer

alai8:
Add a note on managed servers may not be able to start up while the admin server is shut down.

### DIFF
--- a/docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
+++ b/docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
@@ -155,13 +155,12 @@ To shut down a specific standalone server, add it to the Domain and set its `ser
 ```
 {{% notice note %}}
 The Administration Server can be shut down by setting the `serverStartPolicy` of the `adminServer` to `NEVER`.
-Care should be taken when shutting down the Administration Server, as it may not be able to start up 
-any Managed Servers in the WebLogic domain while the Administration Server is not available.
-A Managed Server will try to start up in 
+Care should be taken when shutting down the Administration Server. If a Managed Server cannot connect 
+to the Administration Server during startup, it will try to start up in
 [*Managed Server Independence (MSI)* mode](https://docs.oracle.com/en/middleware/fusion-middleware/weblogic-server/12.2.1.4/start/failures.html#GUID-CA4696B6-B462-4FD8-92A9-F27DEA8A2E87)
-if it cannot connect to the Administration Server during startup, but this could fail due to reasons such as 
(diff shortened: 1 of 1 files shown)


mriccell (approved):

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -153,6 +153,17 @@ To shut down a specific standalone server, add it to the Domain and set its `ser
       serverStartPolicy: "NEVER"
     ...
 ```
+{{% notice note %}}
+The Administration Server can be shut down by setting the `serverStartPolicy` of the `adminServer` to `NEVER`.
+Care should be taken when shutting down the Administration Server, as it may not be able to start up
```
rosemarymarano, line 158: I suggest changing the wording of the note to this, "Care should be taken when shutting down the Administration Server. When not available, the Administration Server cannot start up Managed Servers in the WebLogic domain. If it cannot connect to the Administration Server during startup, a Managed Server will try to start up in Managed Server Independence (MSI) mode but this could fail due to reasons such as no accessible Authentication Provider from the Managed Server pod."

rosemarymarano (commented):
I suggested a re-wording of the note. It's still a bit redundant but I think it's ok and clear (and correct) in its meaning. Do you agree?

docs-source/content/userguide/managing-domains/domain-lifecycle/startup.md
rosemarymarano: Also, do keep the cross-references to the WLS docs.

alai8: Thanks for the suggestion! 

However, this sentence "When not available, the Administration Server cannot start up Managed Servers in the WebLogic domain." is not entirely correct. It gives the impression that the managed servers are started by the admin server which is not right. It is also possible for the managed servers to start up successfully in some cases, such as if the domain is configured to use some external authentication provider that can be accessed by the managed server pods.

What do you think if we remove that sentence and update to: "Care should be taken when shutting down the Administration Server. If a Managed Server cannot connect to the Administration Server during startup, it will try to start up in Managed Server Independence (MSI) mode but this could fail due to reasons such as no accessible Authentication Provider from the Managed Server pod."

Thanks!

rosemarymarano: Yes, perfect.

rosemarymarano (approved):

PR #2089, opened 2020-12-07
when the name in label or annotations containing slash, it needs to b…

jshum2479:
This is to fix the patch string for annotations/label if it contains slash

https://jira.oraclecorp.com/jira/browse/OWLS-86404

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
@@ -75,11 +75,16 @@ public class KubernetesUtils {
         String basePath,
         Map<String, String> current,
         Map<String, String> required) {
+
     for (String name : required.keySet()) {
+      // We must encode each '/' and '~' in a JSON patch token using '~1' and '~0', otherwise
+      // the JSON patch will incorrectly treat '/' and '~' as special delimiters. (RFC 6901).
+      // The resulting patched JSON will have '/' and '~' within the token (not ~0 or ~1).
+      String encodedPath = basePath + name.replace("~","~0").replace("/","~1");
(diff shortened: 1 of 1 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -77,9 +77,9 @@ static void addPatches(
         Map<String, String> required) {
     for (String name : required.keySet()) {
```
tbarnes-us, line 78: The replacement below is subtle - it took me a bit to figure out (a) why it was needed,  (b) what was so special about `~1`.    This is explained in https://tools.ietf.org/html/rfc6901 - which mentions the need to replace '~' in addition to the need to replace '/'.

(1) Should this fix also take the opportunity to replace all '~' with '~0'? (I vote yes.)

(2) Please consider adding a comment that's something like:
// We must encode each '/' and '~' in a JSON patch token using '~1' and '~0', otherwise
// the JSON patch will incorrectly treat '/' and '~' as delimiters between tokens. (RFC 6901).
// The resulting patched JSON will have '/' and '~' within the token (not ~0 or ~1).

tbarnes-us: (3) Minor code tweak to consolidate the logic and comment together, perhaps introduce a local variable - something like `String encodedName=name.replace("~","~0").replace("/","~1");`

jshum2479: Good idea, we can do all at once too

tbarnes-us (approved):

rjeberhard (approved):

PR #2093, opened 2020-12-09
OWLS-86407 - Fixed the logic to start clustered managed server pods when admin server pod not running

ankedia:
Fixed the logic to start clustered managed server pods when admin server pod is not running. The previous logic was assuming that admin server pod is scheduled and ready when starting managed server pods.   Added a unit test for the new behavior. 

Integration test results at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3305

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
@@ -187,10 +187,10 @@ public class ManagedServerUpIteratorStep extends Step {
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return (ignoreConcurrencyLimits() || numPodsNotReady(numReady) < this.maxConcurrency);
+      return (ignoreConcurrencyLimits() || numNotReady(numReady) < this.maxConcurrency);
     }
 
-    private long numPodsNotReady(long numReady) {
+    private long numNotReady(long numReady) {
       return getNumServersStarted() - numReady;
     }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
@@ -24,10 +24,13 @@ import oracle.kubernetes.operator.helpers.PodHelper;
 import oracle.kubernetes.operator.helpers.ServiceHelper;
 import oracle.kubernetes.operator.logging.LoggingFacade;
 import oracle.kubernetes.operator.logging.LoggingFactory;
+import oracle.kubernetes.operator.wlsconfig.WlsDomainConfig;
 import oracle.kubernetes.operator.work.NextAction;
 import oracle.kubernetes.operator.work.Packet;
 import oracle.kubernetes.operator.work.Step;
 import oracle.kubernetes.weblogic.domain.model.Domain;
+
+import static oracle.kubernetes.operator.ProcessingConstants.DOMAIN_TOPOLOGY;
 
 /**
  * A step which will bring up the specified managed servers in parallel.
@@ -168,7 +171,7 @@ public class ManagedServerUpIteratorStep extends Step {
 
       if (startDetailsQueue.isEmpty()) {
         return doNext(new ManagedServerUpAfterStep(getNext()), packet);
-      } else if (hasServerAvailableToStart(packet.getSpi(DomainPresenceInfo.class))) {
+      } else if (hasServerAvailableToStart(packet)) {
         numStarted.getAndIncrement();
         return doForkJoin(this, packet, Collections.singletonList(startDetailsQueue.poll()));
       } else {
@@ -176,13 +179,15 @@ public class ManagedServerUpIteratorStep extends Step {
       }
     }
 
-    private boolean hasServerAvailableToStart(DomainPresenceInfo info) {
-      return ((numStarted.get() < info.getNumScheduledServers(clusterName))
-              && (canStartConcurrently(info.getNumReadyServers(clusterName))));
+    private boolean hasServerAvailableToStart(Packet packet) {
+      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+      String adminServerName = ((WlsDomainConfig) packet.get(DOMAIN_TOPOLOGY)).getAdminServerName();
+      return ((numStarted.get() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
+              && (canStartConcurrently(info.getNumReadyManagedServers(clusterName, adminServerName)))));
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady - 1)))
+      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady)))
           || (this.maxConcurrency == 0);
     }
   }
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
@@ -182,13 +182,21 @@ public class ManagedServerUpIteratorStep extends Step {
     private boolean hasServerAvailableToStart(Packet packet) {
       DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
       String adminServerName = ((WlsDomainConfig) packet.get(DOMAIN_TOPOLOGY)).getAdminServerName();
-      return ((numStarted.get() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
+      return ((numStartedManagedServers() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
               && (canStartConcurrently(info.getNumReadyManagedServers(clusterName, adminServerName)))));
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady)))
-          || (this.maxConcurrency == 0);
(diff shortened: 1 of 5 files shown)


alai8 (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainPresenceInfoTest.java
```diff
@@ -128,9 +128,9 @@ public void countScheduledServers() {
     addServer("MS5", null);
     addScheduledServer("MS6", null);
 
-    assertThat(info.getNumScheduledServers("cluster1"), equalTo(2L));
-    assertThat(info.getNumScheduledServers("cluster2"), equalTo(2L));
-    assertThat(info.getNumScheduledServers(null), equalTo(1L));
+    assertThat(info.getNumScheduledManagedServers("cluster1", "admin-server"), equalTo(2L));
```
russgold, line 131: This is really unclear. The old call seemed to indicate that we were getting the number of managed servers in the specified cluster. Why do we also pass in "admin-server"? The interface is problematic.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -168,21 +171,23 @@ public NextAction apply(Packet packet) {
...
+    private boolean hasServerAvailableToStart(Packet packet) {
+      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+      String adminServerName = ((WlsDomainConfig) packet.get(DOMAIN_TOPOLOGY)).getAdminServerName();
+      return ((numStarted.get() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
+              && (canStartConcurrently(info.getNumReadyManagedServers(clusterName, adminServerName)))));
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady - 1)))
+      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady)))
```
russgold, line 190: This would be a good time to do some refactoring. The expression is not so easy to read and understand.

I would start by extracting maxConcurrency == 0 into a method named 'ignoreConcurrencyLimits' and list that first in the expression, not last. Then you don't really need `this.maxConcurrency > 0` as that is implied (unless there is some support for negative values). 

The remaining part is still too complex and not obvious. So extract `numStarted.get()` into its own method, `getNumServersStarted` as there is no reason for all of the callers to know that there is an Atomic backing the value. Then can you explain what the rest of the logic is?

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -168,21 +171,23 @@ public NextAction apply(Packet packet) {
 
       if (startDetailsQueue.isEmpty()) {
         return doNext(new ManagedServerUpAfterStep(getNext()), packet);
-      } else if (hasServerAvailableToStart(packet.getSpi(DomainPresenceInfo.class))) {
+      } else if (hasServerAvailableToStart(packet)) {
```
russgold, line 174: good change. There was no need to expose the implementation as it had been.

russgold (changes requested):

anpanigr (approved):
I verified the following integration test usecases in domain-on-pv model

   Start a domain with with two managed servers in a dynamic cluster
   Stop the admin serer
   Stop managed serer2 
   Start the manged server2

I see manged server2 is coming up in the absence of Admin server.

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStepTest.java
```diff
@@ -249,6 +248,19 @@ public void whenConcurrencyLimitIs1_secondClusteredServerDoesNotStartIfFirstIsNo
     assertThat(getStartedManagedServers(), hasSize(1));
   }
```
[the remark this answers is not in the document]
ankedia: Changed the unit test method name and added new method `createDomainPresenceInfoWithNoAdminServer` as suggested.

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainPresenceInfoTest.java
ankedia: Previous call used to get both admin server (if scheduled) and managed servers that are scheduled. I was passing in the admin-server so that admin server can be filtered out from list and only scheduled managed servers are returned. My latest change adds a new method so original unit test stays the same.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
ankedia: Made the refactoring changes. The remaining logic is to check if the number of started servers is less than (maxConcurrency + ready managed servers for cluster "clusterName"). So if maxConcurrency is `1` and one server is started, it'll wait for that server to come to ready state before starting a new server. Please see https://github.com/oracle/weblogic-kubernetes-operator/pull/1855#issuecomment-671665520 for the pseudo code of this logic.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -182,13 +182,21 @@ public NextAction apply(Packet packet) {
...
-      return ((numStarted.get() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
+      return ((getNumServersStarted() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
               && (canStartConcurrently(info.getNumReadyManagedServers(clusterName, adminServerName)))));
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady)))
-          || (this.maxConcurrency == 0);
+      return (ignoreConcurrencyLimits() ||
+              (this.maxConcurrency > 0) && (getNumServersStarted() < (this.maxConcurrency + numReady)));
```
russgold, line 191: You should no longer need `(this.maxConcurrency > 0)` here, as you cannot reach this condition when it is 0. 

Can you explain the logic in the rest of the expression? Why can you start a new server when the number of servers started is less than the concurrency limit added to the number ready?

ankedia: ok, I'll make the change. As per OWLS-83431, we want to start one clustered server at a time and wait for it be scheduled before starting a new server in same cluster. So number of started servers in a cluster should be less than or equal to number of scheduled servers in that cluster. Initially when there are no servers scheduled, it'll start one server and wait for it be scheduled before starting second server. The scheduling happens very fast (few milli-seconds). Please see https://github.com/oracle/weblogic-kubernetes-operator/pull/1855#issuecomment-671665520 for more details.

russgold:
You've made the code change, but not answered the question. Can you explain the logic? It seems rather opaque to me. It is seriously problematic to have code in the system that we don't actually understand.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
russgold: I looked at the comment. The pseudo code doesn't explain the logic, either. 

It appears to me that the intent of `numStarted < maxConcurrency + numReady` is to look at the number of pods currently starting up (I think K8S calls that `waiting`) and compare that how many may be started at once. If so, the actual comparison should be:

`(numStarted - numReady) < maxConcurrency` or

`numWaiting < maxConcurrency`.

If so, I would extract a `numPodsWaiting` method:

```
// returns the number of pods currently starting up
private int numWaiting(int numReady) {
  return getNumStarted() - numReady;
}
```

and compare that to the maxConcurrency.

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainPresenceInfoTest.java
russgold: So the problem is that DomainPresenceInfo doesn't know which server is the admin server. I think we should address that, but not right now, perhaps, as I don't see an easy and obvious solution.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
ankedia: Yes, that's correct. I think k8s calls pods that are currently starting up as `not ready` or `inactive`. We can go with `waiting` if that clarifies the intention of code. I'll go ahead and push above change.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpIteratorStep.java
```diff
@@ -168,22 +171,35 @@ public NextAction apply(Packet packet) {
...
+      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+      String adminServerName = ((WlsDomainConfig) packet.get(DOMAIN_TOPOLOGY)).getAdminServerName();
+      return ((getNumServersStarted() <= info.getNumScheduledManagedServers(clusterName, adminServerName)
+              && (canStartConcurrently(info.getNumReadyManagedServers(clusterName, adminServerName)))));
     }
 
     private boolean canStartConcurrently(long numReady) {
-      return ((this.maxConcurrency > 0) && (numStarted.get() < (this.maxConcurrency + numReady - 1)))
-          || (this.maxConcurrency == 0);
+      return (ignoreConcurrencyLimits() || numPodsNotReady(numReady) < this.maxConcurrency);
```
russgold, line 190: Shouldn't the parentheses be around the expression on the right, not the entire condition? Does Intellij not highlight it for you?

russgold: Huh. Apparently not.

russgold (approved):
LGTM

PR #2094, opened 2020-12-10
Plugin updates to be compatible with latest JDK

rjeberhard:
Very minor updates to plugins to be compatible with JDK 15.0.1. When I updated the checkstyle plugin, this then required a few minor changes to Java files because of issues that were previously not detected.

### DIFF
--- a/operator/pom.xml
+++ b/operator/pom.xml
@@ -30,7 +30,7 @@
       <plugin>
         <groupId>pl.project13.maven</groupId>
         <artifactId>git-commit-id-plugin</artifactId>
-        <version>2.2.6</version>
+        <version>${git-commit-id-plugin-version}</version>
         <executions>
           <execution>
             <id>get-the-git-infos</id>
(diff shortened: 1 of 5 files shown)


operator/pom.xml
```diff
@@ -30,7 +30,7 @@
       <plugin>
         <groupId>pl.project13.maven</groupId>
         <artifactId>git-commit-id-plugin</artifactId>
-        <version>2.2.6</version>
+        <version>${git-commit-id-plugin-version}</version>
```
ddsharpe, line 33: nit - shouldn't this be in the plugin management section instead of properties?

ddsharpe (approved):

operator/pom.xml
rjeberhard: Honestly, not sure. I just found another version number that wasn't with the others and wanted to fix that. I'm happy to see if I can move this to plugin management.

ankedia (approved):

operator/pom.xml
rjeberhard: Going to punt on this for now.

PR #2099, opened 2020-12-15
Remove istio enablement for Operator namespace

anpanigr:
Recently after Upgrade to istio 1.7.3, we are observing intermittent failures in istio usecases when  the Operator container fails to comes up.  We have a tracking JIRA for this issue.   https://jira.oraclecorp.com/jira/browse/OWLS-86461

Since the current istio usecase does not using mTLS, there is no need enable istio for Operator Namespace. This may be considered as workaround till we resolve the issue OWLS-86461 or upgrade the istio version.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3353/  
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3354/


rjeberhard (changes requested):
I don't think that we should check-in a workaround like this. Instead, let's disable the test instead and then work on the plan proposed by Johnny.

PR #2100, opened 2020-12-15
Address locale and potential count issues with combining aplit values

russgold:
This addresses two issues:

1. In some environments, copying values split across multiple volumes was joining the pieces in the wrong order. Setting the LOCALE as part of the copy addresses that.

2. There was a concern that if we ever needed to split a file across more than ten volumes, the numbering would not result in a proper copy. That is addressed by using 0-padding with three digits. It seems very unlikely that we'd ever need more than 1000 config maps.


tbarnes-us (approved):
Looks fine to me

PR #2101, opened 2020-12-15
Shell unit test fix

russgold:
This addresses two issues:

1. In some environments, copying values split across multiple volumes was joining the pieces in the wrong order. Setting the LOCALE as part of the copy addresses that.

2. The shunit2 plugin now honors the `skip.unit.tests` property


tbarnes-us (approved):

anpanigr (approved):
I verified the following in my hosted Linux Box.
(a) unit test passed in the branch hell-unit-test-fix
(b) the -Dskip.unit.tests is honored

PR #2103, opened 2020-12-16
Support Kind 0.9.0 and testing with Kubernetes 1.19

rjeberhard:
This change updates the Kind-based test script and the accompanying properties file to support running with Kind 0.9.0. This version of Kind is the first version to support Kubernetes 1.19.

The release notes for Kind (https://github.com/kubernetes-sigs/kind/releases) state that there are different Docker images for the Kubernetes nodes that must be used with each version of Kind. Previously, we were always using the images associated with Kind 0.7.0, but this change updates the metadata so that we use the appropriate images with Kind 0.7.0, 0.8.0, or 0.9.0.

There are a few Kubernetes version deltas between the sets of images supported with each Kind release. For instance, Kind 0.7.0 and 0.8.0 support Kubernetes 1.15.11, but Kind 0.9.0 supports Kubernetes 1.15.12. Similarly, the earlier Kind releases support Kubernetes 1.16.9 while Kind 0.9.0 supports Kubernetes 1.16.15. Kind 0.9.0 is the only Kind version that supports Kubernetes 1.19 -- specifically, 1.19.1.

Based on these version changes, the default Kubernetes version is changed to "1.16", which will pick the correct 1.16.x based on the version of Kind used.

After this change is merged, I intend to change the Jenkins configuration for the Kind-based job to default to Kind 0.9.0, Kubernetes 1.16, and kubectl 1.16.15.

Finally, this change follows the Kind guidance to create a ConfigMap containing the details of the local registry. A follow-on change to the integration tests could read this ConfigMap to discover that there is a local registry rather than relying on the current set of environment variables.

Clean tests: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3371/


ankedia (approved):
LGTM

tbarnes-us (approved):
I'm not familiar with the mechanics of kind, but the overall bash, configmap, etc, all LGTM.

PR #2104, opened 2020-12-16
Parameterize the istio version based on environment varaiable

anpanigr:
Due to recent intermittent Jenkin failures in Istio Usecases, modify the integration test infra to install a given version of ISTIO based on a environment variable ISTIO_VERSION.  The default is v 1.7.3  that matches with Verrazzano usage.

  public static final String ISTIO_VERSION =
        Optional.ofNullable(System.getenv("ISTIO_VERSION")).orElse("1.7.3");

This PR gives a flexibility to run the integration test against a custom version of ISTIO which is being updated frequently.

https://build.weblogick8s.org:8443/job/wko-tmp/138/  ( v1.7.3 ) 
https://build.weblogick8s.org:8443/job/wko-tmp/139/  ( v1.7.6 )

None of the results are clean. We have similar failure in both runs.  
We have the JIRA  OWLS-86461 open to track the intermittent issue


jshum2479:
The change looks ok to me.

jshum2479 (approved):

rjeberhard (approved):

PR #2106, opened 2020-12-16
OWLS 86406: WKO - scaling with webhook invoked from Prometheus not working

lennyphan:
This PR contains the following changes:

1. Changes to scalingAction.sh for creating temporary per request python scripts for processing JSON results from K8s API requests.  These changes were for OFSS scaling.
2. Support in scalingAction.sh to use 'jq' to parse JSON when running on oraclelinux:8-slim, since python is not included.
3. Change to initialize-internal-operator-identity.sh to use openssl, instead of keytool, for generating private key and self-signed certificate for the operator's internal REST endpoint.  curl clients running on oraclelinux:8-slim would fail verification of the keytool generated self-signed certificate.


src/scripts/scaling/scalingAction.sh
```diff
@@ -2,24 +2,272 @@
...
+  echo "$num_ms"
+}
+
+# Get the current replica count for the WLS cluster if defined in the CRD's Cluster
+# configuration.  If WLS cluster is not defined in the CRD then return the Domain
+# scoped replica value, if present.  Returns replica count = 0 if no replica count found.
+# args:
+# $1 "True" if WLS cluster configuration defined in CRD, "False" otherwise
+# $2 Custom Resource Domain
+function get_replica_count() {
```
tbarnes-us, line 200: Should this take into account 'min replicas' when minimum replicas is enabled?

lennyphan:
Jenkins run on kind cluster for ItParameterizedDomain integration test as it's the only test that scales the cluster using WLDF and scalingAction.sh script:

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3361/

src/scripts/scaling/scalingAction.sh
```diff
@@ -2,24 +2,272 @@
...
+print(json.load(sys.stdin)["spec"]["version"])
+INPUT
+domain_api_version=`echo ${CRD} | python cmds-$$.py`
+  fi
+  echo "$domain_api_version"
+}
+
+# Retrieve Custom Resource Domain
+function get_custom_resource_domain() {
+  local DOMAIN=`curl -v --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" $kubernetes_master/apis/weblogic.oracle/$domain_api_version/namespaces/$wls_domain_namespace/domains/$domain_uid`
```
tbarnes-us, line 109: This is one of a few examples of long curl commands in this pull that is on a single line.  Can this line and the others be broken up into multiple lines in order to aid readability?

src/scripts/scaling/scalingAction.sh
```diff
@@ -2,24 +2,272 @@
 # Copyright (c) 2017, 2020, Oracle Corporation and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
-echo "called scalingAction.sh" >> scalingAction.log
+currdate=`date`
```
tbarnes-us, line 5: 1) Nice to have: See resources/scripts/utils.sh for a timestamp bash function that generates a timestamp in a format that exactly matches the operator logs timestamp format and the timestamp format we use in startServer.sh, etc.  (It's nice to have all of our major scripts use the same timestamp format.)
2) Nice to have: Consider defining a 'trace' function to replace your use of echo throughout. The trace can then automatically inject the timestamp...

russgold (commented):
General comment: now that we have the capability of writing unit tests for bash (see `operator/src/test/sh/modelInImageTest.sh` for an example), please consider adding unit tests for this changed functionality as well.

src/scripts/scaling/scalingAction.sh
lennyphan: Would you be ok if I just pull in the trace and timestamp functions since this is a standalone script, that's not deployed or run as part of the operator, where user's have to copy to use it either in a webhook or with WLDF to use  for scaling?

tbarnes-us: > Would you be ok if I just pull in the trace and timestamp functions since this is a standalone script, that's not deployed or run as part of the operator, where user's have to copy to use it either in a webhook or with WLDF to use for scaling?

Sure. That's definitely better IMO.

Feel free to copy the timestamp function. 

I have some doubt that you want to copy over the full trace function - something like the following would be fine IMO: 
```
function traceInfo() {
   echo "@[$(timestamp)][INFO]" "$@"
}
function traceError() {
   echo "@[$(timestamp)][ERROR]" "$@"
}
```

src/scripts/initialize-internal-operator-identity.sh
```diff
@@ -34,56 +34,24 @@ function generateInternalIdentity {
...
+    req \
+    -newkey rsa:2048 \
     -nodes \
-    -nocerts \
-    -out ${OP_KEY_PEM}
+    -keyout ${OP_KEY_PEM} \
+    -x509 \
+    -days ${DAYS_VALID} \
+    -out ${OP_CERT_PEM} \
+    -subj "/C=US/ST=CALIFORNIA/L=REDWOOD CITY/O=WebLogic/OU=Development/CN=weblogic-operator" \
```
rjeberhard, line 53: Ha, ha... Austin, TX. (Don't bother making that change -- I just found it funny that you chose California)

rjeberhard (approved):

rjeberhard:
@russgold, we probably need to have you give a brown-bag on the shell script unit-testing work that you did. This is a single script with (generally) a singular purpose and it is still fairly complex. It will likely be a significant effort to get it fully unit-tested and I'm not sure where we would start.

src/scripts/scaling/scalingAction.sh
lennyphan: Can you clarify where we should take 'min replicas' into account?  At the moment, we're just scaling up/down based on the currently configured replica count.  I see that the operator will enforce the 'min replicas' at runtime during ManagedServersUpStep.

lennyphan:
I've created the following JIRA for creating unit test for scalingAction.sh using the shell script unit-testing framework Russ implemented.

https://jira.oraclecorp.com/jira/browse/OWLS-86589

src/scripts/scaling/scalingAction.sh
tbarnes-us: 1 - If the 'true' replica count is 2, but this method returns 0, then a scale-up request will scale up to 1, and therefore it won't truly scale-up the cluster.  

2 - I haven't checked if this script already does this already, but I assume it'd be helpful if this script 'failed-fast' on an attempt to scale too low or too high, instead of indirectly relying on an administrator to figure out what went wrong.

The 'life cycle' scripts that Anil recently added have code for getting the min replica count (see getReplicaCount and getMinReplicas 'helper.sh').

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,291 @@ scaling_size=1
...
+  # convert nano to milli
+  local milli="`echo $timestamp | awk '{ print $2 }' | sed 's/\(^...\).*/\1/'`"
+  local secs_since_epoch="`echo $timestamp | awk '{ print $3 }'`"
+  local millis_since_opoch="${secs_since_epoch}${milli}"
+  local timezone="`echo $timestamp | awk '{ print $4 }'`"
+  echo "${ymdhms}.${milli} ${timezone}"
+}
+
+function trace() {
+  echo "@[$(timestamp)][INFO]" "$@"
```
tbarnes-us, line 36: It looks like every single trace is redirected to 'scalingAction.log', so the redirect can be centralized into this function.

Also, can the same log file be shared by multiple different clusters and domains?  If so, then it'd be helpful to decorate each trace with the current namespace, domain-uid, and cluster-name "[ns=xxx][uid=xxx][cl=xxx]".

src/scripts/scaling/scalingAction.sh
lennyphan: Added verification of scaling request count against configured minimum replica count for cluster.

src/scripts/scaling/scalingAction.sh
lennyphan: Great suggestion.  Updated the trace logging to include the above suggestions

src/scripts/scaling/scalingAction.sh
lennyphan: Also, scaling above configured max cluster size is verified by the REST backend implementation and exception (HTTP status code and message) returned to client.

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+
+# Retrieve the api version of the deployed Custom Resource Domain
+function get_domain_api_version() {
+  # Retrieve Custom Resource Definition for WebLogic domain
+  local CRD=$(curl \
+    -v \
+    --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+    -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+    -X GET \
+    $kubernetes_master/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/domains.weblogic.oracle)
```
rjeberhard, line 112: This URL path is wrong for K8s 1.16 and above as CRD's graduated from "v1beta1" to "v1". You could either determine the version of the k8s master or I have another idea. If you curl the "$kubernetes_master/apis" path, you get back a list of all API's supported by the server. This is equivalent to `kubectl api-versions`. The code could pick out the highest version of "weblogic.oracle" returned. For instance, on my system I get back "weblogic.oracle/v7" and "weblogic.oracle/v8".

rjeberhard: We need to test on k8s 1.16 -- it's likely that I'm wrong that this is broken now as 1.16 probably honors the "v1beta1" path. If so, that will be dropped in some k8s release -- I'll try and figure out which one.

rjeberhard: I found the answer... Your current code will work until k8s 1.22.

lennyphan: "v1beta1" is still honored as I'm running 1.18.12:

bash-4.2$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.12", GitCommit:"7cd5e9086de8ae25d6a1514d0c87bac67ca4a481", GitTreeState:"clean", BuildDate:"2020-11-12T09:18:55Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.12", GitCommit:"7cd5e9086de8ae25d6a1514d0c87bac67ca4a481", GitTreeState:"clean", BuildDate:"2020-11-12T09:11:15Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+# args:
+# $1 Managed server count
+# $2 Custom Resource Domain
+# $3 Cluster name
+function verify_minimum_ms_count_for_cluster() {
+  local new_ms="$1"
+  local domainJson="$2"
+  local clusterName="$3"
+  # check if replica count is less than minimum replicas
+  getMinReplicas "${domainJson}" "${clusterName}" minReplicas
```
tbarnes-us, line 297: I assume this should be "get_min_replicas".

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+      print j[index]["replicas"]
+INPUT
+  num_ms=`echo ${DOMAIN} | python cmds-$$.py`
+  fi
+
+  if [ "${num_ms}" == "null" ]; then
+    num_ms=0
+  fi
+
+  echo "$num_ms"
```
tbarnes-us, line 201: This should return 'min_replica_count' if min_replica_ count is higher than the discovered value.

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+for i in json.load(sys.stdin)["items"]:
+  print i["spec"]["replicas"]
+INPUT
+  num_ms=`echo ${DOMAIN} | python cmds-$$.py`
+  fi
+
+  if [ "${num_ms}" == "null" ]; then
+    num_ms=0
+  fi
+
```
tbarnes-us, line 221: This should return 'min_replica_count' if min_replica_ count is higher than the discovered value.

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+  local num_ms
+  if [ "$in_cluster_startup" == "True" ]
+  then
+    trace "$wls_cluster_name defined in clusters"
+    num_ms=$(get_num_ms_in_cluster "$DOMAIN")
+  else
+    trace "$wls_cluster_name NOT defined in clusters"
+    num_ms=$(get_num_ms_domain_scope "$DOMAIN")
+  fi
+  echo "$num_ms"
```
tbarnes-us, line 261: This should return min_replica_count if discovered value is less than the min

src/scripts/scaling/scalingAction.sh
tbarnes-us: LGTM

src/scripts/scaling/scalingAction.sh
```diff
@@ -16,10 +14,328 @@ scaling_size=1
...
+    # old shell versions don't support %N or --utc
+    timestamp="`date -u '+%Y-%m-%dT%H:%M:%S 000000 %s %Z' 2>&1`"
+  fi
+  local ymdhms="`echo $timestamp | awk '{ print $1 }'`"
+  # convert nano to milli
+  local milli="`echo $timestamp | awk '{ print $2 }' | sed 's/\(^...\).*/\1/'`"
+  local secs_since_epoch="`echo $timestamp | awk '{ print $3 }'`"
+  local millis_since_opoch="${secs_since_epoch}${milli}"
+  local timezone="`echo $timestamp | awk '{ print $4 }'`"
+  echo "${ymdhms}.${milli} ${timezone}"
```
tbarnes-us, line 32: Minor:  It looks like secs_since_epoch & millis_since_epoch are unused - leftover from an older iteration of the function.  They can be removed from here and from the same function in 'utils.sh'.

src/scripts/scaling/scalingAction.sh
tbarnes-us: LGTM

src/scripts/scaling/scalingAction.sh
lennyphan: Great catch!  Thank you

src/scripts/scaling/scalingAction.sh
lennyphan: This function is only called from get_replica_count() function so made the change in one place to return 'min_replica_count' if min_replica_ count is higher than the discovered value.

src/scripts/scaling/scalingAction.sh
lennyphan: This function is only called from get_replica_count() function so made the change in one place to return 'min_replica_count' if min_replica_ count is higher than the discovered value.

src/scripts/scaling/scalingAction.sh
lennyphan: Added code to return 'min_replica_count' if min_replica_ count is higher than the discovered value.

src/scripts/scaling/scalingAction.sh
lennyphan: Removed the unused local vars.

tbarnes-us (approved):

PR #2108, opened 2020-12-17
Operator version 3.1.1

rjeberhard:
This is a draft PR tracking the preparation for a 3.1.1 bugfix release.

Let's make sure now that we have all of the necessary "3.1.0" to "3.1.1" changes and that the description of this bug fix change in this release is sufficient.

The bug fix itself is not yet present in this PR.


rosemarymarano (approved):
LGTM 
(In a side note (and I imagine that you are aware of this), the documentation for prior releases of the operator stops at 2.6.0. It's a bummer but I think we need the most recent past docs much more so than the older ones.)

ankedia (approved):

mriccell (approved):

PR #2118, opened 2020-12-23
Retry behavior for synchronous calls during initialization

rjeberhard:
As you know, we've been seeing fairly regular integration test failures for the ItIstio* series of tests. The tests fail because the operator fails to start being unable to connect to the Kubernetes master. In each case I analyzed, the operator failed just before the Istio Envoy proxy finished initialization. 

Most operator calls are done using the async pattern, which has built-in delay and retry; however, the synchronous calls done during operator initialization do not have this functionality. Therefore, I've added a wrapper method that can do this retry. I've put it down at the CallBuilder level because this gave the easiest location to add the unit-test.

Creating this as draft so that you can take a look. I think I need to use some configured value for the retry delay rather than hardcoding 5 seconds.

Istio tests are clean with this change: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3454/console


operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -594,6 +600,40 @@ public VersionInfo readVersionCode() throws ApiException {
...
+   * @throws Exception Exception types other than ApiException, which will cause failure
+   */
+  public <T> T executeSynchronousCallWithRetry(Callable<T> call, int retryDelaySeconds) throws Exception {
+    T result = null;
+    boolean complete = false;
+    do {
+      try {
+        result = call.call();
+        complete = true;
+      } catch (RuntimeException re) {
```
jshum2479, line 618: What happened if it is a RuntimeException but not an ApiException?  Do we want to limit the number of retries or make it retry forever (hopefully the condition can be resolved)?

rjeberhard: I'll think about other exceptions... I didn't want to limit the number of retries here because the operator cannot go on until it can connect to the master and the operator will eventually be killed by the liveness probe.

jshum2479 (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -594,6 +600,40 @@ public VersionInfo readVersionCode() throws ApiException {
...
+        Throwable cause = re.getCause();
```
ddsharpe, line 619: Does the ApiException ever get nested further down?  Is there ever a case where you need to loop through the causes looking for ApiException?

operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -594,6 +600,40 @@ public VersionInfo readVersionCode() throws ApiException {
...
+        if (cause instanceof ApiException) {
+          LOGGER.warning(MessageKeys.EXCEPTION, cause);
+        }
+      } catch (Throwable t) {
+        LOGGER.warning(MessageKeys.EXCEPTION, t);
```
ddsharpe, line 624: Please add a comment here that we expect the liveness probe to cancel this process if it "retries forever".  The next coder may not understand the assumption/expectation.

ddsharpe (approved):

PR #2120, opened 2020-12-24
OWLS-86606 - Integration test for multiple pod restarts regression issue in OWLS-86552

ankedia:
Integration test to verify the rolling restart behavior in a domain with multiple WLS clusters and making sure that pods are restarted only once. This is to catch the regression issue reported by OFSS team in OWLS-86552. 

Test performs rolling restart of a MII domain with 2 clusters by changing the `spec.imagePullPolicy` to `Never`  (from IfNotPresent). It verifies that DomainChanged, DomainProcessingStarted and DomainProcessingCompleted events are generated. Additionally it verifies that managed server pods in both clusters are killed and started exactly once by checking pod events. 
Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3463/


alai8 (approved):

sankarpn (approved):
LGTM

doxiao (approved):

PR #2122, opened 2021-01-04
Update copyright statements for 2021 and adjust language

rjeberhard:
Normally, we would not need to bulk update the copyright statements for a new year. We would instead only adjust the copyright year on each file as it otherwise changed; however, since the copyright language has changed (removing "Corporation") it was easier to do this all in bulk.


rosemarymarano (approved):
I can certify that Corporation has been removed from the copyright statements. As for whether every file in which the copyright date needed to be updated from 2020 to 2021 was done, I have no way of knowing that.

PR #2124, opened 2021-01-05
OWLS-86816 - Fix to generate domain processing event only when there domain updates

ankedia:
OWLS-86816 - Operator generates domain processing events every time domain presence is rechecked even when there are no domain updates. This happens in cases where cluster's replica count is lower than the max cluster size. In such cases, `serversToStop` list in the `ManagedServersUpStep` contains servers that should not be started. Later in processing, the `ServerDown` step checks if the server pod is null and invokes `DeletePod` step only when the server pod is not null. 

This fix removes the servers that are already stopped (servers with null pod) from `serversToStop` list and it also checks whether server only needs it's service pre-created. In such cases, the domain processing events are not generated.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
@@ -74,28 +74,24 @@ public class ManagedServersUpStep extends Step {
     List<ServerShutdownInfo> serversToStop = getServersToStop(info, factory.shutdownInfos);
 
     if (!serversToStop.isEmpty()) {
-      if (isServiceOnlyServers(serversToStop)) {
-        steps.add(new ServerDownIteratorStep(factory.shutdownInfos, null));
-      } else {
-        insert(steps,
-                Step.chain(createProgressingStartedEventStep(info, MANAGED_SERVERS_STARTING_PROGRESS_REASON, true,
-                        null), new ServerDownIteratorStep(factory.shutdownInfos, null)));
-      }
+      insert(steps,
+              Step.chain(createProgressingStartedEventStep(info, MANAGED_SERVERS_STARTING_PROGRESS_REASON, true,
+                      null), new ServerDownIteratorStep(factory.shutdownInfos, null)));
     }
 
     return Step.chain(steps.toArray(new Step[0]));
   }
 
-  private static boolean isServiceOnlyServers(List<ServerShutdownInfo> serversToStop) {
-    List<ServerShutdownInfo> nonServiceOnlyServers = serversToStop.stream()
-            .filter(ssi -> !ssi.isServiceOnly()).collect(Collectors.toList());
-    return nonServiceOnlyServers.isEmpty();
-  }
-
   private static List<ServerShutdownInfo> getServersToStop(
           DomainPresenceInfo info, List<ServerShutdownInfo> shutdownInfos) {
     return shutdownInfos.stream()
-            .filter(ssi -> info.getServerPod(ssi.getServerName()) != null).collect(Collectors.toList());
+            .filter(ssi -> isNotAlreadyStoppedOrServiceOnly(info, ssi)).collect(Collectors.toList());
+  }
+
+  private static boolean isNotAlreadyStoppedOrServiceOnly(DomainPresenceInfo info, ServerShutdownInfo ssi) {
+    return (info.getServerPod(ssi.getServerName()) != null
+            && !info.isServerPodBeingDeleted(ssi.getServerName()))
+            || (ssi.isServiceOnly() && info.getServerService(ssi.getServerName()) == null);
   }
 
   private static Step createAvailableHookStep() {
--- a/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
@@ -87,9 +87,9 @@ public class ManagedServersUpStep extends Step {
   }
 
   private static boolean isServiceOnlyServers(List<ServerShutdownInfo> serversToStop) {
-    List<ServerShutdownInfo> list = serversToStop.stream()
+    List<ServerShutdownInfo> nonServiceOnlyServers = serversToStop.stream()
             .filter(ssi -> !ssi.isServiceOnly()).collect(Collectors.toList());
-    return list.isEmpty();
+    return nonServiceOnlyServers.isEmpty();
   }
 
   private static List<ServerShutdownInfo> getServersToStop(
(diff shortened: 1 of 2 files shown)


operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -74,18 +74,28 @@ private static Step scaleDownIfNecessary(
...
+        insert(steps,
+                Step.chain(createProgressingStartedEventStep(info, MANAGED_SERVERS_STARTING_PROGRESS_REASON, true,
+                        null), new ServerDownIteratorStep(factory.shutdownInfos, null)));
+      }
     }
 
     return Step.chain(steps.toArray(new Step[0]));
   }
 
+  private static boolean isServiceOnlyServers(List<ServerShutdownInfo> serversToStop) {
```
alai8, line 89: i think it would make the code more easy to read if this method is renamed to `isAllServersServiceOnly`

alai8:
Please don't forget to include unit test

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
rjeberhard: The logic makes sense to me, but I agree that the name needs to be longer and/or have a comment that describes the intention. I think this is looking for the condition where for all of the servers in the stop list we are only interested in deleting the service rather than deleting a pod. Is that right?

rjeberhard (approved):
Approving since my only comment is already covered by Anthony's review.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -74,18 +74,28 @@ private static Step scaleDownIfNecessary(
     List<ServerShutdownInfo> serversToStop = getServersToStop(info, factory.shutdownInfos);
 
     if (!serversToStop.isEmpty()) {
-      insert(steps,
-          Step.chain(createProgressingStartedEventStep(info, MANAGED_SERVERS_STARTING_PROGRESS_REASON, true,
-          null), new ServerDownIteratorStep(factory.shutdownInfos, null)));
+      if (isServiceOnlyServers(serversToStop)) {
```
doxiao, line 77: I am a little confused about the logic here. If the server pod is not null, the pod will be taken down no matter if the server is serviceOnly or not. If there are servers (pod or/and service) to be shutdown, the domain's progressing status needs to be updated, and the DomainProcessingStarting event should be generated. Only if there is no pod/service that needs to be taken down, the domain status update and event generation can be skipped.

ankedia:
> Please don't forget to include unit test

@alai8 @doxiao I'm trying to write a unit test for this. However, it seems domain processing event are being generated in two places (once in `DomainIntrospectorJobStep` and once in `ManagedServerUpStep`). So even if `ManagedServerUpStep` doesn't generate domain processing event, unit test still passes. Additionally, I'm seeing that even when same set of events (with same reason etc.) is generated twice, the `testSupport.getResources(KubernetesTestSupport.EVENT)` still return only one set of events. Not sure if I'm missing something. I'll keep looking into it but wanted to let you know.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
ankedia: > The logic makes sense to me, but I agree that the name needs to be longer and/or have a comment that describes the intention. I think this is looking for the condition where for all of the servers in the stop list we are only interested in deleting the service rather than deleting a pod. Is that right?

Yes. Additionally if precreateServerService option is set to `true` and service doesn't exist (for the first time), ManagedServerDown step creates the service with `isPreserveService` option set to `true` (instead of deleting the service and the pod). Please see below code snippet in `ServerDownIteratorStep.java`. Since I was only trying to address the issue of domain processing events being generated for every domain presence recheck, I decided not to change this logic. Please let me know if this needs to be changed.

```
  private StepAndPacket createManagedServerDownDetails(Packet packet, DomainPresenceInfo.ServerShutdownInfo ssi) {
    if (ssi.isServiceOnly()) {
      return new StepAndPacket(createServiceStep(ssi),
              createPacketForServer(packet, ssi));
    } else {
      return new StepAndPacket(new ServerDownStep(ssi.getName(), null), packet.clone());
    }
  }
```

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
ankedia: The service is deleted only if `isPreserveServices` option is not set to `true`. I believe when `precreateServerServices` is `true`, `isPreserveServices` option is also set to `true`. So in this case, the service will not be deleted and the whole step will be a `noop`. When `precreateServerServices` is `true` and service doesn't exist (which will be the case during first time), this step will create the service. Please see my comment above - https://github.com/oracle/weblogic-kubernetes-operator/pull/2124#discussion_r552255595

doxiao: I think we need to update the domain status and generate the event unless ServerDownIteratorStep is a noop.

doxiao:
> > Please don't forget to include unit test
> 
> @alai8 @doxiao I'm trying to write a unit test for this. However, it seems domain processing event are being generated in two places (once in `DomainIntrospectorJobStep` and once in `ManagedServerUpStep`). So even if `ManagedServerUpStep` doesn't generate domain processing event, unit test still passes. Additionally, I'm seeing that even when same set of events (with same reason etc.) is generated twice, the `testSupport.getResources(KubernetesTestSupport.EVENT)` still return only one set of events. Not sure if I'm missing something. I'll keep looking into it but wanted to let you know.
@ankedia The operator should not generate multiple Starting event in the same round of domain processing even though you see create event steps are put into the step chain in different parts of the workflow. There is a logic in EventHelper to skip duplicated createEventStep.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
ankedia: > i think it would make the code more easy to read if this method is renamed to `isAllServersServiceOnly`

I have refactored the code and removed the `isServiceOnlyServers` method in c26deeb. Please let me know if it's not clear.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
ankedia: Latest changes in c26deeb will generate the event unless the ServerDownIteratorStep is a noop. Thanks.

ankedia:
> > > Please don't forget to include unit test
> > @ankedia The operator should not generate multiple Starting event in the same round of domain processing even though you see create event steps are put into the step chain in different parts of the workflow. There is a logic in EventHelper to skip duplicated createEventStep.

Added unit tests in c26deeb.

alai8 (approved):

PR #2134, opened 2021-01-08
Owls-86939 - Generate domain processing completed event when scaling up a cluster

ankedia:
OWLS-86939 - The domain processing completed event is not generated during a scale up operation. This is because event type in the packet is not `domain processing starting` . The domain processing started event is generated during managed server pod creation which happens in a child fiber and packet in the child fiber is different from packet in the parent fiber. In ManagedServerIteratorUpStep, the packet is cloned and passed to each child fiber with additional info such as SERVER_NAME, SCAN etc.. Hence the `domain processing started` event type in not set in the parent packet. 

This proposed fix stores the event type/item in DomainPresenceInfo instead of storing it in the packet. I have added a unit test which reproduces the issue. Integration test results are at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3571/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
@@ -55,7 +55,7 @@ public class DomainPresenceInfo {
   private final ConcurrentMap<String, V1Service> clusters = new ConcurrentHashMap<>();
 
   private final List<String> validationWarnings = Collections.synchronizedList(new ArrayList<>());
-  private EventItem eventType;
+  private EventItem lastEventItem;
 
   /**
    * Create presence for a domain.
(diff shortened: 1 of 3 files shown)


rjeberhard:
This seems like a reasonable approach to me.

ankedia:
> This seems like a reasonable approach to me.

Thanks for the feedback. I have run the integration tests and manual tests in my environment with this change and the tests passed.

rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
```diff
@@ -54,6 +55,7 @@
   private final ConcurrentMap<String, V1Service> clusters = new ConcurrentHashMap<>();
 
   private final List<String> validationWarnings = Collections.synchronizedList(new ArrayList<>());
+  private EventItem eventType;
```
doxiao, line 58: This records the last EventType, so it would be clearer if we name the variable lastEventType.

doxiao (commented):
LGTM. just had one minor comment.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
ankedia: ok, I'll change the variable and method names. I'm thinking to change the variable name to `lastEventItem` since it's of type `EventItem`. Any thoughts? Thanks.

doxiao: that is fine.

doxiao (approved):

PR #2135, opened 2021-01-11
OWLS-86589 add unit test for scalingAction.sh

alai8:
Added new unit test under src directory for testing scalingAction.sh. This includes creating a new pom.xml in src directory with plugin for shunit2 testing. A few new unit tests have been added. 

Some changes were made to scalingAction.sh to support the unit tests, including redirecting stderr from jq and python commands to scalingAction.log file, and some minor fixes found during testing.

jenkins test result https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3590/testReport/


pom.xml
```diff
@@ -16,6 +16,7 @@
     <module>kubernetes</module>
     <module>operator-build-maven-plugin</module>
     <module>json-schema</module>
+    <module>src</module>
```
russgold, line 19: It makes me very uncomfortable to have a module called, "src."  I wonder if it is now time to rethink this directory. It does appear to be a module, but its purpose is not obvious.

russgold (commented):
Overall, it's looking very good - I would like to have a discussion over the issue I raised, though @rjeberhard

pom.xml
markxnelson: +1

src/scripts/scaling/scalingAction.sh
```diff
@@ -183,7 +194,8 @@ in_cluster_startup=`echo ${DOMAIN} | python cmds-$$.py`
 function get_num_ms_in_cluster() {
   local DOMAIN="$1"
   local num_ms
-  if [ -x "$(command -v jq)" ]; then
+  trace "zzz- wls_cluster_name is ${wls_cluster_name}"
```
lennyphan, line 197: Does this trace need to be cleaned up?

alai8: Yes, it will be cleaned up. This is WIP and I am still adding more tests and making changes to scalingAction.sh as needed.

pom.xml
rjeberhard: Agreed... I don't remember the history of how we created a top-level directory called "src", but it's time to move it to a more appropriate name. If we use `git mv` then git will preserve the history.

rjeberhard: It looks like everything in src is related to the operator's packaging, e.g. the operator's own scripts and tests of the same.

rjeberhard: What do people think about "packaging" or "assembly"?

russgold: Those names have meanings in the Maven world, so are probably not the best choice.

I'd want to understand, though, why this simply be moved under the operator module? The reasons for that might clarify the name.

rjeberhard: As we discussed, I'm going to look at moving the files under the operator module.

alai8:
will resubmit pull request after PR 2139 is merged

PR #2137, opened 2021-01-13
OWLS-87041 - Fix to start namespaces when namespaces are listed in multiple chunks

ankedia:
OWLS-87041 - OKD (Openshift) creates a large number of namespaces and domain is not started when namespaces are listed in multiple chunks. This is caused because the last chunk of namespaces (partial list) overwrites any previous chunks and only namespaces contained in the last chunk are started if they also happen to be part of the operator's target namespaces. 

This change collects the namespaces returned in all chunks in a bigger list so that all namespaces are started. Integration test link - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3595/ . There's a single test failure which is unrelated to this change.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
@@ -136,7 +136,7 @@ class DomainRecheck {
 
   private class NamespaceListResponseStep extends DefaultResponseStep<V1NamespaceList> {
 
-    private List<String> namespacesToStart = Collections.synchronizedList(new ArrayList<>());
+    private final List<String> namespacesToStart = Collections.synchronizedList(new ArrayList<>());
 
     private NamespaceListResponseStep() {
       super(new Namespaces.NamespaceListAfterStep(domainNamespaces));
@@ -164,11 +164,11 @@ class DomainRecheck {
       return doContinueListOrNext(callResponse, packet, createNextSteps(domainNamespaces));
     }
 
-    private Step createNextSteps(Set<String> namespacesToStartNow) {
+    private Step createNextSteps(Set<String> currentBatchOfNamespacesToStart) {
       List<Step> nextSteps = new ArrayList<>();
-      if (!namespacesToStartNow.isEmpty()) {
-        namespacesToStart.addAll(namespacesToStartNow);
-      }
+
+      // Add current batch of namespaces to start to the bigger list
+      namespacesToStart.addAll(currentBatchOfNamespacesToStart);
       if (!namespacesToStart.isEmpty()) {
         nextSteps.add(createStartNamespacesStep(namespacesToStart));
         if (Namespaces.getConfiguredDomainNamespaces() == null) {
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
@@ -136,6 +136,8 @@ class DomainRecheck {
 
   private class NamespaceListResponseStep extends DefaultResponseStep<V1NamespaceList> {
 
+    private List<String> namespacesToStart = Collections.synchronizedList(new ArrayList<>());
+
     private NamespaceListResponseStep() {
       super(new Namespaces.NamespaceListAfterStep(domainNamespaces));
     }
@@ -165,10 +167,13 @@ class DomainRecheck {
     private Step createNextSteps(Set<String> namespacesToStartNow) {
       List<Step> nextSteps = new ArrayList<>();
       if (!namespacesToStartNow.isEmpty()) {
-        nextSteps.add(createStartNamespacesStep(namespacesToStartNow));
+        namespacesToStart.addAll(namespacesToStartNow);
+      }
+      if (!namespacesToStart.isEmpty()) {
+        nextSteps.add(createStartNamespacesStep(namespacesToStart));
         if (Namespaces.getConfiguredDomainNamespaces() == null) {
           nextSteps.add(
-                RunInParallel.perNamespace(namespacesToStartNow, DomainRecheck.this::createNamespaceReview));
+                RunInParallel.perNamespace(namespacesToStart, DomainRecheck.this::createNamespaceReview));
         }
       }
       nextSteps.add(getNext());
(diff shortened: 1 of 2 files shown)


rjeberhard (approved):

doxiao (commented):
The change does resolve the problem. But I am wondering if it would have any potential scalability issues when there are a lot of namespace to start at the same time. If I understand the code correctly, the original logic starts namespaces in parallel in bunches. The change causes all of the namespaces to be start in parallel at the same time.

rjeberhard:
I think this is okay... The original code, if it had worked, would have indeed started the namespaces in bunches, but those bunches would have only been submitted milliseconds apart (the time it would take to complete the next list continuation call). Now, they are all submitted at once, but that submission is to put them all into the same asynchronous queue. Basically, if we ever hit a scaling problem where there were too many namespaces then I think that both the original code and Anil's proposed fix would have performed identically.

ankedia:
> The change does resolve the problem. But I am wondering if it would have any potential scalability issues when there are a lot of namespace to start at the same time. If I understand the code correctly, the original logic starts namespaces in parallel in bunches. The change causes all of the namespaces to be start in parallel at the same time.
[ANIL] My understanding of the original logic was that it was creating the steps to start namespaces in bunches (steps that are intended to be executed in parallel). However all those steps were executed only after the `NamespaceListResponse` step (and other related steps) finish execution i.e. after all the namespaces have been retrieved in different chunks. We can have short zoom call to sync up and make sure if this matches with your understanding. Thanks.

doxiao:
My understanding is that a subsequent bunch will not start to be processed until the previous bunch completes. If my understanding is correct, the behavior change will be noticeable.

ankedia:
The `onSuccess` method of `NamespaceListResponseStep`  returns  `doContinueListOrNext` next action and the next step (return value of `createNextSteps` method) is passed as a parameter to doContinueListOrNext.

```
    public NextAction onSuccess(Packet packet, CallResponse<V1NamespaceList> callResponse) {
      final Set<String> domainNamespaces = getNamespacesToStart(getNames(callResponse.getResult()));
      Namespaces.getFoundDomainNamespaces(packet).addAll(domainNamespaces);

      return doContinueListOrNext(callResponse, packet, createNextSteps(domainNamespaces));
    }
```
The `doContinueListOrNext` method reinvokes the `listNamespaces` call and doesn't return/execute next action (return value of createNextSteps) until `continue` flag is null (or callResponse is null). The `continue` flag will be null only when all namespaces have been retrieved and steps to start all those namespaces in parallel have been created.

```
protected final NextAction doContinueListOrNext(CallResponse<T> callResponse, Packet packet, Step next) {
    String cont = accessContinue(callResponse.getResult());
    if (callResponse != null && cont != null) {
      packet.put(CONTINUE, cont);
      // Since the continue value is present, invoking the original request will return
      // the next window of data.
      return resetRetryStrategyAndReinvokeRequest(packet);
    }
    return doNext(next, packet);
  }
```

doxiao:
My understanding is that each bunch has a different step to actually start the namespace in the bunch,  and the steps for all of the bunches will be chained together. The problem is that the steps for the bunches are not chained correctly.

doxiao:
I am okay with going ahead to get this change in because it does fix a critical issue, and the original logic was not working. We can look into the potential scalability issue in a different task.

doxiao (approved):
Can you create a task to track the potential scalability issue?

ankedia:
> Can you create a task to track the potential scalability issue?

Created OWLS-87091 to track this.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -136,6 +136,8 @@ Step readExistingNamespaces() {
 
   private class NamespaceListResponseStep extends DefaultResponseStep<V1NamespaceList> {
 
+    private List<String> namespacesToStart = Collections.synchronizedList(new ArrayList<>());
```
russgold, line 139: `namespacesToStart` should be final.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -165,10 +167,13 @@ public NextAction onSuccess(Packet packet, CallResponse<V1NamespaceList> callRes
     private Step createNextSteps(Set<String> namespacesToStartNow) {
       List<Step> nextSteps = new ArrayList<>();
       if (!namespacesToStartNow.isEmpty()) {
```
russgold, line 169: The distinction between `namespacesToStart` and `namespacesToStartNow` is not obvious. Can you describe the intent? It may be that changing a name or names will make it clearer.

Why test for `namespacesToStartNow` being not empty if all you're going to do is add its elements. If it's empty, the `addAll` call won't do anything.

russgold (commented):
Looks good overall; just looking for a bit better understandability.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
ankedia: Changed the name of `namespacesToStartNow` to `currentBatchOfNamespacesToStart` to clarify the intention and removed `isEmpty` test. The `namespacesToStartNow` name was used in previous logic and I neglected to change it. It contains the list of namespaces fetched in the current chunk (or batch).

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
ankedia: Changed `namespacesToStart` to final.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -164,11 +164,11 @@ public NextAction onSuccess(Packet packet, CallResponse<V1NamespaceList> callRes
...
     }
 
-    private Step createNextSteps(Set<String> namespacesToStartNow) {
+    private Step createNextSteps(Set<String> currentBatchOfNamespacesToStart) {
       List<Step> nextSteps = new ArrayList<>();
-      if (!namespacesToStartNow.isEmpty()) {
-        namespacesToStart.addAll(namespacesToStartNow);
-      }
+
+      // Add current batch of namespaces to start to the bigger list
```
russgold, line 170: Looks better - this comment is not needed.

(most inline comments indicate that something isn't clear - your changes make the comment not needed)

russgold (approved):

PR #2138, opened 2021-01-13
Separate authorization checks for the operator and domain namespaces

rjeberhard:
We have functionality in the HealthCheckHelper class that verifies that the operator has all of the necessary privileges in each namespace in order to do its work. This code had been previously cleaned-up, but in that rewrite an important distinction was lost.

Specifically, that the privileges needed in the operator's own namespace (as opposed to those needed in some domain namespace) are different and that if the operator's own namespace is also a domain namespace then both sets of privileges are required.

I noticed that most operator logs had a lot of WARNING statements from these authorization checks... The reason was that the HealthCheckHelper was looking for domain namespace privileges for the operator's namespace.

Finally, I cleaned-up just a little the set of roles needed for the operator's namespace.


lennyphan:
Are docs/charts/index.yaml and  docs/charts/weblogic-operator-3.2.0.tgz supposed to be part of the PR?

lennyphan (approved):
Other than my question on whether 'docs/charts/index.yaml' and 'docs/charts/weblogic-operator-3.2.0.tgz' should be part of the PR, the rest of the changes LGTM.

rjeberhard:
> Are docs/charts/index.yaml and docs/charts/weblogic-operator-3.2.0.tgz supposed to be part of the PR?
@lennyphan, yeah they are... I made a change to files under kubernetes/charts/weblogic-operator and this is sufficient to trigger the build of these files.

ankedia (approved):
I'm wondering if there should be a unit test added for operator-only namespace access (when `isDomainNamespace` is false). Otherwise it looks good to me.

PR #2139, opened 2021-01-13
Remove top-level src directory

rjeberhard:
Moved the contents of the top-level src directory under the existing operator directory and updated the various references.


alai8:
There is one reference in integration-tests/README.md that might also need to be updated:

`| `SKIP_CLEANUP`  | Test infrastructure removes all Kubernetes objects created during test execution. To retain all such objects for triaging test failures, this environment variable should be set to true. Users need to run `weblogic-kubernetes-operator/src/integration-tests/bash/cleanup.sh` to clean up Kubernetes objects after traiging.    | false`

rjeberhard:
Thanks, @alai8. I've updated that file.

alai8 (approved):

lennyphan (approved):
LGTM

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioDomainInImage.java
```diff
@@ -215,7 +215,7 @@ public void testIstioDomainHomeInImage() {
     assertTrue(checkConsole, "Failed to access WebLogic console");
     logger.info("WebLogic console is accessible");
 
-    Path archivePath = Paths.get(ITTESTS_DIR, "../src/integration-tests/apps/testwebapp.war");
+    Path archivePath = Paths.get(ITTESTS_DIR, "../operator/integration-tests/apps/testwebapp.war");
```
alai8, line 218: should testwebapp.war be moved to under integration-tests instead of under operator?

rjeberhard: Probably... I was assuming that there was some reason that these files were already in this odd directory. I don't know a lot about these integration-tests. They are separate from the tests in the integration-tests module and bash-based.

rjeberhard: I'm going to leave these alone for now and open a follow-on tracking issue. We're not supposed to have checked-in jars, ears, or wars. I see that other code in the main integration-tests module are building fairly similar applications. These pre-built applications need to be deleted and replaced with references to an application that we build.

PR #2143, opened 2021-01-15
Adding test for onNonDynamicChanges using CancelUpdate

vanajamukkara:
Adding a test to verify non dynamic change(changing DS JDBCDriverParams) using onNonDynamicChanges=CancelUpdate should not roll the domain and the domain status conditions is updated accordingly.

Jenkins results - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3617/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3616/


jshum2479 (approved):
LGTM,  The DomainCondition will be changed in the next few days, probably next week will have the final agreed terms and implementation, but it shouldn't be a big change.

PR #2146, opened 2021-01-19
Move tests to JUnit 5

russgold:
This moves virtually all of the unit tests for the operator module to JUnit 5. There are two exceptions:

- DomainTopologyTest - this test has a lot of dependencies on JUnit 4 assertions and should probably be cleaned up to follow the style that the rest of our tests do, with many small tests using Hamcrest assertions.

- RestTest - this depends on JerseyTest, which is not compatible with JUnit 5. A bug to that effect was filed three years ago, so it is possible that there has been a new release that we're not currently using.


integration-tests/pom.xml
```diff
@@ -1,4 +1,4 @@
-<!-- Copyright (c) 2020, 2021, Oracle and/or its affiliates.
+<!-- Copyright (c)  2020, Oracle Corporation and/or its affiliates.
```
rjeberhard, line 1: Need to undo this change to copyright statement.

operator/pom.xml
```diff
@@ -183,17 +183,6 @@
               <sourceDirectory>src/main/resources/scripts</sourceDirectory>
             </configuration>
           </execution>
-          <execution>
```
rjeberhard, line 186: What is the replacement for this test execution? Is this test still being run?

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -10,15 +10,13 @@
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
```
rjeberhard, line 13: You've accidentally backed out a real change. Looks like you can just revert this file.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -677,7 +677,7 @@ function generateMergedModel() {
   export __WLSDEPLOY_STORE_MODEL__="${NEW_MERGED_MODEL}"
 
   ${WDT_BINDIR}/validateModel.sh -oracle_home ${ORACLE_HOME} ${model_list} \
-    ${archive_list} ${variable_list}  -domain_type ${WDT_DOMAIN_TYPE}  > ${WDT_OUTPUT} 2>&1
```
rjeberhard, line 680: Same thing... This is accidentally backing out a real change.

operator/src/test/java/oracle/kubernetes/job/CreateDomainTest.java
```diff
@@ -1,67 +0,0 @@
-// Copyright (c) 2018, 2021, Oracle and/or its affiliates.
```
rjeberhard, line 1: I can't tell if this was an intentional deletion of a test or another accidental revert.

operator/src/test/java/oracle/kubernetes/operator/helpers/HealthCheckHelperTest.java
```diff
@@ -117,37 +116,22 @@ public void whenRulesReviewSupported_accessGrantedForEverything() {
...
     for (String ns : TARGET_NAMESPACES) {
-      V1SubjectRulesReviewStatus status = HealthCheckHelper.getSelfSubjectRulesReviewStatus(ns);
-      HealthCheckHelper.verifyAccess(status, ns, true);
+      HealthCheckHelper.getAccessAuthorizations(ns);
     }
 
     assertThat(logRecords, containsWarning(VERIFY_ACCESS_DENIED_WITH_NS));
   }
 
-  // HERE
```
rjeberhard, line 138: Whoops! I extraneous "HERE" message was checked-in, but otherwise this is another accidental revert of a real change.

rjeberhard (changes requested):
The vast majority of this change is great. Thanks for taking this on. Please undo the accidental reverts and it should be good to go.

integration-tests/pom.xml
russgold: Oops.

operator/pom.xml
russgold: Another oops. Somehow I wound up with an integration error. I thought I had caught and fixed them all. Clearly not.

operator/src/test/java/oracle/kubernetes/job/CreateDomainTest.java
russgold: This was intentional. It's been @Ignored for more than two years, so I figured we didn't really need it.

operator/src/test/java/oracle/kubernetes/operator/helpers/HealthCheckHelperTest.java
russgold: Think I have it fixed

rjeberhard (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/AdminPodHelperTest.java
```diff
@@ -340,7 +340,7 @@ public void whenAdminServerHasAdditionalVolumesWithReservedVariables_createAdmin
         .withAdditionalVolume("volume2", "/source-$(DOMAIN_NAME)");
 
     assertThat(
-        getCreatedPod().getSpec().getVolumes(),
+        Objects.requireNonNull(getCreatedPod().getSpec()).getVolumes(),
```
alai8, line 343: What is the advantage of using Objects.requireNonNull() here? Thanks

operator/src/test/java/oracle/kubernetes/operator/helpers/CallBuilderTest.java
```diff
@@ -10,7 +10,7 @@
 import java.util.function.Consumer;
 
 import com.google.gson.GsonBuilder;
-import com.meterware.pseudoserver.HttpUserAgentTest;
+import com.meterware.pseudoserver.PseudoServer;
```
alai8, line 13: where do we find out more about PseudoServer?

alai8 (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/AdminPodHelperTest.java
russgold: It avoids the superfluous warnings that method such as `getSpec()` are capable of returning null. That wrapper informs the compiler and code checker that _in this particular case_, we know it will not return null.

operator/src/test/java/oracle/kubernetes/operator/helpers/CallBuilderTest.java
russgold: It's part of HttpUnit. I hadn't remembered that I had put it in its own package hierarchy rather than <com.meterware.httpunit.pseudoserver>.

PR #2163, opened 2021-01-26
Added support to create 2 OKE clusters via terraform

marinakog:
Added support to create 2 OKE clusters via terraform. Supplied 2 testclasses to setup ISTIO based cross domain transaction test. 

Each test class create all needed WLS artifacts on each cluster. 
Istio support for MultiCluster environment will be added later (see OWLS-82415).  

These tests are excluded from kindtest.sh
see run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3774

and activated only if env var TWO_CLUSTERS != false, 
see Jenkins https://build.weblogick8s.org:8443/job/wko-oke-test/227/


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioCrossClusters.java
```diff
@@ -0,0 +1,467 @@
...
+import static oracle.weblogic.kubernetes.utils.ExecCommand.exec;
+import static oracle.weblogic.kubernetes.utils.ThreadSafeLogger.getLogger;
+import static org.awaitility.Awaitility.with;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+@DisplayName("Verify cross domain transaction is successful")
```
anpanigr, line 87: Verify Cross Cluster/Cross Domain transaction works

anpanigr (commented):
Why we need to ItClasses to verify cross domain translation ?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioCrossClusters.java
bhavaniravichandran: maybe say Cross Kubernetes Cluster to make it less ambiguous.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioCrossClusters.java
```diff
@@ -0,0 +1,467 @@
...
+    FileOutputStream out = new FileOutputStream(PROPS_TEMP_DIR + "/" + propFileName);
+    props.setProperty("NAMESPACE", domainNamespace);
+    props.setProperty("K8S_NODEPORT_HOST", host);
+    props.setProperty("DBPORT", Integer.toString(dbNodePort));
+    props.store(out, null);
+    out.close();
+  }
+
+  /*
+   * This test verifies cross domain transaction is successful. domain in image using wdt is used
```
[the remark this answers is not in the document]
bhavaniravichandran: the unique namespaces do not span clusters, correct?
The test uses Domain-in-image model (with WDT) to start a domain in a unique namespace in each of the  kubernetes clusters.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioCrossClusters.java
```diff
@@ -0,0 +1,467 @@
...
+
+    // Now that we got the namespaces for both the domains, we need to update the model properties
+    // file with the namespaces. For cross domain transaction to work, we need to have the externalDNSName
+    // set in the config file. Cannot set this after the domain is up since a server restart is
+    // required for this to take effect. So, copying the property file to RESULT_ROOT and updating the
+    // property file
+    updatePropertyFile();
+  }
+
+  private static void createDomainAndApps() {
```
bhavaniravichandran, line 188: This seems to start domain2. Can't see where you are starting domain1

bhavaniravichandran (commented):
have a couple of questions...

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIstioCrossClusters.java
marinakog: all related to domain1 is configured in ItIstioCrossClustersSetup.java

bhavaniravichandran: Sorry, missed seeing that file. Is it too difficult to combine both in 1 java file?

marinakog: yes it is :(,  to create objects in different clusters you need to use different KUBECONFIG files for k8s client apis, I can't make a switch it in one java program, we load apiClient static block to create Kubernetes object, so even I change env var it loads kubeconfig only once.

bhavaniravichandran: Fair enough. I figured KUBECONFIG would have to change, but wasn't sure  what else was the issue.

bhavaniravichandran (commented):
So, the test doesn't work as is right now, correct? This just sets up the domains in each of the k8s clusters.

marinakog:
> So, the test doesn't work as is right now, correct? This just sets up the domains in each of the k8s clusters.

yes, it has same test logic as original test, but curl invoke will fail, you will need to add extra Istio config to support multi clusters

bhavaniravichandran (approved):

anpanigr (approved):

PR #2169, opened 2021-01-28
OWLS-86750 - remove dependency on org.apache.commons.io from application_deployment.py

alai8:
Updated application_deployment.py test script to remove its usage of methods from Apache Commons IO package, which is causing InvalidJarIndexException when using linux-8-slim images. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3840/

Testing ItIntrospectVersion with change and with test modified to use deployUsingWlst instead of deployUsingRest:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3837/testReport/oracle.weblogic.kubernetes/ItIntrospectVersion/

### DIFF
--- a/integration-tests/src/test/resources/python-scripts/application_deployment.py
+++ b/integration-tests/src/test/resources/python-scripts/application_deployment.py
@@ -6,8 +6,10 @@ import os
 import traceback
 
 from java.util import Base64
-from org.apache.commons.io import FileUtils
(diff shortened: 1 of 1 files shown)


sankarpn (approved):
LGTM

rjeberhard (approved):

PR #2180, opened 2021-02-04
Owls87091 process namespaces in batches

doxiao:
This is for owls-87091; the jira number in the branch name is wrong. 

When the operators retrieving and processing the namespaces from the servers, it keeps the namespaces from the responses in batches instead of aggregating them into one big batch.

Note: the changes are only meant to complete the original logic before the fix for OWLS-87041, so that when the namespaces are retrieved from the K8S server in batches, namespaces in the same batch are started in parallel and the processing of the next batch of namespaces does not start until the previous batch completes. With this change, the operator still does not start processing any namespaces before all namespaces are retrieved.

Jenkins run with this branch had 2 failures: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3928.
Jenkins run with develop branch had 4 failures (including the 2 failures in the job above) : https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3929.


russgold (changes requested):
I'm finding it hard to follow the logic at this point, which makes it difficult to review - and I wrote much of this, originally, so I take much of the blame.

I would suggest trying to write out the logic in pseudo code now, and seeing how we can restructure the code to make it clearer. For example, it's hard to tell now the purpose of the condition:
```Namespaces.getConfiguredDomainNamespaces() == null```

doxiao:
@russgold I agree with you that the code logic is not very clear. But the changes themselves in this PR are very simple: 1) reverting the changes in https://github.com/oracle/weblogic-kubernetes-operator/pull/2137, so that we don't aggregate all namespaces into one big batch; 2) added code that define and use "current" to link steps in individual batches of namespaces into the step chain.

russgold (approved):

ankedia (approved):

PR #2183, opened 2021-02-09
default WLST_PROPERTIES 

tbarnes-us:
WLST_PROPERTIES will now include new env var WLST_EXTRA_PROPERTIES which in turn sets 'urandom'.

The new env var will be honored for life-cycle scripts that call 'wlst.sh', including the introspector job and start server script.

The idea is to speedup any WLST calls to 'encrypt' on low entropy systems - as is done by the introspector job when it encrypts override variables.


jshum2479 (approved):
LGTM

PR #2185, opened 2021-02-10
Owls87514 add refresh/retry to event replace operation and populate events maps on operator startup

doxiao:
This PR contains two features plus unit tests: 
1) leverage existing refresh-and-retry logic in AsyncRequestStep in replaceEvent operation;
2) populate the cached event objects with existing events on operator restart.

The integration test run in Jenkins https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/3990.


russgold (approved):
I'm approving this, as I don't see any actual problems, but I would like to make some process recommendations.

Based on the change history, it appears that you are making your changes and testing manually - and then adding unit tests. That defeats a large portion of the reason for unit testing. A better work cycle is to add a unit test that forces the change you want to make, verify that it fails, and then change the production code to make the test pass - and then go on to the next bit of functionality.

That ensures, first of all, that your tests are actually doing something significant, and distinguish between working code and non-working code. Obviously you don't need to do that to refactor, including renames.

I'd also recommend trying to have your changes happen on top of the latest develop, rather than merging develop into your branch. When you do the merge as you have, your changes are not in a sequence, which makes it harder to review them. Git provides a `rebase` feature to facilitate this. https://www.atlassian.com/git/tutorials/merging-vs-rebasing

rjeberhard (approved):

doxiao:
@russgold thanks for your suggestions.

PR #2188, opened 2021-02-10
OWLS-87753 - Fix to start all managed server pods when watch event notifications not received in jumbo k8s cluster

ankedia:
OWLS-87753- Fix for starting managed servers when watch event notifications are missed.

This change creates waiters for each managed server pod in ManagedServerUpIterator step. These fibers are started in parallel with fibers that schedule and actually start the managed server pods. The waiter fibers invoke WaitForReady step for the corresponding managed server. At the time when WaitFotReady is invoked, managed server pods would not have been created. Hence it passes resourceName to the WaitForReady step instead of the initial resource. The managed server pod state is updated periodically in the DomainPresenceInfo when the callback onSuccess is invoked (every 5 seconds) in waitForReady step. The createReadAndIfReadyCheckStep uses domain-id and namespace from the packet instead of getting it from the initial resource.

The link for integration test results is at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4024/ . There are 5 failures but they seem unrelated to this change.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
@@ -168,18 +168,16 @@ abstract class WaitForReadyStep<T> extends Step {
     fiber
         .createChildFiber()
         .start(
-            createReadAndIfReadyCheckStep(callback, packet),
+            createReadAndIfReadyCheckStep(callback),
             packet.clone(),
             null);
   }
 
-  private Step createReadAndIfReadyCheckStep(Callback callback, Packet packet) {
+  private Step createReadAndIfReadyCheckStep(Callback callback) {
     if (initialResource != null) {
       return createReadAsyncStep(getName(), getNamespace(), getDomainUid(), resumeIfReady(callback));
     } else {
-      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
-      return createReadAsyncStep(getName(), Optional.ofNullable(info).map(i -> i.getNamespace()).orElse(null),
-              Optional.ofNullable(info).map(i -> i.getDomainUid()).orElse(null), resumeIfReady(callback));
+      return new ReadAndIfReadyCheckStep(getName(), callback, getNext());
     }
   }
 
@@ -192,8 +190,7 @@ abstract class WaitForReadyStep<T> extends Step {
   }
 
   public String getName() {
-    return initialResource != null ? getMetadata(initialResource).getName() :
-            Optional.ofNullable(resourceName).orElse(null);
+    return initialResource != null ? getMetadata(initialResource).getName() : resourceName;
   }
 
   private DefaultResponseStep<T> resumeIfReady(Callback callback) {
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
```diff
@@ -160,13 +168,19 @@ private void checkUpdatedResource(Packet packet, AsyncFiber fiber, Callback call
...
   }
 
-  private Step createReadAndIfReadyCheckStep(Callback callback) {
-    return createReadAsyncStep(getName(), getNamespace(), getDomainUid(), resumeIfReady(callback));
+  private Step createReadAndIfReadyCheckStep(Callback callback, Packet packet) {
+    if (initialResource != null) {
+      return createReadAsyncStep(getName(), getNamespace(), getDomainUid(), resumeIfReady(callback));
+    } else {
+      DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+      return createReadAsyncStep(resourceName, Optional.ofNullable(info).map(i -> i.getNamespace()).orElse(null),
```
alai8, line 181: do we need to add check to ensure resourceName is not null?

alai8:
Will there be new tests added for this change? If not in this PR, please file a jira so we won't forget to add tests. Thanks

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
ankedia: The resourceName and initialResource are final variables and are initialized in the constructor. I'm assuming that either initialResource or resourceName will be non-null for the resource that needs to be checked. I can add a non-null check to be safe but I wasn't sure if that's going to be necessary. Thanks.

ankedia:
> Will there be new tests added for this change? If not in this PR, please file a jira so we won't forget to add tests. Thanks
Yes, I will file JIRA for adding unit and integration tests. It's a bit hard to unit test this change because the issue is caused by missed events but I'll look into it as time permits.

alai8 (approved):
Not sure how the latest change address the case if resourceName is null. But it is a minor issue.

lennyphan (approved):

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
```diff
@@ -160,13 +168,19 @@ private void checkUpdatedResource(Packet packet, AsyncFiber fiber, Callback call
...
         .start(
-            createReadAndIfReadyCheckStep(callback),
+            createReadAndIfReadyCheckStep(callback, packet),
             packet.clone(),
             null);
   }
 
-  private Step createReadAndIfReadyCheckStep(Callback callback) {
-    return createReadAsyncStep(getName(), getNamespace(), getDomainUid(), resumeIfReady(callback));
+  private Step createReadAndIfReadyCheckStep(Callback callback, Packet packet) {
```
rjeberhard, line 176: This doesn't follow our Step/Packet pattern. The Step is typically stateless or, at least, can be used for many different calls. The Packet is given on the Step.apply call. The logic in the else should happen in the Step.apply().

rjeberhard (changes requested):
Please move the behavior in the "else" in to a small Step class that returns the correct next Step from its apply() method. This way, you don't need to pass the Packet as a parameter.

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
ankedia: I have made the suggested change, please let me know if I missed something. Thanks.

rjeberhard (approved):

PR #2197, opened 2021-02-17
Operator 3.1.4

rjeberhard:
This release is coming quickly on the tail of 3.1.3, but the good news is that customers are using the tool and keeping us honest. Included are a fix from Anil and an update to a set of dependencies that were highlighted by a scan of CVE's.


rosemarymarano (approved):
Got this down to a science.

ankedia (commented):
LGTM. I'm wondering if we should wait for Fermin to confirm the fix in 3.1.4 branch before cutting the release. He was previously testing with 3.0.1 release and there was a slight change to the fix as well. In case, he finds other issues, we might want to include them in 3.1.4 release as well.

ankedia (approved):

PR #2198, opened 2021-02-17
Use Kubernetes Java Client 11.0.0

rjeberhard:
Next major version of the Kubernetes Java Client supporting K8s 1.19 symbols.

Biggest changes:
- The client changed "V1Event" to "CoreV1Event" as they restored a different set of symbols for the beta version of the Event symbols.
- Updated Dockerfile to use Java 15.0.2
- Updated Kind testing support for Kind 0.10 and to capture Kind logs (e.g. kubelet process) on failure
- Add hash annotation to our CRD similarly to what we do for Pods as I noticed that the operator was constantly spamming an updated version of the CRD when using the equals() algorithm.

Update -- two more significant changes:
- The Kubernetes Java Client tightened up their transitive dependencies, such as removing the dependency on Guava or Apache Commons. Rather than introduce those as direct dependencies, I looked at their changes and made similar changes so that we could also remove the dependencies.
- There is a bug in the Kubernetes client's Copy class on any method that copies local files to a Pod. This is identical with the "socket closed" issues we've seen with Exec. Since we only use this from test code, I've replaced the use of Copy with calls to `kubectl`.


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItStickySession.java
```diff
@@ -537,24 +536,24 @@ private void sendHttpRequestsToTestSessionStickinessAndVerify(String hostname, i
...
-        ingressServiceNodePort, webServiceGetUrl, " -b ");
+        getServerAndSessionInfoAndVerify(hostname,
+            ingressServiceNodePort, webServiceGetUrl, " -b ");
     // get server and session info from web service deployed on the cluster
     String serverName2 = httpDataInfo.get(serverNameAttr);
     String sessionId2 = httpDataInfo.get(sessionIdAttr);
     String countStr = httpDataInfo.get(countAttr);
-    int count = Optional.ofNullable(countStr).map(Ints::tryParse).orElse(0);
+    int count = Optional.ofNullable(countStr).map(Integer::valueOf).orElse(0);
     logger.info("Got the server {0}, session ID {1} and session state {2} "
```
alai8, line 546: Ints.tryParse() doesn't throw exception if the string is not a valid integer, but Integer.valueOf can throw NumberFormatException. Would that be an issue here?

alai8 (approved):

lennyphan (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItStickySession.java
rjeberhard: I don't think that it would be an issue as we would just see the test fail.

PR #2201, opened 2021-02-17
Check job condition to determine if job has failed in addition to checking job status

ankedia:
This PR evaluates the job condition in addition to job status to determine if the job has failed. In "etcd restore" test, when job remains in the Pending state for more than `ActiveDeadlineSeconds`, the job condition is changed to "failed" but the job status remains null. This causes the fiber to get stuck in the `WaitForJobReady` step.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
@@ -3,6 +3,7 @@
 
 package oracle.kubernetes.operator;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


alai8 (commented):
looks good. Don't forget to add test case to JobWatcherTest.

tbarnes-us:
The job timeout for a normally running job can be higher than 15 minutes (especially for the JRF case).  
Is this an issue?
If so, rather than choosing "15 minutes", perhaps do something like `domain.spec.configuration.introspectorJobActiveDeadlineSeconds` + 13 minutes.  (default is 2 minutes)

ankedia:
> looks good. Don't forget to add test case to JobWatcherTest.

Thanks. I have added test cases to JobWatcherTest in 4e59b9c .

alai8 (approved):

rjeberhard (approved):

PR #2204, opened 2021-02-18
minor change plus code cleanup and refactoring

doxiao:
Changes in this PR:
1) let operator continue, instead of fail, when it hits Unauthorized error on creating NamespaceWatchingStopped event in the domain's namespace. This only happens when a namespace is not managed by the operator any more, and the permissions that are required to create an event in that namespace have been removed; the operator already logs this error.
2) clean up some classes based on warnings that IntelliJ points out.
3) minor refactoring in DomainRecheck as a followup to a review comment to PR#2180. 

Integration test run on Jenkins (completed with 2 known failures): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4102/.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -526,17 +526,14 @@ public class DomainStatusUpdater {
       }
 
       private boolean stillHasPodPendingRestart(DomainStatus status) {
-        boolean found = false;
+        return status.getServers().stream()
+            .map(this::getServerPod)
+            .map(this::getLabels)
+            .anyMatch(m -> m.containsKey(LabelConstants.MII_UPDATED_RESTART_REQUIRED_LABEL));
+      }
 
-        for (ServerStatus serverStatus : status.getServers()) {
-          V1Pod pod = super.getInfo().getServerPod(serverStatus.getServerName());
-          Map<String, String> labels = getLabels(pod);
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -529,17 +530,22 @@ private boolean stillHasPodPendingRestart(DomainStatus status) {
```
russgold, line 530: How can you be certain that the pod will never be null?

There are a number of steps here, and I'd be inclined to use a stream to resolve them, such as:

```
      private boolean stillHasPodPendingRestart(DomainStatus status) {
        return status.getServers().stream()
              .map(this::getServerPod)
              .map(this::getLabels)
              .anyMatch(m -> m.containsKey(LabelConstants.MII_UPDATED_RESTART_REQUIRED_LABEL));
      }

      private V1Pod getServerPod(ServerStatus serverStatus) {
        return getInfo().getServerPod(serverStatus.getServerName());
      }
```

russgold (changes requested):
Thanks for doing some cleanup. There are a couple more things I think you should do here.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
doxiao: If the pod is null getLabels() will return an empty map, and the check for containsKey will be false.
But I like your suggestion

ankedia (approved):
LGTM

russgold (approved):

PR #2208, opened 2021-02-19
owls-87625 - Remove stranded old named node port service for the domain when deleting and replacing service

ankedia:
owls-87625 - Operator 3.0.1 version created external NodePort service with "-external" suffix. Operator 3.1.x creates the external NodePort service with shorter name with "-ext" suffix. This is causing an issue when upgrading the Operator from 3.0.1 to 3.1.1 and changing the `restartVersion`.  This is because the port for the external NodePort service created by Operator 3.1.1 is already in use by the NodePort service created by Operator 3.0.1 with the old name (with "-external" suffix).
This change removes stranded old named node port service during the service replace call by listing services of NodePort type created by the Operator for this particular domain. It deletes the node port service returned by the list call and then replaces it with the new name. Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4105/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
@@ -48,6 +48,7 @@ import org.apache.commons.lang3.builder.EqualsBuilder;
 import static oracle.kubernetes.operator.LabelConstants.forDomainUidSelector;
 import static oracle.kubernetes.operator.LabelConstants.getCreatedbyOperatorSelector;
 import static oracle.kubernetes.operator.helpers.KubernetesUtils.getDomainUidLabel;
+import static oracle.kubernetes.operator.helpers.OperatorServiceType.EXTERNAL;
 import static oracle.kubernetes.operator.logging.MessageKeys.ADMIN_SERVICE_CREATED;
 import static oracle.kubernetes.operator.logging.MessageKeys.ADMIN_SERVICE_EXISTS;
 import static oracle.kubernetes.operator.logging.MessageKeys.ADMIN_SERVICE_REPLACED;
@@ -534,7 +535,7 @@ public class ServiceHelper {
     protected abstract String getServiceCreatedMessageKey();
 
     private Step deleteAndReplaceService(Step next) {
-      if (getSpecType().equals(NODE_PORT_TYPE)) {
+      if (serviceType == EXTERNAL) {
         return deleteAndReplaceNodePortService();
       } else {
         V1DeleteOptions deleteOptions = new V1DeleteOptions();
@@ -824,7 +825,7 @@ public class ServiceHelper {
     private final String adminServerName;
 
     ExternalServiceStepContext(Step conflictStep, Packet packet) {
-      super(conflictStep, packet, OperatorServiceType.EXTERNAL);
+      super(conflictStep, packet, EXTERNAL);
       adminServerName = (String) packet.get(ProcessingConstants.SERVER_NAME);
     }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
@@ -4,16 +4,19 @@
 package oracle.kubernetes.operator.helpers;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -528,9 +535,33 @@ private Step createNewService(Step next) {
...
+                      getNamespace(),
+                      new ActionResponseStep<V1ServiceList>() {
+                      public Step createSuccessStep(V1ServiceList result, Step next) {
+                        Collection<V1Service> c = Optional.ofNullable(result).map(list -> list.getItems().stream()
+                                  .filter(s -> isNodePortService(s))
+                                  .collect(Collectors.toList())).orElse(new ArrayList<>());
+                        return new DeleteServiceListStep(c, createReplacementService(next));
+                      }
+
+                      private boolean isNodePortService(V1Service svc) {
```
doxiao, line 560: Can we use the existing isNodePortType method instead of introducing a new method?

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -528,9 +535,33 @@ private Step createNewService(Step next) {
...
+    private Step deleteAndReplaceNodePortService() {
       return new CallBuilder()
-          .deleteServiceAsync(
-              createServiceName(), getNamespace(), getDomainUid(), deleteOptions, new DeleteServiceResponse(next));
+              .withLabelSelectors(forDomainUidSelector(info.getDomainUid()), getCreatedbyOperatorSelector())
+              .listServiceAsync(
+                      getNamespace(),
+                      new ActionResponseStep<V1ServiceList>() {
+                      public Step createSuccessStep(V1ServiceList result, Step next) {
+                        Collection<V1Service> c = Optional.ofNullable(result).map(list -> list.getItems().stream()
```
doxiao, line 554: We are only supposed to have one of these service for a particular domain, so in theory we should check that and only delete and replace that one service.

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -528,9 +535,33 @@ private Step createNewService(Step next) {
 
     private Step deleteAndReplaceService(Step next) {
       V1DeleteOptions deleteOptions = new V1DeleteOptions();
```
doxiao, line 537: this is only used in the else clause, shall we move it there?

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
ankedia: Thanks. For some reason, I thought it was a no-arg method. I have made the suggested change.

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
doxiao: I was thinking if we should check if there is only one service that qualifies the checks in the end and only remove that one service just to be safe, so we don't accidentally remove and replace other services. But on the other hand, we don't know if we can do anything different if we do end up having multiple qualified services. I am fine with this then.

doxiao (approved):
LGTM.

operator/src/main/java/oracle/kubernetes/operator/helpers/ServiceHelper.java
```diff
@@ -527,10 +534,29 @@ private Step createNewService(Step next) {
     protected abstract String getServiceCreatedMessageKey();
 
     private Step deleteAndReplaceService(Step next) {
-      V1DeleteOptions deleteOptions = new V1DeleteOptions();
+      if (getSpecType().equals(NODE_PORT_TYPE)) {
```
alai8, line 537: Since we are replacing external type service, should we be checking serviceType == EXTERNAL here?

ankedia: I thought they the same thing since only external services are of type `NodePort`. I'll go ahead and make the change.

alai8 (approved):

PR #2222, opened 2021-02-24
Owls 87209 - Domain lifecycle scripts fix in cases where domain/cluster specs are overridden at the individual managed server level

ankedia:
Owls 87209 - Domain lifecycle scripts fix in cases where domain/cluster specs are overridden at the individual managed server level. The script doesn't work correctly when `.spec.managedServers` section has pre-existing details for a particular managed server. For e.g Consider the below stanza where the `serverStartState` attribute is overridden for soa_server1.
spec:
  managedServers:    
  - serverName: soa_server1
    serverStartState: RUNNING

This change adds the server start policy element for the managed server if the policy doesn't exist or replaces the policy if the policy already exists. Similarly, it unsets the policy and/or deletes the managed server element if managed server spec only has serverStartPolicy under it.

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -148,10 +148,7 @@ function createPatchJsonToUnsetPolicyAndUpdateReplica {
   local replicaPatch=$3
   local __result=$4
 
-  unsetCmd="(.spec.managedServers[] | select (.serverName == \"${serverName}\") | del (.serverStartPolicy))"
-  replacePolicyCmd=$(echo ${domainJson} | jq -cr "${unsetCmd}")
-  mapCmd=". |= map(if .serverName == \"${serverName}\" then . = ${replacePolicyCmd} else . end)"
-  serverStartPolicyPatch=$(echo ${domainJson} | jq "(.spec.managedServers)" | jq "${mapCmd}")
+  unsetServerStartPolicy "${domainJson}" "${serverName}" serverStartPolicyPatch
   patchJson="{\"spec\": {\"clusters\": "${replicaPatch}",\"managedServers\": "${serverStartPolicyPatch}"}}"
   eval $__result="'${patchJson}'"
 }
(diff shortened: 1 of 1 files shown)


kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -111,14 +111,20 @@ function createServerStartPolicyPatch {
...
-  if [ -z "${currentStartPolicy}" ]; then
+  managedServers=$(echo ${domainJson} | jq -cr '(.spec.managedServers)')
+  if [[ -z "${currentStartPolicy}" && "${managedServers}" == "null" ]]; then
     # Server start policy doesn't exist, add a new policy
     addPolicyCmd=".[.| length] |= . + {\"serverName\":\"${serverName}\", \
       \"serverStartPolicy\":\"${policy}\"}"
     serverStartPolicyPatch=$(echo ${domainJson} | jq .spec.managedServers | jq -c "${addPolicyCmd}")
+  elif [ "${managedServers}" != "null" ]; then
+    extractSpecCmd="(.spec.managedServers)"
+    mapCmd=". |= (map(.serverName) | index (\"${serverName}\")) as \$idx | if \$idx then .[\$idx][\"serverStartPolicy\"] = \"${policy}\" else .+  [{serverName: \"${serverName}\" , serverStartPolicy: \"${policy}\"}] end"
```
tbarnes-us, line 126: This is a long line. Something like the following would be easier to read:
```
mapCmd="\
  . |= (map(.serverName) | index (\"${serverName}\")) as \$idx | \
  if \$idx \
  then .[\$idx][\"serverStartPolicy\"] = \"${policy}\" \
  else .+  [{serverName: \"${serverName}\" , serverStartPolicy: \"${policy}\"}] end "
```
(I may have a syntax error in above - it's just intended as an example.)

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -199,13 +203,34 @@ function createPatchJsonToUnsetPolicy {
...
+
+  unsetCmd="(.spec.managedServers[] | select (.serverName == \"${serverName}\") | del (.serverStartPolicy))"
+  replacePolicyCmd=$(echo ${domainJson} | jq -cr "${unsetCmd}")
+  replacePolicyCmdLen=$(echo "${replacePolicyCmd}" | jq -e keys_unsorted | jq length)
+  if [ ${replacePolicyCmdLen} == 1 ]; then
+    mapCmd=". |= map(if .serverName == \"${serverName}\" then del(.) else . end)"
+  else
+    mapCmd=". |= map(if .serverName == \"${serverName}\" then . = ${replacePolicyCmd} else . end)"
+  fi
+  serverStartPolicyPatch=$(echo ${domainJson} | jq "(.spec.managedServers)" | jq "${mapCmd}")
```
tbarnes-us, line 230: It looks like the intent is for serverStartPolicyPatch to be a local variable, but it has the same name as the variable used for __result by the caller(s).  Maybe give it a different name and declare it with 'local'.

tbarnes-us:
General comment: IIRC, there are integration tests for these scripts. If so, please file a JIRA to enhance them to cover the new use case...

anpanigr (commented):
Can you enhance integration test (ItServerStartPolicy.java) to add this usecase.

ankedia:
> General comment: IIRC, there are integration tests for these scripts. If so, please file a JIRA to enhance them to cover the new use case...

Thanks. I have made the suggested changes and also created JIRA OWLS-87819 to track the enhancement of the integration tests to cover the new use case.

ankedia:
> Can you enhance integration test (ItServerStartPolicy.java) to add this usecase.
I have created JIRA OWLS-87819 to track the integration test enhancement and it can be done in a separate PR. Do we have any other integration tests where domain/cluster specs are overridden at the managed server level?

tbarnes-us:
> >Can you enhance integration test (ItServerStartPolicy.java) to add this usecase.
> I have created JIRA OWLS-87819 to track the integration test enhancement and it can be done in a separate PR. 
Thanks.
> Do we have any other integration tests where domain/cluster specs are overridden at the managed server level?
I don't know, Pani might.

tbarnes-us (approved):

rjeberhard (approved):

PR #2232, opened 2021-02-26
OWLS-87849 - scaleCluster.sh script to patch the 'replicas' attribute of the WLS cluster

ankedia:
OWLS-87849 - The `scaleCluster.sh` script is similar to the cluster stop/start scripts and it patches the replicas attribute for the cluster with the specified value. It performs validations to ensure that clusterName exists in domain topology and that the specified value of replica count is in the allowed range. If the cluster doesn't exist under `.spec.clusters` in the domain resource, it adds the cluster. 
The PR also changes the logic for the cluster stop/start scripts to add the cluster under `.spec.clusters` in the domain resource if it doesn't exist.

Integration test run URL - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4171/

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -271,7 +271,7 @@ function createPatchJsonToUpdateClusterPolicy {
 #
 # Function to create patch json to update cluster replicas
 # $1 - Domain resource in json format
-# $2 - Name of cluster whose policy will be patched
+# $2 - Name of cluster whose replicas will be patched
 # $3 - replica count
 # $4 - Return value containing patch json string
 #
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -223,6 +223,7 @@ function unsetServerStartPolicy {
   local serverName=$2
   local __result=$3
   local unsetStartPolicyPatch=""
+  local mapCmd=""
 
   unsetCmd="(.spec.managedServers[] | select (.serverName == \"${serverName}\") | del (.serverStartPolicy))"
   replacePolicyCmd=$(echo ${domainJson} | jq -cr "${unsetCmd}")
@@ -248,24 +249,29 @@ function createPatchJsonToUpdateClusterPolicy {
   local clusterName=$2
   local policy=$3
   local __result=$4
+  local addClusterStartPolicyCmd=""
+  local mapCmd=""
+  local existingClusters=""
+  local patchJsonVal=""
+  local startPolicyPatch=""
   
-  clusters=$(echo ${domainJson} | jq -cr '(.spec.clusters)')
-  if [ "${clusters}" == "null" ]; then
+  existingClusters=$(echo ${domainJson} | jq -cr '(.spec.clusters)')
+  if [ "${existingClusters}" == "null" ]; then
     # cluster doesn't exist, add cluster with server start policy
     addClusterStartPolicyCmd=".[.| length] |= . + {\"clusterName\":\"${clusterName}\", \
       \"serverStartPolicy\":\"${policy}\"}"
-    startPolicy=$(echo ${clusters} | jq -c "${addClusterStartPolicyCmd}")
+    startPolicyPatch=$(echo ${existingClusters} | jq -c "${addClusterStartPolicyCmd}")
   else
     mapCmd="\
       . |= (map(.clusterName) | index (\"${clusterName}\")) as \$idx | \
       if \$idx then \
       .[\$idx][\"serverStartPolicy\"] = \"${policy}\" \
       else .+  [{clusterName: \"${clusterName}\" , serverStartPolicy: \"${policy}\"}] end"
-    startPolicy=$(echo ${clusters} | jq "${mapCmd}")
+    startPolicyPatch=$(echo ${existingClusters} | jq "${mapCmd}")
   fi
 
-  patchJson="{\"spec\": {\"clusters\": "${startPolicy}"}}"
-  eval $__result="'${patchJson}'"
+  patchJsonVal="{\"spec\": {\"clusters\": "${startPolicyPatch}"}}"
+  eval $__result="'${patchJsonVal}'"
 }
 
 #
@@ -280,23 +286,28 @@ function createPatchJsonToUpdateReplicas {
   local clusterName=$2
   local replicas=$3
   local __result=$4
+  local existingClusters=""
+  local addClusterReplicasCmd=""
+  local replicasPatch=""
+  local mapCmd=""
+  local patchJsonVal=""
 
-  clusters=$(echo ${domainJson} | jq -cr '(.spec.clusters)')
-  if [ "${clusters}" == "null" ]; then
+  existingClusters=$(echo ${domainJson} | jq -cr '(.spec.clusters)')
+  if [ "${existingClusters}" == "null" ]; then
     # cluster doesn't exist, add cluster with replicas
     addClusterReplicasCmd=".[.| length] |= . + {\"clusterName\":\"${clusterName}\", \
       \"replicas\":${replicas}}"
-    replicasPatch=$(echo ${clusters} | jq -c "${addClusterReplicasCmd}")
+    replicasPatch=$(echo ${existingClusters} | jq -c "${addClusterReplicasCmd}")
   else
     mapCmd="\
       . |= (map(.clusterName) | index (\"${clusterName}\")) as \$idx | \
       if \$idx then \
       .[\$idx][\"replicas\"] = ${replicas} \
       else .+  [{clusterName: \"${clusterName}\" , replicas: ${replicas}}] end"
-    replicasPatch=$(echo ${clusters} | jq "${mapCmd}")
+    replicasPatch=$(echo ${existingClusters} | jq "${mapCmd}")
   fi
-  patchJson="{\"spec\": {\"clusters\": "${replicasPatch}"}}"
-  eval $__result="'${patchJson}'"
+  patchJsonVal="{\"spec\": {\"clusters\": "${replicasPatch}"}}"
+  eval $__result="'${patchJsonVal}'"
 }
 
 #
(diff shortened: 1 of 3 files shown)


kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -478,6 +547,24 @@ function getMinReplicas {
...
+#
+function getMaxReplicas {
+  local domainJson=$1
+  local clusterName=$2
+  local __result=$3
+
+  eval $__result=0
+  maxReplicaCmd="(.status.clusters[] | select (.clusterName == \"${clusterName}\")) \
+    | .maximumReplicas"
+  maxReplicas=$(echo ${domainJson} | jq "${maxReplicaCmd}")
```
tbarnes-us, line 564: please declare maxReplicaCmd & maxReplicas as local

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -478,6 +547,24 @@ function getMinReplicas {
...
+# $1 - Domain resource in json format
+# $2 - Name of the cluster
+# $3 - Return value containing maximum replica count
+#
+function getMaxReplicas {
+  local domainJson=$1
+  local clusterName=$2
+  local __result=$3
+
+  eval $__result=0
```
tbarnes-us, line 561: What is the purpose of this eval? The next eval will replace __result...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -478,6 +547,24 @@ function getMinReplicas {
...
+  maxReplicaCmd="(.status.clusters[] | select (.clusterName == \"${clusterName}\")) \
+    | .maximumReplicas"
+  maxReplicas=$(echo ${domainJson} | jq "${maxReplicaCmd}")
+  eval $__result=${maxReplicas}
```
tbarnes-us, line 565: Should this be ${maxReplicas:-0} ?

tbarnes-us, line 565: there is a problem here if $__result happens to be maxReplicas -- since 'maxReplicas' should be a local var the two will complete -- choose something more unique for the local var...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -460,6 +503,32 @@ function isReplicaCountEqualToMinReplicas {
...
+  local __range=$5
+
+  eval $__result=true
+  getMinReplicas "${domainJson}" "${clusterName}" minReplicas
+  getMaxReplicas "${domainJson}" "${clusterName}" maxReplicas
+  range="${minReplicas} to ${maxReplicas}"
+  if [ ${replicas} -lt ${minReplicas} ] || [ ${replicas} -gt ${maxReplicas} ]; then
+    eval $__result=false
+    eval $__range="'${range}'"
+  fi
```
tbarnes-us, line 529: (1) It looks like range is supposed to be a local var.  Make it local and give it a unique name that doesn't overlap with any global.
(2) It looks like there's a missing 'eval' for $__range if we fall into the if-then stanza

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -249,13 +249,56 @@ function createPatchJsonToUpdateClusterPolicy {
...
+      \"serverStartPolicy\":\"${policy}\"}"
+    startPolicy=$(echo ${clusters} | jq -c "${addClusterStartPolicyCmd}")
+  else
+    mapCmd="\
+      . |= (map(.clusterName) | index (\"${clusterName}\")) as \$idx | \
+      if \$idx then \
+      .[\$idx][\"serverStartPolicy\"] = \"${policy}\" \
+      else .+  [{clusterName: \"${clusterName}\" , serverStartPolicy: \"${policy}\"}] end"
+    startPolicy=$(echo ${clusters} | jq "${mapCmd}")
+  fi
```
tbarnes-us, line 265: Why is this change here? It looks like it relates to 'start policy' not 'replicas', but this pull is all about replicas.  Maybe a bad merge?

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -249,13 +249,56 @@ function createPatchJsonToUpdateClusterPolicy {
...
+    replicasPatch=$(echo ${clusters} | jq -c "${addClusterReplicasCmd}")
+  else
+    mapCmd="\
+      . |= (map(.clusterName) | index (\"${clusterName}\")) as \$idx | \
+      if \$idx then \
+      .[\$idx][\"replicas\"] = ${replicas} \
+      else .+  [{clusterName: \"${clusterName}\" , replicas: ${replicas}}] end"
+    replicasPatch=$(echo ${clusters} | jq "${mapCmd}")
+  fi
+  patchJson="{\"spec\": {\"clusters\": "${replicasPatch}"}}"
```
tbarnes-us, line 298: addClustersReplicasCmd, mapCmd, replicasPath, and patchJson should all be declared as local.  They also should be sufficiently uniquely named so that it's unlikely their names can collide with the value that might be in "$__result".

overall, it might be good overall pattern to have all 'local' variables start with __, and then avoid using __ for global vars...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
tbarnes-us: Also, there's a lot of vars here that should be declared local...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
tbarnes-us: overall, it'd probably be a nice 'code cleanup' to update every function's local vars to use an __ pattern - this would eliminate the potential for them to collide with globals...

kubernetes/samples/scripts/domain-lifecycle/scaleCluster.sh
```diff
@@ -0,0 +1,123 @@
...
+  This script scales a WebLogic cluster in a domain by patching
+  'spec.clusters[<cluster-name>].replicas' attribute of the domain
+  resource. This change will cause the operator to perform a scaling
+  operation for the WebLogic cluster based on the value of replica count.
+ 
+  Usage:
+ 
+    $(basename $0) -c mycluster -r replicas [-n mynamespace] [-d mydomainuid] [-m kubecli]
+  
+    -c <cluster-name>   : Cluster name parameter is required.
```
tbarnes-us, line 24: Can we consider defaulting this to 'cluster-1'? That's what we use in the samples...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -478,6 +547,24 @@ function getMinReplicas {
   eval $__result=${minReplicas}
```
tbarnes-us, line 547: Like the other commands, minReplicaCmd & minReplicas should be declared local. Furthermore, they should be unique enough so that there's potential collision with $__result...

kubernetes/samples/scripts/domain-lifecycle/scaleCluster.sh
```diff
@@ -0,0 +1,123 @@
+# !/bin/sh
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
tbarnes-us, line 2: Technically, you can remove the '2020,' since this is new code for 2021.

kubernetes/samples/scripts/domain-lifecycle/scaleCluster.sh
```diff
@@ -0,0 +1,123 @@
...
+script="${BASH_SOURCE[0]}"
+scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
+source ${scriptDir}/helper.sh
+if [ "${debug}" == "true" ]; then set -x; fi;
+
+function usage() {
+
+  cat << EOF
+
+  This script scales a WebLogic cluster in a domain by patching
```
tbarnes-us, line 15: by patching --> by patching the

kubernetes/samples/scripts/domain-lifecycle/helper.sh
ankedia: It's not a bad merge. If the cluster name is not under `.spec.clusters` for the `stopCluster.sh` or `startCluster.sh` script then it'll not work correctly. Previously I was assuming that cluster will always be present under `.spec.clusters`. Since we removed that assumption for the `scaleCluster.sh` script, I'm making a similar change for `stopCluster.sh` and `startCluster.sh` scripts.

kubernetes/samples/scripts/domain-lifecycle/helper.sh
ankedia: I'm using the __ pattern for the return parameters. For now, I'll make sure that local parameters are not used in the global scope.

kubernetes/samples/scripts/domain-lifecycle/scaleCluster.sh
ankedia: We don't default cluster-name to `cluster-1` for `stopCluster.sh` and `startCluster.sh` scripts. If there are multiple clusters in the domain and the user intended to change replicas for `cluster-2` but forgot to specify the cluster-name parameter then it might cause unintended behavior.

rjeberhard (approved):

ankedia:
@tbarnes-us I have addressed the review comments, please let me know if I missed anything. I also ran the integration tests with the changes and test run passed - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4171 . Thanks.

kubernetes/samples/scripts/domain-lifecycle/helper.sh
tbarnes-us: OK. Should this be back-ported?

kubernetes/samples/scripts/domain-lifecycle/scaleCluster.sh
tbarnes-us: Ok.

kubernetes/samples/scripts/domain-lifecycle/helper.sh
tbarnes-us: Thinking about this some more, I assume they occlude if they're declared local - which would make them safe...

tbarnes-us: ... as long as they're declared local...

kubernetes/samples/scripts/domain-lifecycle/helper.sh
ankedia: SOA team has tested the cluster start/stop scripts and didn't run into any issues with the existing version in 3.1.4. My understanding is that it's an extremely rare use-case to not have any cluster defined under `.spec.clusters` in domain resource. There's a workaround to add the cluster with IF_NEEDED serverStartPolicy. IMO it's not a very high priority to backport since SOA team is also planning to move to 3.2.0 release in May/June.

tbarnes-us (approved):
LGTM. See also comment/question about back-porting due to the included fix.

kubernetes/samples/scripts/domain-lifecycle/helper.sh
tbarnes-us: > My understanding is that it's an extremely rare use-case to not have any cluster defined under .spec.clusters in domain resource. 

I think it depends on whether this is during in development or production. I suspect it's the more common case in development.

PR #2235, opened 2021-03-02
OWLS-87876 - Cluster status script to view the domain resource cluster status of any WebLogic cluster.

ankedia:
OWLS-87876 - The cluster status script to view the domain resource cluster status of any WebLogic cluster. This script is a companion to the `scaleCluster.sh` script and is part of the domain lifecycle scripts.

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/README.md
+++ b/kubernetes/samples/scripts/domain-lifecycle/README.md
@@ -1,4 +1,4 @@
-### Domain lifecycle sample scripts
+### Domain life cycle sample scripts
 
 The operator provides sample scripts to start up or shut down a specific Managed Server or cluster in a deployed domain, or the entire deployed domain.
 
@@ -78,7 +78,7 @@ domain.weblogic.oracle/domain1 patched
 
 ### Script to scale a WebLogic cluster
 
-The `scaleCluster.sh` script scales a WebLogic cluster by patching the `spec.clusters[<cluster-name>].replicas` attribute of the domain resource to the specified value. The operator will perform the scaling operation for the WebLogic cluster based on the specified value of the `replicas` attribute after it's value is updated. See the script `usage` information by using the `-h` option.
+The `scaleCluster.sh` script scales a WebLogic cluster by patching the `spec.clusters[<cluster-name>].replicas` attribute of the domain resource to the specified value. The operator will perform the scaling operation for the WebLogic cluster based on the specified value of the `replicas` attribute after its value is updated. See the script `usage` information by using the `-h` option.
 ```
 $ scaleCluster.sh -d domain1 -n weblogic-domain-1 -c cluster-1 -r 3
 [2021-02-26T19:04:14.335 UTC][INFO] Patching replicas for cluster 'cluster-1' to '3'.
@@ -86,9 +86,9 @@ domain.weblogic.oracle/domain1 patched
 [2021-02-26T19:04:14.466 UTC][INFO] Successfully patched replicas for cluster 'cluster-1'!
 ```
 
-### Script to view the status of any WebLogic cluster
+### Script to view the status of a WebLogic cluster
 
-The `clusterStatus.sh` script can be used to view the status of any WebLogic cluster in the WebLogic domain managed by the operator. The WebLogic Cluster Status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script dumps a table containing the status for WebLogic cluster(s) in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
+The `clusterStatus.sh` script can be used to view the status of a WebLogic cluster in the WebLogic domain managed by the operator. The WebLogic Cluster Status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script displays a table containing the status for WebLogic clusters in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
 
 Use the following command to view the status of all WebLogic clusters in all domains across all namespaces.
 ```
--- a/kubernetes/samples/scripts/domain-lifecycle/README.md
+++ b/kubernetes/samples/scripts/domain-lifecycle/README.md
@@ -88,13 +88,13 @@ domain.weblogic.oracle/domain1 patched
 
 ### Script to view the status of any WebLogic cluster
 
-The `clusterStatus.sh` script can be used to view the status of any WebLogic cluster in the WebLogic domain managed by WKO. The status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script dumps a table containing the status for WebLogic cluster(s) in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
+The `clusterStatus.sh` script can be used to view the status of any WebLogic cluster in the WebLogic domain managed by the operator. The WebLogic Cluster Status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script dumps a table containing the status for WebLogic cluster(s) in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
 
 Use the following command to view the status of all WebLogic clusters in all domains across all namespaces.
 ```
 $ clusterStatus.sh
 
-WKO Cluster Status -n "" -d "" -c "":
+WebLogic Cluster Status -n "" -d "" -c "":
 
 namespace          domain            cluster    min  max  goal  current  ready
 ---------          ------            -------    ---  ---  ----  -------  -----
@@ -107,7 +107,7 @@ Use the following command to view the status of all WebLogic clusters in 'domain
 ```
 $ clusterStatus.sh -d domain1 -n weblogic-domain-1
 
-WKO Cluster Status -n "weblogic-domain-1" -d "domain1" -c "":
+WebLogic Cluster Status -n "weblogic-domain-1" -d "domain1" -c "":
 
 namespace          domain   cluster    min  max  goal  current  ready
 ---------          ------   -------    ---  ---  ----  -------  -----
--- a/kubernetes/samples/scripts/domain-lifecycle/README.md
+++ b/kubernetes/samples/scripts/domain-lifecycle/README.md
@@ -100,7 +100,7 @@ namespace          domain            cluster    min  max  goal  current  ready
 ---------          ------            -------    ---  ---  ----  -------  -----
 ns-kvmt            mii-domain1       cluster-1  1    5    5     5        5
 weblogic-domain-1  domain1           cluster-1  0    4    2     2        2
-weblogic-domain-1  domain1           cluster-2  0    4    0     NA       NA
+weblogic-domain-1  domain1           cluster-2  0    4    0     0        0
 ```
 
 Use the following command to view the status of all WebLogic clusters in 'domain1' in 'weblogic-domain-1' namespace.
@@ -112,6 +112,6 @@ WebLogic Cluster Status -n "weblogic-domain-1" -d "domain1" -c "":
 namespace          domain   cluster    min  max  goal  current  ready
 ---------          ------   -------    ---  ---  ----  -------  -----
 weblogic-domain-1  domain1  cluster-1  0    4    2     2        2
-weblogic-domain-1  domain1  cluster-2  0    4    0     NA       NA
+weblogic-domain-1  domain1  cluster-2  0    4    0     0        0
 
 ```
--- a/kubernetes/samples/scripts/domain-lifecycle/README.md
+++ b/kubernetes/samples/scripts/domain-lifecycle/README.md
@@ -76,7 +76,7 @@ domain.weblogic.oracle/domain1 patched
 [INFO] Successfully patched domain 'domain1' in namespace 'weblogic-domain-1' with 'NEVER' start policy!
 ```
 
-### Scripts to scale a WebLogic cluster
+### Script to scale a WebLogic cluster
 
 The `scaleCluster.sh` script scales a WebLogic cluster by patching the `spec.clusters[<cluster-name>].replicas` attribute of the domain resource to the specified value. The operator will perform the scaling operation for the WebLogic cluster based on the specified value of the `replicas` attribute after it's value is updated. See the script `usage` information by using the `-h` option.
 ```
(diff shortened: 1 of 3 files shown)


kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -85,3 +85,33 @@ $ scaleCluster.sh -d domain1 -n weblogic-domain-1 -c cluster-1 -r 3
 domain.weblogic.oracle/domain1 patched
 [2021-02-26T19:04:14.466 UTC][INFO] Successfully patched replicas for cluster 'cluster-1'!
 ```
+
+### Script to view the status of any WebLogic cluster
+
+The `clusterStatus.sh` script can be used to view the status of any WebLogic cluster in the WebLogic domain managed by WKO. The status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script dumps a table containing the status for WebLogic cluster(s) in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
```
rjeberhard, line 91: I don't think we've generally used the "WKO" acronym in our documentation and instead just said "the operator" or "WebLogic", as appropriate. In this line, I think it's "the operator" and then later, "WebLogic Cluster Status."

rjeberhard (approved):
Minor quibble on the documentation but all else looks good.

kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -85,3 +85,33 @@ $ scaleCluster.sh -d domain1 -n weblogic-domain-1 -c cluster-1 -r 3
...
+```
+$ clusterStatus.sh
+
+WKO Cluster Status -n "" -d "" -c "":
+
+namespace          domain            cluster    min  max  goal  current  ready
+---------          ------            -------    ---  ---  ----  -------  -----
+ns-kvmt            mii-domain1       cluster-1  1    5    5     5        5
+weblogic-domain-1  domain1           cluster-1  0    4    2     2        2
+weblogic-domain-1  domain1           cluster-2  0    4    0     NA       NA
```
tbarnes-us, line 103: We are showing NA here because the operator did not report a value for the given attribute. But it looks like we can safely interpret NA is '0' in this case, which would be far more intuitive IMO.  Assuming NA always can be safely interpreted means 0, should we either (a) correct the operator code to report a value for this field, or (b) modify the script to report '0' instead of NA when an attribute is missing, or (c) do both?  Or maybe NA truly means 'don't know'?

@rjeberhard Would you like to weigh in?

ankedia: I'm not very sure but since the replica count goal is 0, displaying NA (not applicable) for current and ready fields might be self-explanatory. Displaying 0 for current/ready might mean that pods are not yet created/ready.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  ) | column --table
+ 
+  echo
+}
+
+domainNS=
+domainUID=
+clusterName=
+kubernetesCli=
```
tbarnes-us, line 116: Please follow the convention of the other domain lifecycle scripts and default this to $KUBERNETES_CLI env var (plus change the help accordingly).

kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -774,6 +774,9 @@ function validatePythonAvailable {
 
 # try to execute kubernetes cli to see whether cli is available
 function validateKubernetesCliAvailable {
+  if [ ${#} -gt 0 ]; then
+    kubernetesCli=$1
```
tbarnes-us, line 778: Consider prepending with 'local' so this function doesn't have the (somewhat surprising) side effect of setting a global.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+  else
+    local __ns_filter="-n $__ns"
+  fi
+
+  local __uid="${2:-}"
+  local __cluster_name="${3:-}"
+  local __kubernetes_cli="${4:-kubectl}"
+
+  validateKubernetesCliAvailable "$__kubernetes_cli"
+  failIfValidationErrors
```
tbarnes-us, line 64: These two lines look like the only code grabbed from 'helper.sh'. To make this a completely stand-alone script, perhaps replace with 
```
  if ! [ -x "$(command -v ${__kubernetesCli})" ]; then
     echo "@@Error: Kuberenetes CLI '${kubernetesCli}' is not installed."
     exit 1
  fi
```

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+      local __uid_cur=$( echo $__val | cut -d ',' -f 3)
+      local __uid_cur=${uid_cur:-$__name_cur} # uid defaults to name when not set
+      local __jp=''
+
+      [ -n "$__uid" ] && [ ! "$__uid" = "$__uid_cur" ] && continue
+
+      if [ -z "$__cluster_name" ]; then
+        __jp+='{range .status.clusters[*]}'
+      else 
+         __jp+='{range .status.clusters[?(@.clusterName=='\"$__cluster_name\"')]}'
```
tbarnes-us, line 92: Minor nit: remove the extra indent space.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+  replica information for each given cluster.
+
+  Omit '-n' or set it to "" to indicate all namespaces.
+  Omit '-d' or set it to "" to indicate all domains in the given namespace(s).
+  Omit '-c' or set it to "" to indicate all clusters in the given domain(s) and the given namespace(s).
+  Use '-m' to specify a Kubernetes command line interface (defaults to 'kubectl').
+  Use '-h' to get this help.
+
+Examples:
+  - Dump cluster 'c1' in uid 'myuid' in ns 'myns': $(basename $0) -n myns -d myuid -c c1
```
tbarnes-us, line 25: in uid --> with uid (same for below)

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  - Dump all clusters in uid 'myuid' in ns 'myns': $(basename $0) -n myns -d myuid
+                                                   -or-
+                                                   $(basename $0) -n myns -d "" -c ""
```
tbarnes-us, line 29: Th -or- is incorrect, it should have `-d myuid` instead of `-d ""`.  
Regardless I think we can just remove this particular -or- all-together - it's only there to demonstrate the use of `""` but the following sample command already demonstrates `""`.   Maybe even remove the following '-or-' too -- I put it in there myself, but, thinking about it some more, the demonstration of `""` is trivially helpful.

kubernetes/samples/scripts/domain-lifecycle/README.md
tbarnes-us: The goal is '0' - so if displaying 0 for current/ready "might mean that pods are not yet created/ready", why is that a bad thing?

If it turns out that we can safely substitute 0 for NA, not only would it be more intuitive, but also any script that uses this script as input wouldn't have to add special case handling for the NA string...

kubernetes/samples/scripts/domain-lifecycle/README.md
rosemarymarano: Definitely true. We do not use the acronym WKO in the docs; let's not start (there are enough acronyms in this world). Much better to use the term, "the operator," as Ryan suggested.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+set -o pipefail
+script="${BASH_SOURCE[0]}"
+scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
+source ${scriptDir}/helper.sh
+
+function usage() {
+cat<<EOF
+
+Usage:
+  $(basename $0) [-n DomainNS] [-d DomainUID] [-c clsterName] [-m kubernetesCli] [-h]
```
tbarnes-us, line 13: clstrName --> clusterName

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  This script dumps a table of WKO domain status
+  replica information for each given cluster.
```
tbarnes-us, line 16: given cluster --> given WebLogic cluster.

kubernetes/samples/scripts/domain-lifecycle/README.md
ankedia: I'm not saying it's a bad thing, I just thought that displaying NA might be OK.

kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -76,7 +76,7 @@ domain.weblogic.oracle/domain1 patched
 [INFO] Successfully patched domain 'domain1' in namespace 'weblogic-domain-1' with 'NEVER' start policy!
 ```
 
-### Scripts to scale a WebLogic cluster
+### Script to scale a WebLogic cluster
```
rosemarymarano, line 80: The title of this doc should be: Domain life cycle sample scripts. (life cycle (n), lifecycle (adj))
after it's value -> after its value

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,128 @@
...
+  ) | column --table
+
+  echo
+}
+
+domainNS=
+domainUID=
+clusterName=
+__kubernetesCli=${KUBERNETES_CLI:-kubectl}
+validateErrors=false
```
tbarnes-us, line 112: Unused variable.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,128 @@
...
+
+  ) | column --table
+
+  echo
+}
+
+domainNS=
+domainUID=
+clusterName=
+__kubernetesCli=${KUBERNETES_CLI:-kubectl}
```
tbarnes-us, line 111: Minor nit: it'd be more clear to remove the preceding underscores here. The convention in this particular script is to use "__" to indicate the use of internal function local vars...

kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -85,3 +85,33 @@ $ scaleCluster.sh -d domain1 -n weblogic-domain-1 -c cluster-1 -r 3
 domain.weblogic.oracle/domain1 patched
 [2021-02-26T19:04:14.466 UTC][INFO] Successfully patched replicas for cluster 'cluster-1'!
 ```
+
+### Script to view the status of any WebLogic cluster
```
rosemarymarano, line 89: of any WebLogic cluster -> of a WebLogic cluster (it's understood that it could be any cluster)

kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -85,3 +85,33 @@ $ scaleCluster.sh -d domain1 -n weblogic-domain-1 -c cluster-1 -r 3
...
+
+The `clusterStatus.sh` script can be used to view the status of any WebLogic cluster in the WebLogic domain managed by the operator. The WebLogic Cluster Status contains information about the minimum, maximum, goal, current, and ready replica count for a WebLogic cluster. This script dumps a table containing the status for WebLogic cluster(s) in one or more domains across one or more namespaces. See the script `usage` information by using the `-h` option.
```
rosemarymarano, line 91: dumps a table -> displays a table 
cluster(s) -> clusters (it's understood that it could be one or more)

rosemarymarano (commented):
Some edits and suggestions.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,125 @@
...
+set -eu
+set -o pipefail
+
+function usage() {
+cat<<EOF
+
+Usage:
+  $(basename $0) [-n DomainNS] [-d DomainUID] [-c clusterName] [-m kubernetesCli] [-h]
+
+  This script dumps a table of WebLogic Cluster status
```
tbarnes-us, line 12: As per @rosemarymarano suggestion for the ".md" doc, replace occurrences of "dump" with "display" throughout.

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,125 @@
...
+  Use '-m' to specify a Kubernetes command line interface. Default is 'kubectl' if KUBERNETES_CLI env
+  variable is not set. Otherwise default is the value of KUBERNETES_CLI env variable.
+  Use '-h' to get this help.
+
+Examples:
+  - Dump all uid across all ns                       : $(basename $0)
+  - Dump all clusters with uid 'myuid' across all ns : $(basename $0) -d myuid
+  - Dump all clusters and all uid in ns 'myns'       : $(basename $0) -n myns
+  - Dump all clusters with uid 'myuid' in ns 'myns'  : $(basename $0) -n myns -d myuid
+  - Dump cluster 'c1' with uid 'myuid' in ns 'myns'  : $(basename $0) -n myns -d myuid -c c1
```
tbarnes-us, line 27: Dump --> Display

rosemarymarano (approved):

kubernetes/samples/scripts/domain-lifecycle/README.md
tbarnes-us: Done.

tbarnes-us (approved):

kubernetes/samples/scripts/domain-lifecycle/clusterStatus.sh
```diff
@@ -0,0 +1,128 @@
...
+      __jp+='{"'$__ns_cur' "}'
+      __jp+='{"'$__uid_cur' "}'
+      __jp+='{.clusterName}{" ~!"}'
+      __jp+='{.minimumReplicas}{" ~!"}'
+      __jp+='{.maximumReplicas}{" ~!"}'
+      __jp+='{.replicasGoal}{" ~!"}'
+      __jp+='{.replicas}{" ~!"}'
+      __jp+='{.readyReplicas}{" "}'
+      __jp+='{"\n"}'
+      __jp+='{end}'
```
tbarnes-us, line 96: Very minor nit (still approved):  This would be more clear:
```
      __jp+='{"'$__ns_cur'"}'
      __jp+='{" "}{"'$__uid_cur'"}'
      __jp+='{" "}{.clusterName}'
      __jp+='{" ~!"}{.minimumReplicas}'
      __jp+='{" ~!"}{.maximumReplicas}'
      __jp+='{" ~!"}{.replicasGoal}'
      __jp+='{" ~!"}{.replicas}'
      __jp+='{" ~!"}{.readyReplicas}'
      __jp+='{"\n"}'
      __jp+='{end}'
```

ankedia: Done in 4ab94ac .

PR #2236, opened 2021-03-02
fix intermittent test issue in ITMiiDynamicUpdate testMiiChangeListenPort

xiancao:
In the nightly Jenkins run, sometime we will get the test failure from ITMiiDynamicUpdate#testMiiChangeListenPort, testMiiChangeListenAddress and testMiiChangeSSL. 
The reason of the failure:
Those tests are negative tests and at the end of tests, I updated the config map with correct contents and verify the introspector pod is created, runs and deleted. However if the content change in the config map is too small and the introspector runs very fast, the test will not detect the introspector got created. 
How to fix it:
Just verify the introspector pod was deleted after the config map is updated.

Jenkins result:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4207/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4206/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4205/


rjeberhard (approved):

vanajamukkara (approved):
LGTM

PR #2237, opened 2021-03-02
Update unit tests and simplify logic  for dynamic update

jshum2479:
Russ and I worked on adding unit tests and simplify the makeright logic for dynamic update.


operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
```diff
@@ -1520,6 +1590,13 @@ public void whenPodCreated_createPodWithOwnerReference() {
     assertThat(getCreatedPod().getMetadata().getOwnerReferences(), contains(expectedReference));
   }
 
+  // TODO: add tests for dynamic updates
```
rjeberhard, line 1593: How hard would it be to create these tests? I'm not a fan of leaving a TODO in the code.

jshum2479: actually the tests are there, removing TODO

rjeberhard (approved):

jshum2479:
There is regression in the branch, withdrawing now.

jshum2479:
I found the problem, actually it's a bug rather than a regression in mii v1 model diff code

russgold (approved):

jshum2479:
@rjeberhard  Actually, it's not a regression, I found a bug in v1 and fixed that.  It should be good to go

PR #2242, opened 2021-03-03
Owls 87789 istio session replication

jshum2479:
This is to cover the case when customer wanted to use session replication in their application, the weblogic core returns the channel 'default' which will not work as the splitter channel disabled outbound traffic.  User has the option to create their own custom network channel for session replication or with this fix specify the 'tcp-default' as the replication channel in the cluster configuration.


lennyphan (approved):
LGTM

rjeberhard (approved):

PR #2245, opened 2021-03-03
Add support for operator sidecar

russgold:
Adds support for integration with the operator sidecar.

This makes an addition to the domain resource by adding a `monitoring-exporter` element, which can specify both the exporter image and the configuration. 

Note that, at present, only connecting to the WLS plaintext port is actually working. We need to update the client truststore to handle certificates.


kubernetes/crd/domain-v1beta1-crd.yaml
```diff
@@ -31,6 +31,30 @@ spec:
...
+                  enum:
+                  - Always
+                  - Never
+                  - IfNotPresent
+                configuration:
+                  type: object
+                  description: The configuration for the WebLogic Monitoring Exporter
+                    sidecar. If specified, the operator will deploy a sidecar alongside
+                    each server instance. See https://github.com/oracle/weblogic-monitoring-exporter
+                  xKubernetesPreserveUnknownFields: true
```
rjeberhard, line 57: Isn't this supposed to be: `x-kubernetes-preserve-unknown-fields: true`

operator/pom.xml
```diff
@@ -348,6 +348,11 @@
       <artifactId>jersey-test-framework-provider-inmemory</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>com.networknt</groupId>
+        <artifactId>json-schema-validator</artifactId>
+        <version>1.0.49</version>
```
rjeberhard, line 354: This new dependency should be okay. Apache license and earlier versions have been approved for other teams. Please move the version to a property in the top-level pom with the other version properties.

operator/src/main/java/oracle/kubernetes/operator/KubernetesConstants.java
```diff
@@ -6,6 +6,8 @@
 /** Kubernetes constants. */
 public interface KubernetesConstants {
   String DEFAULT_IMAGE = "container-registry.oracle.com/middleware/weblogic:12.2.1.4";
+  String DEFAULT_EXPORTER_IMAGE = "oracle/weblogic-monitoring-exporter:2.0";
```
rjeberhard, line 9: We can no longer publish images to Docker Hub. Therefore, we've chosen to preference defaults from ghcr.io rather than OCR. Let's use `ghcr.io/oracle/weblogic-monitoring-exporter:2.0`.

operator/src/main/java/oracle/kubernetes/operator/helpers/AnnotationHelper.java
```diff
@@ -25,14 +25,14 @@
    * Marks metadata with annotations that let Prometheus know how to retrieve metrics from the
    * wls-exporter web-app. The specified httpPort should be the listen port of the WebLogic server
    * running in the pod.
-   *
-   * @param meta Metadata
+   *  @param meta Metadata
```
rjeberhard, line 28: Why the extra space?

operator/src/main/java/oracle/kubernetes/operator/helpers/AnnotationHelper.java
```diff
@@ -25,14 +25,14 @@
...
+   * @param applicationContext the context for the exporter application
    * @param httpPort HTTP listen port
    */
-  static void annotateForPrometheus(V1ObjectMeta meta, int httpPort) {
+  static void annotateForPrometheus(V1ObjectMeta meta, String applicationContext, int httpPort) {
     meta.putAnnotationsItem(
         "prometheus.io/port", "" + httpPort); // should be the ListenPort of the server in the pod
-    meta.putAnnotationsItem("prometheus.io/path", "/wls-exporter/metrics");
+    meta.putAnnotationsItem("prometheus.io/path", applicationContext + "/metrics");
```
rjeberhard, line 35: What is the default `applicationContext`? Unfortunately, these annotations were added at the wrong code point and are present when the model is hashed; therefore, any change to these values causes existing pods to roll.

kubernetes/crd/domain-v1beta1-crd.yaml
russgold: In the code it is - and then Kubernetes changes it to camel case, which is confusing. I spent a good while trying to understand that. Recall that this is a generated file.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -833,6 +871,7 @@ public String getModelHome() {
         .map(Configuration::getModel).map(Model::getModelHome).orElse(DEFAULT_WDT_MODEL_HOME);
   }
 
+  // todo - should these be put into alphabetical order? It is getting increasingly difficult to find items
```
rjeberhard, line 874: They were originally in the order that they were defined in the class. I don't think there is any technical reason to oppose that they be alphabetically sorted.

operator/src/main/java/oracle/kubernetes/operator/helpers/AnnotationHelper.java
russgold: Oops - didn't intend to do that.

operator/src/main/java/oracle/kubernetes/operator/helpers/AnnotationHelper.java
russgold: The application context is returned from `ExporterContext.getBasePath()` and depends on whether the annotations are generated for the webapp or the sidecar, and will only change if we swap between those - which is what we want to have happen. For the webapp, it is "/wls-exporter"; for the sidecar, it is "".

kubernetes/crd/domain-v1beta1-crd.yaml
rjeberhard: I'm worried that this won't work. For the dedicated mode use case, customers create the CRD in advance by `kubectl create -f` with this document. This misnamed field likely won't do what it needs to. When the operator creates the CRD by directly creating the CRD does the CRD get the correct contents?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
russgold: I'm alphabetizing `toString` and `hashCode` but not `equals` for now, as the latter, unlike the first two, mixes field references with getters, and I don't know why that was done.

kubernetes/crd/domain-v1beta1-crd.yaml
russgold: Kubernetes is definitely getting the correct value when the operator generates the CRD this way - without this field in the CRD, it doesn't preserve the exporter configuration.

operator/pom.xml
russgold: Done.

operator/src/main/java/oracle/kubernetes/operator/KubernetesConstants.java
russgold: Done.

kubernetes/crd/domain-v1beta1-crd.yaml
rjeberhard: You can simulate that test this way:
1. Uninstall the operator and delete the CRD
2. Install the CRD using `kubectl create -f <crd.yaml>`
3. Install the operator
4. (Optional) make sure that the operator doesn't overwrite the CRD -- it shouldn't because of the hash label.
5. See if Kubernetes drops the exporter configuration now.

rjeberhard: Or, you could just do the equivalent of a `sed` before writing the CRD to file. :)

docs/domains/Domain.json
```diff
@@ -573,6 +586,29 @@
...
       }
     },
+    "MonitoringExporterSpecification": {
+      "type": "object",
+      "properties": {
+        "image": {
+          "description": "The WebLogic Monitoring Exporter sidecar image name. Defaults to ghcr.io/oracle/weblogic-monitoring-exporter:2.0",
+          "type": "string"
+        },
+        "imagePullPolicy": {
```
jshum2479, line 596: Is it always public open?  Will there be a case that it needs imagePullSecrets?

rjeberhard: That's a good point. The image would be available publicly available on ghcr.io or OCR. It is possible that the customer would have a custom image for the exporter that is available on a repo that requires image pull secrets.

russgold: I assume that such a thing is possible, given that the exporter is open source.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -505,6 +538,120 @@ protected String getClusterName() {
...
+      void addContainer(List<V1Container> containers) {
+        containers.add(createMonitoringExporterContainer());
+      }
+
+      private V1Container createMonitoringExporterContainer() {
+        return new V1Container()
+              .name(EXPORTER_CONTAINER_NAME)
+              .image(getDomain().getMonitoringExporterImage())
+              .imagePullPolicy(getDomain().getMonitoringExporterImagePullPolicy())
+              .addEnvItem(new V1EnvVar().name("JAVA_OPTS").value(createJavaOptions()))
```
ankedia, line 630: I'm wondering if there's a need to populate server and domain-related environment variables that we populate in the main container and the init containers. I recently worked on populating these env variables in an init container in JIRA OWLS-77957.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -505,6 +538,120 @@ protected String getClusterName() {
...
+    class WebAppExporterContext extends ExporterContext {
+
+      @Override
+      boolean isEnabled() {
+        return getListenPort() != null;
+      }
+
+      @Override
+      int getPort() {
+        return getListenPort();
```
ankedia, line 583: Does this need to be SSL port if the admin channel is enabled or SSL is enabled?

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
```diff
@@ -250,6 +255,10 @@ public void setSslListenPort(Integer listenPort) {
     this.sslListenPort = listenPort;
   }
 
+  public PortDetails getSslListenPortDetails() {
```
ankedia, line 258: Is this method added for future use, it doesn't seem to be used anywhere?

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
```diff
@@ -308,6 +317,10 @@ public WlsServerConfig setAdminPort(int adminPort) {
     return this;
   }
 
+  public PortDetails getAdminPortDetails() {
```
ankedia, line 320: Is this method added for future use, it doesn't seem to be used anywhere?

kubernetes/crd/domain-v1beta1-crd.yaml
russgold: You're correct. If I create the CRD from this file, it ignores the annotation. If I edit the field to turn it back to dash case, it works. So I need to fix the generator to bypass whatever has been converting it.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -496,6 +522,13 @@ protected String getClusterName() {
       return new ArrayList<>(super.getContainerCommand());
     }
 
+    @Override
+    protected List<V1Container> getContainers() {
+      List<V1Container> containers = new ArrayList<>(super.getContainers());
+      exporterContext.addContainer(containers);
```
ankedia, line 528: It might be more clear to add the sidecar container in `createPodSpec` since the method name `getContainers` is a bit misleading here.

ankedia (commented):
I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
rjeberhard: The "sidecar" name is really a design pattern, but the pod spec type only has a list of containers. There really isn't any distinction between the main and sidecar containers when you create the pod. So, I think that this name makes sense since it's generating the content for the containers field of the pod spec.

ankedia: ok, yes. Currently, we add the primary container in `createPodSpec` in BasePodStepContext and getContainers() method name implies it's only getting existing containers. I was wondering if the container item for the sidecar container can be added in `createPodSpec` similar to the primary container or change the method name to something like addAndGetContainer().

```
  protected V1PodSpec createPodSpec(TuningParameters tuningParameters) {
    return new V1PodSpec()
        .containers(getContainers())
        .addContainersItem(createPrimaryContainer(tuningParameters))
```

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
russgold: Shouldn't need them, as far as I know. The exporter only uses a small number of such variables.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
russgold: It gets complicated, as Prometheus is currently assuming plainport access to the exporter. But this is only for the web app case, which we are hoping to deprecate within the operator.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
russgold: I think I used it at one point, but no longer do, so I'll delete it. Thanks for catching this.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
russgold: same as before.

russgold:
> I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?

I'm not familiar with our ELK integration, but this pattern appears to work and is easy. There is nothing that Helm needs to be involved in, and the operator does need to know about the monitoring exporter to configure it.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
russgold: No, because it's not under user control. It is an internal implementation detail of the operator.

I agree that `getContainers` is a confusing name, but I wasn't sure what to change it to. I did rename `createContainer` to `createPrimaryContainer` to add clarity. Maybe `getContainers` should be `getUserContainers` or something along those lines.

ankedia: I thought the "sidecar" container for monitoring exporter is not a user container but something that the operator injects. The users might have their own init and/or sidecar containers. In that case, `getUserContainers` name will still be confusing.

rjeberhard:
> > I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?
> 
> I'm not familiar with our ELK integration, but this pattern appears to work and is easy. There is nothing that Helm needs to be involved in, and the operator does need to know about the monitoring exporter to configure it.

The Helm install is, of course, for the operator. @ankedia, are you thinking of a sort of global switch to enable the monitoring exporter with every domain? I'd like to keep what we have and then see if customers are asking for a configuration option that spans domains.

ankedia:
> > > I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?
> > 
> > 
> > I'm not familiar with our ELK integration, but this pattern appears to work and is easy. There is nothing that Helm needs to be involved in, and the operator does need to know about the monitoring exporter to configure it.
> 
> The Helm install is, of course, for the operator. @ankedia, are you thinking of a sort of global switch to enable the monitoring exporter with every domain? I'd like to keep what we have and then see if customers are asking for a configuration option that spans domains.

[ANIL] Yes, I was thinking of a sort of global switch to enable monitoring exporter with basic configuration (similar to ELK integration). The users can then customize/add to the existing configuration for monitoring exporter.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
russgold: Correct, and that's why this is confusing. the `getContainers` method returns the list of containers the user has specified in the domain - they have nothing to do with the exporter sidecar container.

russgold:
> > > > I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?
> > > 
> > > 
> > > I'm not familiar with our ELK integration, but this pattern appears to work and is easy. There is nothing that Helm needs to be involved in, and the operator does need to know about the monitoring exporter to configure it.
> > 
> > 
> > The Helm install is, of course, for the operator. @ankedia, are you thinking of a sort of global switch to enable the monitoring exporter with every domain? I'd like to keep what we have and then see if customers are asking for a configuration option that spans domains.
> 
> [ANIL] Yes, I was thinking of a sort of global switch to enable monitoring exporter with basic configuration (similar to ELK integration). The users can then customize/add to the existing configuration for monitoring exporter if they wanted.

As it works now, the exporter is only enabled IF the user specifies a configuration. What other customizations are you envisioning?

ankedia:
> > > > > I'm curious if it's possible to enable the monitoring exporter configuration using helm chart values similar to how it's done for the ELK integration?
> > > > 
> > > > 
> > > > I'm not familiar with our ELK integration, but this pattern appears to work and is easy. There is nothing that Helm needs to be involved in, and the operator does need to know about the monitoring exporter to configure it.
> > > 
> > > 
> > > The Helm install is, of course, for the operator. @ankedia, are you thinking of a sort of global switch to enable the monitoring exporter with every domain? I'd like to keep what we have and then see if customers are asking for a configuration option that spans domains.
> > 
> > 
> > [ANIL] Yes, I was thinking of a sort of global switch to enable monitoring exporter with basic configuration (similar to ELK integration). The users can then customize/add to the existing configuration for monitoring exporter if they wanted.
> 
> As it works now, the exporter is only enabled IF the user specifies a configuration. What other customizations are you envisioning?
I was thinking that the user will only set some value (say enableMonitoringExporter) to `true` in values.yaml when installing the operator Helm chart to enable the exporter. This will automatically add the required configuration to the domain resource. However, if a user specifies the configuration then that value in values.yaml will be ignored. 

Anyways, I'm ok with what Ryan suggested above to keep what we have and then see if customers are asking for a configuration option that spans domains.

kubernetes/crd/domain-v1beta1-crd.yaml
russgold: The problem is that the way the Yaml is written involves classes with maps of V1JsonSchemaProps objects, which contains fields like this
```
  public static final String SERIALIZED_NAME_X_KUBERNETES_PRESERVE_UNKNOWN_FIELDS = "x-kubernetes-preserve-unknown-fields";
  @SerializedName("x-kubernetes-preserve-unknown-fields")
  private Boolean xKubernetesPreserveUnknownFields;
```
When serialized by Gson, the serialized name is used - but Kubernetes uses SnakeYaml to write this, and it doesn't know about that annotation. So I do in fact need to process the lines as they are written, as though simulating `sed`. ::sigh::

russgold: Alternately, I can change the YAML generation to generate _JSON_ and then convert that to YAML. That would probably be more obvious and safer.

russgold: And that two-step process works, although it changes the order of some fields in the output.

rjeberhard: The changed order makes it a little hard to review, but I did my best... I think it looks good.

rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
ankedia: It looks like you deleted the other method but not this one. Is this intentional?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
ankedia: Should the method be called `getUserContainersAndAddExporterContainer`? I'm ok either way, just thought it might be slightly more clear for someone trying to read the code for the first time.

operator/src/main/java/oracle/kubernetes/operator/wlsconfig/WlsServerConfig.java
russgold: I thought I'd gotten them all, but I only deleted two of the three. I'm trying to get this merged, and afraid if I keep making changes it keeps requiring more reviews. @rjeberhard should we defer deleting this?

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
russgold: That sounds excessively implementation specific, and problematic. The intent is to add all the containers other than the primary one, and it therefore overrides the default, which is just to add the user-specified containers.

ankedia: Well, in that case, can the method be renamed `getAndAddContainers` or just `addContainers`?

ankedia (approved):
LGTM except for a couple of minor comments above.

jshum2479 (approved):

PR #2247, opened 2021-03-05
The sip and sips protocols support both TCP and UPD on the same channel

rjeberhard:
We received an enhancement request from the mobile communications business unit to better support the sip and sips protocols by updating our container and Service ports to specify both TCP and UDP ports when a network access point supports one of these protocols.

The ports will share a listenAddress, which is legal, but can't share a name, so I've chosen to prefix the UDP port's name with "udp-" since this is also consistent with the Istio port naming convention.

Update:
I figured out why we previously needed the testNodePort filtering behavior... The use of a ports variable created state in the context. If I remove this variable and check for duplication in the generated ports of the model then we can both prevent duplication and allow for the two different SIP service ports that share a listen port.

Outdated:
> Note: I removed calls to a method called "testNodePort" which was written to prevent a container or Service from having ports with the same listenPort value. No unit-tests failed when I removed that code and the behavior this check was preventing was exactly the behavior that I now need.  I also verified with several manual tests with both SIP and non-SIP protocol network access points that there is no duplication of ports otherwise.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -233,6 +233,17 @@ public abstract class PodStepContext extends BasePodStepContext {
     return "sip".equals(nap.getProtocol()) || "sips".equals(nap.getProtocol());
   }
 
+  private V1ContainerPort createContainerPort(String name, Integer port, String protocol) {
+    return new V1ContainerPort()
+        .name(LegalNames.toDns1123LegalName(name))
+        .containerPort(port)
+        .protocol(protocol);
+  }
+
+  private V1ContainerPort prefixName(String prefix, V1ContainerPort port) {
+    return port.name(prefix + port.getName());
+  }
+
   private List<V1ContainerPort> getContainerPorts() {
     if (scan != null) {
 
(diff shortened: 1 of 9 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/ServiceHelperTest.java
```diff
@@ -714,17 +752,20 @@ protected boolean matchesSafely(V1Service item, Description mismatchDescription)
         }
         mismatchDescription.appendText("contains no port with name ").appendValue(expectedName);
       } else {
-        if (matchSelectedPort(matchingPort)) {
+        if (matchesSelectedProtocol(matchingPort) && matchSelectedPort(matchingPort)) {
```
russgold, line 755: Consistency - please use either "match" or "matches" for both methods.

operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTestBase.java
```diff
@@ -247,7 +249,9 @@ public void setUp() throws Exception {
     WlsDomainConfigSupport configSupport = new WlsDomainConfigSupport(DOMAIN_NAME);
     configSupport.addWlsServer(ADMIN_SERVER, ADMIN_PORT);
     if (!ADMIN_SERVER.equals(serverName)) {
```
russgold, line 251: I hadn't noticed this code before. Maybe it would be better to move it to the ManagedServer version of the setUp method rather than leaving it in the common code and testing the name.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -229,19 +229,33 @@ private String getRuntimeEncryptionSecret() {
...
+  }
+
   private List<V1ContainerPort> getContainerPorts() {
     if (scan != null) {
 
       List<V1ContainerPort> ports = new ArrayList<>();
       if (scan.getNetworkAccessPoints() != null) {
         for (NetworkAccessPoint nap : scan.getNetworkAccessPoints()) {
           String napName = nap.getName();
+
```
russgold, line 243: This method is starting to get large. Might be a good idea to factor out the loop body. The two cases of adding a port seem similar enough that common code could be extracted from them.

russgold (commented):
Overall, looks good. I have made some recommendations in improvements to readability.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
rjeberhard: done

russgold (approved):

ankedia (approved):
LGTM

PR #2248, opened 2021-03-05
Added tests for monitoring exporter sidecar

marinakog:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4228/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4229


russgold (approved):
As much as I can understand these tests, they look fine. I think I see some that you might want to consider for the future:

- admin port enabled (doesn't currently work)
- when domain qualifier specified, should use the name of the operator domain rather than the WLS domain, as apparently many customers use the same name for WLS domains. Perhaps we have to verify this with Monica
- once we release the sidecar as a docker image, verify using the released one rather than building as part of the test

rjeberhard (approved):

PR #2250, opened 2021-03-08
Owls 87307 fix and change to store the domain topology in json format (for OWLS-87478) 

ankedia:
Fix for Owls 87307 and change to store the domain topology in JSON format (in addition to YAML format). This change removes the dependency on `yq` and/or `python` when running scripts on Mac OS or with versions of python that don't have YAML module support. This change helps in the documentation of pre-requisites for running domain lifecycle scripts (OWLS-87478).

Integration test run URL - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4235/

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/README.md
+++ b/kubernetes/samples/scripts/domain-lifecycle/README.md
@@ -2,7 +2,7 @@
 
 The operator provides sample scripts to start up or shut down a specific Managed Server or cluster in a deployed domain, or the entire deployed domain.
 
-**Note**: Prior to running these scripts, you must have previously created and deployed the domain. These scripts make use of [jq](https://stedolan.github.io/jq/) for processing JSON, you must have `jq 1.5 or higher` installed in order to run these scripts. See the installation options on the [jq downlod](https://stedolan.github.io/jq/download/) page.
+**Note**: Prior to running these scripts, you must have previously created and deployed the domain. These scripts make use of [jq](https://stedolan.github.io/jq/) for processing JSON. You must have `jq 1.5 or higher` installed in order to run these scripts. See the installation options on the [jq downlod](https://stedolan.github.io/jq/download/) page.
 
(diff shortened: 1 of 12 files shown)


tbarnes-us:
So the domain-lifecycle scripts will no longer work on older WKO versions?

ankedia:
> So the domain-lifecycle scripts will no longer work on older WKO versions?

I'm not sure if we need to support interop with the domain-lifecycle scripts. The domain lifecycle scripts were introduced in Operator 3.1.0 release and the older version of domain life-cycle scripts will still work with older versions. SOA QA teams have run into issues using these scripts since they had the python version (OWLS-87478) which didn't support `yaml` module. If we need to support interop then I can add logic to check json topology and then fall back to yaml topology if json topology doesn't exist. In that case, we still need to document which version of python/yq is required to run the script.

tbarnes-us (approved):
Pre-approving. Up to you if you want to follow my suggestion to update the validation error messages.  I'm not sure what the next version would be those two messages -- substitute whatever Ryan tells you would be good for '3.2.?'.

ankedia:
> Pre-approving. Up to you if you want to follow my suggestion to update the validation error messages. I'm not sure what the next version would be those two messages -- substitute whatever Ryan tells you would be good for '3.2.?'.

Thanks, I have changed the error messages as per your suggestion. @rjeberhard Please let me know the Operator version in the error message or any other contents that need to be changed.

kubernetes/samples/scripts/domain-lifecycle/README.md
```diff
@@ -2,7 +2,7 @@
 
 The operator provides sample scripts to start up or shut down a specific Managed Server or cluster in a deployed domain, or the entire deployed domain.
 
-**Note**: Prior to running these scripts, you must have previously created and deployed the domain.
+**Note**: Prior to running these scripts, you must have previously created and deployed the domain. These scripts make use of [jq](https://stedolan.github.io/jq/) for processing JSON, you must have `jq 1.5 or higher` installed in order to run these scripts. See the installation options on the [jq downlod](https://stedolan.github.io/jq/download/) page.
```
rjeberhard, line 5: ", you must" -> ". You must" This should be a new sentence.

rjeberhard (approved):

PR #2251, opened 2021-03-09
Fix http throwable responses

doxiao:
The PR contains a fix for a problem in http error handling that is discovered during testing domain secure mode enabled without configuring SSL use case, plus some code cleanup based on warnings in IntelliJ. 

Problem: when a HttpAsyncRequestStep's AsyncProcessing is resumed without a response object, it was always considered as a timeout although sometimes resume is called with an Exception that should be made visible to the user. One example of such exceptions is javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target. 

With the fix, operator logs a warning message and fails the corresponding step.

(Note: the changes originally contained a fix for a NPE on unboxing the chosen port number for Prometheus. The code change for this has been removed after merging with the develop branch yesterday; apparently the problematic code has been changed by a PR that is merged yesterday.)

Integration test result in Jenkins before the last merge with develop: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4250/.
Integration test run with the latest: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4255/.
Jenkins job with review changes: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4256/.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
@@ -53,7 +53,7 @@ public class HttpAsyncRequestStep extends Step {
    * @param responseStep the step to handle the response
    * @return a new step to run as part of a fiber, linked to the response step
    */
-  public static HttpAsyncRequestStep createGetRequest(String url, HttpResponseStep responseStep) {
+  static HttpAsyncRequestStep createGetRequest(String url, HttpResponseStep responseStep) {
     HttpRequest request = HttpRequest.newBuilder(URI.create(url)).GET().build();
     return create(request, responseStep);
   }
@@ -108,7 +108,11 @@ public class HttpAsyncRequestStep extends Step {
 
     private void resume(AsyncFiber fiber, HttpResponse<String> response, Throwable throwable) {
       if (throwable != null) {
-        LOGGER.fine(MessageKeys.HTTP_REQUEST_TIMED_OUT, request.method(), request.uri(), throwable);
+        if (throwable instanceof HttpTimeoutException) {
+          LOGGER.fine(MessageKeys.HTTP_REQUEST_TIMED_OUT, throwable.getMessage());
+        } else if (response == null) {
+          recordThrowableResponse(throwable);
+        }
       }
       
       Optional.ofNullable(response).ifPresent(this::recordResponse);
@@ -121,8 +125,12 @@ public class HttpAsyncRequestStep extends Step {
       }
       HttpResponseStep.addToPacket(packet, response);
     }
-  }
 
+    private void recordThrowableResponse(Throwable throwable) {
+      LOGGER.warning(MessageKeys.HTTP_REQUEST_GOT_THROWABLE, request.method(), request.uri(), throwable.getMessage());
+      HttpResponseStep.addToPacket(packet, throwable);
+    }
+  }
 
   private static CompletableFuture<HttpResponse<String>> createFuture(HttpRequest request) {
     return httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofString());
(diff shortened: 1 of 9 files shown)


operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
```diff
@@ -108,7 +108,11 @@ private void checkTimeout(AsyncFiber fiber) {
 
     private void resume(AsyncFiber fiber, HttpResponse<String> response, Throwable throwable) {
       if (throwable != null) {
```
russgold, line 110: This method now has two different tests for response being null - one in the block with throwable != null and one in the Optional.ofNullable(response) logic.

It would probably be better to merge them or otherwise simplify the logic. There are two different things being tested for null, here: `throwable` and `response`. Are all four combinations meaningful?

operator/src/main/java/oracle/kubernetes/operator/http/HttpResponseStep.java
```diff
@@ -21,7 +21,15 @@ public HttpResponseStep(Step next) {
 
   @Override
   public NextAction apply(Packet packet) {
-    return Optional.ofNullable(getResponse(packet)).map(r -> doApply(packet, r)).orElse(doNext(packet));
+    return Optional.ofNullable(getResponse(packet))
+        .map(r -> doApply(packet, r))
```
russgold, line 25: This logic has gotten complex, with two separate `orElse` cases. Can the contents of the first one be extracted to a new method?

operator/src/test/java/oracle/kubernetes/operator/http/HttpAsyncRequestStepTest.java
```diff
@@ -118,6 +126,19 @@ public void whenErrorResponseReceived_logMessage() {
...
+    Memento m = TestUtils.silenceOperatorLogger()
+        .collectLogMessages(logRecords, HTTP_REQUEST_GOT_THROWABLE)
+        .withLogLevel(Level.WARNING)
+        .ignoringLoggedExceptions(HttpAsyncRequestStep.HttpTimeoutException.class);
+    final NextAction nextAction = requestStep.apply(packet);
+
+    completeWithThrowableBeforeTimeout(nextAction, new Throwable("Test"));
+
+    assertThat(logRecords, containsWarning(HTTP_REQUEST_GOT_THROWABLE));
+    m.revert();
```
russgold, line 140: All mementos should be reverted in the `tearDown` method; among other things, that ensures that if the test exceptions out, the revert still happens, so other tests are not affected. It appears that you want to collect an additional log messages. The simplest thing would be simply to add it to the original least created during the `setUp` call. If for some reason you cannot do that, take advantage of the fact that the memento itself is saved as `consoleMemento` and just call its `collectLogMessages` method, here.

operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
doxiao: > This method now has two different tests for response being null - one in the block with throwable != null and one in the Optional.ofNullable(response) logic.
> 
> It would probably be better to merge them or otherwise simplify the logic. There are two different things being tested for null, here: `throwable` and `response`. Are all four combinations meaningful?

The only permutation that does not meaningful is both null. Let me try to refactor this method a little.

operator/src/test/java/oracle/kubernetes/operator/http/HttpAsyncRequestStepTest.java
doxiao: I started with adding this new memento to the list in setUp method, but it did not work. Let me try consoleMemento suggestion.

russgold (approved):

rjeberhard (approved):

PR #2254, opened 2021-03-09
Trust all certificates when connecting to the management REST interface from the operator

rjeberhard:
Creating as a draft for visibility to @russgold and @doxiao.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/http/HttpAsyncRequestStep.java
@@ -8,9 +8,14 @@ import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.security.KeyManagementException;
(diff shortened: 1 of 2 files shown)


russgold (commented):
It looks a bit more than is needed. This is what I have seen recommended (and use now in the exporter):

```
        private SSLContext createSSLContext() throws GeneralSecurityException {
            return SSLContexts.custom()
                .loadTrustMaterial(null, createAcceptingTrustStrategy())
                .build();
        }

        private TrustStrategy createAcceptingTrustStrategy() {
            return (cert, authType) -> true;
        }
```

rjeberhard:
@russgold, I thought that the exporter was using Apache client. This is using the JDK's client. I've tried with less than what I have here and it didn't work.

russgold:
Yes, the exporter uses the Apache client - which uses the JDK's security mechanisms. Did you try what I suggested, just setting up the SSLContext as described?

rjeberhard:
The SSLContexts class is from the Apache library. I looked through the code behind that call and it's fairly similar to what I have.

russgold (approved):

PR #2257, opened 2021-03-10
Owls 87956 - Generate shorter volume name when override secret name is too long

ankedia:
OWLS-87956 - K8s volume name has a limit of 63 characters. For override secrets, the operator generates volume name by appending "-volume" to the secret name. If the volume name length exceeds the limit, the volume fails to be mounted.  
This change creates a shorter volume name (based on the ordinal suffix) for the override secrets when the secret name exceeds the limit and uses the old naming scheme when the secret name length is less than the limit. 
Added new unit tests in the JobHelperTest and integration test results are at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4264/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
@@ -9,6 +9,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import io.kubernetes.client.openapi.models.V1ConfigMapVolumeSource;
 import io.kubernetes.client.openapi.models.V1Container;
@@ -44,6 +45,10 @@ public abstract class JobStepContext extends BasePodStepContext {
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
   private static final String WEBLOGIC_OPERATOR_SCRIPTS_INTROSPECT_DOMAIN_SH =
         "/weblogic-operator/scripts/introspectDomain.sh";
+  private static final int MAX_ALLOWED_VOLUME_NAME_LENGTH = 63;
+  private final AtomicInteger volumeIndex = new AtomicInteger(1);
+  private final AtomicInteger mountIndex = new AtomicInteger(1);
+  public static final String VOLUME_NAME_SUFFIX = "-volume";
   private V1Job jobModel;
 
   JobStepContext(Packet packet) {
@@ -313,14 +318,14 @@ public abstract class JobStepContext extends BasePodStepContext {
   private void addConfigOverrideSecretVolume(V1PodSpec podSpec, String secretName) {
     podSpec.addVolumesItem(
           new V1Volume()
-                .name(secretName + "-volume")
+                .name(getVolumeName(secretName))
                 .secret(getOverrideSecretVolumeSource(secretName)));
   }
 
   private void addConfigOverrideVolume(V1PodSpec podSpec, String configOverrides) {
     podSpec.addVolumesItem(
           new V1Volume()
-                .name(configOverrides + "-volume")
+                .name(configOverrides + VOLUME_NAME_SUFFIX)
                 .configMap(getOverridesVolumeSource(configOverrides)));
   }
 
@@ -331,7 +336,7 @@ public abstract class JobStepContext extends BasePodStepContext {
   private void addWdtConfigMapVolume(V1PodSpec podSpec, String configMapName) {
     podSpec.addVolumesItem(
         new V1Volume()
-            .name(configMapName + "-volume")
+            .name(configMapName + VOLUME_NAME_SUFFIX)
             .configMap(getWdtConfigMapVolumeSource(configMapName)));
   }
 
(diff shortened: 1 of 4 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -313,14 +318,14 @@ protected V1PodSpec createPodSpec(TuningParameters tuningParameters) {
...
-                .name(secretName + "-volume")
+                .name(getVolumeName(secretName))
                 .secret(getOverrideSecretVolumeSource(secretName)));
   }
 
   private void addConfigOverrideVolume(V1PodSpec podSpec, String configOverrides) {
     podSpec.addVolumesItem(
           new V1Volume()
-                .name(configOverrides + "-volume")
+                .name(configOverrides + VOLUME_NAME_SUFFIX)
```
rjeberhard, line 328: In the spirit of fixing all of the related problems, I think we should make the same change to volume items that include the name of a config map as well. Config maps are another example of a resource that can have names longer than 63 characters.

Let's please look at PodHelper/PodStepContext too to see if we automatically generate volume/volumeMount names based on resource names.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -389,6 +394,19 @@ protected V1Container createPrimaryContainer(TuningParameters tuningParameters)
...
+    return getName(secretName, volumeIndex) + VOLUME_NAME_SUFFIX;
+  }
+
+  private String getVolumeMountName(String secretName) {
+    return getName(secretName, mountIndex) + VOLUME_NAME_SUFFIX;
+  }
+
+  private String getName(String secretName, AtomicInteger index) {
+    return secretName.length() > (MAX_ALLOWED_VOLUME_NAME_LENGTH - VOLUME_NAME_SUFFIX.length())
+            ? "long-secret-name-" + index.getAndIncrement() : secretName;
```
rjeberhard, line 407: Instead of "long-secret-name-" can we please use something simpler like "volume-" or as much of the secretName that we have room for (eliminating the end of the name).

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -500,6 +517,57 @@ public void whenDomainHasAdditionalVolumesWithCustomVariablesInvalidValue_jobNot
...
+
+    assertThat(getJobVolumes(), hasSecretVolume(VOLUME_NAME_FOR_LONG_SECRET_NAME, LONG_SECRET_NAME, DEFAULT_MODE));
+    assertThat(getJobVolumeMounts(), hasVolumeMount(VOLUME_NAME_FOR_LONG_SECRET_NAME,
+            "/weblogic-operator/config-overrides-secrets/" + LONG_SECRET_NAME, true));
+  }
+
+  @Test
+  public void whenDomainHasMultipleConfigOverrideSecretsWithLongNames_volumeCreatedWithShorterNames() {
+    resourceLookup.defineResource(LONG_SECRET_NAME, KubernetesResourceType.Secret, NS);
+    resourceLookup.defineResource(SECOND_LONG_SECRET_NAME, KubernetesResourceType.Secret, NS);
```
jshum2479, line 537: Probably want to add multiple long named secrets where the first 63 characters are the same.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
ankedia: We suffix the volume names with "-volume" and I thought to keep the same suffix for the secrets volume names where the name exceeds  63 characters limit (for consistency). I will take the first 50 characters of the secret name and then leave the remaining characters for the ordinal/index of the volume in case there are multiple secrets having the same first 50 characters.

russgold: And you have to make sure that modifying the domain doesn't create a different set up mappings based on the order in which the secret/volume names are shortened. One way to do that might be to compute a checksum for the original name, and then use that as part of the generated name, rather than an index.

ankedia: I have made the change to compute a checksum for the original name and use that as part of the generated name.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
ankedia: Done

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
ankedia: Done. I looked PodHelper/PodStepContext but didn't find any place where we automatically generate volume/volumeMount names based on resource names.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -389,6 +394,26 @@ protected V1Container createPrimaryContainer(TuningParameters tuningParameters)
     return container;
   }
 
+  private String getVolumeName(String resourceName, String type) {
```
russgold, line 397: If these methods do the same thing, why do you need two of them?

russgold (approved):
Just had one comment; otherwise, looks good.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
ankedia: It was leftover from the previous implementation and I thought having methods with different names might provide better readability. I went ahead and removed it.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -389,6 +394,22 @@ protected V1Container createPrimaryContainer(TuningParameters tuningParameters)
...
+
+  private String getName(String resourceName, String type) {
+    return resourceName.length() > (MAX_ALLOWED_VOLUME_NAME_LENGTH - VOLUME_NAME_SUFFIX.length())
+            ? getShortName(resourceName, type)
+            : resourceName + VOLUME_NAME_SUFFIX;
+  }
+
+  private String getShortName(String resourceName, String type) {
+    String volumeSuffix = VOLUME_NAME_SUFFIX + "-" + type + "-"
+            + Optional.ofNullable(ChecksumUtils.getMD5Hash(resourceName)).orElse("");
```
rjeberhard, line 409: When would the result of getMD5Hash be null?

ankedia: It can be null when `MessageDigest.getInstance("MD5")` method (in ChecksumUtils) throws `NoSuchAlgorithmException` or `String.getBytes()` method throws `UnsupportedEncodingException`.

rjeberhard (approved):

PR #2259, opened 2021-03-11
add retry count and maximum retry count to retrying event

doxiao:
Add retry count and maximum retry count to DomainProcessingRetrying event message, and add necessary unit test case in EventHelperTest.java.

Here are examples of generated events with the change and without the change.

48s         Normal    DomainProcessingRetrying   domain/domain1                   Retrying processing of domain resource domain1, retry 2 of a maximum 5

95s         Normal    DomainProcessingRetrying   domain/domain1                   Retrying the processing of domain resource domain1 after one or more failed attempts

Note that operator aggregates "same" events into a single event object with a count that indicates how many such events have been generated. Before this change, all DomainProcessingRetrying events for a particular domain are the "same" because they all have the same event message. With this change, the DomainProcessingRetrying events with different retry counts are considered "different" so we will see more event objects. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4260/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/EventConstants.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/EventConstants.java
@@ -29,9 +29,9 @@ public interface EventConstants {
   String DOMAIN_PROCESSING_FAILED_PATTERN
       = "Failed to complete processing domain resource %s due to: %s, the processing will be retried if needed";
   String DOMAIN_PROCESSING_RETRYING_PATTERN
-      = "Retrying the processing of domain resource %s after one or more failed attempts";
+      = "Retrying processing of domain resource %s, retry %s of a maximum %s";
   String DOMAIN_PROCESSING_ABORTED_PATTERN
-      = "Aborting the processing of domain resource %s permanently due to: %s";
+      = "Aborting processing of domain resource %s permanently due to: %s";
   String DOMAIN_VALIDATION_ERROR_PATTERN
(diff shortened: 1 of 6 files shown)


operator/src/main/java/oracle/kubernetes/operator/EventConstants.java
```diff
@@ -29,9 +29,9 @@
   String DOMAIN_PROCESSING_FAILED_PATTERN
       = "Failed to complete processing domain resource %s due to: %s, the processing will be retried if needed";
```
tbarnes-us, line 30: "if needed" seems vague

when isn't it needed?
how many retry attempts will there be?

tbarnes-us: I've been informed in slack that current behaviors are not well known and are subject to change.
If the above comment can't easily be addressed, can you file a follow on JIRA that (a) references the pending retry work, this retry work, and this pull, (b) describes enhancing this particular error message, and finally (c) asks to enhance  events so that they include the expected retry interval?

doxiao: Apparently this task depends on OWLS-84438. It seems premature to work on this. Based on a discussion offline, we decided to hold on proceeding this PR, and resume working on this after OWLS-84438 is resolved. I am closing this PR.

PR #2265, opened 2021-03-15
Owls 87012 - Domain processing completed event generated after domain scaling operation integration test

ankedia:
Integration test to verify that domain processing completed event is generated after the domain is scaled. The test performs scale down and scale up Integration test run URL - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4285. I'm rerunning the integration test and update the latest URL when the test finishes.

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
@@ -89,6 +89,7 @@ import static oracle.weblogic.kubernetes.utils.K8sEvents.DOMAIN_VALIDATION_ERROR
 import static oracle.weblogic.kubernetes.utils.K8sEvents.NAMESPACE_WATCHING_STARTED;
 import static oracle.weblogic.kubernetes.utils.K8sEvents.NAMESPACE_WATCHING_STOPPED;
 import static oracle.weblogic.kubernetes.utils.K8sEvents.checkDomainEvent;
+import static oracle.weblogic.kubernetes.utils.K8sEvents.getDomainEventCount;
 import static oracle.weblogic.kubernetes.utils.K8sEvents.getEventCount;
 import static oracle.weblogic.kubernetes.utils.ThreadSafeLogger.getLogger;
 import static oracle.weblogic.kubernetes.utils.WLSTUtils.executeWLSTScript;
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
```diff
@@ -374,11 +374,44 @@ public void testDomainK8sEventsScalePastMax() {
...
+    try {
+      scaleDomainAndVerifyCompletedEvent(1, "scale down", true);
+      scaleDomainAndVerifyCompletedEvent(2, "scale up", true);
+    } finally {
+      scaleDomain(2);
+    }
+  }
+
+  private void scaleDomainAndVerifyCompletedEvent(int replicaCount, String testType, boolean verify) {
+    DateTime timestamp = new DateTime(Instant.now().getEpochSecond() * 1000L);
```
doxiao, line 394: I am not sure if this timestamp is too coarse-grained to make the verification of the event's timestamp after the two scale operations reliable. We could also get the event's count before and after each scale operation, and make sure that the count is increased by 1 each time.

ankedia: We get the current timestamp after each scaling operation. The test logic is as follows -
1. Get the current `timestamp`.
2. Scale down the domain 
3. Verify that `DomainProcessingComplete` event is generated after `timestamp` from step 1.
4. Get the current `timestamp` again.
5. Scale up the domain
6. Verify that `DomainProcessingComplete` event is generated after `timestamp` from step 4.

I thought we follow a similar validation approach in other test methods and am not sure if getting the event count is necessary.

doxiao: My concern is that the timestamp at step #1 and step #4 could be the same since the timestamp only gets the seconds. Either way, adding count in the verification is a good idea. I don't see it is tested anywhere.

ankedia: I have adding count verification in b0079f4

doxiao (approved):
LGTM. Thanks.

rjeberhard (approved):

PR #2267, opened 2021-03-17
Use JDK 16 and Kubernetes Java Client 12.0.0

rjeberhard:
* JDK 16 is now GA
* Kubernetes Java Client 12.0.0 now available: the big change is that use of Joda library is removed in favor of standard java.time.* types

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
@@ -3,6 +3,8 @@
 
 package oracle.kubernetes.operator;
 
+import java.time.OffsetDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -223,7 +225,7 @@ public class JobWatcher extends Watcher<V1Job> implements WatchListener<V1Job>, 
   }
 
   private class WaitForJobReadyStep extends WaitForReadyStep<V1Job> {
-    private final long jobCreationTime;
+    private final OffsetDateTime jobCreationTime;
 
     private WaitForJobReadyStep(V1Job job, Step next) {
       super(job, next);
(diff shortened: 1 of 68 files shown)


operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -319,7 +319,7 @@ public String toString() {
 
     private long getJobStartedSeconds() {
       if (job.getStatus() != null && job.getStatus().getStartTime() != null) {
-        return (System.currentTimeMillis() - job.getStatus().getStartTime().getMillis()) / 1000;
+        return (System.currentTimeMillis() - job.getStatus().getStartTime().toEpochSecond()) / 1000;
```
rjeberhard, line 322: Need to fix this... Don't need to divide by 1000.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ObjectPatch.java
```diff
@@ -250,15 +247,12 @@ void addField(JsonPatchBuilder builder, String path, String newValue) {
...
-
-    DateTimeField(String name, Function<T, DateTime> getter) {
+    DateTimeField(String name, Function<T, OffsetDateTime> getter) {
       super(name, a -> toString(getter.apply(a)));
     }
 
-    private static String toString(DateTime dateTime) {
-      return Optional.ofNullable(dateTime).map(DATE_FORMAT::print).orElse(null);
+    private static String toString(OffsetDateTime dateTime) {
+      return Optional.ofNullable(dateTime).map(OffsetDateTime::toString).orElse(null);
```
rjeberhard, line 255: I think I did this, but double check that OffsetDateTime toString() is the correct RFC 3339 format.

ddsharpe (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesUtilsTest.java
```diff
@@ -16,34 +18,40 @@
 
   @Test
   public void whenCreationTimesDiffer_metadataWithLaterTimeIsNewer() {
-    V1ObjectMeta meta1 = new V1ObjectMeta().creationTimestamp(new DateTime(2)).resourceVersion("2");
-    V1ObjectMeta meta2 = new V1ObjectMeta().creationTimestamp(new DateTime(1)).resourceVersion("1");
+    V1ObjectMeta meta1 = new V1ObjectMeta().creationTimestamp(
```
russgold, line 21: This is now more difficult to read than before. I'd recommend extracting  a method, such as:
```
private OffsetDateTime createTimeStamp(long time) {
    return OffsetDateTime.ofInstant(Instant.ofEpochSecond(time), ZoneId.of("UTC"));
}
``` 
and use that in these expressions.

PR #2279, opened 2021-03-25
Prevent pod roll on operator upgrade

russgold:
This is a revised implementation of the means by which the operator will avoid rolling pods when upgrading. Instead of preserving the old SHA256 hash values, it adds a version label to pods and removes the MII zip hash and Prometheus annotations from the SHA256 computation - and then uses the presence of the version label to recognize a pre-3.2 pod. This requires an alternative computation of the hash, to duplicate what was used in the previous release.

In such a case, if it can reuse the pod, it patches it to look like one created by the current operator.


jshum2479 (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -104,13 +109,20 @@
   }
 
   private static boolean isPatchableItem(Map.Entry<String, String> entry) {
-    return isCustomerItem(entry) || entry.getKey().equals(INTROSPECTION_STATE_LABEL);
+    return isCustomerItem(entry) || PATCHABLE_OPERATOR_KEYS.contains(entry.getKey());
   }
 
+  private static final Set<String> PATCHABLE_OPERATOR_KEYS
+        = Set.of(INTROSPECTION_STATE_LABEL, OPERATOR_VERSION, MODEL_IN_IMAGE_DOMAINZIP_HASH, SHA256_ANNOTATION);
```
rjeberhard, line 116: @jshum2479, can you please confirm whether we would want to patch MODEL_IN_IMAGE_DOMAINZIP_HASH? The use case would be that we've decided not to roll an existing pod, but the pod's value for this label doesn't match the pod model. I believe this means that we are in a dynamic update use case and the dynamic update was successful. Given that, it's probably fine to either patch this value or not patch this value; however, I'd like your opinion. I'm confident that Russ' change to patch OPERATOR_VERSION and SHA256_ANNOTATION are appropriate.

rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
jshum2479: We need patch it for the use case you described,  the code that checks for whether domain zip has changed to reuse the pod / patch the pod depends on the domain zip hash label and the result coming from the introspector

PR #2280, opened 2021-03-25
fix script to show online update is disabled and update sample script…

jshum2479:
This will fix ItSample update 4 cases

### DIFF
--- a/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/patch-enable-online-update.sh
+++ b/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/patch-enable-online-update.sh
@@ -54,7 +54,7 @@ set -o pipefail
 
 echo "@@ Info: Patching domain '${DOMAIN_UID}' in namespace '${DOMAIN_NAMESPACE}' to enable onlineUpdate to 'true'."
 kubectl -n ${DOMAIN_NAMESPACE} patch domain ${DOMAIN_UID} --type='json' \
-  -p='[{"op": "replace", "path": "/spec/configuration/model/onlineUpdate", "value": {"enabled" : 'true'} }]'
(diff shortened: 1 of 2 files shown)


rjeberhard (approved):

anpanigr (commented):
Please update the PR description with Jenkin result

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/patch-enable-online-update.sh
```diff
@@ -54,7 +54,7 @@ set -o pipefail
 
 echo "@@ Info: Patching domain '${DOMAIN_UID}' in namespace '${DOMAIN_NAMESPACE}' to enable onlineUpdate to 'true'."
 kubectl -n ${DOMAIN_NAMESPACE} patch domain ${DOMAIN_UID} --type='json' \
-  -p='[{"op": "replace", "path": "/spec/configuration/model/onlineUpdate", "value": {"enabled" : 'true'} }]'
+  -p='[{"op": "add", "path": "/spec/configuration/model/onlineUpdate", "value": {"enabled" : 'true'} }]'
```
tbarnes-us, line 57: (1) Why is this change needed?  "replace" worked for me in all three cases:
```
   model:

   model:
      onlineUpdate:

   model:
      onlineUpdate: false
``` 

(2) The documentation at ./content/samples/simple/domains/model-in-image/update4.md also needs to be updated to match whatever we change here.  (The doc mentions the same 'patch domain' command.)

jshum2479: ok, then we don't need this anymore, I was testing few days ago and changing add from replace let the integration tests passed.  I will back this out

jshum2479: revered

tbarnes-us (approved):

PR #2285, opened 2021-03-27
OWLS 86115: Model in Image should not use operator generated configuration overrides

lennyphan:
---

**NOTE:** This pull is superseded by pull https://github.com/oracle/weblogic-kubernetes-operator/pull/2344

---

In this PR:

1. For Domain home source type MII, configuration overrides are generated using WDT model filter.  Model filter implementation follows the same logic for generating situational configuration in introspectDomain.py's SitConfigGenerator. Domain home source types Domain-on-PV and Domain-in-Image will continue to use situational configuration overrides.

2. Added ability to run python based unit tests for model filter implementation.

Passed all MII integration tests except for the single test failure: oracle.weblogic.kubernetes.ItMiiDynamicUpdate.testMiiRemoveTarget  which matches Integration Tests Nightly Results as of (03/26)


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -811,6 +811,10 @@ void addStartupEnvVars(List<V1EnvVar> vars) {
     addEnvVar(vars, ServerEnvVars.AS_SERVICE_NAME, LegalNames.toServerServiceName(getDomainUid(), getAsName()));
     Optional.ofNullable(getDataHome()).ifPresent(v -> addEnvVar(vars, ServerEnvVars.DATA_HOME, v));
     addEnvVarIfTrue(mockWls(), vars, "MOCK_WLS");
+
+    if (getDomainHomeSourceType() == DomainSourceType.FromModel) {
+      addEnvVar(vars, ServerEnvVars.FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR, "false");
```
rjeberhard, line 816: This is going to change our pod hash (`weblogic.sha256`) which means that any MII pod that is running when the operator is upgraded will be rolled. Can this instead be set in the pod start script (`startServer.sh`).

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -767,6 +767,10 @@ def addNetworkAccessPoint(self, server, nap, is_server_template):
     nap_protocol = getNAPProtocol(nap, server, self.env.getDomain(), is_server_template)
 
     if istio_enabled == 'true':
+      name = nap.getName()
+      if name.startswith('http-') or name.startswith('tcp-') or name.startswith('tls-') or name.startswith('https-'):
```
rjeberhard, line 771: Could this be fragile? The customer could have a nap name that starts with one of these prefixes. Do we have some earlier validation that ensures that the only names with these prefixes would be naps that were edited by our Istio functionality?

operator/src/test/python/test_wdt_mii_filter.py
```diff
@@ -0,0 +1,243 @@
+import os
```
rjeberhard, line 1: Putting comment here... Is __init__.py supposed to be empty?

rjeberhard:
Wow, @lennyphan, this is a ton of work. I think I'm going to need a "show-and-tell" since I don't know enough to review the filter and changes to the introspector script. I want to make sure that you've accounted for all of the situational configuration and Istio changes. I suspect that you have, but can't tell.

On the sit cfg environment variable, unfortunately, we've just been through learning and relearning why kind of changes cause operator upgrade bugs (this is where upgrading the operator causes problems to already running servers, including unexpected restarts). Anything that changes the generated pod model, including adding environment variables to the container, will do this. As noted, you could instead change the startServer.sh script that we inject. 

I've delayed releasing 3.2.0 until we resolve one other upgrade bug, which will likely be done late Monday or Tuesday, so we could potentially get this change in :). I will be really happy to tell people about this improvement!

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
lennyphan: Just to clarify the scenario you're describing for my understanding, this is when upgrading the Operator from a previous MII domain that uses sit config overrides to this version which uses the model filter?  The one place that uses this environment is in startNodeManager.sh (as a server startup argument) and I couldn't determine how we know the domain source type is 'FromModel' (MII).  serverStart.sh does:

if [ -f /weblogic-operator/introspector/domainzip.secure ]; then
...
fi

I think this should work.  Let me make the change and verify.

operator/src/test/python/test_wdt_mii_filter.py
lennyphan: I'm not a python expert but my understanding is that an empty file named "__init__.py" file marks the directory as a python package/module and I added it for python unittest discovery and test execution.  Not currently using any third party tools for unit tests and version of python in our env seems to be 2.7 so using the most basic constructs to run unit tests.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
rjeberhard: Right, a customer is using operator 3.1.4 and has a MII domain. The admin server and some number of managed servers are running. The customer upgrades to 3.2.0 containing your work. I would expect that pods do not roll simply because I upgrade the operator.

Thinking more deeply about this... The change to add the extra environment variable would have caused the existing pods to roll. I think that we have a workaround in that we can set this in startServer.sh. However, we likely have an earlier issue... Your change to not generate config overrides will mean that the generated domain home will be different. This change would also cause existing pods to roll, correct? If so, we should talk together with @jshum2479 and see if we can come up with a way to roll this out so that existing pods are not rolled or restarted, but newly created pods come up with the new pattern.

operator/src/test/python/test_wdt_mii_filter.py
rjeberhard: Sounds good.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
jshum2479: We can add logic in start script to reset the env when the DOMAIN_SOURCE_TYPE is FromModel.

tbarnes-us: It doesn't look like it'd hurt to leave it alone. It'd become a NO-OP.

operator/src/test/java/oracle/kubernetes/operator/helpers/ConfigMapHelperTest.java
```diff
@@ -69,6 +69,7 @@
     "model-diff-v1.py",
     "modelInImage.sh",
     "model-wdt-create-filter.py",
+    "wdt_mii_filter.py",
```
tbarnes-us, line 72: consider renaming to start with "model" and to use "-" instead of "_". This follows the pattern of other mii related scripts, and ensures model related scripts sort next to each-other in an `ls`.

operator/src/test/python/test_wdt_mii_filter.py
```diff
@@ -0,0 +1,243 @@
...
+import unittest
+import yaml
```
jshum2479, line 3: Which interpreter are we using here for package yaml?

operator/src/main/resources/scripts/wdt_mii_filter.py
```diff
@@ -0,0 +1,546 @@
...
+    self.ACCESS_LOG_IN_LOG_HOME   = self.getEnvOrDef('ACCESS_LOG_IN_LOG_HOME', 'true')
+    self.DATA_HOME                = self.getEnvOrDef('DATA_HOME', "")
+    self.CREDENTIALS_SECRET_NAME  = self.getEnv('CREDENTIALS_SECRET_NAME')
+
+    # initialize globals
+    self.INTROSPECT_HOME          = '/tmp/introspect/' + self.DOMAIN_UID
+    self.CREDENTIALS_SECRET_PATH = self.getEnvOrDef('CREDENTIALS_SECRET_PATH', '/weblogic-operator/secrets')
+
+    if model and 'topology' in model:
+      self.DOMAIN_NAME = model['topology']['Name']
```
tbarnes-us, line 40: I assume we always expects domain-name for the MII case?  If so, maybe have this script error out if it's not found...

operator/src/main/resources/scripts/wdt_mii_filter.py
```diff
@@ -0,0 +1,546 @@
+# Copyright (c) 2018, 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+# ------------
+# Description:
+# ------------
+#  This is a WDT filter for handling WLS configuration overrides.
```
tbarnes-us, line 7: (1) Minor comment: This is a model-in-image WDT filter for overriding WLS configuration, it replaces 'situational configuration overrides'.

(2) Less minor comment:  Please expand the description to describe (at a relatively high level) this script's command line, env var, and file inputs, plus to describe its expected output in the success and error cases...

operator/src/main/resources/scripts/wdt_mii_filter.py
```diff
@@ -0,0 +1,546 @@
...
+#
+
+import inspect
+import os
+import sys
```
tbarnes-us, line 12: General comment #1: I suspect the life-cycle and/or architecture documentation describes the propagation of internal sit-cfg.  If so, then these docs should be updated.

General comment #2: Does this change lift any of the restrictions detailed in 'runtime-updates.md' for MII?  If so, then the doc should be updated. Note that if the update includes adding support for shrinking a cluster we (a) may want to have integration tests before officially documenting support, (b) may want to make sure the documentation clearly states that JTA and JMS are not particularly amenable to such shrinks.

operator/src/test/python/test_wdt_mii_filter.py
lennyphan: I've been running with 2.7.1 and I believe that's the version running with WLST.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
jshum2479: The `FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR` is for the core config override logic to failed to start the server if the override file is wrong, the start server script already have the logic to set it to true if the environment is not set,  the env is for the case if user choose to override it.  

I am not seeing a reason to set it for mii since we are using override and it will cause the roll because of the pod spec hash changed because the newly introduced environment variable.

tbarnes-us: The overall point is that there's no need to _change_ the current setting in this pull.  We don't care what it's set to in the new path...

lennyphan: in startNodeManager.sh, we set FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR system property to 'true' by default.  Managed servers fail to start/boot with the following error:

<Mar 29, 2021 8:22:23 PM GMT> <Critical> <Management> <BEA-141327> <The SitConfigRequired configuration element is set to true, but no situational config files were found.> 
<Mar 29, 2021 8:22:23 PM GMT> <Critical> <WebLogicServer> <BEA-000362> <Server failed. Reason: 141327> 
<Mar 29, 2021 8:22:23 PM GMT> <Notice> <WebLogicServer> <BEA-000365> <Server state changed to FAILED.> 

The validation is here:
if (isSitConfigRequired() && unexpiredFiles.isEmpty()) {
        if (debugLogger.isDebugEnabled()) {
          debugLogger.debug( "[SitConfig] isSitConfigRequired is true and no Sit Config Files exist. Failing boot.");
        }
        String l = ManagementLogger.logSituationalConfigRequired();
        throw new ServiceFailureException(l);
}

So we should either set it to 'false' or not define it for MII.

operator/src/test/java/oracle/kubernetes/operator/helpers/ConfigMapHelperTest.java
lennyphan: Seems python doesn't like packages/modules using hyphen/dashes in the name when attempting to import:

ERROR: test_wdt_mii_filter (unittest.loader.ModuleImportFailure)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_wdt_mii_filter
Traceback (most recent call last):
  File "/usr/lib64/python2.7/unittest/loader.py", line 252, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib64/python2.7/unittest/loader.py", line 230, in _get_module_from_name
    __import__(name)
  File "/scratch/lphan/weblogic-kubernetes-operator/operator/src/test/python/test_wdt_mii_filter.py", line 5
    import model-wdt-mii-filter
                ^
SyntaxError: invalid syntax

What do you think about renaming to 'model_wdt_mii_filter.py'?

operator/src/test/python/test_wdt_mii_filter.py
jshum2479: I think anyone who tries to build it needs to update the python to install the yaml package.  I got import error - No module named yaml.

lennyphan: Thanks.  Any suggestions on alternative?  I know there is a yaml parser in WDT.

rjeberhard: In addition to `topology.yaml`, we now also have `topology.json`. @ankedia added this to make it easier to consume the topology from scripts. Could you use this and `jq`?

jshum2479: I think this is the first time we introduce unit testing python into the test. WDT has it's own yaml parser,  I don't think we want to introduce the dependency between two projects.  If we agreed to build time python requirement, then we can just make it a requirement.  @rjeberhard ?

rjeberhard: Can the build or test suite verify that you have the right dependency installed? Can we verify that it's possible to build on both Oracle Linux and MacOS?

rjeberhard: MacOS comes with Python 2.7 by default and it's possible, but a pain, to replace this with the Homebrew install of Python 3. What does this work require?

operator/src/main/resources/scripts/introspectDomain.py
lennyphan: I could be wrong but I couldn't find any validation for naps that have these prefixes edited by our Istio functionality. @jshum2479 is there validation for these naps that I didn't see?

operator/src/main/resources/scripts/wdt_mii_filter.py
lennyphan: For online updates, I noticed that WDT updateDomain command is processing a sparse model fragment where the topology doesn't specify the 'Name' attribute (domain-name).  The only configuration override that this is verified in is for the domain level log file/path.  I just recently ran into this while running ItMiiDynamicUpdate integration test and so we don't generate the override for this sparse model processing.  Unfortunately, I don't see the domain name set as an environment variable (or I missed it) when running introspection and the current introspectDomain.py gets the domain name from the domain config.xml.  Any suggestion for an alternative place to get the domain name?

tbarnes-us: For the update case, I think the operator pod may be providing the domain name in the DOMAIN_NAME env var, as I think the Operator sets the env var when re-running the introspector (it gets the information from the 'topology' that is output from the first introspectdor run).  Alternatively, I think the 'topology' is available in the introspect cm, which is generated based on the output of the first run of the introspector...

tbarnes-us:
General comment: We should update the 'runtime update' user doc to reflect that this will lift some of the restrictions.  For example sit-cfg conflict with logHome (and not documented it turns out for dataHome).  Also perhaps other restrictions. Johnny, you, and I should probably read through the 'update' doc together to see what should be changed. We will need integration tests for each lifted restriction before we document support.

operator/src/test/java/oracle/kubernetes/operator/helpers/ConfigMapHelperTest.java
tbarnes-us: Sure.

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -102,9 +102,11 @@ if [ $? != 0 ]; then
   exit $RETVAL
 fi
 
-copySitCfgWhileRunning /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
-copySitCfgWhileRunning /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jms         'Sit-Cfg-JMS--'
-copySitCfgWhileRunning /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jdbc        'Sit-Cfg-JDBC--'
-copySitCfgWhileRunning /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/diagnostics 'Sit-Cfg-WLDF--'
+if [ ! -f /weblogic-operator/introspector/domainzip.secure ]; then
```
tbarnes-us, line 105: This looks like a fragile way to check if MII is in the picture - it relies on a relatively obscure internal behavior. Instead, how about checking the DOMAIN_SOURCE_TYPE env var? (You'll need to make sure that this env var is passed to all servers - the operator might only be passing it to the introspector pod right now...)

rjeberhard: Agreed to check if the env var is present and use that... If it is, unfortunately, not present then I won't want to add it as that would cause servers to roll on operator upgrade (unless we are very careful to add the env. var after the hash... sigh -- hopefully, it is already there)

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -793,6 +794,13 @@ function generateMergedModel() {
...
 
+  local wdtArgs=""
+  wdtArgs+=" -oracle_home ${ORACLE_HOME}"
+  wdtArgs+=" ${model_list} ${archive_list} ${variable_list}"
+  wdtArgs+=" -domain_type ${WDT_DOMAIN_TYPE}"
+
+  trace "About to call '${WDT_BINDIR}/validateModel.sh ${wdtArgs}'."
+
   ${WDT_BINDIR}/validateModel.sh -oracle_home ${ORACLE_HOME} ${model_list} \
     ${archive_list} ${variable_list}  -domain_type ${WDT_DOMAIN_TYPE}  > ${WDT_OUTPUT} 2>&1
```
tbarnes-us, line 805: If the call used the same 'wdtArgs' as the trace, then this would ensure they're in sync.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -917,6 +925,15 @@ function wdtUpdateModelDomain() {
...
+  wdtArgs+=" -oracle_home ${ORACLE_HOME}"
+  wdtArgs+=" -domain_home ${DOMAIN_HOME}"
+  wdtArgs+=" ${model_list} ${archive_list} ${variable_list}"
+  wdtArgs+=" -domain_type ${WDT_DOMAIN_TYPE}"
+  wdtArgs+=" ${UPDATE_RCUPWD_FLAG}"
+
+  trace "About to call '${WDT_BINDIR}/updateDomain.sh ${wdtArgs}'."
+
   ${WDT_BINDIR}/updateDomain.sh -oracle_home ${ORACLE_HOME} -domain_home ${DOMAIN_HOME} $model_list \
   ${archive_list} ${variable_list}  -domain_type ${WDT_DOMAIN_TYPE}  ${UPDATE_RCUPWD_FLAG}  >  ${WDT_OUTPUT} 2>&1
```
tbarnes-us, line 938: If the call used the same 'wdtArgs' as the trace, then this would ensure they're in sync.

operator/src/main/resources/scripts/model_wdt_mii_filter.py
```diff
@@ -0,0 +1,560 @@
...
+def customizeServerTemplates(model):
+  topology = model['topology']
+  if 'ServerTemplate' not in topology:
+    return
+
+  serverTemplates = topology['ServerTemplate']
+  template_names = serverTemplates.keys()
+  if template_names is not None:
+    for template_name in template_names:
+     template = serverTemplates[template_name]
```
tbarnes-us, line 165: Minor comment: indents should use two spaces throughout.

operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -241,13 +241,20 @@ if [ ! "${SERVER_NAME}" = "introspector" ]; then
     [ ! $? -eq 0 ] && trace SEVERE "Could not remove stale file '$wl_state_file'." && exit 1
   fi
 
+  if [ -f /weblogic-operator/introspector/domainzip.secure ]; then
```
tbarnes-us, line 244: This check looks a little obscure/fragile to me. Instead, how about checking the DOMAIN_SOURCE_TYPE env var? (You'll need to make sure that this env var is passed to all servers - the operator might only be passing it to the introspector pod right now...)

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -102,8 +102,12 @@ function startWLS() {
 
   FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR=${FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR:-true}
   SERVER_OUT_MONITOR_INTERVAL=${SERVER_OUT_MONITOR_INTERVAL:-3}
-  if [ ${FAIL_BOOT_ON_SITUATIONAL_CONFIG_ERROR} == 'true' ] ; then
-    ${SCRIPTPATH}/monitorLog.sh ${SERVER_OUT_FILE} ${SERVER_OUT_MONITOR_INTERVAL} &
+  if [ ! -f /weblogic-operator/introspector/domainzip.secure ]; then
```
tbarnes-us, line 105: This check looks a little obscure/fragile to me. Instead, how about checking the DOMAIN_SOURCE_TYPE env var? (You'll need to make sure that this env var is passed to all servers - the operator might only be passing it to the introspector pod right now...)

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -281,10 +285,13 @@ createFolder ${DOMAIN_HOME}/servers/${SERVER_NAME}/security
 copyIfChanged /weblogic-operator/introspector/boot.properties \
               ${DOMAIN_HOME}/servers/${SERVER_NAME}/security/boot.properties
 
-copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
-copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jms         'Sit-Cfg-JMS--'
-copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jdbc        'Sit-Cfg-JDBC--'
-copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/diagnostics 'Sit-Cfg-WLDF--'
+if [ ! -f /weblogic-operator/introspector/domainzip.secure ]; then
```
tbarnes-us, line 288: This check looks a little obscure/fragile to me. Instead, how about checking the DOMAIN_SOURCE_TYPE env var? (You'll need to make sure that this env var is passed to all servers - the operator might only be passing it to the introspector pod right now...)

operator/src/main/resources/scripts/wdt_mii_filter.py
tbarnes-us: Thanks for addressing (1), can you take a look at (2)? The idea is to doc how the script fits into the flow for anyone that comes along later that needs to make a modification...

lennyphan:
> Can this cause a roll of an existing domain on upgrade? (I only ask because @rjeberhard mentioned that could happen.)

This environment property gets added through the PodStepContext.withNonHashedElements(V1Pod pod) method so it's added after the pod's hash has been calculated.  Hopefully it won't cause a roll.   I'm currently running all the integration tests to verify.  Also, currently setting up the upgrade test to verify locally.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
tbarnes-us: The point is that we're not supporting custom or internal sit-cfg in MII regardless, so there won't be any sit-cfg that can fail, so the flag should have no effect. If removing it is safe (doesn't cause a roll), then feel free to remove it. Or feel free to  leave it untouched.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -319,9 +320,9 @@ function createWLDomain() {
   fi
 
   # copy the filter related files to the wdt lib
-
   cp ${WDT_FILTER_JSON} ${WDT_ROOT}/lib/model_filters.json
   cp ${WDT_CREATE_FILTER} ${WDT_ROOT}/lib
+  cp ${WDT_MII_FILTER} ${WDT_ROOT}/lib
```
tbarnes-us, line 325: consider appending `|| trace SEVERE "cp failed" && exitOrLoop` to each of the cp commands (fail-fast)

operator/src/main/resources/scripts/wdt_mii_filter.py
tbarnes-us: Nice! Thanks!

tbarnes-us (approved):
Way cool!

rjeberhard:
@lennyphan, you need to adjust or recreate the PR to target the `main` branch.

rjeberhard:
I merged your branch with main and pushed to `OWLS-86115-main-merge`. You can merge that branch into yours and then update the target of this PR.

operator/src/main/resources/scripts/introspectDomain.py
jshum2479: This function is to write out the topology in the introspector output (not config override), it is trying to format the name in istio ways.  It's fine to leave it here if the name used by  customer already had that convention and skipping it.

rjeberhard:
This PR was replaced.

PR #2289, opened 2021-03-30
add xref

rosemarymarano:
Add x-ref to Oracle Fusion Middleware products on Kubernetes.


rjeberhard:
You will want to copy this change to the "3.2" version as well. When we go live with 3.2, it will be that content (under "3.2") that is visible on the doc. site and the content under "staging" will be for the "next" release. I'm happy to do the copy work, but want you to know what is happening :).

rjeberhard (approved):

rosemarymarano:
I will copy it to 3.2. Thanks for the heads up. I was a bit confused about exactly in which branch (under develop) to put it.

rjeberhard:
> ... confused ...

Yeah, I'm trying to lock this down, but it's also been a work in progress. My intention is for the `main` branch to be the successor to `develop` (soon!) and for both the production documentation and ongoing development to be found in that branch. The production documentation would be the highest number valued directory and ongoing, but not released, documentation would be in "staging". Of course, it's odd today because I'd already created "3.2" when Derek asked for a staging directory. In the future, I'll create that production directory as part of releasing.

rosemarymarano:
ok, pushed to 3.2 also.

PR #2292, opened 2021-03-31
Don't roll on upgrades from 3.0

russgold:
Takes into consideration that in 3.0, the prometheus annotations were computed differently than in 3.1.


rjeberhard (approved):

jshum2479 (commented):
LGTM,  the only comment I have is when a pod needs to be replaced because of domain zip hash changed, the operator logs a message "Replacing pod ... because of null"  - is there a way to put the actual reason there?

rjeberhard:
> LGTM, the only comment I have is when a pod needs to be replaced because of domain zip hash changed, the operator logs a message "Replacing pod ... because of null" - is there a way to put the actual reason there?

I'll file a bug, but I don't want to hold up an initial release. We've implemented a series of compatibility checks, which must not be complete. If none of them match then the result is "null".

russgold:
> > LGTM, the only comment I have is when a pod needs to be replaced because of domain zip hash changed, the operator logs a message "Replacing pod ... because of null" - is there a way to put the actual reason there?
> 
> I'll file a bug, but I don't want to hold up an initial release. We've implemented a series of compatibility checks, which must not be complete. If none of them match then the result is "null".

Then do file a bug and we can deal with it in the next release. I have no test for this and did not notice it.

PR #2297, opened 2021-04-04
Investigate an intermittent issue from testSameSessionStickinessUsingVoyager

hzhao-github:
Haven't had a clear run yet. However, the results from below matches nightly result from 4/06

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4598 (parallel) 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4612 (sequential)

After synced up with main branch, there is only one failure, 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4638 (parallel) 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4639 (parallel)


anpanigr (approved):
Update the description with Parallel Jenkin Result

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItStickySession.java
```diff
@@ -209,6 +210,18 @@ public void testSameSessionStickinessUsingVoyager() {
     List<String>  hostNames =
         installVoyagerIngressAndVerify(domainUid, domainNamespace, ingressName, clusterNameMsPortMap);
 
+    withStandardRetryPolicy
```
vanajamukkara, line 213: I looked at https://jira.oraclecorp.com/jira/browse/OWLS-87825. It's failing while creating ingress, my suggestion was to call create ingress in loop.

vanajamukkara (changes requested):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItStickySession.java
hzhao-github: changed

vanajamukkara (approved):
LG

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItStickySession.java
```diff
@@ -202,12 +203,41 @@ public void testSameSessionStickinessUsingVoyager() {
...
+      hostNames =
+          installVoyagerIngressAndVerify(domainUid, domainNamespace, ingressName, clusterNameMsPortMap);
+
+      if (hostNames != null && !hostNames.isEmpty()) {
+        break;
+      }
+
+      try {
+        // sometimes the ingress may not be ready even the condition check is ready, sleep a little bit
+        Thread.sleep(1000);
```
rjeberhard, line 224: Why are you using a for-loop and sleep pattern here instead of just using the Awaitility pattern used elsewhere?

hzhao-github: we are using Awaitility in installVoyagerIngressAndVerify in CommonTestUtils.java already. We just wanted to wait a little more time for the ingress to be created just for this test case without affecting all others tests, plus we need a "Callable<Boolean>" return type to use  Awaitility (I may be wrong here) but installVoyagerIngressAndVerify returns List<String>. That's why Vanaja and I chose to use a for loop here :(

PR #2301, opened 2021-04-07
Owls88611 make sure REST calls are made using the expected port and protocol

doxiao:
The changes in this PR fix an issue where, in some configurations, the operator may choose a wrong port and/or a wrong protocol (mistakenly considering a non-sure port to be secure) for making REST calls to get WLS server health states. 

The issue is caused by the following bugs in operator.
1. The code relies on the port names in a server service and WlsServerConfig#getAdminProtocolChannelName to determine if a port has "admin" privilege. This is not reliable because for example, the ssl port is exposed by the server service as "default-secure" by default, but as "https-secure" if istio is enabled. As a result, when istio is enabled, the server SSL listenport is not chosen when it should be the one to use. 
2. If there is no "admin" port to use for a server, the logic picks the first port in the server service, instead of considering the one that exposes the server listen port first. As a result, the port of a channel may be chosen even the server listen port exists. When this happens, a non-secure port may be treated as a secure port because the code treat all non default listen port as secure.

The changes fix these issues by 
1. removing the dependence to server services;
2. Use getLocalAdminProtocolChannelPort to get the right port;
3. add a validation check for the case where the default listen port, SSL port, admin port are disabled/unconfigured, and there is not a channel with admin privileges. 

Unit test cases are added to cover different configuration permutations. 

Integrations test run on Jenkins (had no additional failures except for the known failures in main): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4623/.

Jenkins run with the  latest version: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4658/.


rosemarymarano (approved):
LGTM

russgold:
I could use some more explanations, here. There are a fair number of places where we need to select the REST port, including for the health check and the metrics. WlsServerConfig.getLocalAdminProtocolChannelPort as its own logic, which may not match that used in other places. I see fairly high level tests, but nothing that pertains directly to the algorithm for doing this selection. I could be missing something.

doxiao:
> I could use some more explanations, here. There are a fair number of places where we need to select the REST port, including for the health check and the metrics. WlsServerConfig.getLocalAdminProtocolChannelPort as its own logic, which may not match that used in other places. I see fairly high level tests, but nothing that pertains directly to the algorithm for doing this selection. I could be missing something.

WlsServerConfig.getLocalAdminProtocolChannelPort has exactly what we need here: it chooses a port to use in the following order:
1. the channel that has admin privileges ;
2. the server's admin port;
3. the server's SSL port
4. the server's default listenport.

The original logic uses the ports on the server's K8S service, which is not needed.

As to the test cases, they do test if the correct port is picked/used given a server configuration. The expected port is specified in the line like this: defineResponse(200, "", "http://dyn-managed-server2.Test:7001");. This means the test expects that the default listenport 7001 is picked.   I noticed that the test names do not make this obvious. I can make some changes to make it clearer.

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainValidationTest.java
```diff
@@ -1086,6 +1087,142 @@ public void whenDomainUidPlusClusterNameExceedMaxAllowed_reportError() {
         "DomainUID ", domainUID, "cluster name", clusterName, "exceeds maximum allowed length")));
   }
 
+  @Test
+  public void whenDomainServerHasListenPort_dontReportError() {
+    String domainUID = "TestDomainForRest";
+    String clusterName = "TestClusterForFRest";
```
ankedia, line 1093: Looks like typo - TestClusterForFRest

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainValidationTest.java
```diff
@@ -1086,6 +1087,142 @@ public void whenDomainUidPlusClusterNameExceedMaxAllowed_reportError() {
...
+    configureDomain(myDomain)
+        .withDomainHomeSourceType(Image)
+        .withWebLogicCredentialsSecret(SECRET_NAME, null)
+        .withDomainType("WLS")
+        .configureAdminServer()
+        .configureAdminService()
+        .withChannel("default");
+
+    testSupport.addToPacket(DOMAIN_TOPOLOGY, domainConfigWithCluster);
+    System.out.println(myDomain.getAfterIntrospectValidationFailures(testSupport.getPacket()));
```
ankedia, line 1113: Should this use logger instead of "System.out.println" or removed?

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainValidationTest.java
```diff
@@ -1086,6 +1087,142 @@ public void whenDomainUidPlusClusterNameExceedMaxAllowed_reportError() {
...
```
ankedia, line 1140: Same as above, "System.out.println" may not be needed.

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainValidationTest.java
```diff
@@ -1086,6 +1087,142 @@ public void whenDomainUidPlusClusterNameExceedMaxAllowed_reportError() {
...
```
ankedia, line 1167: Same as above, "System.out.println" may not be needed.

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainValidationTest.java
```diff
@@ -1086,6 +1087,142 @@ public void whenDomainUidPlusClusterNameExceedMaxAllowed_reportError() {
...
```
ankedia, line 1195: Same as above, "System.out.println" may not be needed.

operator/src/main/java/oracle/kubernetes/operator/steps/ReadHealthStep.java
```diff
@@ -108,6 +106,7 @@ public NextAction apply(Packet packet) {
     }
 
     private HttpRequest createRequest() {
+      LOGGER.fine("Create REST request to service URL: " + getRequestUrl());
```
ankedia, line 109: Not sure if this is going to be logged for every request which is executed every 3 seconds? This causes log files to be verbose, can it be FINER instead of FINE?

ankedia (approved):
LGTM

ankedia:
One minor comment - It looks like there are some test methods with unused clusterName variable -

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainValidationMessages.java
```diff
@@ -131,4 +131,8 @@ public static String exceedMaxExternalServiceName(
       String domainUid, String adminServerName, String result, int limit) {
     return getMessage(MessageKeys.ILLEGAL_EXTERNAL_SERVICE_NAME_LENGTH, domainUid, adminServerName, result, limit);
   }
+
+  public static String noAvaiablePortToUse(String domainUid, String serverName) {
```
russgold, line 135: Spelling: "available" not "avaiable"

operator/src/main/resources/Operator.properties
```diff
@@ -167,9 +167,11 @@ WLSDO-0019=DomainUID ''{0}'' and admin server name ''{1}'' combination ''{2}'' e
 WLSDO-0020=Online WebLogic configuration updates complete \
   but there are pending non-dynamic changes that require \
   pod restarts to take effect. The changes are:
-
+WLSDO-0021=DomainUID ''{0}'' server ''{1}'' does not have a port available for the operator to send REST calls. \
+  The default listen port and SSL port are disabled, the admin port is not configured and there is not a channel with the admin privileges.\
```
russgold, line 171: Phrasing is clumsy. You could check with Rosemary, but I would recommend, "... and there is no channel with admin privileges."

russgold:
Mostly looks good; some questions/comments:

- You said that the names weren't a reliable way to select the right port, but I didn't see code/tests that explicitly dealt with that. The tests I see appear to be done at a much higher level. I'd have expected tests on WlsServerConfig at the least.
- The monitoring exporter also needs to compute the REST port, but I don't see any changes/tests that deal with that.

rjeberhard:
@doxiao, @russgold, @ankedia, @jshum2479, can someone provide the list of all code points that need this information? I can think of the monitoring exporter, the server health check through REST, and the dynamic update in the introspector (this likely will be a different code base since it's behavior from WDT). Are there other cases? These then need to give the correct answer with and without Istio.

doxiao:
The problem exists in code paths that use the result of WlsServerConfig#getAdminProtocolChannelName method to match a port in server service's ports. As far as I can tell, only REST (ReadHealthStep) and exporter (PodHelper) do this. I have changed both places to NOT use server service ports to determine which port to use, instead just call getLocalAdminProtocolChannelPort method.

doxiao:
> Mostly looks good; some questions/comments:
> 
> * You said that the names weren't a reliable way to select the right port, but I didn't see code/tests that explicitly dealt with that. The tests I see appear to be done at a much higher level. I'd have expected tests on WlsServerConfig at the least.

The problem and fix are not in WlsServerConfig layer, instead, in ReadHealthStep. Therefore the newly added test cases are at the ReadHealthStep layer. 

> * The monitoring exporter also needs to compute the REST port, but I don't see any changes/tests that deal with that.

doxiao:
@russgold I noticed that in PodHelper#createMonitoringExporterContainer method, the logic of choosing a port for the Java options -DWLS_PORT is different from what is for choosing a port for the ContainerPort. The former looks for a channel with admin privileges before looking for server's admin port, ssl port and listen port, but the latter does not. I am not sure if it is intentional. Changes in this PR only impacts the former where WlsServerConfig#getAdminProtocolChannelName is used.

russgold (approved):

PR #2302, opened 2021-04-07
fix domain in image sample to start the initial replica servers

bhavaniravichandran:
When updating the domain-on-pv using wdt sample, DynamicClusterSize was mistakenly set to DynamicClusterSize to initialReplicaCountServers. This caused only 2 servers to be configured. But setting it to ConfiguredManagedServerCount, all the servers were started. The fix is to set DynamicClusterSize to ConfiguredManagedServerCount in the topology section of the model file. Also, set replicas to 2 and added a cluster section with the initial cluster (cluster-1) and its replicas in the kubernetes section. This started the replica count of cluster-1 servers and when a second cluster was added, started 2 server in the second cluster. I also verified scaling up and down both the clusters and they work as expected.

### DIFF
--- a/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain.sh
+++ b/kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/create-domain.sh
@@ -213,11 +213,9 @@ function createDomainHome {
     sleep 30
     max=30
     count=0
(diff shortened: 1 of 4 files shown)


bhavaniravichandran:
Started jenkins run. Will update here when it finishes

bhavaniravichandran:
Closing this PR. Opened another one against main

PR #2303, opened 2021-04-08
fix broken doc link

doxiao:
Fix the link to scalingAction.sh in scaling doc.


rosemarymarano (approved):
I will do a search through the docs for this part of the path, main/src/scripts/, to make sure that there are no other instances of it, unless you have already done it, @doxiao

doxiao:
> I will do a search through the docs for this part of the path, main/src/scripts/, to make sure that there are no other instances of it, unless you have already done it, @doxiao

I have done a quick search and did not find any other occurrence. But it would be nice if we have an extra pair of eyes on this.  Thanks.

rosemarymarano:
@doxiao My seaarch finds it in the scaling doc only, in 3.2 and under staging. If you are fixing it in both those branches, then I think we got them all. Thank you.

rjeberhard (approved):

PR #2305, opened 2021-04-08
OWLS-88571 - Potential fixes for domain startup issues in large k8s cluster when watch events are not delivered.

ankedia:
Potential fixes for the domain startup issue in GBU CNE large k8s cluster where watch events are not reliably delivered by k8s. Currently, the introspection pod is not started when the domain is deleted and recreated using the `kubectl apply` command if there are previously running fibers for the domain. In such cases, domain status is null and this change starts a new fiber if the domain status is null. The second change is to discard the API client and associated HTTP client if the client gets into a bad state due to ProtocolException (and also due to a potential bug in okhttp3/okio library) and create a new client instance. Creating this draft PR to review these changes so that they can be provided to the GBU CNE team for further testing while we discuss other approaches. We're also discussing the unit testing approach to diagnose this type of issues in case of missed watch events and verify fixes.


rjeberhard (approved):

rjeberhard:
Plan to merge this as we worked on it together and reviewed each other's work.

PR #2312, opened 2021-04-13
Remove the exitcode check and add retry

sankarpn:
Remove the exit code status check for the curl command. A for loop is used to recursively send http requests to the weblogic servers to populate the http access logs. Retry until the HTTP response code is 200.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4699/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
@@ -13,6 +13,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 import io.kubernetes.client.custom.V1Patch;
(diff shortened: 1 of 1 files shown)


anpanigr (approved):
LGTM

xiancao (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
```diff
@@ -278,14 +277,9 @@ public void testMiiHttpServerLogsAreOnPV() {
...
           "/bin/sh", "-c", curlCmd));
-      if (execResult.exitValue() == 0) {
-        logger.info("\n HTTP response is \n " + execResult.toString());
-        assertAll("Check that the HTTP response is 200",
-            () -> assertTrue(execResult.toString().contains("HTTP/1.1 200 OK"))
-        );
-      } else {
-        fail("Failed to access sample application " + execResult.stderr());
-      }
+      assertFalse(execResult.exitValue() != 0 && execResult.stderr() != null && !execResult.stderr().isEmpty(),
```
tbarnes-us, line 280: 1) IMO, this is easier to understand as `assertTrue(execResult.exitValue() == 0 || execResult.stderr() == null || execResult.stdError().isEmpty, ...`.

2) Are there any failure cases where a non-zero exit code plus "no stderr" is _not_ safely assumed to be a success? It seems hard to know this for sure.

3) Instead of assuming 2, maybe this code can retry?

rjeberhard: I'd like to better understand under what circumstances curl can succeed but return a non-zero return?

tbarnes-us: Good point - I doubt that was intended. It's my understanding the goal of this fix is to continue and assume the curl succeeded when curl returns non-zero with no stderr?

sankarpn: Process proc = kubernetesExec.exec(command);
proc.exitValue()

More precisely the exit code is from the Process object which at times happens to return a non-zero value though I don't see any issue with curl commands executed by the exec() method.

tbarnes-us: I don't see it can be asserted that the non-zero code with an empty-stderr always means success - there are many many reasons that a call can fail, and I assume it's unpredictable whether the failure will populate stderr. 

I also don't quite see that a zero exit code without checking for an expected return from the call always means success.  It seems possible for a call to return success but not have the expected value in the return string. 

So, instead of assuming, can the code retry?

sankarpn: Agree with Tom 
assertTrue(execResult.exitValue() == 0 || execResult.stderr() == null || execResult.stdError().isEmpty()) 
is easier to understand. Fixed it.

tbarnes-us: The 'assertTrue' refactor makes it easier to understand the change, but it does not address the questions about the change in the first place.

sankarpn: Retry is used if the HTTP response code is not 200.

tbarnes-us (approved):

PR #2319, opened 2021-04-20
Special characters in Admin credential and mii encryption credential

anpanigr:
Ref.  https://jira.oraclecorp.com/jira/browse/OWLS-88359

The current PR covers the following usecases

  Tests special character combination in weblogic admin secret, mii encryption secret and db secret
  Add non-conventional admin user such as 'wlsadmin'  (other than standard 'weblogic')
  Usage of custom environment variables with special character combination and its usage in model yaml file. Verify the 
     macro expansion using Runtime MBean 

  Usage of pre-defined Environment variable ( custom env is not supported) in situational config  usecase such as DOMAIN_UID and Verify the macro
      expansion using Runtime MBean 
  ( secret macro in sitconfig is already covered in existing test)

Parallel Run:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4793/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4825


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
```diff
@@ -111,18 +110,18 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
     // install and verify operator
     installAndVerifyOperator(opNamespace, domainNamespace);
 
-    // create secret for admin credentials
+    // create secret for admin credential with special characters
     logger.info("Create secret for admin credentials");
     String adminSecretName = "weblogic-credentials";
     assertDoesNotThrow(() -> createDomainSecret(adminSecretName, 
-            ADMIN_USERNAME_DEFAULT, ADMIN_PASSWORD_DEFAULT, domainNamespace),
+            ADMIN_USERNAME_DEFAULT, "W##%*}!`''1~3x", domainNamespace),
```
vanajamukkara, line 117: can you add all the characters mentioned in the JIRA - Testing a single quote, a double quote, two single quotes, two double quotes, single slash, back slash, double slash, ‘$’, back-quote, double back-quote, ‘!’, etc, etc...

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
```diff
@@ -111,18 +110,18 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
...
             String.format("createSecret failed for %s", adminSecretName));
 
-    // create encryption secret
+    // create encryption secret with special characters
     logger.info("Create encryption secret");
     String encryptionSecretName = "encryptionsecret";
     assertDoesNotThrow(() -> createDomainSecret(encryptionSecretName, "weblogicenc",
-            "weblogicenc", domainNamespace),
+            "W##%*}!`''1~3x", domainNamespace),
```
tbarnes-us, line 124: It'd be good to test similar for internal situational config (DII pass), custom situational config maps (such as for a database password), and custom model config maps (such as for database password).  In addition, it'd be good to test the use in environment variables (sit-cfg and model files can reference env vars instead of secrets).

tbarnes-us: Finally, it'd be good to test for something that takes an arbitrary unencrypted string - maybe the notes field of some mbean would be fine...  (For sit-cfg, model YAML, and model-config-map YAML).

anpanigr: situational config Feature is overwritten by Dynamic Mii Upgrade.  Let me try custom model config maps (such as for database password) and ENV variable in domain resources. Here I do not want to create a whole new domain/class  to verify all edge (nice-to-have) cases. Let me see if I can insert these usecases in existing test class.

tbarnes-us: Sit-cfg still applies to the other domain home source types (DII, and  DIPV), and so should also be tested.

tbarnes-us: Also, please test with a single quote instead of two quotes right next to each-other.  Two consecutive quotes could parse correctly even if there's a bug in macro parsing.  

In addition, please put the comment '#' symbol first - I think that'd be more likely to be parsed incorrectly if there's a bug.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
anpanigr: done, Now new password is [ W##%*}!""`'']\\//1$$~x ]

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
tbarnes-us: In addition, please add something like `${DOMAIN_UID}` and `$(ls)` to the sit-cfg/model X secrets/env-vars/props - as this catches more cases.  Plus the set values should be checked using mbeans that the values are as expected -- if you don't check via mbeans, it's possible that the set values will be parsed using shell on _both_ the configuration and the usage paths in WKO and seem to work fine (which would be bad).

anpanigr: Now the password is changed to  [ ##W%*}!"'"`']\\\\//1$$~x ] Added test for config-map with db password with special characters

anpanigr: Updated the test to use "${DOMAIN_UID} $(ls)" in encryption secret. Add a test to define a env variable with special characters in its value.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
```diff
@@ -111,18 +111,19 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
...
     installAndVerifyOperator(opNamespace, domainNamespace);
 
-    // create secret for admin credentials
+    // create secret for admin credential with special characters
+    // the resultant password is ##W%*}!"'"`']\\\\//1$$~x 
     logger.info("Create secret for admin credentials");
     String adminSecretName = "weblogic-credentials";
     assertDoesNotThrow(() -> createDomainSecret(adminSecretName, 
-            ADMIN_USERNAME_DEFAULT, ADMIN_PASSWORD_DEFAULT, domainNamespace),
+            ADMIN_USERNAME_DEFAULT, "##W%*}!\"'\"`']\\\\//1$$~x", domainNamespace),
```
vanajamukkara, line 119: create a constant in TestConstants for password with special characters and use that, instead of hard coding in all places

vanajamukkara (approved):
left a comment which needs to be addressed, otherwise LG

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
```diff
@@ -248,15 +249,39 @@ public void beforeEach() {
...
+
+    boolean found = false;
+    for (int i = 0; i < envList.size(); i++) {
+      logger.info("The name is: {0}, value is: {1}", envList.get(i).getName(), envList.get(i).getValue());
+      if (envList.get(i).getName().equalsIgnoreCase("SPECIAL_ENV")) {
+        assertTrue(
+            envList.get(i).getValue().equalsIgnoreCase("${DOMAIN_UID}~##!'%*$(ls)"),
+            "Expected value for SPECIAL_ENV variable does not mtach");
+        found = true;
+      }
```
tbarnes-us, line 272: To complete the testing of env we need to see if WL configuration that references the env at runtime ends up with expected value:
- Reference the env in a macro within WL configuration (such as in the notes field of some bean), and verify using WLST or JMX that the deployed domain's bean attribute contains the expected value.
- Do this for both 'model YAML' and 'sit-cfg'.  (They have a different macro format, but each can reference an env var.)
(See the JIRA for the combinatorial matrix of test use cases.)

anpanigr: Added configuration check for custom environment inside domain configuration. 
We will address situational config usecase in different PR

anpanigr (commented):
done

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
anpanigr: This is one-off variable, not being used in multiple classes.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiCustomSslStore.java
anpanigr: Added test for non standard admin user such as 'wlsadmin'.  
Add sitconfig test with supported environment variables  ( https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/configoverrides/#override-template-macros)
We have already sitconfig tests with secret macro.

tbarnes-us (approved):

PR #2321, opened 2021-04-20
Update config maps based on hash like other resources

rjeberhard:
We're seeing an upgrade issue where the operator continuously updates the script config map. On diagnosis, the script config map is updated successfully; however, on next check, the operator still calculates the the script needs to be updated and the cycle continues.

The proposed fix is to check the contents of the config map using the same strategy we are using for pods and the CRD, etc.


jshum2479 (approved):
LGTM

ankedia (approved):

PR #2326, opened 2021-04-21
Cache secret for a configurable time period and reset on auth/authz failure

rjeberhard:
This resolves an observed increase in the operator's network utilization by not reading the webLogicCredentialsSecret every time that it is necessary and, instead, caching the value for a configurable time period. Additionally, the secret is cleared (and then reread on the next iteration) following an auth or authz failure.


ankedia (approved):
LGTM

russgold:
Where are the tests for this new functionality? 

I would have expected something which tried the read the secret, and tried again both within and after the timeout, seeing a cached value returned in the first case and a new read from k8s in the latter case.

rjeberhard:
@russgold, I found that whenAuthorizedToReadHealthAndThenWait_verifySecretCleared was sufficient. This tests that the clearing behavior worked. The earlier tests then validate that the read will happen when the cache is empty.

russgold (approved):

PR #2330, opened 2021-04-22
Add NS and domainUID to log messages in more code paths

doxiao:
Make sure the domain namespace and domainUID (when applicable) are in the log messages that are originated from the following code paths:
1) HttpAsyncRequest response code path;
2) Code path to create start namespaces steps when there is no fiber on the current thread;
3) StuckPodProcessing code path;
4) AsyncRequestStep where operator's namespace is accidentally used in the log messages.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4833/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
@@ -216,7 +216,7 @@ class DomainRecheck {
 
   private Step startNamespaceSteps(String ns) {
     try (LoggingContext ignored =
-             setThreadContext().namespace(ns).domainUid("")) {
+             setThreadContext().namespace(ns)) {
       return Step.chain(
           createNamespaceReview(ns),
           new StartNamespaceBeforeStep(ns),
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -211,10 +215,13 @@ private Step createNamespaceReviewStep(Set<String> namespacesToStartNow) {
   }
 
   private Step startNamespaceSteps(String ns) {
-    return Step.chain(
+    try (LoggingContext ignored =
+             setThreadContext().namespace(ns).domainUid("")) {
```
rjeberhard, line 219: I don't understand this line. Why are you setting the domainUid to empty?

doxiao: This is the beginning of starting a namespace and we are creating the necessary steps; we don't have a domainUID and we don't have a fiber on the thread. So we set the logging context in the thread local, and we don't want any left over doainUID to take effect. Once we have the DomainPresenceImpl later on, we'll set the info in the packet, which will take precedence over the thread local.

doxiao: It is more of a defense measure to set the domainUID to empty.

doxiao: LoggingContext is AutoCloseable, and the thread local is cleared on close. I removed the part of setting domainUID to "".

rjeberhard (approved):

PR #2337, opened 2021-04-25
Use metadata.generation and remove beta CRD support

rjeberhard:
Now that we are requiring a minimum 1.16 Kubernetes version, we can use metadata.generation to detect changes to the Domain spec (rather than relying on equals) and remove support for the beta CRD API version.

Anil,

I still suspect that you are seeing unexpected introspector runs because your new changes are failing the equals check... if so, I'd like to fix that, but these changes will help.


ankedia (approved):

doxiao (approved):
LGTM.

PR #2338, opened 2021-04-26
Owls 87757 - common mount feature to separate WDT model and archive from WLS binary image.

ankedia:
The initial set of changes to the operator for the common mounts feature to separate WDT model and archive from WLS binary image. Added commonMount to the serverPod section of the schema and it allows mounting an array of containers. The change also removes hard-coded value for WDT binary home and makes it configurable. I have tested it with jit_feature WIT branch.  I have started an integration test run at https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4850 .


documentation/domains/Domain.json
```diff
@@ -184,6 +184,38 @@
...
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/Container"
+          }
+        },
+        "medium": {
+          "description": "The emptyDir volume medium. Defaults to unset.",
+          "type": "string"
+        },
+        "emptyDirVolumeName": {
```
tbarnes-us, line 213: Does the volume name need to be configurable? I thought it just needs to be sufficiently unique.

ankedia: It's set to the value `operator-common-volume` that we decided on in the meeting. I thought of leaving it configurable for increased flexibility. Do you foresee any issues due to this?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -213,6 +213,13 @@
       + "See `kubectl explain pods.spec.containers.volumeMounts`.")
   private final List<V1VolumeMount> volumeMounts = new ArrayList<>();
 
+  /**
+   * The common mount.
+   *
+   */
+  @Description("The common mount with the containers hosting files to be copied to the common volume.")
```
tbarnes-us, line 220: IMO, this could use some word-smithing. The 'volume' is internal from a user perspective, as is the copy operation. Also, IMO, it'd help to explain the use case for the feature here, plus describe how/why 'emptyDir' is mentioned in some of the feature's settings. How about something like:

Use a common mount to automatically include directory content from additional images. This is a useful alternative for including Model in Image model files, or other types of files, in a pod without requiring modifying the pod's base 'domain.spec.image'. This feature internally uses a Kubernetes emptyDir volume and Kubernetes init containers to share the files from the additional images with the pod.

kubernetes/crd/domain-v1beta1-crd.yaml
```diff
@@ -9160,6 +9307,54 @@ spec:
...
+                                  type: string
+                                  enum:
+                                  - Always
+                                  - Never
+                                  - IfNotPresent
+                                name:
+                                  description: The name for the operator created init
+                                    container. Set to 'operator-common-container-${CONTAINER_NUM}'.
+                                  type: string
+                                command:
```
jshum2479, line 9348: An init container can have command and args.   Do we allow user to change the  command (which shell or others)? Also, we need to document how this is finally in the container spec.  For example,  user if coding the yaml manually can have:

command: ["/bin/sh"]
args: ["-c", "cp something somewhere"]

How this command translate into the above?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
ankedia: ok, I'll use the above wording. ```Use a common mount to automatically include directory content from additional images. This is a useful alternative for including Model in Image model files, or other types of files, in a pod without requiring modifications to the pod's base image 'domain.spec.image'. This feature internally uses a Kubernetes emptyDir volume and Kubernetes init containers to share the files from the additional images with the pod.```

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Container.java
```diff
@@ -0,0 +1,128 @@
...
+          "The image pull policy for the common mount container image. "
+                  + "Legal values are Always, Never, and IfNotPresent. "
+                  + "Defaults to Always if image ends in :latest; IfNotPresent, otherwise.")
+  @EnumClass(ImagePullPolicy.class)
+  private String imagePullPolicy;
+
+  @Description("The name for the operator created init container. Set to 'operator-common-container-${CONTAINER_NUM}'.")
+  private String name;
+
+  @Description("The init container command. Defaults to 'cp -R $COMMON_DIR/* $TARGET_DIR'.")
```
tbarnes-us, line 37: How about:

The command for this init container. Defaults to 'cp -R $COMMON_DIR/* $TARGET_DIR'. This is an advanced setting for customizing the container command for copying files from the container image to the common mount emptyDir volume. Use the 'COMMON_DIR' environment variable to reference the value configured in 'commonMount.mountPath' (which  defaults to '/common'). Use 'TARGET_DIR' to refer to the temporary directory created by the Operator that resolves to the common mount's internal emptyDir volume.

documentation/domains/Domain.json
tbarnes-us: Yes. The more configuration, the more complicated the feature looks to the end user, and the more chance they make a mistake when tuning the configuration. If we choose a 'safe' unique hard coded value, then there'd be no such worries. We use hard coded values in several other similar paths in the Operator for the same reasons.

kubernetes/crd/domain-v1beta1-crd.yaml
ankedia: The command that the user provides using this schema element translates to "cp something somewhere" (in args). I was envisioning use-cases where a user wants to perform additional tasks such as unzipping a file, copying files in a different location, or executing a script. We populate the `COMMON_MOUNT_PATH` and `COMMON_TARGET_PATH` env variables in the init container for this. I wanted to avoid having multiple moving parts where users can pass additional arguments to commands or change the command to something other than executing a shell command or script. I will look into the documentation part to clarify how the specified command finally translates in the container spec.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
```diff
@@ -49,6 +61,37 @@ protected V1Container createPrimaryContainer(TuningParameters tuningParameters)
...
+    return new V1Volume().name(cm.getEmptyDirVolumeName()).emptyDir(emptyDirVolumeSource);
+  }
+
+  protected V1Container createInitContainerForCommonMount(Container container, int index, CommonMount cm) {
+    return new V1Container().name(getName(container, index))
+        .image(container.getImage())
+            .imagePullPolicy(container.getImagePullPolicy())
+            .command(Collections.singletonList(SHELL))
+            .env(createEnv(cm))
+            .args(Arrays.asList("-c", container.getCommand()))
```
tbarnes-us, line 78: I suspect more informative error handling would be helpful here.  In detail, I suspect an error here would result in a somewhat obscure failure and error message at the moment.  

So how about invoking a small wrapper script that provides detailed debugging and failure messages? And that uses the same 'utils' helper trace methods as the main scripts (so that the Operator can parse the output and include any errors it scrapes in its error messages)?  You can put the commands in a shell script in the resources directory.

The script could do something like:

```
common-mount-copy.sh
# Init container script for the common mount feature. 
# See 'domain.spec.serverPod.commonMount' for details.
load 'utils.sh'
trace "Common Mount: About to copy common mount container image=$CONTAINER_IMAGE files in COMMON_DIR=$COMMON_DIR TARGET_DIR=xxx"
trace messages that show current 'id', directory content, and file permissions for the COMMON_DIR in the image (there's a one-line utils.sh helper method for this)
trace ERROR message if COMMON_DIR doesn't exist or is empty (and error out)
trace FINE about to exec COPY_COMMAND=$COPY_COMMAND
results=$(exec "$COPY_COMMAND" 2>&1)
trace ERROR message if fails err=$results
trace FINE message on success out=$results
exit non-zero on failure
```

in addition, there should be integration tests for error paths in ^^^ - if the above errors happen in the introspector, then the introspector is expected to fail, and the ^^^ error messages should show automatically show up in the operator log and the domain status

kubernetes/crd/domain-v1beta1-crd.yaml
jshum2479: I am trying to avoid confusion, if user tries to put the command as "/bin/bash -c 'blah blah'" .  If the user knows how it is translated then they can adjust accordingly.

tbarnes-us: For a variety of reasons, I think we should have a wrapper script that in turn calls the specified commands.  If the user wants to see how a custom command is translated, they can examine the wrapper script. I just made a separate comment that details wrapper script ideas.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Model.java
```diff
@@ -109,6 +127,7 @@ public String toString() {
             .append("domainType", domainType)
             .append("configMap", configMap)
             .append("modelHome", modelHome)
+            .append("wdtBinaryHome", wdtInstallHome)
```
tbarnes-us, line 130: "wdtBinaryHome" -> "wdtInstallHome"

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,34 @@
...
+
+scriptDir="$( cd "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
+
+if [ "${debug}" == "true" ]; then set -x; fi;
+
+source ${scriptDir}/utils.sh
+[ $? -ne 0 ] && echo "[SEVERE] Missing file ${scriptDir}/utils.sh" && exit 1
+
+if [ -z ${COMMON_MOUNT_COMMAND} ]; then
+  trace ERROR "The common mount command $COMMON_MOUNT_COMMAND is empty. Exiting"
```
tbarnes-us, line 14: '$COMMON_MOUNT_COMMAND' will resolve to '' in this case, and so isn't needed. 

Also, maybe make the error message a little more informative with "The 'serverPod.commonMount.container.mountCommand' is empty."

Finally, we don't know which image has the error, so maybe move the following trace above this trace.

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+fi
+
+trace INFO "Common Mount: About to execute command $COMMON_MOUNT_COMMAND in container image=$CONTAINER_IMAGE. COMMON_MOUNT_PATH is $COMMON_MOUNT_PATH and COMMON_TARGET_PATH is ${COMMON_TARGET_PATH}."
```
tbarnes-us, line 18: Minor: 1) For this and all other traces, it's more clear IMO to put single quotes around each env var.
2) You could prefix all trace text with "Common Mount:"

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+fi
+
+trace FINE "About to execute COMMON_MOUNT_COMMAND=$COMMON_MOUNT_COMMAND ."
+results=$(eval $COMMON_MOUNT_COMMAND 2>&1)
+if [ $? -ne 0 ]; then
+  trace ERROR "Execution of common mount command $COMMON_MOUNT_COMMAND failed. Error -> $results ."
+  exit 1
+else
+  trace FINE "Command $COMMON_MOUNT_COMMAND executed successfully ."
```
tbarnes-us, line 32: Append "Output -> '$results'." to help with debugging the success case.

operator/src/main/resources/scripts/commonMount.sh
tbarnes-us: Two more things:
- Better yet, include the name of the container's image in the error message. 
- We probably don't want to exit 1 as  this would short circuit the checks/tracing to pod log in the main container.   
- Ditto for (all?) of the other exit 1 failures...

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Model.java
ankedia: Done.

operator/src/main/resources/scripts/commonMount.sh
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommonMount.java
```diff
@@ -0,0 +1,159 @@
...
+
+  public static final String COMMON_MOUNT_PATH = "/common";
+  public static final String COMMON_TARGET_PATH = "/tmpCommonMount";
+  public static final String COMMON_VOLUME_NAME = "operator-common-volume";
+
+  @Description("The common mount containers.")
+  private List<Container> containers;
+
+  @Description("The common mount path. Defaults to /common.")
+  private String mountPath;
```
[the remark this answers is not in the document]
ankedia: @rjeberhard @jshum2479 - Do you have any thoughts or suggestions on this? I think we had discussed it during the meeting but didn't come to a conclusion. I was originally thinking about doing the "double copy" (1) copy from mountPath to temp dir in the init container and (2) from temp dir to mountPath in the main container for this. But we decided against that approach in the meeting.

operator/src/main/resources/scripts/commonMount.sh
ankedia: Done

operator/src/main/resources/scripts/commonMount.sh
ankedia: Done

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,32 @@
...
+
+source ${scriptDir}/utils.sh
+[ $? -ne 0 ] && echo "[SEVERE] Missing file ${scriptDir}/utils.sh" && exit 1
+
+if [ -z ${COMMON_MOUNT_COMMAND} ]; then
+  trace ERROR "Common Mount: The 'serverPod.commonMount.container.mountCommand' is empty for the container image='$CONTAINER_IMAGE'. Exiting"
+  exit
+fi
+
+trace INFO "Common Mount: About to execute command '$COMMON_MOUNT_COMMAND' in container image='$CONTAINER_IMAGE'. COMMON_MOUNT_PATH is '$COMMON_MOUNT_PATH' and COMMON_TARGET_PATH is '${COMMON_TARGET_PATH}'."
```
tbarnes-us, line 18: We should probably avoid INFO. It's too verbose because INFO and above are automatically copied into the operator log by default. Please use FINE instead, plus modify the trace ERROR at line 29 to also mention the COMMON_MOUNT_PATH.

I should have thought of this earlier (sorry).

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Container.java
```diff
@@ -0,0 +1,121 @@
...
+import oracle.kubernetes.operator.helpers.KubernetesUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+public class Container {
+
+  public static final String INIT_CONTAINER_NAME_PREFIX = "operator-common-container";
+  public static final String DEFAULT_INIT_CONTAINER_COMMAND = "cp -R $COMMON_MOUNT_PATH/* $COMMON_TARGET_PATH";
+  public static final String INIT_CONTAINER_WRAPPER_SCRIPT = "/weblogic-operator/scripts/commonMount.sh";
```
tbarnes-us, line 20: It would help IMO to have this script avoid doing an 'exit 1' (as exit 1 handling is known to be problematic at the Operator level), and to have this script record its output so the main introspector pod log, plus the pod logs for every weblogic pod, can include its tracing in its stdout


```
/weblogic-operator/scripts/commonMount.sh > /tmp/cm.out 2>&1 ; cat /tmp/cm.out ; cp /tmp/cm.out  $COMMON_MOUNT_DIR/commonMountLogs/commonMount-1.out ; exit 0
```
in addition the wl pods (and probably also the introspector pod script), should sanity check the script output to make sure nothing went wrong. Particularly as the operator is not involved with checking the results for each wl pod.

tbarnes-us: If the ^^^ command is too complex, the script itself could be adjusted to do the same steps - which might be more elegant. (You'll need to pass in teh container# to use for decorating the output.)

ankedia: If the init container fails for WebLogic pods, pods will be in `Init:Error` status, and users can always get the logs of the init container using `kubectl logs -c` command. So it's not an issue for WebLogic pods.For introspector pod, I'm are copying the output of script in operator logs and if the introspector pod is still around, users can get the logs using `kubectl logs`. I think it would be good to discuss this in a meeting since this was missed during the original design/implementation meeting.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommonMount.java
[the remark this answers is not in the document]
ankedia: FYI - If the main image has files in the mountPath then those files will be "hidden" and not accessible. We will need to document this so customers can avoid having files in mountPath.

operator/src/main/resources/scripts/commonMount.sh
ankedia: Done.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Container.java
ankedia: I have made the changes to capture the output of wrapper script in the main introspector pod log and the pod logs for every weblogic pod. Please let me know if missed something.

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# The operator fails the introspector if it detects an
+# ERROR/SEVERE, and succeeds if it detects
+# 'executed successfully'.
+#
+# The main introspector and pod scripts will echo
+# the contents of /${COMMON_MOUNT_PATH}/common-mount-logs/
+# and fail if they are missing, or if any do not
+# include 'executed successfully', or if the scripts
+# cannot create (touch) files in /${COMMON_MOUNT_PATH}.
```
tbarnes-us, line 20: Minor: Add "# (See also utils.sh checkCommonMount function)

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,39 @@ function adjustPath() {
...
 }
+
+#
+# checkCommonMount
+#   purpose: If the COMMON_MOUNT_PATH directory exists, it echoes the contents of output files
+#            in ${COMMON_MOUNT_PATH}/commonMountLogs dir. It returns 1 if a SEVERE message
+#            is found in any of the output files in ${COMMON_MOUNT_PATH}/commonMountLogs dirs.
+#            It also returns 1 if 'successfully' message is not found in the output files
+#            or if the COMMON_MOUNT_PATH directory is empty. Otherwise it returns 0 (success).
+#
```
tbarnes-us, line 900: Minor: Add something like "# see also 'commonMount.sh'"

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommonMount.java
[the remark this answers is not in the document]
tbarnes-us: Got it. Thanks.

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,57 @@
...
+
+scriptDir="$( cd "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
+
+if [ "${debug}" == "true" ]; then set -x; fi;
+
+source ${scriptDir}/utils.sh
+[ $? -ne 0 ] && echo "[SEVERE] Missing file ${scriptDir}/utils.sh" && exit 1
+
+function copyFiles() {
```
tbarnes-us, line 29: Minor: Consider renaming this function to something like "initCommonMount" and moving the function into utils.sh next to its sister checkMount function. That'll keep the related code close together, plus named similarly...

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommomEnvVars.java
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.weblogic.domain.model;
+
+public class CommomEnvVars {
```
jshum2479, line 6: Is this for common mount only?  The class name seems to suggest a generic environment variable class

jshum2479 (commented):
Overall LGTM,  can you verify upgrade won't cause any rolling?

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,81 @@ function adjustPath() {
...
+    for out_file in $out_files; do
+      if [ "$(grep -c SEVERE $out_file)" != "0" ]; then
+        trace SEVERE "Common Mount: Error found in file '${out_file}' while initializing commonMount."
+        severe_found=true
+      elif [ "$(grep -c successfully $out_file)" = "0" ]; then
+        trace SEVERE "Common Mount: Command execution was unsuccessful in file '${out_file}' while initializing commonMount."
+        severe_found=true
+      else
+        trace "Common Mount: Contents of '${out_file}':"
+        trace "Common Mount: End of '${out_file}' contents"
```
tbarnes-us, line 925: The 'End' trace belongs after teh cat.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommomEnvVars.java
tbarnes-us: +1 Suggest 'CommonMountEnvVars'

operator/src/main/resources/scripts/introspectDomain.sh
```diff
@@ -56,6 +56,11 @@ source ${SCRIPTPATH}/utils.sh
 
 traceTiming "INTROSPECTOR '${DOMAIN_UID}' MAIN START"
 
+# 
+# if the common mount feature is active, verify the mount, and log mount information
+#
+checkCommonMount || exit 1
+
```
tbarnes-us, line 63: Minor: Consider adding "checkEnv WDT_MODEL_HOME WDT_INSTALL_HOME || exit 1" at line 166, inside the 'FromModel' if clause.  This will assert that the env vars are set and implicitly create FINE trace statements for each.

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,81 @@ function adjustPath() {
...
+#            See also 'commonMount.sh'.
+#
+function checkCommonMount() {
+  # check commonMount results (if any)
+  if [ ! -z "$COMMON_MOUNT_PATH" ]; then
+    trace FINE "Common Mount: COMMON_MOUNT_PATH is '$COMMON_MOUNT_PATH'."
+    traceDirs $COMMON_MOUNT_PATH
+    touch ${COMMON_MOUNT_PATH}/testaccess.tmp
+    if [ $? -ne 0 ]; then
+      trace SEVERE "Common Mount: Cannot write to the COMMON_MOUNT_PATH '${COMMON_MOUNT_PATH}'" && return 1
```
tbarnes-us, line 909: minor: Since COMMON_MOUNT_PATH is an internal var name from an admin/end-user perspective, consider adding ". This path is configurable using the domain resource 'serverPod.commonMount.mountPath' attribute."

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,81 @@ function adjustPath() {
...
+      else
+        trace "Common Mount: Contents of '${out_file}':"
+        trace "Common Mount: End of '${out_file}' contents"
+      fi
+      cat $out_file
+    done
+    [ "${severe_found}" = "true" ] && return 1
+    rm -fr $COMMON_MOUNT_PATH/commonMountLogs
+    [ -z "$(ls -A $COMMON_MOUNT_PATH)" ] \
+      && trace SEVERE "Common Mount: No files found in '$COMMON_MOUNT_PATH'. Do your commonMount images have files in their '$COMMON_MOUNT_PATH' directories?" \
```
tbarnes-us, line 932: 1 - For more context, consider adding "This path is configurable using the domain resource 'serverPod.commonMount.mountPath' attribute."

2 - Long SEVERE lines can be broken up into multiple strings:
```
trace SEVERE  "foo " \
              "bar " \
              "baz"
```

operator/src/main/resources/scripts/commonMount.sh
```diff
@@ -0,0 +1,34 @@
...
+# include 'executed successfully', or if the scripts
+# cannot create (touch) files in /${COMMON_MOUNT_PATH}.
+# (See also utils.sh checkCommonMount function)
+
+scriptDir="$( cd "$(dirname "$0")" > /dev/null 2>&1 ; pwd -P )"
+
+if [ "${debug}" == "true" ]; then set -x; fi;
+
+source ${scriptDir}/utils.sh
+[ $? -ne 0 ] && echo "[SEVERE] Missing file ${scriptDir}/utils.sh" && exit 1
```
tbarnes-us, line 28: minor: add an SEVEREassertion/tracing for each of the env vars that are directly referenced. To do this, you can call `checkEnv COMMON_TARGET_PATH COMMON_MOUNT_CONTAINER_NAME`. (Also COMMON_TARGET_PATH should renamed COMMON_MOUNT_TARGET_PATH).

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommomEnvVars.java
ankedia: I was envisioning this class to have all the environment variables which are populated in both the introspector job pod and WebLogic server pods. Currently, we have `IntrospectorJobEnvVars` and `ServerEnvVars` but I didn't find a place to add env variables that are common to both "introspector job" and "server pod". One example of this is WDT_INSTALL_HOME which needs to be populated in all the pods. Is there an existing place for common env variables and I missed it? Thanks.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -213,6 +213,17 @@
...
+  /**
+   * The common mount.
+   *
+   */
+  @Description("Use a common mount to automatically include directory content from additional images. "
+          + "This is a useful alternative for including Model in Image model files, or other types of files, in a pod "
+          + "without requiring modifications to the pod's base image 'domain.spec.image'. "
+          + "This feature internally uses a Kubernetes emptyDir volume and Kubernetes init containers to share "
+          + "the files from the additional images with the pod.")
+  private CommonMount commonMount;
```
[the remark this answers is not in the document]
ankedia: > Overall LGTM, can you verify upgrade won't cause any rolling?

I ran the upgrade tests from 3.1.4 to the latest branch (including my changes) and it didn't cause a rolling restart. I believe Russ has added unit tests to catch the unexpected rolling issue and I believe we should be good here.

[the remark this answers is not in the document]
ankedia: Regarding the validation error "when a commonMount is defined but no containers are defined within it", I have added validation checks and made the change to not set the common mount path env var. After this check, it won't fail with the surprising error since env var will not be set when no containers are defined. Thanks.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/CommomEnvVars.java
ankedia: I have changed the class name to CommonMountEnvVars and moved the WDT_INSTALL_HOME env variable to `IntrospectorJobEnvVars` (to be colocated with WDT_MODEL_HOME).

tbarnes-us:
> Johnny: Overall LGTM, can you verify upgrade won't cause any rolling?

Johnny, Anil replied to your question in a sub-thread, so in case you missed it:

> Anil: I ran the upgrade tests from 3.1.4 to the latest branch (including my changes) and it didn't cause a rolling restart. I believe Russ has added unit tests to catch the unexpected rolling issue and I believe we should be good here.

@jshum2479 @ankedia

operator/src/main/resources/scripts/utils.sh
```diff
@@ -947,11 +951,13 @@ function checkCommonMount() {
 function initCommonMount() {
 
   if [ -z "${COMMON_MOUNT_COMMAND}" ]; then
```
tbarnes-us, line 953: commom-->common in comment ^^^

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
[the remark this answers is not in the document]
tbarnes-us: Can you point me to the code where you added validation checks? I somehow missed it.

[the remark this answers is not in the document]
ankedia: The check is in 3 different files (1) JobStepContext.java, (2) JobHelper.java and (3) PodStepContext.java. I added below check to see if commonMount.getContainers is not null using `Optional.ofNullable(cm.getContainers()).ifPresent()` -

```
  protected V1PodSpec addCommonMountInitContainerAndVolume(V1PodSpec podSpec, CommonMount cm) {
    Optional.ofNullable(cm.getContainers()).ifPresent(cl -> addInitContainerAndVolume(podSpec, cm, cl));
    return podSpec;
  }			
```
```
   Optional.ofNullable(cm.getContainers()).ifPresent(containerList ->
            addEnvVar(vars, CommomMountEnvVars.COMMON_MOUNT_PATH, cm.getMountPath()));					
```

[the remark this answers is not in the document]
tbarnes-us: Sorry - there was a bit of terminology confusion. 

I generally assume a 'validation check' in the operator context means that the operator will generate an error message that's reflected in the status. But no error message is generated here, which might be fine. 

It looks like the operator now silently ignores the case where a commonMount is configured with no containers? 

I'm OK with this - it's better than generating error messages that don't point out the underlying problem  - but wonder if it'd be even better to generate a validation error (plus document that at least one container must be specified). This is similar to the way we error out when 'model' is specified but its required 'model.runtimeEncryptionSecret' setting is missing.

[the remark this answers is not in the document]
ankedia: ok, I have added the logic to generate a validation error message when no containers are defined. Thanks.

operator/src/main/resources/Operator.properties
```diff
@@ -170,6 +170,9 @@ WLSDO-0020=Online WebLogic configuration updates complete \
 WLSDO-0021=DomainUID ''{0}'' server ''{1}'' does not have a port available for the operator to send REST calls. \
   The default listen port and SSL port are disabled, the admin port is not configured and there is no channel with admin privileges.\
 
+WLSDO-0022=The common mount configured by 'spec.serverPod.commonMount' in the domain resource does not define any containers. \
+  The common mount must specify at least one container.
```
tbarnes-us, line 174: serverPod can also scope to the cluster level or even the pod level, so 'spec.serverPod.commonMount', so:
- Change 'The common mount' to 'A common mount'
- Change 'spec.serverPod.commonMount' to 'serverPod.commonMount'.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -821,6 +822,12 @@ private void verifyIntrospectorJobName() {
       }
     }
 
+    private void verifyCommonMountContainers() {
+      // if the common mount is specified, verify that at least one container is defined.
```
tbarnes-us, line 826: serverPod can scope to the cluster and pod level, is this the place to check those too?

documentation/domains/Domain.md
```diff
@@ -17,6 +17,7 @@ The specification of the operation of the WebLogic domain. Required.
...
 | `allowReplicasBelowMinDynClusterSize` | Boolean | Whether to allow the number of running cluster member Managed Server instances to drop below the minimum dynamic cluster size configured in the WebLogic domain configuration, if this is not specified for a specific cluster under the `clusters` field. Defaults to true. |
 | `clusters` | array of [Cluster](#cluster) | Lifecycle options for all of the Managed Server members of a WebLogic cluster, including Java options, environment variables, additional Pod content, and the ability to explicitly start, stop, or restart cluster members. The `clusterName` field of each entry must match a cluster that already exists in the WebLogic domain configuration. |
+| `commonMountVolumes` | array of [Common Mount Volume](#common-mount-volume) | The common mount tuning. Allows overriding the global default values for mountPath, medium and sizeLimits for the common mount. See spec.commonMountTuning.globalDefaults for more details. |
```
tbarnes-us, line 20: I thought commonMountTuning is going away?  The description should be something like "Configure common mount volumes including their respective mount paths. Common mount volumes are in turn referenced by one or more `serverPod.commonMount` mounts, and are internally implemented using a Kubernetes 'emptyDir' volume."

documentation/domains/Domain.md
```diff
@@ -89,6 +90,15 @@ The current status of the operation of the WebLogic domain. Updated automaticall
 | `serverStartPolicy` | string | The strategy for deciding whether to start a WebLogic Server instance. Legal values are NEVER, or IF_NEEDED. Defaults to IF_NEEDED. More info: https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/domain-lifecycle/startup/#starting-and-stopping-servers. |
 | `serverStartState` | string | The WebLogic runtime state in which the server is to be started. Use ADMIN if the server should start in the admin state. Defaults to RUNNING. |
 
+### Common Mount Volume
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `medium` | string | The emptyDir volume medium. This is an advanced setting that rarely needs to be configured. Defaults to unset, which means the volume's files are stored on the local node's file system for the life of the pod. |
+| `mountPath` | string | The common mount path. The files in the path are populated from the same named directory in the images supplied by each container in 'commonMount.containers'. Defaults to '/common'. |
```
tbarnes-us, line 98: 1) commonMount containers --> 'serverPod.commonMount'
2) Add sentence:  "Each common mount volume must be configured with a different common mount path."

documentation/domains/Domain.md
```diff
@@ -89,6 +90,15 @@ The current status of the operation of the WebLogic domain. Updated automaticall
...
+| `name` | string | The name of common mount volume. |
```
tbarnes-us, line 99: Consider moving `name` first - it's the most important.  Also, consider making `mountPath` second.

tbarnes-us: Add "Required."

documentation/domains/Domain.md
ankedia: sorry, the current set of changes are a work in progress based on new schema, I still need to do code cleanup. I have converted the PR to be draft.

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,90 @@ function adjustPath() {
...
+        trace "Common Mount: Contents of '${out_file}':"
+        cat $out_file
+        trace "Common Mount: End of '${out_file}' contents"
+      done
+      [ "${severe_found}" = "true" ] && return 1
+      rm -fr $COMMON_MOUNT_PATH/commonMountLogs
+      [ -z "$(ls -A $COMMON_MOUNT_PATH)" ] \
+        && trace SEVERE "Common Mount: No files found in '$COMMON_MOUNT_PATH'. " \
+         "Do your commonMount images have files in their '$COMMON_MOUNT_PATH' directories? " \
+         "This path is configurable using the domain resource 'serverPod.commonMount.mountPath' attribute." \
```
tbarnes-us, line 937: --> 'spec.commonMountVolumes.mountPath'

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -279,6 +280,10 @@
   @Description("Models and overrides affecting the WebLogic domain configuration.")
   private Configuration configuration;
 
+  @Description("The common mount tuning. Allows overriding the global default values for mountPath, medium "
+          + "and sizeLimits for the common mount. See spec.commonMountTuning.globalDefaults for more details.")
```
tbarnes-us, line 284: Change this to match the recommended Domain.md wording.

documentation/domains/Domain.md
ankedia: I have name and mountPath defined as first and second in CommonMountVolume.java but somehow generated `Domain.md ` is not adding it in that order (it seems to be sorted alphabetically). Any ideas on how to make name and mountPath appear first in the generated file?

documentation/domains/Domain.md
```diff
@@ -228,6 +242,17 @@ The current status of the operation of the WebLogic domain. Updated automaticall
...
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `command` | string | The command for this init container. Defaults to 'cp -R $COMMON_MOUNT_PATH/* $TARGET_MOUNT_PATH'. This is an advanced setting for customizing the container command for copying files from the container image to the common mount emptyDir volume. Use the '$COMMON_MOUNT_PATH' environment variable to reference the value configured in 'commonMount.mountPath' (which defaults to '/common'). Use 'TARGET_MOUNT_PATH' to refer to the temporary directory created by the Operator that resolves to the common mount's internal emptyDir volume. |
+| `image` | string | The name of an image with files located in directory specified by 'spec.commonMountVolumes.mountPath' of the common mount volume referenced by serverPod.commonMounts.volume (which defaults to '/common'). |
+| `imagePullPolicy` | string | The image pull policy for the common mount container image. Legal values are Always, Never, and IfNotPresent. Defaults to Always if image ends in :latest; IfNotPresent, otherwise. |
+| `volume` | string | The name of common mount volume defined in 'spec.commonMountVolumes'. Required. |
```
tbarnes-us, line 254: Add an 'a': The name of a common mount volume...

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -982,6 +1002,59 @@ private boolean skipValidation(String mountPath) {
...
+    }
+
+    private void checkDuplicateCommomMountVolume(CommonMountVolume commonMountVolume) {
+      if (commonMountVolumes.stream().anyMatch(cmv -> cmv.getMountPath().equals(commonMountVolume.getMountPath()))) {
+        failures.add(DomainValidationMessages.duplicateCMVMountPath(commonMountVolume.getMountPath()));
+      } else if (commonMountVolumes.stream().anyMatch(cmv ->
+              cmv.getName().equals(toDns1123LegalName(commonMountVolume.getName())))) {
+        failures.add(DomainValidationMessages.duplicateCommonMountVolumeName(commonMountVolume.getName()));
+      } else {
+        commonMountVolumes.add(commonMountVolume);
```
tbarnes-us, line 1054: Minor: IMO, a check function should not have side effects unrelated to the check. This check adds cMV to commonMountVolumes...

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainValidationMessages.java
```diff
@@ -27,6 +27,25 @@ static String duplicateServerName(@Nonnull String serverName) {
...
+    return getMessage(MessageKeys.DUPLICATE_COMMON_MOUNT_VOLUME_FOUND, volumeName);
+  }
+
+  /**
+   * Returns a validation message indicating that more than one common mount volume has same mountPath.
+   * @param mountPath the duplicate mount path
+   * @return the localized message
+   */
+  static String duplicateCMVMountPath(@Nonnull String mountPath) {
+    return getMessage(MessageKeys.DUPLICATE_MOUNT_PATH_FOUND, mountPath);
```
tbarnes-us, line 46: _MOUNT_ --> _COMMON_MOUNT_

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,94 @@ function adjustPath() {
...
+    traceDirs $COMMON_MOUNT_PATH
+    touch ${COMMON_MOUNT_PATH}/testaccess.tmp
+    if [ $? -ne 0 ]; then
+      trace SEVERE "Common Mount: Cannot write to the COMMON_MOUNT_PATH '${COMMON_MOUNT_PATH}'. " \
+                   "This path is configurable using the domain resource 'spec.commonMountVolumes.mountPath' " \
+                   "attribute." && return 1
+    fi
+    rm -f ${COMMON_MOUNT_PATH}/testaccess.tmp || return 1
+
+    out_files=$(set -o pipefail ; ls -1 $COMMON_MOUNT_PATH/commonMountLogs/*.out 2>1 | sort --version-sort) \
```
tbarnes-us, line 921: Minor: 
Add comment at 920: 
`# The container .out files embed their container name, the names will sort in the same order in which the containers ran`

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,94 @@ function adjustPath() {
...
+      || (trace SEVERE "Common Mount: Assertion failure. No files found in '$COMMON_MOUNT_PATH/commonMountLogs/*.out" \
+      && return 1)
+    severe_found=false
+    for out_file in $out_files; do
+      if [ "$(grep -c SEVERE $out_file)" != "0" ]; then
+        trace SEVERE "Common Mount: Error found in file '${out_file}' while initializing commonMount."
+        severe_found=true
+      elif [ "$(grep -c successfully $out_file)" = "0" ]; then
+        trace SEVERE "Common Mount: Command execution was unsuccessful in file '${out_file}' while initializing commonMount."
+        severe_found=true
```
tbarnes-us, line 931: If we arrive here, the `trace "Common Mount: Contents of '${out_file}':"` should be skipped, so that the operator's parse of this SEVERE will include the contents of the .out file.  In addition, add the sentence 'Contents of '${out_file}' to this SEVERE.

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,94 @@ function adjustPath() {
...
+    fi
+    rm -f ${COMMON_MOUNT_PATH}/testaccess.tmp || return 1
+
+    out_files=$(set -o pipefail ; ls -1 $COMMON_MOUNT_PATH/commonMountLogs/*.out 2>1 | sort --version-sort) \
+      || (trace SEVERE "Common Mount: Assertion failure. No files found in '$COMMON_MOUNT_PATH/commonMountLogs/*.out" \
+      && return 1)
+    severe_found=false
+    for out_file in $out_files; do
+      if [ "$(grep -c SEVERE $out_file)" != "0" ]; then
+        trace SEVERE "Common Mount: Error found in file '${out_file}' while initializing commonMount."
```
tbarnes-us, line 927: We can turn this SEVERE into a FINE. There will already be a SEVERE in the .out file that we're about to cat, so this will reduce the verbosity of SEVERE reporting/parsing...

operator/src/main/resources/scripts/utils.sh
```diff
@@ -889,3 +889,94 @@ function adjustPath() {
...
+#            If the specified COMMON_MOUNT_COMMAND is empty, it logs an error message and returns.
+#            If the COMMON_MOUNT_PATH directory doesn't exist or is empty, it logs error and returns.
+#            If the command execution fails, it logs errror message with failure details. Otherwise it
+#            logs a success message with details.
+#            See also 'commonMount.sh'.
+#
+function initCommonMount() {
+
+  if [ -z "${COMMON_MOUNT_COMMAND}" ]; then
+    trace ERROR "Common Mount: The 'serverPod.commonMounts.command' is empty for the " \
```
tbarnes-us, line 960: Minor:  it's more clear to use 'trace SEVERE' instead of 'trace ERROR'  (it's more clear as 'trace ERROR' in this location secretly depends on the fact that trace implicitly converts an ERROR to SEVERE, the operator greps for SEVERE, and the checkCommonMount script greps for it too)

tbarnes-us (changes requested):
Only have minor comments now...

documentation/domains/Domain.md
tbarnes-us: No. This is a minor issue - closing as resolved as we can ignore it.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
ankedia: The `commonMountVolumes` HashSet here is only used for validation purposes (to check for a duplicate name or duplicate mount path). I followed the pattern similar to `checkDuplicateClusterName` method which does the following. I can change the name or create two HashSets (one for common volume name and one for mount path) if that makes it more clear.

    private void checkDuplicateClusterName(String clusterName) {
      if (clusterNames.contains(clusterName)) {
        failures.add(DomainValidationMessages.duplicateClusterName(clusterName));
      } else {
        clusterNames.add(clusterName);
      }
    }

tbarnes-us: Oh - ok. Thank for explaining. Resolving.

ankedia:
@tbarnes-us I have addressed the review comments on this PR. Please let me know if I missed something.

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -0,0 +1,256 @@
+# Copyright (c) 2021, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
```
tbarnes-us, line 3: Add comment something like:

# Important: Functions defined in file this must be able to work with unknown shells,
#                   e.g. pure shell instead of bash, when UNKNOWN_SHELL has been set to true.

documentation/staging/content/userguide/managing-domains/model-in-image/common-mounts/_index.md
```diff
@@ -0,0 +1,56 @@
...
+      volume: commonMountsVolume1
+```
+
+The `volume` field in the common mounts refers to the name of a common mount volume defined in `spec.commonMountVolumes` section. Here's an example configuration for the common mount volumes.
+```
+  spec:
+    commonMountVolumes:
+    - name: commonMountsVolume1
+      mountPath: /common
+      medium: Memory
```
tbarnes-us, line 33: Should not use 'Memory' - this is not the default, and pod memory may be at a premium

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -0,0 +1,256 @@
...
+}
+
+#
+# initCommonMount
+#   purpose: Execute the COMMON_MOUNT_COMMAND specified as part of the common mount init container.
+#            If the specified COMMON_MOUNT_COMMAND is empty, it logs an error message and returns.
+#            If the COMMON_MOUNT_PATH directory doesn't exist or is empty, it logs error and returns.
+#            If the command execution fails, it logs errror message with failure details. Otherwise it
+#            logs a success message with details.
+#            See also 'commonMount.sh'.
```
tbarnes-us, line 229: Minor: Add something like: See also checkCommonMount in 'utils.sh.'
Minor: Similarly update the checkCommonMount comment to point  to initCommonMount in 'base_utils.sh'...

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -213,6 +214,18 @@
...
+  /**
+   * The common mount.
+   *
+   */
+  @Description("Use a common mount to automatically include directory content from additional images. "
+          + "This is a useful alternative for including Model in Image model files, or other types of files, in a pod "
+          + "without requiring modifications to the pod's base image 'domain.spec.image'. "
+          + "This feature internally uses a Kubernetes emptyDir volume and Kubernetes init containers to share "
+          + "the files from the additional images with the pod.")
+  @Feature("CommonMounts")
```
rjeberhard, line 226: In DomainSpec, you had @Feature("CommonMount"). Here you have the plural.

.github/workflows/publish-github-pages.yml
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2021, Oracle and/or its affiliates.
```
tbarnes-us: This file looks unrelated to the common mount feature - looks like an incorrect merge?

tbarnes-us: Ditto for a number of the recent doc file changes.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
ankedia: I have changed the DomainSpec to use plural @Feature("Common Mounts") in 2865293 . Thanks,

rjeberhard (approved):

tbarnes-us (approved):
Cool!

PR #2343, opened 2021-04-27
update patching doc

rosemarymarano:
As per Monica, OWLS-89082 Add new WKO documentation which describe how to patch WebLogic images using WIT

### DIFF
--- a/documentation/staging/content/userguide/base-images/_index.md
+++ b/documentation/staging/content/userguide/base-images/_index.md
@@ -162,9 +162,11 @@ WebLogic Server images with quarterly Patch Set Updates (PSUs), which include se
 
 Use either the WIT [`create`](https://github.com/oracle/weblogic-image-tool/blob/master/site/create-image.md) or
 [`update`](https://github.com/oracle/weblogic-image-tool/blob/master/site/update-image.md) command, however,
-patching using the `create` command results in a smaller WebLogic Server image size.
+patching using the `create` command results in a smaller WebLogic Server image size. Note that you will need to
+download the WebLogic Server 12.2.1.4.0 installer and JDK installer prior to running the `create` command. For details, see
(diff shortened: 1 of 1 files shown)


documentation/staging/content/userguide/base-images/_index.md
```diff
@@ -154,7 +155,21 @@ including keys and credentials that are used to access external resources
...
 
+Use the [WebLogic Image Tool](https://github.com/oracle/weblogic-image-tool) (WIT) to patch
+WebLogic Server images with quarterly Patch Set Updates (PSUs), which include security fixes, or with one-off patches.
+
+Use either the WIT [`create`](https://github.com/oracle/weblogic-image-tool/blob/master/site/create-image.md) or
+[`update`](https://github.com/oracle/weblogic-image-tool/blob/master/site/update-image.md) command, however,
+patching using the `create` command results in a smaller WebLogic Server image size.
+
+Example: Create an image named `sample:wls` with the WebLogic Server 12.2.1.4.0 slim installer, JDK 8u291, latest Oracle 7-slim,
+and latest PSU and recommended CPU and SPU patches applied.
```
rjeberhard, line 168: This sample command presupposes that the customer has downloaded the installer and the patches and placed them in the image tool's cache. We could debate if we should document all of those steps here, but if we don't then we do need to document that downloading these artifacts in advance is required.

ddsharpe: +1, can we just add a "...(you will need to download the WebLogic Server 12.2.1.4.0 installer and JDK installer prior to running the create command)."

rjeberhard: Yes, let's do that for now.

rosemarymarano: Yes. For details, see the WIT Quick Start guide, https://github.com/oracle/weblogic-image-tool/blob/master/site/quickstart.md). Does that work?

PR #2344, opened 2021-04-27
Recreate OWLS-86115 work

rjeberhard:
I'm not sure what happened with GitHub, but when I tried to change the base for @lennyphan's work GitHub created a PR with hundreds of changes. This PR is recreated with just his deltas, but targeted to the main branch. (Original pull: https://github.com/oracle/weblogic-kubernetes-operator/pull/2285).

In this PR:

- For Domain home source type MII, configuration overrides are now generated using WDT model filter. The model filter implementation follows the same logic for generating situational configuration in introspectDomain.py's SitConfigGenerator. Domain home source types Domain-on-PV and Domain-in-Image will continue to use situational configuration overrides.

- Added ability to run python based unit tests for model filter implementation.


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -319,9 +320,9 @@ function createWLDomain() {
   fi
 
   # copy the filter related files to the wdt lib
-
   cp ${WDT_FILTER_JSON} ${WDT_ROOT}/lib/model_filters.json
   cp ${WDT_CREATE_FILTER} ${WDT_ROOT}/lib
+  cp ${WDT_MII_FILTER} ${WDT_ROOT}/lib
```
tbarnes-us, line 325: consider appending || trace SEVERE "cp failed" && exitOrLoop to each of the cp commands (fail-fast)

tbarnes-us (approved):
Way cool!

tbarnes-us:
General comment: I think this change will likely lift a restriction or two that we document in the user guide 'model updates' chapter. We (@jshum2479 @lennyphan and I) should meet to discuss updating the chapter, and whether we should test first.

lennyphan:
Successful run of all integration tests on this branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4895/

PR #2345, opened 2021-04-27
Owls88697 stop namespaces that no longer managed by operator when change to dedicated strategy

doxiao:
For List, RegExp and LabelSelector domain selection strategy, the operator cleans up the namespaces that it does not managed any more in Namespaces.NamespaceListAfterStep. The cleanup is missed for the dedicated case because the operator does not go through the steps to list the existing namespaces. The fix is to add the Namespaces.NamespaceListAfterStep to the dedicated case and customize the getFoundDomainNamespaces to return the operator namespace.

The PR also makes the operator perform more necessary clean upon stopping managing a namespace or when a domain is deleted.

Unit test cases are added to cover the scenarios associated with the code changes. Some existing unit test cases where a particular domain namespaces selection strategy is tested are modified to run CreateReadNamespacesStep instead of readExsitingNamespaces because the latter does not exercise the strategy specific code path.

The test case in ItKuerbenetesEvents.java for dedicated strategy that used to fail because of this issue is turned on. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4870/

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/MainTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/MainTest.java
@@ -838,6 +838,60 @@ public class MainTest extends ThreadFactoryTestBase {
   }
 
   @Test
+  public void withNamespaceList_changeToDedicatedAfterReadExistingNSs_onRecheck_nsWatchStoppedEventCreated() {
+    defineSelectionStrategy(SelectionStrategy.List);
+    String namespaceString = "NS1";
+    HelmAccessStub.defineVariable(HelmAccess.OPERATOR_DOMAIN_NAMESPACES, namespaceString);
+    createNamespaces(4);
+
+    testSupport.runSteps(createDomainRecheck().readExistingNamespaces());
+
+    defineSelectionStrategy(SelectionStrategy.Dedicated);
+
+    recheckDomains();
+
+    assertThat("Found NAMESPACE_WATCHING_STOPPED event with expected message",
+        containsEventWithMessageForNamespaces(getEvents(testSupport),
+            NAMESPACE_WATCHING_STOPPED, Collections.singletonList("NS1")), is(true));
+  }
+
+  @Test
+  public void withNamespaceList_changeToDedicatedAfterReadExistingNSs_onRecheck_nsWatchStartedEventCreatedInOpNS() {
+    defineSelectionStrategy(SelectionStrategy.List);
+    String namespaceString = "NS1";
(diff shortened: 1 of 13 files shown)


rjeberhard (approved):

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
```diff
@@ -833,10 +836,83 @@ public void withNamespaceList_onReadExistingNamespaces_whenDomainNamespaceRemove
...
+    runCreateReadNamespacesStep();
     assertThat(logRecords, containsInfo(MessageKeys.END_MANAGING_NAMESPACE, "NS3"));
   }
 
+  @Test
+  public void withNamespaceList_changeToDedicated_onCreateReadNamespaces_nsWatchStoppedEventCreated() {
+    defineSelectionStrategy(SelectionStrategy.List);
+    String namespaceString = "NS1";
+    HelmAccessStub.defineVariable(HelmAccess.OPERATOR_DOMAIN_NAMESPACES, namespaceString);
+    createNamespaces(4);
```
russgold, line 848: I have found that organizing tests can help readability:

1. general setup (what the world looks like before the operation)
2. the action to test
3. the assertion.


You can do that with judicious use of spaces. Here, lines 845-850 establish your starting point, and 852-854 are the action, so I would remove the blank lines in between them, and reserve those as separators only between the three sections.

If those are complex, you might consider extract them to their own methods, especially if you do the same setup in multiple tests.

russgold (commented):
Generally looks good; I have left a style recommendation.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
```diff
@@ -833,10 +836,66 @@ public void withNamespaceList_onReadExistingNamespaces_whenDomainNamespaceRemove
...
+    assertThat("Found NAMESPACE_WATCHING_STOPPED event with expected message",
+        containsEventWithMessageForNamespaces(getEvents(testSupport),
+            NAMESPACE_WATCHING_STOPPED, Collections.singletonList("NS1")), is(true));
+  }
+
+  @Test
+  public void withNamespaceList_changeToDedicated_onCreateReadNamespaces_nsWatchStartedEventCreatedInOpNS() {
+    setupDomainWithNamespaceListStrategy("NS1");
+    runCreateReadNamespacesStep();
+    defineSelectionStrategy(SelectionStrategy.Dedicated);
```
russgold, line 859: Here you are changing to a new strategy, so line 859 begins the 'action' part of the test and should be separated by a blank line from the 'setup'

PR #2347, opened 2021-04-29
Remove unnecessary Kubernetes version check from dedicated mode test

rjeberhard:
We no longer have the "beta" CRD documents. Additionally, the "kubectl version" check was misleading because later Kubernetes versions still contain "1.15" in the output because of some of the other fields.

I have started a jenkin run  
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4912/


doxiao (approved):
LGTM.

PR #2351, opened 2021-05-03
Fix missing error output from WDT compareModel

jshum2479:
This fixes the missing output from WDT compareModel tool when the tool failed itself and not creating the compare_model_output file, using redirected terminal output instead.

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -609,11 +609,11 @@ function diff_model() {
   # $1 - new model, $2 original model
 
   # no need to redirect output, -output_dir for compareModel will generate the output file
-  ${WDT_BINDIR}/compareModel.sh -oracle_home ${ORACLE_HOME} -output_dir /tmp $1 $2 > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


tbarnes-us (approved):
LGTM. I deleted the outdated comment above the change.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -608,12 +608,11 @@ function diff_model() {
   export __WLSDEPLOY_STORE_MODEL__=1
   # $1 - new model, $2 original model
 
-  # no need to redirect output, -output_dir for compareModel will generate the output file
-  ${WDT_BINDIR}/compareModel.sh -oracle_home ${ORACLE_HOME} -output_dir /tmp $1 $2 > /dev/null 2>&1
+  ${WDT_BINDIR}/compareModel.sh -oracle_home ${ORACLE_HOME} -output_dir /tmp $1 $2 >  ${WDT_OUTPUT} 2>&1
```
rjeberhard, line 611: Is the double space after the > intentional?

jshum2479: no, I can remove it.

rjeberhard (approved):

PR #2352, opened 2021-05-03
minor event doc update

tbarnes-us:
use ',' instead of '&&' in kubectl selector as '&&' is only valid in new versions of kubectl


doxiao (approved):
LGTM. thanks

rjeberhard (approved):

PR #2355, opened 2021-05-04
change the assert for admin port change

sankarpn:
Some times the rest api call returns an incorrect exit code and this change checks if the test needs to be failed in case if the error stream is not null and non empty

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/4945/


anpanigr (approved):
LGTM.  Run the parallel tests

sankarpn:
> 
> 
> LGTM. Run the parallel tests

Updated it with parallel run results

xiancao (approved):

PR #2361, opened 2021-05-14
Feature gates

rjeberhard:
The intention is to create a reusable feature gating functionality that lets us check in other features (e.g. common mounts) into the `main` branch prior to that other feature being ready for release. For instance, the feature may lack documentation or a complete set of integration tests. 

The goal is to preserve our ability to release the main branch as a bug-fix release at any time and not have to resort to backports to a previous release branch. 

The customer can then enable a feature as part of the Helm install for the operator, which is also the pattern that we will follow in our integration tests. Sometime later, the specific feature can either be set to be enabled by default or the checks for the feature gate can be removed.

The feature gating functionality includes a Helm value where the customer (or test cases) can enable a feature, an annotation that can be used on fields in our model classes to designate them as related to the feature, and API's to check if a given feature is enabled prior to invoking feature-related code paths.

Note: I also cleaned up a few spelling errors and removed a few unused strings in the message bundle.


jshum2479 (approved):

ankedia (approved):
LGTM

PR #2365, opened 2021-05-18
Include monitoring exporter sidecar and Prometheus annotations for admin server

rjeberhard:
This is a fairly simple change based on an update to the requirements and expected behavior. Previously, the operator injected the monitoring exporter sidecar and created Prometheus annotations only for managed server instances. Now, all servers are included, specifically including the admin server.

The necessary changes were to 1) move tests up from the managed pod suite to the base class so that the tests are included for all servers, and then 2) make the similar change to push behavior up from the managed pod step context to the general pod step context classes.


doxiao (approved):
LGTM.

ddsharpe (approved):

PR #2367, opened 2021-05-18
add note for create-domain.sh

rosemarymarano:
As per OWLS-89511: Update doc for samples that use create-domain.sh


tbarnes-us:
I'm leery of making the notice too verbose, but it could be a little more precise:

The `create-domain.sh` script and its inputs file are for demonstration purposes _only_; its contents might change without notice.

Could be:

The `create-domain.sh` script and its inputs file are for demonstration purposes _only_; its contents and the domain resource file that it generates for you might change without notice.

tbarnes-us (approved):
LGTM - Approving - I've suggested a small change but am not sure it should go in - leaving it up to you.

rosemarymarano:
I am ok with the suggestion and have taken it.

rjeberhard (approved):

PR #2372, opened 2021-05-20
Removing the cluster from the kubernetes spec section of the wdt k8s template file

bhavaniravichandran:
Our samples create a domain with 1 cluster by default, but one of our customers want to be able to create a domain with more than 1 cluster. If the cluster is specified in kubernetes spec section of the wdt model file, then the domain resource spec section only lists that 1 cluster even though MS's belonging to all the clusters do start. So, removing the cluster from the spec section, so that all the clusters will be listed in the domain resource spec section. 
I am commenting out the cluster instead of removing it in case anyone wants to know how to specify clusters in the k8s spec section of the wdt model file.


bhavaniravichandran:
submitted jenkins job with the change

ddsharpe (approved):

bhavaniravichandran:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5066/

rjeberhard (approved):

PR #2373, opened 2021-05-20
Script to do the set up and run integration tests from Jenkins

vanajamukkara:
Currently Jenkins build contains commands to set up helm, kubectl, kind and to run the integration tests. This script will replace those commands.

Tested on https://build.weblogick8s.org:8443/job/wko-tmp/29/


jenkinsScript.sh
```diff
@@ -0,0 +1,112 @@
...
+# OCR_PASSWORD
+# OCR_USERNAME
+# OCIR_USERNAME
+# OCIR_PASSWORD
+# OCIR_EMAIL
+
+set -o errexit
+set -o pipefail
+
+function checkEnvVar {
```
tbarnes-us, line 24: Multiple issues with the env var checks:
- the function is relying on $0, but $0 is not the the first parm, it's $1 ($0 is always set to the filename of the script)
- it looks like the intent is to pass the name of the var, not the val of the var to this check
  so the script should be using ${!1} to dereference the val from the var name, and
  the caller of teh script should be passing VARNAME not "$VARNAME"
- it'd be nice to display errors for all bad env vars, not just the first one encountered
- it'd be nice to display the values of good env vars as long as they are not passwords

Fix by:
```
function checkEnvVars {
  local has_errors=false
  while [ ! -z "${1}" ]; do
    if [ -z "${!1}" ]; then
      echo "Error: '${1}' env variable is not set"
      has_errors=true
    else
      if [ "${1/PASSWORD//}" = "${1}" ]; then
        echo "Info: env var ${1}='${!1}'"
      else
        echo "Info: env var ${1}='***'"
      fi
    fi
    shift
  done
  if [ ! "$has_errors" = "false" ]; then
    echo "Error: Missing env vars, exiting."
    exit 1
  fi
}

checkEnvVars  \
   APACHE_MAVEN_HOME  \
   HELM_VERSION  \
   KUBECTL_VERSION \
   KIND_VERSION  \
   IT_TEST  \
   WDT_DOWNLOAD_URL  \
   WIT_DOWNLOAD_URL  \
   NUMBER_OF_THREADS  \
   JAVA_HOME  \
   OCR_PASSWORD  \
   OCR_USERNAME  \
   OCIR_USERNAME  \
   OCIR_PASSWORD  \
   OCIR_EMAIL
```

tbarnes-us:
LGTM apart from the minor bug which is now fixed. I'm not familiar with this area overall, so am leaving final approvals up to @anpanigr and @anpanigr ...

vanajamukkara:
> LGTM apart from the minor bug which is now fixed. I'm not familiar with this area overall, so am leaving final approvals up to @anpanigr and @rjeberhard ...

Thanks for reviewing @tbarnes-us

rjeberhard (approved):

anpanigr (approved):

PR #2374, opened 2021-05-21
Restore role definition to match documentation

rjeberhard:
There was a previous effort to limit the permissions given to the operator. We apparently cut too close to the bone. This change restores some permissions. It happens that this makes the Helm chart match what we currently document ;). With these restored permissions then operator can record the internal certificate that it generates.


ddsharpe (approved):

ankedia (approved):
LGTM

PR #2379, opened 2021-05-26
Fix mii wdt filterpy

jshum2479:
We need to set the default listenPort/ssl listenPort if one is not provided, otherwise the istio nap injected will use 0 for the port, resulting in validation error during introspection.

### DIFF
--- a/operator/src/main/resources/scripts/model_wdt_mii_filter.py
+++ b/operator/src/main/resources/scripts/model_wdt_mii_filter.py
@@ -418,8 +418,11 @@ def customizeServerIstioNetworkAccessPoint(server, listen_address):
   model = env.getModel()
   if ssl is not None and 'Enabled' in ssl and ssl['Enabled'] == 'true':
     ssl_listen_port = ssl['ListenPort']
+    if ssl_listen_port is None:
+      ssl_listen_port = 7002
   elif ssl is None and isSecureModeEnabledForDomain(model['topology']):
     ssl_listen_port = "7002"
+
 
   if ssl_listen_port is not None:
     _writeIstioNAP(name='https-secure', server=server, listen_address=listen_address,
@@ -481,6 +484,8 @@ def customizeManagedIstioNetworkAccessPoint(template, listen_address):
   model = env.getModel()
   if ssl is not None and 'Enabled' in ssl and ssl['Enabled'] == 'true':
     ssl_listen_port = ssl['ListenPort']
+    if ssl_listen_port is None:
+      ssl_listen_port = 7002
   elif ssl is None and isSecureModeEnabledForDomain(model['topology']):
     ssl_listen_port = "7002"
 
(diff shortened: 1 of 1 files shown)


tbarnes-us:
General comment #1:  Does introspectDomain.py have a similar issue?

tbarnes-us:
General comment #2: IIRC, the default changes based on whether security is enabled for the domain.

jshum2479:
> General comment #2: IIRC, the default behavior changes based on whether security is enabled for the domain. Is that applicable here?

The change is for user not providing listenPort and/or ssl listenPort when it is enabled which causes the injected nap with no value - that becomes 0  which is not valid.

lennyphan (approved):

operator/src/main/resources/scripts/model_wdt_mii_filter.py
```diff
@@ -385,6 +385,11 @@ def customizeServerIstioNetworkAccessPoint(server, listen_address):
   if istio_readiness_port is None:
     return
   admin_server_port = server['ListenPort']
+  # Set the default if it is not provided to avoid nap default to 0 which fails validation.
+
+  if admin_server_port is None:
+    admin_server_port = 7001
```
ddsharpe, line 391: What happens if the domain is in SecuredMode?

jshum2479: istio won't work with secure mode yet

rjeberhard: Oh! That seems like an issue. Do we have a JIRA to complete that support? I'd like to resolve that gap before the Verrazzano team hits it.

rjeberhard: @ddsharpe @jshum2479, what would be the correct behavior for secure mode? Does this "None" check let us differentiate between when the customer has left it blank or if the admin server's default port is disabled because of secure mode?

jshum2479: Note:  the filter code already have provision to setup the ssl port  if the secure mode is enabled (line 423).    

But, ultimately it will not work in istio environment. 

1) whenever admin port (same for secure mode) is enabled, the readiness probe /weblogic/ready is treated as management function because it started with /weblogic and it must be accessed directly read-address:adminport and not proxied it through localhost:adminport
2) Istio always proxy it through unless there is annotation to forbid the rewrite the port traffic (essentially take it out of the mesh).
3) operator current implementation always use the plain readiness port in the domain spec,  while it can be fixed in PodStepContext but it won't fix (2).

jshum2479: As for the correct behavior,  I am not sure what's the correct behavior.  If secure mode is enabled, does it mean regular listen port is disabled?

rjeberhard: Yes, that's right: https://docs.oracle.com/en/middleware/fusion-middleware/weblogic-server/12.2.1.3/lockd/secure.html#GUID-F980DB67-7CDE-4EF8-986D-D188D4EDB706

jshum2479: So the core should automatically disabled the listenPort.  The question is whether the operator needs to do anything special about it.   I suggest creating another issue to handle secure mode as I suspect there maybe issue in non-istio case.  This PR is for missing listenPort only.  For adminport/secure mode it's a dead end for istio for now.

rjeberhard (approved):

PR #2381, opened 2021-05-26
FAQ update for external clients

lennyphan:
Documentation changes for OWLS-89509:

1. Add Monica's suggestion to add documentation note to emphasize using 'Load balancer tunneling' over Node Port
2. Add documentation for configuring 'External Listen Address' for cross domain transactions within same K8s cluster.
3. Add documentation for configuring 'Server Affinity' load balancing algorithm so tunneling clients using smart stubs are routed to the same pod.


documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+```
+How to implement Server Affinity in Cluster: 
+=========================================== 
+1. Go to  admin console 
+2. Expand the Clusters link on the left panel 
+3. Click on 'cluster-1' link 
+    i.   look on the right panel. 
+    ii.  look for Default Load Algorithm property 
+    iii. change the value to "RoundRobinAffinity" instead of "RoundRobin". 
+4. Restart the domain/weblogic server instances.
```
anpanigr, line 399: In WKO environment all the changes to config.xml is lost if we do not use the Domain-on-PV model. So this is not a suggested way to modify the cluster attribute.

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+1. Go to Admin console 
+2. Expand the Environment link on the left panel 
+3. Click on the Servers link on the sub-panel
+4. On the right panel, click on AdminServer -> Configuration -> General tabs. 
+5. Click on the Advanced link to expand the server configuration page.
+6. Enter the service name for the WLS server pod in the External Listen Address field.
+   For example:
+     i. For Admin Server the service name would be of the form <DOMAIN_UID>-admin-server.<NAMESPACE> 
+     ii. For a managed server the service name would be of the form <DOMAIN_UID>-managed-server${id}.<NAMESPACE>
+```
```
anpanigr, line 444: In WKO environment any change to config.xml is lost if we do not use the Domain-on-PV model. So this is not a suggested way to modify the cluster attribute.

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+2. Expand the Environment link on the left panel 
+3. Click on the Clusters link on the sub-panel
+4. Click on the Server Templates link on the sub-panel
+5. On the right panel, click on cluster-1-template -> Configuration -> General tabs. 
+6. Click on the Advanced link to expand the server configuration page.
+7. Enter the service name for the WLS server pod in the External Listen Address field.
+   For example:
+     i. For Admin Server the service name would be of the form <DOMAIN_UID>-admin-server.<NAMESPACE> 
+     ii. For a managed server the service name would be of the form <DOMAIN_UID>-managed-server${id}.<NAMESPACE>
+```
```
anpanigr, line 462: In WKO environment any change to config.xml is lost if we do not use the Domain-on-PV model. So this is not a suggested way to modify the cluster attribute.

documentation/staging/content/faq/external-clients.md
tbarnes-us: I think you clearly document that this is for teh domain-on-PV model, which mitigates Pani's concern, but I also think it isn't the preferred way to make a change even for domain-on-PV.  The WLST offline & WDT snippets are sufficient IMO.  The WL console instructions can be removed. 

Ditto for the other locations Pani made the same comment...

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+4. Click on the Server Templates link on the sub-panel
+5. On the right panel, click on cluster-1-template -> Configuration -> General tabs. 
+6. Click on the Advanced link to expand the server configuration page.
+7. Enter the service name for the WLS server pod in the External Listen Address field.
+   For example:
+     i. For Admin Server the service name would be of the form <DOMAIN_UID>-admin-server.<NAMESPACE> 
+     ii. For a managed server the service name would be of the form <DOMAIN_UID>-managed-server${id}.<NAMESPACE>
+``` 
+
+Here is a snippet of offline WLST code for setting the 'External Listen Address' for a stand-alone server named 'AdminServer':
```
tbarnes-us, line 464: add "in namespace 'weblogic-domain'".

tbarnes-us: and add for domain-uid `domain1`.

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+
+```javascript
+serverName = "AdminServer"
+cd('/Servers/%s' % serverName)
+set('ExternalDNSName', 'domain1-adminserver.weblogic-domain')
```
tbarnes-us, line 469: Please:
Add namespace = "weblogic.domain".
Add domainUid = "domain1"
Add code for creating the dns name from the three vars and making the DNSName DNS1123 compliant
change the set to use the calculate dns name

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+```
+
+Here is a snippet of offline WLST code for setting the 'External Listen Address' for a server template named 'cluster-1-template':
```
tbarnes-us, line 472: Please update intro sentence to mention namespace and domain-uid and cluster-name, update the script to set vars for same, update the script to calc DNS1123 name, etc.

tbarnes-us: Also update the ExternalDNSName to use the managed-server${id} instead of managed-server1

tbarnes-us: (WebLogic itself will substitute the right value for ${id} at runtime.)

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
   * Also apply patch 30656708 on each target WebLogic Server for versions 12.2.1.4 (PS4) or earlier.
 
+#### Configuring WebLogic Server Affinity Load Balancing Algorithm
+
+When providing external clients access to EJB and JMS resources using load balancer tunneling, it is recommended to configure a server affinity load balancing algorithm for the cluster in which the resources are targeted.  Using a server affinity based algorithm turns off load balancing between external Java clients and server instances causing method calls to 'stick' to a server instance to which the external client is communicating with.  For more information see [Load Balancing for EJBs and RMI Objects](https://docs.oracle.com/en/middleware/fusion-middleware/weblogic-server/12.2.1.4/clust/load_balancing.html#GUID-2470EEE9-F6F9-44EF-BA54-671728E93DE6) in _Load Balancing in a Cluster_.
```
tbarnes-us, line 386: IMO this doesn't really 'turn off load balancing' from a customer perspective, since RMI stub stub load balancing is a somewhat hidden feature, and there was already load balancing decision made for the JNDI context itself when the JNDI context was created.

How about something like:

Using a server affinity based algorithm reduces the amount of time it takes for EJB and JMS stand-alone or server hosted clients to establish a connection through a tunneling port.  It does this by ensuring that the clients prefer communicating with a target server instance which already has an established connection to the client, where this server instance is implicitly picked when such clients create their JNDI context.

documentation/staging/content/faq/external-clients.md
```diff
@@ -114,6 +120,8 @@ is hosted outside of the Kubernetes cluster, then:
 - You may need to [enable unknown host access](#enabling-unknown-host-access)
   on the WebLogic Servers that host the EJB or JMS resources.
 
+- You may need to configure a default load balancer algorithm that provides server affinity, for the cluster in which EJB or JMS resources are targted, to enable appropriate routing of requests via replica-aware stubs.
```
tbarnes-us, line 123: 1) You may -->  "If tunneling, "

2) Please add a link to the new section for changing the algorithm.   

3) "replica-aware stub" is an internal term IMO - so I recommend changing "
to enable appropriate routing of requests via replica-aware stubs" to 
something like "to speedup connection creation for EJB and JMS clients."

documentation/staging/content/faq/external-clients.md
```diff
@@ -124,11 +132,16 @@ that call the resource are located within the cluster, then:
     * [Set up an HTTP tunneling-enabled custom channel](#adding-a-weblogic-custom-channel) on the external WebLogic Servers.
     * Specify URLs on the source server that resolve to the load balancer's address and that start with `http` instead of `t3`.
     * Ensure the load balancer configures the HTTP flow to be 'sticky'.
+    * Configure a WebLogic load balancer method that provides server affinity for the cluster's default load balancing algorithm to enable appropriate routing of requests via replica-aware stubs.
```
tbarnes-us, line 135: 1 - Please add a link to the new section for changing the algorithm.

2 - "replica-aware stub" is an internal term IMO - so I recommend changing "
to enable appropriate routing of requests via replica-aware stubs" to
something like "to speedup connection creation for EJB and JMS clients."

documentation/staging/content/faq/external-clients.md
```diff
@@ -141,6 +154,8 @@ Here are the steps:
 
 - __Important__: Ensure that the load balancer configures the HTTP flow to be 'sticky' - for example, a Traefik load balancer has a `sticky sessions` option. This ensures that all of the packets of a tunneling client connection flow to the same pod, otherwise the connection will stall when its packets are load balanced to a different pod.
 
+- __Important__: For EJB and JMS resources that are targeted to a cluster, it is recommended that you configure a default load balancer algorithm that provides server affinity (round-robin-affinity, weight-based-affinity, random-affinity) to enable appropriate routing of requests via replica-aware stubs. For more information see [Load Balancing for EJBs and RMI Objects](https://docs.oracle.com/en/middleware/fusion-middleware/weblogic-server/12.2.1.4/clust/load_balancing.html#GUID-2470EEE9-F6F9-44EF-BA54-671728E93DE6) in _Load Balancing in a Cluster_.
```
tbarnes-us, line 157: "replica-aware stub" is an internal term IMO - so I recommend changing "
to enable appropriate routing of requests via replica-aware stubs" to
something like "to speedup connection creation for EJB and JMS clients."

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+                ServerTemplate:  'cluster-1-template'
+                ServerNamePrefix: 'managed-server'
+                DynamicClusterSize: '5'
+                MaxDynamicClusterSize: '5'
+                MinDynamicClusterSize: '0'
+                CalculatedListenPorts: false
+            DefaultLoadAlgorithm: 'round-robin-affinity'
+```
+
+#### Configuring External Listen Address for WebLogic Default Channel
```
tbarnes-us, line 425: Address --> Addresses

documentation/staging/content/faq/external-clients.md
```diff
@@ -366,6 +381,129 @@ To enable an 'unknown host' source WebLogic Server to initiate EJB, JMS, or JTA
...
+
+WebLogic Server provides an External Listen Address/DNS Name for use in RMI stubs to allow external clients to connect to the server through a Kubernetes external service name. For more information on 'External Listen Address' (i.e. ExternalDNSName attribute) on the ServerMBean, see ExternalDNSName attribute in [ ServerMBean reference ](https://docs.oracle.com/en/middleware/fusion-middleware/weblogic-server/12.2.1.4/wlmbr/core/index.html).
```
tbarnes-us, line 427: "RMI stub" is sort of an internal/opaque concept to customers -- and the external LA feature was not created specifically for "external kubernetes services".

how about rewording to "WebLogic Server provides an External Listen Address/DNS Name feature to enable external JMS and EJB clients to properly reestablish and/or load balance connections to servers with listen addresses that aren't directly available in the client's DNS (such as through a firewall or a load-balancer, or between different namespaces in Kubernetes).

documentation/staging/content/faq/external-clients.md
tbarnes-us: and Channel --> Channels

documentation/staging/content/faq/external-clients.md
```diff
@@ -23,6 +23,8 @@ description: "This FAQ describes approaches for giving WebLogic applications acc
 - [Enabling unknown host access](#enabling-unknown-host-access)
    - [When is it necessary to enable unknown host access?](#when-is-it-necessary-to-enable-unknown-host-access)
    - [How to enable unknown host access](#how-to-enable-unknown-host-access)
+- [Configuring WebLogic Server Affinity Load Balancing Algorithm](#configuring-weblogic-server-affinity-load-balancing-algorithm)
+- [Configuring External Listen Address for WebLogic Default Channel](#configuring-external-listen-address-for-weblogic-default-channel)
```
tbarnes-us, line 27: I just had a customer ask how to put together a cross-namespace URL for the stand-alone server case.

Please update 
```
- The URL used by the application can begin with `t3` or `t3s`, and the  DNS address in the URL must be fully specified to differentiate the namespace. ...
```
to (a) use 'svc.cluster.local' or drop it all-together, and (b) to specifically mention the stand-alone server case.

Something like:
```
- The URL used by the application can begin with `t3` or `t3s`, and the  DNS address in the URL must be fully specified to differentiate the namespace.  For example:
   - If the target resource is in WebLogic cluster `mycluster` with    domain UID `myuid` in namespace `myns`, and the cluster is    listening on port `8001`, then the cluster service name     that is generated by the operator will    be `myuid-cluster-mycluster` and the full URL will be `t3://myuid-cluster-mycluster.myns:8001`.
   - If the target resource is on stand-alone WebLogic server `myserver` with domain UID `myuid` in namespace `myns`, and the server is   listening on port `7001`, then the server service name   that is generated by the operator will be `myuid-myserver` and the full URL will be `t3://myuid-myserver.myns:8001`.
```

rosemarymarano (approved):
I made the edits I thought necessary. Anything you want changed, just let me know.

tbarnes-us (approved):

documentation/staging/content/faq/external-clients.md
```diff
@@ -112,23 +141,50 @@ is hosted outside of the Kubernetes cluster, then:
...
 - You may need to [enable unknown host access](#enabling-unknown-host-access)
-  on the WebLogic Servers that host the EJB or JMS resources.
+  on the WebLogic Server instances that host the EJB or JMS resources.
+
+- If tunneling, then you may need to configure clusters that host EJB or
+  JMS resources with a 'server affinity' default load balancer algorithm.
+  This can significantly speedup connection creation for EJB and JMS clients.
+  See [Configuring WebLogic Server affinity load balancing algorithms](#configuring-weblogic-server-affinity-load-balancing-algorithms).
+
+- __JTA NOTE__: The operator does not currently support external WebLogic JTA access
```
mriccell, line 151: I do not understand this note.  Once Anthony's change of RJVM Forwarding the fact that every managed server has the same port number should not be a cross domain communication limitation.  Each server has a different service name which makes the CoordinatorURL or SubCoordinatorURL unique even when the ports are the same.    Is this not the case?  Please see Alex's blog post https://blogs.oracle.com/weblogicserver/weblogic-server-jta-in-a-kubernetes-environment

The original limitation of cross domain transaction communication between domains where one is inside of K8s and one is outside of k8s is due to load balancing.  With Anthony's change RJVM Forwarding any 2PC calls will always end up being processed by the same Managed Server.

documentation/staging/content/faq/external-clients.md
```diff
@@ -112,23 +141,50 @@ is hosted outside of the Kubernetes cluster, then:
...
+
+- Plus, if the target server or servers can be accessed only by tunneling through a load balancer using HTTP:
+  - [Set up an HTTP tunneling-enabled custom channel](#adding-a-weblogic-custom-channel) on the external WebLogic Server instances.
+  - Specify URLs on the source server that resolve to the load balancer's address and that start with `http` instead of `t3`.
+  - Ensure the load balancer configures the HTTP flow to be 'sticky'.
+  - You may need to configure clusters that host EJB or
+    JMS resources with a 'server affinity' default load balancer algorithm.
+    This can significantly speedup tunneling connection creation for EJB and JMS clients.
+    See [Configuring WebLogic Server affinity load balancing algorithms](#configuring-weblogic-server-affinity-load-balancing-algorithms).
+  - __JTA NOTE__: WebLogic does not support JTA access to a cluster that is exposed using a single port.
```
mriccell, line 172: Same comment as above, I disagree with this note.

mriccell (changes requested):
Excellent documentation, thanks for writing this!  I have a couple of comments that would like to check with Tom and Alex because once we have documented Anthony's change is documented there no longer should be a limitation for cross domain communication in JTA.

documentation/staging/content/faq/external-clients.md
lennyphan: Removed the JTA Notes as RJVM Forwarding documentation will address these JTA limitations.

mriccell (approved):

PR #2383, opened 2021-05-27
Support clusterService.sessionAffinity

rjeberhard:
A customer asked for us to support the `sessionAffinity` field for the per-cluster Service. While we've recommended setting session affinity at the ingress controller or load balancer, there is nothing inappropriate about a customer wanting to expose the per-cluster service and doing session affinity at this level. 

While we've had the `cluster.clusterService` field for a long time, this is the first time that it has fields that aren't present under the `serverService` field, so the implementation is introducing a ClusterService class for the first time.


jshum2479 (approved):
LGTM

lennyphan (approved):
LGTM

PR #2386, opened 2021-05-28
Consistently use ISO-8601 date time format

rjeberhard:


tbarnes-us:
The shell script logging 'timestamp' utils function is designed to match the operator logs timestamp format and timezone. So:

How does the change in this pull differ from the current format?

Is the time still reported as UTC time?

Is the utils function 'close enough' or should it be changed to match?

rjeberhard:
@tbarnes-us, I've updated the scripts and WLST to as closely as possible match the timestamps that the operator generates. @ankedia, I know you are away today, but next week can you please take a look at this? Tom noted that you may have timing scripts that depend on the format of timestamps.

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFormatter.java
```diff
@@ -134,14 +132,13 @@ public String format(LogRecord record) {
...
       String tmp =
-          "{\"@timestamp\":%1$s,\"level\":%2$s, \"class\":%3$s, \"method\":\"format\", \"timeInMillis\":%4$d, "
+          "{\"@timestamp\":%1$s,\"level\":%2$s, \"class\":%3$s, \"method\":\"format\", "
               + "\"@message\":\"Exception while preparing json object\",\"exception\":%5$s}\n";
       return String.format(
           tmp,
           dateString,
           level,
           LoggingFormatter.class.getName(),
-          rawTime,
```
tbarnes-us, line 144: why remove 'rawTime'? (it doesn't seem to hurt to include it, and it's sometimes useful)

operator/src/main/resources/scripts/utils.py
```diff
@@ -52,8 +52,8 @@ def traceInner(logLevel,object):
   }
   # use FINE as logLevel if logLevel is not a known type
   logLevel=switcher.get(logLevel.upper(),'FINE')
-  print("@[%d-%.2d-%.2dT%.2d:%.2d:%.2d.%.3d UTC][%s:%s][%s] %s"
-        % (dt.year,dt.month,dt.day,dt.hour,dt.minute,dt.second,dt.microsecond/1000,
+  print("@[%d-%.2d-%.2dT%.2d:%.2d:%.2d.%dZ][%s:%s][%s] %s"
```
tbarnes-us, line 55: Should '%dZ' be '%.6dZ'  (zero fill the micro-seconds).

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -33,17 +33,17 @@
 #
 #   examples:
 #     trace "Situation normal."
-#     @[2018-09-28T18:10:52.417 UTC][myscript.sh:91][FINE] Situation normal.
+#     @[2018-09-28T18:10:52.417Z][myscript.sh:91][FINE] Situation normal.
```
tbarnes-us, line 36: Should 417 --> 417000 ? (6 digit microseconds?)

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -167,21 +167,15 @@ function traceDirs() {
...
-#   example:  20181001T14:00:00.001 UTC
+#   purpose:  echo timestamp in the form yyyy-mm-ddThh:mm:ss.nnnnnnZ
+#   example:  2018-10-01T14:00:00.001Z
 function timestamp() {
-  local timestamp="`date --utc '+%Y-%m-%dT%H:%M:%S %N %s %Z' 2>&1`"
+  local timestamp="`date --utc '+%Y-%m-%dT%H:%M:%S.%NZ' 2>&1`"
   if [ ! "${timestamp/illegal/xyz}" = "${timestamp}" ]; then
     # old shell versions don't support %N or --utc
-    timestamp="`date -u '+%Y-%m-%dT%H:%M:%S 000000 %s %Z' 2>&1`"
+    timestamp="`date -u '+%Y-%m-%dT%H:%M:%SZ' 2>&1`"
```
tbarnes-us, line 176: '%SZ' --> '%S.000000Z' ? (more closely match format of new shell versions)

operator/src/main/resources/scripts/utils.py
```diff
@@ -8,9 +8,9 @@
 #    'ERROR' is converted to 'SEVERE' 
 #    Unknown logLevels are converted to 'FINE'.
 #
-# This matches format of bash utils.sh trace, and rougly matches the operator's log format.
+# This matches format of bash utils.sh trace and the operator's log format.
 #
-# Sample output:   @[2018-09-28T17:23:55.335 UTC][introspectDomain.py:614][FINE] Domain introspection complete.
+# Sample output:   @[2018-09-28T17:23:55.335Z][introspectDomain.py:614][FINE] Domain introspection complete.
```
tbarnes-us, line 13: 335 --> 335000

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFormatter.java
rjeberhard: This isn't for ordinary log messages, but is the catch block for what I suspect is the rare occurrence where we have a JSON exception during the generation of the normal messages. Is it helpful in that use case? Unless I missed it, this raw time doesn't seem to be part of the normal log message.

operator/src/main/resources/scripts/utils.py
rjeberhard: Yup, you're right.

operator/src/main/java/oracle/kubernetes/operator/logging/LoggingFormatter.java
tbarnes-us: `    map.put(TIME_IN_MILLIS, rawTime);` is removed in this pull, which presumably has a side effect on the code in the following try
```
     ObjectMapper mapper = new ObjectMapper();
      json = mapper.writeValueAsString(map);
...
    return json + "\n";

```

tbarnes-us: In other words, it looks to me like it shows up in the 'normal' path too when the `map.put` is in place...

rjeberhard: Okay, I'll re-add it... I apparently need more ☕ for my reading comprehension these days.

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -167,21 +167,15 @@ function traceDirs() {
 }
 
 # timestamp
-#   purpose:  echo timestamp in the form yyyymmddThh:mm:ss.mmm ZZZ
-#   example:  20181001T14:00:00.001 UTC
+#   purpose:  echo timestamp in the form yyyy-mm-ddThh:mm:ss.nnnnnnZ
+#   example:  2018-10-01T14:00:00.001Z
```
tbarnes-us, line 171: 001 --> 000001

ankedia (approved):
LGTM. As I mentioned, the timing scripts are not actively being used and I can tweak them to match the new format if needed.

tbarnes-us (approved):

PR #2388, opened 2021-05-29
OWLS-89535 - scalingAction.sh script fails if multiple domains are running in the same namespace

alai8:
It was caused by a typo in the variable name for domain uid in the Kubernetes API call to retrieve the domain info from Kubernetes. With the typo, info for all domains in the same namespace is being retrieved instead of that of a specified domain. This PR fixes the typo in the API call, and the parsing of the retrieved json.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5130/


lennyphan (approved):
LGTM

rjeberhard (approved):

PR #2390, opened 2021-06-02
Monitoring exporter improvements, including port configuration, Istio support, and validation

rjeberhard:
Several improvements to the monitoring exporter integration:

1. Support Istio by having the container port and service port names match and conform to the Istio naming requirements.
2. Support configuring the exporter's port and clarifying the defaulting rules
3. Removed the debug endpoint
4. When the customer configures an explicit port for the exporter, then validate that this port is not used by WebLogic channels or NAP's

### DIFF
--- a/documentation/domains/Domain.json
+++ b/documentation/domains/Domain.json
@@ -618,6 +618,10 @@
           "x-kubernetes-preserve-unknown-fields": "true",
           "description": "The configuration for the WebLogic Monitoring Exporter. If WebLogic Server instances are already running and have the monitoring exporter sidecar container, then changes to this field will be propagated to the exporter without requiring the restart of the WebLogic Server instances.",
           "$ref": "#/definitions/Map"
+        },
+        "port": {
+          "description": "The port exposed by the WebLogic Monitoring Exporter running in the sidecar container. Defaults to 8080 unless that port value is in use by a channel or network access point (NAP) of the WebLogic Server instance, in which case the next higher available port is selected. If a port value is specified then it must not conflict with a port used by the WebLogic Server instance.",
+          "type": "number"
         }
       }
     },
--- a/documentation/domains/Domain.json
+++ b/documentation/domains/Domain.json
@@ -620,7 +620,7 @@
           "$ref": "#/definitions/Map"
         },
         "port": {
-          "description": "The port exposed by the WebLogic Monitoring Exporter running in the sidecar container. Defaults to 8080 unless that port value is in use by a channel or network access point (NAP) of the WebLogic Server instance, in which case the next higher available port is selected. If a port value is specified then it must not conflict with a port used by the WebLogic Server instance.",
+          "description": "The port exposed by the WebLogic Monitoring Exporter running in the sidecar container. Defaults to 8080. The port value must not conflict with a port used by any WebLogic Server instance, including the ports of built-in channels or network access points (NAPs).",
           "type": "number"
         }
       }
--- a/operator/src/main/resources/Operator.properties
+++ b/operator/src/main/resources/Operator.properties
@@ -177,8 +177,8 @@ WLSDO-0024=More than one item under ''spec.commonMountVolumes'' in the domain re
 WLSDO-0025=More than one item under ''spec.commonMountVolumes'' in the domain resource has same DNS-1123 name ''{0}''.
 WLSDO-0026=An item under ''spec.commonMountVolumes'' in the domain resource does not have a name. A name is required for \
   the common mount volumes defined under ''spec.commonMountVolumes''.
-WLSDO-0027=Monitoring Exporter port ''{0}'' conflicts with WebLogic Server instance ''{1}'' port ''{2}''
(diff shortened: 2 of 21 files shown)


tbarnes-us:
> When the customer configures an explicit port for the exporter, then validate that this port is not used by WebLogic channels or NAP's

Why? A NAP can support admin QoS.

documentation/domains/Domain.json
```diff
@@ -618,6 +618,10 @@
           "x-kubernetes-preserve-unknown-fields": "true",
           "description": "The configuration for the WebLogic Monitoring Exporter. If WebLogic Server instances are already running and have the monitoring exporter sidecar container, then changes to this field will be propagated to the exporter without requiring the restart of the WebLogic Server instances.",
           "$ref": "#/definitions/Map"
+        },
+        "port": {
+          "description": "The port exposed by the WebLogic Monitoring Exporter running in the sidecar container. Defaults to 8080 unless that port value is in use by a channel or network access point (NAP) of the WebLogic Server instance, in which case the next higher available port is selected. If a port value is specified then it must not conflict with a port used by the WebLogic Server instance.",
```
tbarnes-us, line 623: >  in which case the next higher available port is selected

Is generating a port number heuristically helpful?  Instead perhaps it'd be better just to generate an error message. If I understand correctly, tooling that imports from the exporter must specify this port - so such configuration would need to 'guess' the expected value...

operator/src/main/resources/Operator.properties
```diff
@@ -177,6 +177,8 @@ WLSDO-0024=More than one item under ''spec.commonMountVolumes'' in the domain re
 WLSDO-0025=More than one item under ''spec.commonMountVolumes'' in the domain resource has same DNS-1123 name ''{0}''.
 WLSDO-0026=An item under ''spec.commonMountVolumes'' in the domain resource does not have a name. A name is required for \
   the common mount volumes defined under ''spec.commonMountVolumes''.
+WLSDO-0027=Monitoring Exporter port ''{0}'' conflicts with WebLogic Server instance ''{1}'' port ''{2}''
+WLSDO-0028=Monitoring Exporter port ''{0}'' conflicts with Cluster ''{1}'' dynamic server template port ''{2}''
```
tbarnes-us, line 181: IMO, it'd be helpful if these two messages mentioned the corresponding domain resource setting.

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainValidationStepTest.java
```diff
@@ -361,11 +383,52 @@ public void whenServerDoesNotExistInDomain_logWarning() {
...
+  public void whenMonitoringExporterPortConflictsWithClusterServerTemplatePort_logWarning() {
+    configureDomain(domain).withMonitoringExporterConfiguration("queries:\n").withMonitoringExporterPort(9001);
+    testSupport.addToPacket(DOMAIN_TOPOLOGY, domainConfig);
+
+    testSupport.runSteps(topologyValidationStep);
+
+    assertThat(logRecords, containsWarning(MessageKeys.MONITORING_EXPORTER_CONFLICT_DYNAMIC_CLUSTER));
+    assertThat(info.getValidationWarningsAsString(),
+        stringContainsInOrder(Integer.toString(9001), DYNAMIC_CLUSTER_NAME,
+            Integer.toString(SERVER_TEMPLATE_PORT_NUM)));
```
tbarnes-us, line 424: Do the namespace events and domain resource status also contain validation warnings (in addition to the operator log)?
If not, then they should - perhaps in a separate pull.
If yes, then it'd be a nice-to-have if the unit test checked for them.

documentation/domains/Domain.json
rjeberhard: I felt compelled to keep it because we've already released with this behavior :(. It's not correct that tooling has to guess the value since the port value is published as one of the Prometheus annotations.

operator/src/main/resources/Operator.properties
rjeberhard: Sure, I can do that.

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainValidationStepTest.java
rjeberhard: @doxiao, do these kind of domain validation errors that will appear in the domain status also show up in the event stream?

doxiao: I recall that Anthony added DomainValidationError event. @alai8 Can you confirm?

alai8: The domain validation errors  should also show up in as events.

rjeberhard: @tbarnes-us, I found that some of the other unit-tests have checks that the event is generated. I'll do the same for these new tests.

ddsharpe (approved):

tbarnes-us (approved):

PR #2393, opened 2021-06-02
OWLS-76974: Add more debugging/help when node manager fails to start

lennyphan:
If the NM fails to start after 60 seconds then provide more help  in the log:

1.  process-id of the NM
2.  thread dump of NM using 'jcmd'.  'jstack' not available in $JAVA_HOME/bin directory of the WLS image.
3.  text output suggeting how to fix the problem if entropy is the root cause

### DIFF
--- a/operator/src/main/resources/scripts/startNodeManager.sh
+++ b/operator/src/main/resources/scripts/startNodeManager.sh
@@ -337,13 +337,24 @@ while [ 1 -eq 1 ]; do
     break
   fi
   if [ $((SECONDS - $start_secs)) -ge $max_wait_secs ]; then
-    trace INFO "Trying to put a node manager thread dump in '$nodemgr_out_file'."
-    kill -3 `jps -l | grep weblogic.NodeManager | awk '{ print $1 }'`
+    pid=$(jps | grep NodeManager | awk '{ print $1 }')
+    if [ -z $pid ]; then
+      trace INFO "Node manager process id not found. Cannot create thread dump."
+    else
+      trace INFO "Node manager process id is '$pid'."
+      if [ -x "$(command -v $JAVA_HOME/bin/jcmd)" ]; then
+        trace INFO "Node manager thread dump:"
+        $JAVA_HOME/bin/jcmd $pid Thread.print
+      fi
+    fi
+    trace INFO "Entropy: "
+    cat /proc/sys/kernel/random/entropy_avail
     trace INFO "Contents of node manager log '$nodemgr_log_file':"
     cat ${nodemgr_log_file}
     trace INFO "Contents of node manager out '$nodemgr_out_file':"
-    cat ${nodemgr_out_file}
-    trace SEVERE "Node manager failed to start within $max_wait_secs seconds."
+    cat ${NODEMGR_OUT_FILE}
+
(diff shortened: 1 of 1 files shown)


alai8 (approved):

operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -337,13 +337,24 @@ while [ 1 -eq 1 ]; do
...
-    trace INFO "Trying to put a node manager thread dump in '$nodemgr_out_file'."
-    kill -3 `jps -l | grep weblogic.NodeManager | awk '{ print $1 }'`
+    pid=$(jps | grep NodeManager | awk '{ print $1 }')
+    if [ -z $pid ]; then
+      trace INFO "Node manager process id not found. Cannot create thread dump."
+    else
+      trace INFO "Node manager process id is '$pid'."
+      if [ -x "$(command -v $JAVA_HOME/bin/jcmd)" ]; then
+        trace INFO "Node manager thread dump:"
+        $JAVA_HOME/bin/jcmd $pid Thread.print
```
tbarnes-us, line 347: Minor:  The JIRA was old - probably written before the 'kill -3' was added.  How about we try generate a thread dump two ways in case either the new `jcmd` approach or the old `kill -3` method works better? My guess is the answer varies based on the nature of the NM failure.  If you agree, then please add `trace INFO "Trying to put a node manager thread dump in '$nodemgr_out_file'."` followed by `kill -3 $pid` just after the `trace INFO "Node manager process id is '$pid'."` and just before the `if`.

operator/src/main/resources/scripts/startNodeManager.sh
```diff
@@ -337,13 +337,24 @@ while [ 1 -eq 1 ]; do
...
     cat ${nodemgr_log_file}
     trace INFO "Contents of node manager out '$nodemgr_out_file':"
-    cat ${nodemgr_out_file}
-    trace SEVERE "Node manager failed to start within $max_wait_secs seconds."
+    cat ${NODEMGR_OUT_FILE}
+
+    trace SEVERE "The node manager failed to start within $max_wait_secs seconds. To increase this timeout, define the NODE_MANAGER_MAX_WAIT environment variable in your domain resource, and set it higher than $max_wait_secs. To diagnose the problem, see the above INFO messages for node manager log contents, stdout contents, pid, thread dump, and entropy. If the log and stdout contents are sparse and reveal no errors, then the node manager may be stalled while generating entropy -- especially if entropy is below 500. If entropy is the problem, then for testing purposes you can temporarily work around this problem by specifying '-Djava.security.egd=file:/dev/./urandom' in a USER_MEM_ARGS environment variable defined via your domain resource, but for production purpose the problem should be solved by following the guidance in 'https://docs.oracle.com/middleware/12213/wls/NODEM/starting_nodemgr.htm' (search for keyword 'rngd'). "
```
tbarnes-us, line 357: Can we try make this more readable by shortening the max line length?

I don't know the best approach for multi-line strings in bash, but something like the 
the following will probably work:

```
trace SEVERE   $(cat << EOF
The node manager failed to start within $max_wait_secs seconds.
To increase this timeout, define the NODE_MANAGER_MAX_WAIT
environment variable in your domain resource, and set it higher
than $max_wait_secs. To diagnose the problem, see the above INFO
messages for node manager log contents, stdout contents, pid,
thread dump, and entropy. If the log and stdout contents are
sparse and reveal no errors, then the node manager may be stalled
while generating entropy -- especially if entropy is below 500.
If entropy is the problem, then for testing purposes you can
temporarily work around this problem by specifying
'-Djava.security.egd=file:/dev/./urandom' in a USER_MEM_ARGS
environment variable defined via your domain resource, but
for production purposes the problem should be solved by following
the guidance in
'https://docs.oracle.com/middleware/12213/wls/NODEM/starting_nodemgr.htm'
(search for keyword 'rngd').
EOF
)
```

operator/src/main/resources/scripts/startNodeManager.sh
lennyphan: I've added the additional dump to the output file using 'kill -3' as I agree that getting thread dumps could also be based on the nature of the NM failure.  It doesn't hurt to have multiple methods for getting thread dumps.

operator/src/main/resources/scripts/startNodeManager.sh
lennyphan: I updated with your suggestion above without seeing your suggestion here about putting it in FAQ.  If we go this route then lets wait till 'RJVM Forwarding' information is committed so that we don't conflict.

tbarnes-us: Sure

tbarnes-us (approved):

PR #2398, opened 2021-06-08
OWLS-89106 - Potential fix for pod startup issue in GBU CNE environment after node drain/repave operation

ankedia:
OWLS-89106 - In GBU CNE environment with a large k8s cluster, some pods are not started after a node drain operation. The event notifications are not generated for pod delete/start and fiber gets stuck in a suspended state due to this. This change will interrupt the fiber in the `WaitForPodReady` step after a timeout interval (2 min) and execute a make-right domain operation. 

The `WaitForJobReady` step logic is not changed because the introspector job for the JRF domain is not fully idempotent. We also don't store the state of job pods in DomainPresenceInfo. Hence it doesn't need to be refreshed with an explicit relisting in case of a missed watch event. 

This PR also contains a change to ItParameterizedDomain.testMultiClustersRollingRestart as that test performs extra validation using domain events. With this change, the `DomainProcessingCompleted` event may not always be generated in time during the test validation. We cover domain events testing in `ItKubernetesEvents` and this extra validation is not necessary.

The integration test run is clean with current set of changes - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5310/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
@@ -349,13 +349,13 @@ public class PodWatcher extends Watcher<V1Pod> implements WatchListener<V1Pod>, 
             if (onReadNotFoundForCachedResource(getServerPod(info, serverName), isNotFoundOnRead(callResponse))) {
               LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, serverName, callback.getRecheckCount());
               removeCallback(resource, callback);
-              return doNext(NEXT_STEP_FACTORY.createMakeDomainRightStep(callback, info, resource, getNext()), packet);
+              return doNext(NEXT_STEP_FACTORY.createMakeDomainRightStep(callback, info, getNext()), packet);
             }
           }
 
           if (isReady(callResponse.getResult()) || callback.didResumeFiber()) {
             callback.proceedFromWait(callResponse.getResult());
-            return doNext(packet);
+            return null;
           }
 
           if (shouldWait()) {
@@ -369,7 +369,7 @@ public class PodWatcher extends Watcher<V1Pod> implements WatchListener<V1Pod>, 
             LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, serverName, callback.getRecheckCount());
             removeCallback(resource, callback);
             // Watch backstop recheck count is more than configured recheck count, proceed to make-right step.
-            return doNext(NEXT_STEP_FACTORY.createMakeDomainRightStep(callback, info, resource, getNext()), packet);
+            return doNext(NEXT_STEP_FACTORY.createMakeDomainRightStep(callback, info, getNext()), packet);
           }
         }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
@@ -47,6 +47,7 @@ import oracle.kubernetes.operator.work.Step;
 
 import static oracle.kubernetes.operator.ProcessingConstants.SERVER_NAME;
 import static oracle.kubernetes.operator.logging.MessageKeys.EXECUTE_MAKE_RIGHT_DOMAIN;
+import static oracle.kubernetes.operator.logging.MessageKeys.LOG_WAITING_COUNT;
 
 /**
  * Watches for changes to pods.
@@ -314,6 +315,8 @@ public class PodWatcher extends Watcher<V1Pod> implements WatchListener<V1Pod>, 
 
   private abstract static class WaitForPodStatusStep extends WaitForReadyStep<V1Pod> {
 
+    public static final int RECHECK_DEBUG_COUNT = 10;
+
     private WaitForPodStatusStep(V1Pod pod, Step next) {
       super(pod, next);
     }
@@ -338,12 +341,12 @@ public class PodWatcher extends Watcher<V1Pod> implements WatchListener<V1Pod>, 
         public NextAction onSuccess(Packet packet, CallResponse<V1Pod> callResponse) {
 
           DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
+          String serverName = (String)packet.get(SERVER_NAME);
           if ((info != null) && (callResponse != null)) {
-            String serverName = (String)packet.get(SERVER_NAME);
             Optional.ofNullable(callResponse.getResult()).ifPresent(result ->
                     info.setServerPodFromEvent(getPodLabel(result), result));
             if (onReadNotFoundForCachedResource(getServerPod(info, serverName), isNotFoundOnRead(callResponse))) {
-              LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, callback.getRecheckCount());
+              LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, serverName, callback.getRecheckCount());
               return doNext(new CallBuilder().readDomainAsync(info.getDomainUid(),
                       info.getNamespace(), new MakeRightDomainStep(callback,null)), packet);
             }
@@ -355,11 +358,14 @@ public class PodWatcher extends Watcher<V1Pod> implements WatchListener<V1Pod>, 
           }
 
           if (shouldWait()) {
+            if ((callback.getRecheckCount() % RECHECK_DEBUG_COUNT) == 0) {
+              LOGGER.fine(LOG_WAITING_COUNT,  serverName, callback.getRecheckCount());
+            }
             // Watch backstop recheck count is less than or equal to the configured recheck count, delay.
             return doDelay(createReadAndIfReadyCheckStep(callback), packet,
                     getWatchBackstopRecheckDelaySeconds(), TimeUnit.SECONDS);
           } else {
-            LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, callback.getRecheckCount());
+            LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, serverName, callback.getRecheckCount());
             // Watch backstop recheck count is more than configured recheck count, proceed to make-right step.
             return doNext(new CallBuilder().readDomainAsync(info.getDomainUid(),
                     info.getNamespace(), new MakeRightDomainStep(callback, null)), packet);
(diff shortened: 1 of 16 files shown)


operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
```diff
@@ -383,4 +442,20 @@ protected void removeCallback(String podName, Consumer<V1Pod> callback) {
...
   }
+
+  private static class MakeRightDomainStep extends DefaultResponseStep {
+    MakeRightDomainStep(Step next) {
+      super(next);
+    }
+
+    @Override
+    public NextAction onSuccess(Packet packet, CallResponse callResponse) {
+      MakeRightDomainOperation makeRightDomainOperation =
```
doxiao, line 453: I am not sure if it is always correct for all cases to use the existing MakeRightDomainOperation object. For example, if the original object has an EventData, we may not know if the event has been generated already or not in the previous attempt.

ankedia: Thanks, Dongbo. That's a good point. In the latest integration test run, I'm seeing some new failures in ItKubernetesEvent and ItPodsRestart related to event generation. It looks like these failures are intermittent and timing-dependent. I'm not sure if I can create a new MakeRightDomainOperation object in this step. I'll continue looking into this tomorrow.

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
```diff
@@ -322,6 +331,55 @@ V1ObjectMeta getMetadata(V1Pod pod) {
...
+            resetWatchBackstopRecheckCount(info);
+            return proceedFromWait(packet, callResponse);
+          }
+          if (shouldWait(info)) {
+            // Watch backstop recheck count is less than or equal to the configured recheck count, delay.
+            return doDelay(createReadAndIfReadyCheckStep(callback), packet,
+                    getWatchBackstopRecheckDelaySeconds(), TimeUnit.SECONDS);
+          } else {
+            // Watch backstop recheck count is more than configured recheck count, proceed to make-right step.
+            return doNext(new CallBuilder().readDomainAsync(info.getDomainUid(),
```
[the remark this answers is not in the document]
ankedia: I was following the approach in `handleModifiedDomain` where we call the make-right domain with just the domain object received in the watch event.

```
  private void handleModifiedDomain(Domain domain) {
    LOGGER.fine(MessageKeys.WATCH_DOMAIN, domain.getDomainUid());
    createMakeRightOperation(new DomainPresenceInfo(domain))
        .interrupt()
        .withEventData(EventItem.DOMAIN_CHANGED, null)
        .execute();
  }
```

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
doxiao: Ryan, I agree with you that the operator's processing flow should not depend on in-memory states.  We could review the flow to make sure that is the case. But the issue here is a little different, I think - we are reusing a DomainMakeRightOperation object to do a new execution. I don't think the operator does this prior to this PR. EventData is just one piece of the data in the MakeRightDomainOperation object. We are probably better off creating a new MakeRightDomainOperation; the new code here already overrides some of the data in the existing object prior to calling execute().

doxiao: My suggestion is to create a new MakeRightDomainOperation passing in the new DomainPresenceInfo, instead of reusing the current make right domain operation. This suggestion does not change the logic/spirit of the current changes. I discussed with Anil offline, and found out that apparently he had problem in creating a new make right domain operation instance within PodWatcher's static context. We could refactor the Main class a little and add a static factory method there for PodWatcher to use to create a new make right domain operation instance. What I like about using a new instance is that the code here would have full control of what it needs to do. Alternatively, we could have a clear() method on MakeRightDomainOperation to start clean if we know for sure there is no other fiber running using this make right domain operation instance. Then we have to keep the clear() method up-to-date when we add new variables into it in the future. I personally prefers the first option.

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
```diff
@@ -239,6 +266,25 @@ public NextAction apply(Packet packet) {
...
+      super(next);
+      this.callback = callback;
+    }
+
+    @Override
+    public NextAction onSuccess(Packet packet, CallResponse callResponse) {
+      MakeRightDomainOperation makeRightDomainOperation =
+              (MakeRightDomainOperation)packet.get(MAKE_RIGHT_DOMAIN_OPERATION);
+      makeRightDomainOperation.setLiveInfo(new DomainPresenceInfo((Domain)callResponse.getResult()));
+      callback.fiber.terminate(null, packet);
```
doxiao, line 282: we probably need to call removeCallback here.

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
```diff
@@ -197,17 +209,32 @@ public String getName() {
...
       public NextAction onSuccess(Packet packet, CallResponse<T> callResponse) {
+        DomainPresenceInfo info = packet.getSpi(DomainPresenceInfo.class);
         if ((callResponse != null) && (callResponse.getResult() instanceof V1Pod)) {
           V1Pod pod = (V1Pod) callResponse.getResult();
-          Optional.ofNullable(packet.getSpi(DomainPresenceInfo.class))
+          Optional.ofNullable(info)
                   .ifPresent(i -> i.setServerPodFromEvent(getPodLabel(pod, LabelConstants.SERVERNAME_LABEL), pod));
         }
-        if (isReady(callResponse.getResult())) {
+        if (isReady(callResponse.getResult()) || callback.didResume.get()) {
```
doxiao, line 218: I don't understand why we need to check callback.didResume here. How would the operator get here again if the callback has been resumed already?

ankedia: The check callback.didResume is to avoid the race condition. We do the periodic listing of the introspector job to check if it's ready and also dispatch the callback when the watch event is received. In this scenario, the watch event notifications are flowing. After the intro job completed watch notification is received, the callback is removed and the fiber is resumed. After this, the intro job gets deleted but the child fiber that's listing the job periodically to check for readiness doesn't see the job as ready since it got deleted. Hence the child fiber never finishes and times out. The above check will terminate the child fiber after intro job has been deleted.

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
ankedia: I made the change to remove callback here.

ankedia:
The integration tests are passing with current set of changes - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5310/

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
```diff
@@ -322,6 +334,66 @@ V1ObjectMeta getMetadata(V1Pod pod) {
...
+            Optional.ofNullable(callResponse.getResult()).ifPresent(result ->
+                    info.setServerPodFromEvent(getPodLabel(result), result));
+            if (onReadNotFoundForCachedResource(getServerPod(info, serverName), isNotFoundOnRead(callResponse))) {
+              LOGGER.fine(EXECUTE_MAKE_RIGHT_DOMAIN, serverName, callback.getRecheckCount());
+              return doNext(new CallBuilder().readDomainAsync(info.getDomainUid(),
+                      info.getNamespace(), new MakeRightDomainStep(callback,null)), packet);
+            }
+          }
+
+          if (isReady(callResponse.getResult()) || callback.didResumeFiber()) {
```
doxiao, line 355: Do we need to reset the recheckCount here?

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
doxiao: If I understand it correctly, check for didResume (didResumeFiber now) is only needed for the job case. If so, we need to remove the check in the pod code path,

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
ankedia: I moved the `recheckCount` variable from DomainPresenceInfo to the Callback object since we create a separate callback instance for each resource. The WaitForReady step registers a new Callback instance in the `resumeWhenReady` method and the previous callback will be removed once the fiber is resumed. With this approach, there's no need to reset the recheckCount since previous callback instance will be garbage collected once the resource is ready.

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
ankedia: The `didResumeFiber` will return true only when the fiber is resumed and the resource should be ready at that time. I think there's no harm in keeping the check as that should avoid any potential race condition in pod case as well.

operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
ankedia: I have implemented the clear() method on MakeRightDomainOperation to reset its state before executing the new make-right operation in `MakeRightDomainStep`. Thanks.

doxiao (commented):
Looks good to me. 

We need to add unit test cases to cover the timeout code path.

ankedia:
> Looks good to me.
> 
> We need to add unit test cases to cover the timeout code path.

Added new unit test `whenPodCreatedAndNotReadyAfterTimeout_executeMakeRightDomain` in PodWatcherTest to cover the timeout code path in 1c3b6a0. The integration test run is clean after this change - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5349/

operator/src/main/java/oracle/kubernetes/operator/WaitForReadyStep.java
```diff
@@ -27,7 +31,18 @@
...
 abstract class WaitForReadyStep<T> extends Step {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
   private static final int DEFAULT_RECHECK_SECONDS = 5;
+  private static final int DEFAULT_RECHECK_COUNT = 60;
+
+  static NextStepFactory NEXT_STEP_FACTORY =
+          (callback, info, next) -> createMakeDomainRightStep(callback, info, next);
+
+  protected static Step createMakeDomainRightStep(WaitForReadyStep.Callback callback,
+                                           DomainPresenceInfo info, Step next) {
```
doxiao, line 42: The "next" parameter is not used.

doxiao (approved):
LGTM

rjeberhard (approved):

PR #2401, opened 2021-06-09
updated monexp to 2.0

marinakog:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5242/
Updated to newest wls exporter web app version


anpanigr (commented):
Is there a way we can pull the latest MonitoringExporter version as a default ?

marinakog:
> Is there a way we can pull the latest MonitoringExporter version as a default ?

there is variable for jenkin (MONITORING_EXPORTER_VERSION) to specify version you want to run, by default right now it is 2.0 with this change.

bhavaniravichandran (approved):
LGTM

rjeberhard:
> there is variable for jenkin (MONITORING_EXPORTER_VERSION) to specify version you want to run, by default right now it is 2.0 with this change. The only reason I added extra coding logic because it used to be 3 digits version, but for 2.0 Russel/Ryan published files differently ( used both 2.0.0 and 2.0 for dirs and file names) so just updating that var value did not work

I don't quite understand this answer. The latest monitoring exporter is version 2.0.2, which is what I would want to test for the sidecar integration. There were no changes to the war file in either 2.0.2 or 2.0.2, so the 2.0.0 version is the latest war. Is this the confusion? I've had someone else ask if we could include the war in the released files with those bug fix versions. If I did that, could you use 2.0.2?

marinakog:
added more comments and fixed some var names @rjeberhard can we merge? : latest result:https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5304

PR #2405, opened 2021-06-10
Fix for port in use intermittent nightly failures.

sankarpn:
Many domain in pv tests uses available free port for admin node port at the time of domain creation. But when they are actually used the free ports were occupied causing port in use errors.

This PR allocates a free port for tests from single synchronized static method and once allocated that port is never reused for the entire test run. 

Parallel runs
--------------
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5258/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5257
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5256

Sequential runs
----------------
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5294/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5293/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5271/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5264/


anpanigr (commented):
Can you start  a sequential run, since all the classes are not covered in parallel run.

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/TestUtils.java
```diff
@@ -178,6 +178,28 @@ public static synchronized int getNextFreePort(int from, int to) {
...
+   * Get the next free port between port and END_PORT.
+   *
+   * @return the next free port number, if there is no free port below END_PORT return -1.
+   */
+  public static synchronized int getNextFreePort() {
+    LoggingFacade logger = getLogger();
+    int freePort = 0;
+    while (port <= END_PORT) {
+      freePort = port++;
+      if (isLocalPortFree(freePort)) {
```
rjeberhard, line 194: Even though this method is synchronized, I'm concerned that you will still have timing bugs. I could see this method returning a given port value, but there being a delay before the test uses that port and it no longer being free. That would allow the method to return the same port to multiple callers. What does the isLocalPortFree method do?

rjeberhard: I checked... It's what I thought. I believe that this is still not safe.

sankarpn: Isn't the port value increased every time the method is called? So once allocated that port is never reused in the life of the test run.

sankarpn: If the port was used by some other process before used by the tests yes it won't work.

rjeberhard: It doesn't seem to. The logic always starts at 0 and looks for the next port not currently in use.  Once it returns a certain port number there will be some delay before that port is in use since it's not recorded here. Even though you've had some successful test runs, this code isn't safe. There only seem to be about 50 ports. Why not just assign port values to each test?

rjeberhard (approved):

sankarpn:
> 
> 
> Can you start a sequential run, since all the classes are not covered in parallel run.

Sequential runs results are updated.

anpanigr (approved):
LGTM

PR #2406, opened 2021-06-11
Ensure that API failures are logged when the retry strategy elects to not retry

rjeberhard:
I found this while working with the V8o team. Previously, the ResponseStep was not logging the async failure if there was a retry strategy, but it elected to not retry (e.g. any kind of permanent failure).

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
@@ -135,16 +135,13 @@ public abstract class ResponseStep<T> extends Step {
    * @return Next action for retry or null, if no retry is warranted
    */
   private NextAction doPotentialRetry(Step conflictStep, Packet packet, CallResponse<T> callResponse) {
-    RetryStrategy retryStrategy = packet.getSpi(RetryStrategy.class);
-    if (retryStrategy != null) {
-      return retryStrategy.doPotentialRetry(conflictStep, packet, callResponse.getStatusCode());
-    }
-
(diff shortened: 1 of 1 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
```diff
@@ -135,16 +135,13 @@ protected final NextAction doContinueListOrNext(CallResponse<T> callResponse, Pa
...
-    }
-
-    LOGGER.fine(MessageKeys.ASYNC_NO_RETRY,
-        callResponse.getExceptionString(),
-        callResponse.getStatusCode(),
-        callResponse.getHeadersString());
-    return null;
+    return Optional.ofNullable(packet.getSpi(RetryStrategy.class))
+        .map(rs -> rs.doPotentialRetry(conflictStep, packet, callResponse.getStatusCode()))
+        .orElseGet(() -> {
```
ankedia, line 140: Probably I'm missing something but I'm not able to see how this is different from previous the code i.e. how it logs the failures when the retry strategy is not null?

doxiao (approved):
LGTM.

operator/src/main/java/oracle/kubernetes/operator/helpers/ResponseStep.java
doxiao: If I understand this correctly, the only difference is that when doPotentialRetry returns null (which means no need to retry), the old code does not log and the new code does.

ankedia: ok, got it now. Thanks Dongbo.

ankedia (approved):

PR #2408, opened 2021-06-11
Use tcp prefix for metrics exporter

rjeberhard:
After diagnosing with the V8o team, we've discovered that using the "tcp-" prefix works better because this works when the metrics exporter is addressed with either the DNS address or using the pod's ClusterIP value.


lennyphan (approved):
LGTM

ankedia (approved):

PR #2422, opened 2021-06-22
Add tests to verify MII sample using common mount in WLS and FMW domains

vanajamukkara:
Jenkins Results - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5404/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5403/ (full suite, still running)

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiSample.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiSample.java
@@ -376,7 +376,7 @@ public class ItMiiSample {
   }
 
   /**
-   * Test to verify WLS update1 use case.
+   * Test to verify WLS update1 use case using common mount.
    * Adds a data source to initial domain via a configmap and updates the
    * domain resource restartVersion.
    * Verifies all WebLogic Server pods roll to ready, roll to the expected
@@ -388,14 +388,14 @@ public class ItMiiSample {
   @Test
   @Order(13)
   @DisabledIfEnvironmentVariable(named = "SKIP_WLS_SAMPLES", matches = "true")
-  @DisplayName("Test to verify MII sample WLS update1 use case")
+  @DisplayName("Test to verify MII sample WLS update1 use case using common mount")
   public void testCMWlsUpdate1UseCase() {
     envMap.put("DO_CM", "true");
     execTestScriptAndAssertSuccess("-update1", "Update1 use case failed");
   }
 
   /**
-   * Test to verify WLS update2 use case.
+   * Test to verify WLS update2 use case using common mount.
    * Deploys a second domain 'domain2' with a different domain UID,
    * different secrets, and different datasource config map,
    * but that is otherwise the same as the update1 domain.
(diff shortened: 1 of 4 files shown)


anpanigr (approved):
LGTM

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiSample.java
```diff
@@ -355,6 +355,201 @@ public void testFmwUpdate4UseCase() {
...
+  /**
+   * Test to verify MII sample WLS initial use case using common mount.
+   * Builds image required for the initial use case, creates secrets, and
+   * creates domain resource.
+   * Verifies all WebLogic Server pods are ready, are at the expected
+   * restartVersion, and have the expected image.
+   * Verifies the sample application is running
+   * (response includes "Hello World!").
+   */
+  @Test
```
tbarnes-us, line 367: IMO, this seems like quite a bit of cut-and-paste code - which makes a little error prone and harder to maintain/understand.

Could we instead design the ItMiiSample java test to be modal and then run the same test twice, once the 'old way' and once in 'cm mode'?  For example, would it be possible to have a ItMIISampleCM.java that inherits from ItMIISample.java - where the child class sets a boolean to flag to indicate to the parent class that it should add `envMap("DO_CM", "true")`?

tbarnes-us: If that's not possible, then can the content of tests 2 through 11 be moved to helper functions? If so, then the old tests 2 through 11 and the new CM tests 12 through 21 could then share the same helper functions.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiSample.java
```diff
@@ -381,17 +576,19 @@ public void tearDownAll() {
...
-    if (imageName.equals(MII_SAMPLE_JRF_IMAGE_NAME_V1)) { 
-      imageVer = "JRF-v1"; 
+    if (imageName.equals(MII_SAMPLE_JRF_IMAGE_NAME_V1)) {
+      imageVer = "JRF-" + decoration + "v1";
     }
-    if (imageName.equals(MII_SAMPLE_JRF_IMAGE_NAME_V2)) { 
-      imageVer = "JRF-v2"; 
+    if (imageName.equals(MII_SAMPLE_JRF_IMAGE_NAME_V2)) {
+      imageVer = "JRF-" + decoration + "v2";
     }
```
tbarnes-us, line 592: This code could be mad a little more clear in newer versions of Java which accept strings in case statements - e.g. something like:
```
String decoration = (envMap.get("DO_CM") != null && envMap.get("DO_CM").equalsIgnoreCase("true"))  ? "CM-" : "";  
switch (imageName) {
  case MII_SAMPLE_WLS_IMAGE_NAME_V1 : 
    imageVer = "WLS-" + decoration + "-v1"; 
    break;
  ...
}
```

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiSample.java
vanajamukkara: We have not used inheritance in integration tests so far as both Mark and Ryan suggested not to use in IT's in the past. I can do the helper functions.

tbarnes-us (approved):

PR #2429, opened 2021-06-28
use getNextFreePort() instead of 0 to avoid port in use when running tests in parallel

sankarpn:
Use getNextFreePort() instead of 0 to avoid port in use when running tests in parallel. The getNextFreePort(0 keeps track of the assigned ports and never reuses the same port in the lifetime of the current run. 

This should avoid errors like

message: '0/2 nodes are available: 1 node(s) didn''t have free ports for the requested pod ports, 1 node(s) had taints that the pod didn''t tolerate.'
reason: FailedScheduling


hzhao-github:
LGTM

hzhao-github (approved):

rjeberhard (approved):

PR #2430, opened 2021-06-28
OWLS-90180 Optimize detection of when to run introspection

ankedia:
The changes in this PR are to address the use case where the operator starts a long-running introspector job and then the operator dies and is restarted – the newly restarted operator finds the existing introspector job and waits for those results rather than deleting this job and replacing it. Previously, in this use case, the operator deleted the existing job and replaced it with a new job.

For the case when the operator restarts and finds a failed job with a SEVERE message, it checks if the current time is greater than the sum of job creation time and the `domainPresenceRecheckIntervalSeconds` interval. If so, it deletes the failed job and replaces it with a new job. 
Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5446/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
@@ -6,6 +6,7 @@ package oracle.kubernetes.operator.helpers;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.StringReader;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -49,6 +50,7 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.yaml.snakeyaml.Yaml;
 
 import static java.lang.System.lineSeparator;
+import static java.time.temporal.ChronoUnit.MILLIS;
 import static oracle.kubernetes.operator.DomainStatusUpdater.BAD_TOPOLOGY;
 import static oracle.kubernetes.operator.IntrospectorConfigMapConstants.DOMAINZIP_HASH;
 import static oracle.kubernetes.operator.IntrospectorConfigMapConstants.DOMAIN_INPUTS_HASH;
@@ -508,7 +510,7 @@ public class ConfigMapHelper {
     }
 
     private long timeSinceJobStart(Packet packet) {
-      return System.currentTimeMillis() - ((Long) packet.get(JobHelper.START_TIME));
+      return ((OffsetDateTime)packet.get(JobHelper.START_TIME)).until(OffsetDateTime.now(), MILLIS);
     }
   }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
@@ -504,13 +504,8 @@ public class ConfigMapHelper {
     }
 
     private long timeSinceJobStart(Packet packet) {
-      return System.currentTimeMillis() - getStartTime(packet);
+      return System.currentTimeMillis() - ((Long) packet.get(JobHelper.START_TIME));
     }
-
-    private Long getStartTime(Packet packet) {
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/ConfigMapHelper.java
```diff
@@ -504,7 +504,11 @@ public NextAction apply(Packet packet) {
     }
 
     private long timeSinceJobStart(Packet packet) {
-      return System.currentTimeMillis() - ((Long) packet.get(JobHelper.START_TIME));
+      return System.currentTimeMillis() - getStartTime(packet);
+    }
+
+    private Long getStartTime(Packet packet) {
+      return Optional.ofNullable((Long) packet.get(JobHelper.START_TIME)).orElse(0L);
```
jshum2479, line 511: It probably ok, but if START_TIME is not set, then  the elapsed time will be very long.

ankedia: This check is for the operator restart use case where the introspector job is started before the operator is restarted. The START_TIME will not be set in the packet in this case. It seems the elapsed time is only used to log a FINE level log message and I'm not sure if it's worth storing the START_TIME in the introspector config map if it's only used for logging.

ankedia: I have removed this check by setting the START_TIME to job creation time for operator restart use-case.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -507,17 +542,31 @@ public NextAction onSuccess(Packet packet, CallResponse<String> callResponse) {
...
                 packet);
       }
 
       return doNext(packet);
     }
 
+    private Long getJobLazyDeletionTime(V1Job domainIntrospectorJob) {
+      int retryIntervalSeconds = TuningParameters.getInstance().getMainTuning().domainPresenceRecheckIntervalSeconds;
+      return Optional.ofNullable(domainIntrospectorJob.getMetadata())
+              .map(m -> m.getCreationTimestamp()).map(t -> t.toInstant().toEpochMilli()).orElse(0L)
```
rjeberhard, line 566: Why are you converting these OffsetDateTime instances to epoch millis? Why not return an OffsetDateTime and use the isBefore() or isAfter() comparison methods. Similarly, the class has built-in support for adding seconds or millis, etc.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -392,50 +417,60 @@ public NextAction apply(Packet packet) {
...
+          return doNext(Step.chain(
+                  createProgressingStartedEventStep(info, INSPECTING_DOMAIN_PROGRESS_REASON, true, null),
+                  readDomainIntrospectorPodLogStep(null),
+                  deleteDomainIntrospectorJobStep(null),
+                  ConfigMapHelper.createIntrospectorConfigMapStep(null),
+                  ConfigMapHelper.readExistingIntrospectorConfigMap(namespace, info.getDomainUid()),
+                  new DomainProcessorImpl.IntrospectionRequestStep(info),
+                  createDomainIntrospectorJobStep(getNext())), packet);
+        } else {
+          packet.putIfAbsent(START_TIME, System.currentTimeMillis());
```
rjeberhard, line 468: START_TIME could be an instance of OffsetDateTime.now(). There isn't a good reason to convert to a Long.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
ankedia: Done, made the changes to use OffsetDateTime, isAfter() and plus() methods instead of converting it to epoch millis in fedcece .

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
ankedia: Done in fedcece .

rjeberhard (approved):

jshum2479 (approved):
LGTM

PR #2443, opened 2021-07-06
OWLS-90614 - Handle ApiException with 409 code returned from the replaceEvent API call

ankedia:
Code to handle ApiException with 409 status code in the replaceEvent async API call seen in the GBU large cluster environment.  The code will treat this exception similar to the "404 - Not Found" error so that a new event can be created.


doxiao (approved):
LGTM.

rjeberhard (approved):

PR #2447, opened 2021-07-08
Don't remove other conditions when adding ConfigChangesPendingRestart

russgold:
Adding a condition of type ConfigChangesPendingRestart is currently removing any other statuses. This fixes that and does some code cleanup.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
@@ -13,13 +13,8 @@ public enum DomainConditionType {
   Available,
   ConfigChangesPendingRestart {
     @Override
-    String getStatusMessage(DomainCondition condition) {
-      return condition.getMessage();
-    }
-
-    @Override
-    String getStatusReason(DomainCondition condition) {
-      return condition.getReason();
+    DomainConditionType[] typesToRemove() {
+      return new DomainConditionType[] {};
     }
   },
   Failed {
@@ -34,17 +29,13 @@ public enum DomainConditionType {
     }
 
     @Override
-    DomainConditionType[] typesToRemoveAlways() {
-      return new DomainConditionType[] {Progressing};
+    DomainConditionType[] typesToRemove() {
+      return new DomainConditionType[] {Progressing, Available};
     }
   };
 
   DomainConditionType[] typesToRemove() {
     return new DomainConditionType[] {Progressing, Available, Failed};
-  }
-
-  DomainConditionType[] typesToRemoveAlways() {
-    return new DomainConditionType[] {};
   }
 
   String getStatusMessage(DomainCondition condition) {
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
@@ -5,6 +5,7 @@ package oracle.kubernetes.weblogic.domain.model;
 
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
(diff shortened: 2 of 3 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -116,28 +117,24 @@ public DomainStatus(DomainStatus that) {
    */
   public DomainStatus addCondition(DomainCondition newCondition) {
     if (conditions.contains(newCondition)) {
-      conditions = conditions.stream()
-          .filter(c -> preserve(c, newCondition.getType().typesToRemoveAlways())).collect(Collectors.toList());
```
tbarnes-us, line 120: I am probably misinterpreting, but it looks we can't assume all conditions are preserved in the case where conditions contains the new condition?  The only I reason I say this is that the new 'preserve' function below looks to me like it's careful to remove an old condition from the current set if it is the 'same type' as the new condition and the old condition is not True.

russgold: Yes; most condition types already replace old versions of themselves in any case, but this ensures that those that don't, such as Failed, will still remove old conditions with status False or Unknown.

tbarnes-us:
The change LGTM given Russ' answer to my question. I recommend getting sign-off from Ryan, Dongbo, and Johnny.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -116,28 +117,24 @@ public DomainStatus(DomainStatus that) {
...
   }
 
-  private boolean preserve(DomainCondition condition, DomainConditionType[] types) {
-    for (DomainConditionType type : types) {
-      if (condition.getType() == type) {
-        return false;
-      }
+  // Returns true if adding a condition of the new type should not remove the specified condition.
+  private boolean preserve(DomainCondition condition, DomainConditionType newType) {
+    if (newType == condition.getType() && !"True".equalsIgnoreCase(condition.getStatus())) {
```
doxiao, line 133: this condition will never be true because the code returns in line 120 when newType already exists.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -116,28 +117,24 @@ public DomainStatus(DomainStatus that) {
    */
   public DomainStatus addCondition(DomainCondition newCondition) {
     if (conditions.contains(newCondition)) {
-      conditions = conditions.stream()
```
doxiao, line 119: If we have a sequence of Failed/progressing/failed, the progressing condition remains because of this change. Is this behavior change intentional?

doxiao (changes requested):

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
```diff
@@ -13,13 +13,8 @@
...
   ConfigChangesPendingRestart {
     @Override
-    String getStatusMessage(DomainCondition condition) {
-      return condition.getMessage();
-    }
-
-    @Override
-    String getStatusReason(DomainCondition condition) {
-      return condition.getReason();
+    DomainConditionType[] typesToRemove() {
```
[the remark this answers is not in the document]
russgold: How does it change the behavior for available? In both the old code and new code, the available type simply inherits the default behavior.

It does change the behavior for ConfigChangesPendingRestart, which no longer sets the overall status to match that of the pending condition.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
russgold: That code should only return if the conditions are equal. This tests whether _another_ condition of the same type exists and has status True.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
russgold: Yes - since we can have multiple failures, it is not a given that we want to end the progressing state at that point. It will be removed when we hit the end progressing step.

doxiao: ok

doxiao (approved):
LGTM. thanks for your explanations.

rjeberhard (approved):

PR #2458, opened 2021-07-14
OWLS-90857 - Create copy of init container env variables for each sever pod to avoid the roll issue

ankedia:
OFSS domain has 1 domain with multiple clusters each having 2 replicas. There are init containers defined in each cluster. When the introspectVersion is changed with no other changes, the second pod gets rolled. This is caused because the init container env variables point to the same underlying data structures (and not deep copies). However, we set different values for some variables such as SERVER_NAME and SERVICE_NAME. This causes the server name env variable of the init container for the second server to have an incorrect value of the first server name. This change creates a copy of the init container env variables list so they don't get overwritten when value changes for other pods. Integration test results at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5658/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
@@ -38,12 +38,12 @@ import io.kubernetes.client.openapi.models.V1WeightedPodAffinityTerm;
 import jakarta.validation.Valid;
 import oracle.kubernetes.json.Description;
 import oracle.kubernetes.json.Feature;
+import oracle.kubernetes.operator.helpers.PodHelper;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
 import static java.util.Collections.emptyList;
-import static oracle.kubernetes.operator.helpers.PodHelper.createCopy;
 
 class ServerPod extends KubernetesResource {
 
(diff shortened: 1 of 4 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -492,6 +493,17 @@ private void addAuxiliaryImage(ServerPod serverPod1) {
     }
   }
 
+  private V1Container copyContainer(V1Container c) {
+    return new V1Container().args(c.getArgs()).command(c.getCommand()).env(createCopy(c.getEnv()))
```
doxiao, line 497: Wondering if it is better to do a clone followed with a resetting with a copy.

doxiao: Well, that does not work since V1Container is not a Cloneable.

doxiao (approved):
The change looks good to me, although I hoped that we could have a better deep copy mechanism.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerPod.java
```diff
@@ -492,6 +493,17 @@ private void addAuxiliaryImage(ServerPod serverPod1) {
     }
   }
 
+  private V1Container createWithEnvCopy(V1Container c) {
```
rjeberhard, line 496: I'm nervous about this as it will be difficult to remember to keep this in sync if later versions of K8s add more fields to V1Container. Can you use new V1ContainerBuilder(V1Container)? We should probably update the copy of V1EnvVar too.

ankedia: Change to use V1ContainerBuilder(V1Container) and V1EnvVarBuild(V1EnvVar) in c1cda3c. Please let me know if I missed something. Thanks.

rjeberhard (approved):

PR #2465, opened 2021-07-19
OWLS-90971 - Fix for admin server restart when Domain CR patched with incorrect auxiliary image command.

ankedia:
This change will rerun the introspector if the previous introspection run had failed. The recent changes to optimize the detection of when to run the introspection might prevent introspector from re-running in such cases. 

Integration test run URL - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5738/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImage.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImage.java
@@ -63,6 +63,7 @@ import static oracle.weblogic.kubernetes.actions.TestActions.patchDomainCustomRe
 import static oracle.weblogic.kubernetes.assertions.TestAssertions.secretExists;
 import static oracle.weblogic.kubernetes.assertions.TestAssertions.verifyRollingRestartOccurred;
 import static oracle.weblogic.kubernetes.utils.CommonMiiTestUtils.createDomainResource;
+import static oracle.weblogic.kubernetes.utils.CommonMiiTestUtils.verifyPodsNotRolled;
 import static oracle.weblogic.kubernetes.utils.CommonPatchTestUtils.patchDomainResource;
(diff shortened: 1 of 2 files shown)


jshum2479 (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImage.java
```diff
@@ -708,16 +709,13 @@ public void testErrorPathDomainWithFailCustomMountCommand() {
     checkPodLogContainsString(opNamespace, operatorPodName, expectedErrorMsg);
 
     // verify the domain is not rolled
-    // TODO: enable this check once https://jira.oraclecorp.com/jira/browse/OWLS-90971 is fixed
-    /*
     logger.info("sleep 2 minutes to make sure the domain is not restarted");
     try {
```
rjeberhard, line 713: @ankedia, this doesn't necessarily have to be part of your change, but I've not generally let tests have a long sleep, but have instead wanted some kind of loop that checks for a condition on a cycle. Is there some condition that we can test for (such as a status change) that would tell us that we no longer have to worry about an unexpected restart?

ankedia: Looping in @vanajamukkara and @xiancao who wrote this test. Xian asked me to uncomment the validation in the test once the issue is fixed so I only uncommented that part. I'm not sure if other tests have some pattern that can be used here.

rjeberhard: Started a conversation in Slack. I know that Vanaja is out today so maybe others can help.

rjeberhard: I found other places that do the same thing, so I'm not going to hold Anil's change for this. We should fix this generally later.

rjeberhard (approved):

PR #2466, opened 2021-07-19
Update dependencies including to the Kubernetes Java Client 13

rjeberhard:
Updated dependencies and plugins to latest versions. 

Notes:
* The latest version of Checkstyle had a non backward-compatible change to a property name
* Used "--add-opens java.base/java.time=ALL-UNNAMED" in place of "--illegal-access=permit" because the later is deprecated and the former is more specific -- the issue you would see without either of these options is that the serialization libraries could not access the fields of OffsetDateTime, which is in the java.time module.
* The CRDHelper has a utility method to output (serialize) the schema as YAML. This needed some code that I found in the Kubernetes client to properly wrap the constructors so that it would work with the SnakeYAML library.


lennyphan (approved):
LGTM

ankedia (approved):

PR #2484, opened 2021-07-28
Added Integration Test to test cluster scoped auxiliary Image

hzhao-github:
Use cases:

Use auxiliary Image at cluster scope, create image with plain text files, use them at cluster scope config and make sure the text files are only available on all pods in the cluster.

Two failures from Jenkins job below are not related to this change

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5848/ (parallel)

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImageCluster.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImageCluster.java
@@ -1,0 +1,398 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.weblogic.kubernetes;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.List;
+import java.util.Map;
+
+import oracle.weblogic.domain.Domain;
+import oracle.weblogic.kubernetes.actions.impl.primitive.Command;
+import oracle.weblogic.kubernetes.actions.impl.primitive.CommandParams;
+import oracle.weblogic.kubernetes.annotations.IntegrationTest;
+import oracle.weblogic.kubernetes.annotations.Namespaces;
+import oracle.weblogic.kubernetes.logging.LoggingFacade;
+import oracle.weblogic.kubernetes.utils.ExecResult;
+import org.apache.commons.io.FileUtils;
+import org.awaitility.core.ConditionFactory;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImageCluster.java
```diff
@@ -0,0 +1,398 @@
...
+    // verify that plain text file only copied to servers in the cluster but not copied to admin server
+    logger.info("Verify that plain text file only copied to servers in the cluster but not copied to admin server");
+    verifyFileInPod(Paths.get(auxiliaryImagePath, customDir, fileInCustomDir.getFileName().toString()).toString(),
+        fileContent);
+  }
+
+  /**
+   * A negative test to verify that model files in an auxiliary images at the cluster scope will be ignored.
+   * The test creates an auxiliary image containing JMS model files and use it
+   * to config the cluster scope configuration.
```
anpanigr, line 310: Will there be any event/log message when the Auxiliary Image @ cluster scope contains WLS resource OR  it will be silently ignored ?

anpanigr (commented):
Will there be any event/log message when the Auxiliary Image @ cluster scope contains WLS resource ? Or it will be silently ignored ?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImageCluster.java
hzhao-github: silently ignored  per Ryan

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/CommonMiiTestUtils.java
```diff
@@ -344,6 +349,181 @@ public static Domain createDomainResourceWithAuxiliaryImage(
...
+      int replicaCount,
+      List<String> clusterNames,
+      Map<String, List<String>> auxiliaryImagePathVolume,
+      List<String> auxiliaryImageDomainScopeNames,
+      List<String> auxiliaryImageClusterScopeNames) {
+
+    Domain domainCR = null;
+    int i = 0;
+
+    for (String clusterName : clusterNames) {
```
ankedia, line 389: Instead of calling the `createDomainResourceWithAuxiliaryImageClusterScope` in the for loop, it might be more clear to pass `clusterNames` as an argument to this method and then iterate over the `clusterNames` list in that method.

hzhao-github: there are two createDomainResourceWithAuxiliaryImageClusterScope, one take one cluster, another one takes a list of cluster. This test only used createDomainResourceWithAuxiliaryImageClusterScope with one cluster. createDomainResourceWithAuxiliaryImageClusterScope with a cluster list(multiple cluster) is NOT used by this changes but saved for later tests when we need to config cluster scope configurations in two cluster

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImageCluster.java
anpanigr: Ok.

anpanigr (approved):
LGTM

ankedia (approved):
LGTM

PR #2486, opened 2021-07-29
Owls 91143 - Move internal certificate initialization logic to operator initalization

ankedia:
Changes to move the internal certificate initialization logic from bash script to the java code and with the rest of the operator's initialization. The key pair and certificate are generated using `bouncycastle` library. If the operator secret is created by the `helm install`, it'll be replaced with the new data containing the operator's internal key.  Otherwise, a new secret will be created. The integration test run results are at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5819/  (I'm rerunning the failed test).


operator/src/main/java/oracle/kubernetes/operator/steps/InitializeInternalIdentityStep.java
```diff
@@ -0,0 +1,160 @@
...
+    super(next);
+  }
+
+  @Override
+  public NextAction apply(Packet packet) {
+    try {
+      KeyPair keyPair = createKeyPair();
+      String key = convertToPEM(keyPair.getPrivate());
+      writeToFile(key, new File(INTERNAL_CERTIFICATE_KEY));
+      X509Certificate cert = SelfSignedCertGenerator.generate(keyPair, SHA_256_WITH_RSA, COMMON_NAME, 3650);
```
lennyphan, line 59: Should we represent some of the key/cert parameters as statics, e.g. 3650 as CERTIFICATE_VALIDITY_DAYS ?

lennyphan (approved):
Just a couple trivial comments, otherwise LGTM.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeInternalIdentityStep.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  private static String convertToPEM(Key key) throws IOException {
+    StringWriter writer = new StringWriter();
+    JcaPEMWriter pemWriter = new JcaPEMWriter(writer);
+    pemWriter.writeObject(key);
+    pemWriter.flush();
+    return writer.toString();
+  }
```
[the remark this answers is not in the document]
ankedia: Fixed. I made the change to close the JcaPEMWriter. As per StringWriter java doc "Closing a StringWriter has no effect".

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeInternalIdentityStep.java
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeInternalIdentityStep.java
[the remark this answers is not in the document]
rjeberhard: Can you use try-with-resources pattern? https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html. This will close the stream even if something fails.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeInternalIdentityStep.java
```diff
@@ -0,0 +1,158 @@
...
+    pemWriter.close();
+    return writer.toString();
+  }
+
+  private static void writeToFile(String content, File path) throws IOException {
+    path.getParentFile().mkdirs();
+    Writer wr = Files.newBufferedWriter(path.toPath());
+    wr.write(content);
+    wr.flush();
+    wr.close();
```
rjeberhard, line 83: Same... Please use try-with-resources

rjeberhard:
@ankedia, the existing behavior is that the operator generates an internal certificate on start-up if the certificate and key don't already exist. That is, the key and certificate will be generated the first time that the operator is started. If the operator is killed and restarted then the operator will use the values already present in the config map and secret. Similarly, if the operator is upgraded then the operator will use the existing key and certificate.

I can't tell if the proposed code preserves this behavior.

operator/src/main/java/oracle/kubernetes/operator/utils/SelfSignedCertUtils.java
```diff
@@ -0,0 +1,141 @@
...
+   * @throws OperatorCreationException on creating a key id
+   * @throws CertIOException on building JcaContentSignerBuilder
+   * @throws CertificateException on getting certificate from provider
+   */
+  public static X509Certificate generateCertificate(KeyPair keyPair, String hashAlgorithm, String commonName,
+                                                    int certificateValidityDays)
+          throws OperatorCreationException, CertificateException, CertIOException {
+    Instant now = Instant.now();
+    Date notBefore = Date.from(now);
+    Date notAfter = Date.from(now.plus(Duration.ofDays(certificateValidityDays)));
```
rjeberhard, line 102: Interesting... I guess we need to handle regenerating the certificate within some window of it's becoming invalid.

lennyphan (approved):
LGTM

operator/src/main/java/oracle/kubernetes/operator/utils/SelfSignedCertUtils.java
rjeberhard: Will handle in a separate JIRA.

rjeberhard (approved):

lennyphan (approved):
LGTM

PR #2499, opened 2021-08-10
Owls 91266 - Changes to generate CoherenceMemberConfig.UnicastListenAddress for Istio

ankedia:
Changes to generate CoherenceMemberConfig.UnicastListenAddress for Istio. These include changes to the model WDT filter file for the MII domain and situational config generator for DII and domain-on-pv domains. 

Jenkins results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/5908/

### DIFF
--- a/operator/src/main/resources/scripts/introspectDomain.py
+++ b/operator/src/main/resources/scripts/introspectDomain.py
@@ -1068,6 +1068,8 @@ class SitConfigGenerator(Generator):
     self.writeListenAddress(server.getListenAddress(),listen_address)
     self.customizeNetworkAccessPoints(server,listen_address)
     self.customizeServerIstioNetworkAccessPoint(listen_address, server)
+    if (self.getCoherenceClusterSystemResourceOrNone(server) is not None):
+      self.customizeCoherenceMemberConfig(server.getCoherenceMemberConfig(),listen_address)
     self.undent()
     self.writeln("</d:server>")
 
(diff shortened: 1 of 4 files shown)


lennyphan (commented):
The changes only apply to Server Templates?  Do we need to handle for statically defined managed servers?

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -1163,6 +1167,33 @@ def _writeIstioNAP(self, name, server, listen_address, listen_port, protocol, ht
     self.undent()
     self.writeln('</d:network-access-point>')
 
+  def getCoherenceClusterSystemResourceOrNone(self, serverOrTemplate):
```
jshum2479, line 1170: I don't think this is the right test.  We should add/replace if  the domain is a managed coherence cluster.  That is having a 

resources:
       CoherenceClusterSystemResource:

defined in the WDT model or at the domain level in WLST mean,   then we are going to add/replace the unicast address.  i.e. the user may not have the CoherenceClusterSystemResource in the Server/ServerTemplate level.   Also, you need to check whether the root level CoherenceClusterSystemResource is targeted to the cluster and only do it for those are targeted.

rjeberhard (approved):

jshum2479 (approved):

PR #2500, opened 2021-08-12
get stored secrets and configmaps from packet without dependency on c…

lennyphan:
When listing secrets and configmaps, greater than the configured callRequestLimit (default 50), the previous list in the packet is overridden with the last set of results causing DomainValidation to fail.  When reading multiple results, a 'continue' flag is placed in the packet but is cleared at the beginning of the next async request which results in an empty list returned instead of the aggregated list in the packet.  It is unnecessary to rely on the 'continue' flag to return the results in the packet and hence this fix is to just return the secrets/configmaps list if it exists in the packet.

Unit tests already exist for handling listing of secrets and configmaps greater than the configured callRequestLimit.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
@@ -17,7 +17,6 @@ import io.kubernetes.client.openapi.models.V1SecretList;
 import oracle.kubernetes.operator.DomainStatusUpdater;
 import oracle.kubernetes.operator.MakeRightDomainOperation;
 import oracle.kubernetes.operator.ProcessingConstants;
-import oracle.kubernetes.operator.calls.AsyncRequestStep;
 import oracle.kubernetes.operator.calls.CallResponse;
 import oracle.kubernetes.operator.helpers.EventHelper.EventData;
 import oracle.kubernetes.operator.helpers.EventHelper.EventItem;
@@ -80,12 +79,12 @@ public class DomainValidationSteps {
     }
 
     static List<V1Secret> getSecrets(Packet packet) {
-      return Optional.ofNullable(getSecretsIfContinue(packet)).orElse(new ArrayList<>());
+      return Optional.ofNullable(getSecretsFromPacket(packet)).orElse(new ArrayList<>());
     }
 
     @SuppressWarnings("unchecked")
-    private static List<V1Secret> getSecretsIfContinue(Packet packet) {
-      return packet.get(AsyncRequestStep.CONTINUE) != null ? (List<V1Secret>) packet.get(SECRETS) : null;
+    private static List<V1Secret> getSecretsFromPacket(Packet packet) {
+      return packet.get(SECRETS) != null ? (List<V1Secret>) packet.get(SECRETS) : null;
     }
   }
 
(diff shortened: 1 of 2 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -80,12 +79,12 @@ public NextAction onSuccess(Packet packet, CallResponse<V1SecretList> callRespon
     }
 
     static List<V1Secret> getSecrets(Packet packet) {
-      return Optional.ofNullable(getSecretsIfContinue(packet)).orElse(new ArrayList<>());
+      return Optional.ofNullable(getSecretsFromPacket(packet)).orElse(new ArrayList<>());
```
doxiao, line 82: We can just inline (List<V1Secret>) packet.get(SECRETS) here and remove the getSecretsFromPacket method.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -105,12 +104,12 @@ public NextAction onSuccess(Packet packet, CallResponse<V1ConfigMapList> callRes
     }
 
     static List<V1ConfigMap> getConfigMaps(Packet packet) {
-      return Optional.ofNullable(getConfigMapsIfContinue(packet)).orElse(new ArrayList<>());
+      return Optional.ofNullable(getConfigMapsFromPacket(packet)).orElse(new ArrayList<>());
```
doxiao, line 107: The same as above.

doxiao (commented):
LGTM. Just had two minor comments.

russgold:
Please add a unit test that requires this change, to make it clear why and to ensure that it is not inadvertently undone.

rjeberhard:
So, with this change do you see the aggregate list? How did this regress? I think we have a slightly different pattern for other lists.

lennyphan:
> So, with this change do you see the aggregate list? How did this regress? I think we have a slightly different pattern for other lists.

Yes, with this change we see the aggregate list.  The issue is that retrieval of the list from subsequent query is dependent on the CONTINUE flag in the packet.  Unfortunately, for the subsequent request the CONTINUE flag is removed from the packet, in AsyncRequestStep.apply() (see line 263), and hence we return a new empty list for the SECRETS instead of the list from the packet.  We then overwrite the SECRETS entry in the packet with the last result set, losing the previous result set.  I tried looking at other lists for resources, like namespace, and it looks very similar to the proposed change.  Can you clarify the pattern used for other lists in comparison to my proposed change?  I would like to align it so that we’re handling lists the same.  Also, our unit tests pass since the unit test support stubs out the AsyncRequestStep and we aren't removing the CONTINUE flag from the packet.  I have a change that incorporates the same behavior (removing CONTINUE flag from packet) and this exposed the same failure in the unit tests that Rakshit was seeing.

rjeberhard:
> Can you clarify the pattern used for other lists in comparison to my proposed change?

Everything is fine with your change. I see that the SECRETS and CONFIGMAP Packet entries are populated with code that follows the pattern I was expecting.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -105,12 +104,12 @@ public NextAction onSuccess(Packet packet, CallResponse<V1ConfigMapList> callRes
...
     }
 
     @SuppressWarnings("unchecked")
-    private static List<V1ConfigMap> getConfigMapsIfContinue(Packet packet) {
-      return packet.get(AsyncRequestStep.CONTINUE) != null ? (List<V1ConfigMap>) packet.get(CONFIGMAPS) : null;
+    private static List<V1ConfigMap> getConfigMapsFromPacket(Packet packet) {
+      return packet.get(CONFIGMAPS) != null ? (List<V1ConfigMap>) packet.get(CONFIGMAPS) : null;
```
rjeberhard, line 112: It looks like Dongbo's comment above is correct -- this code could be simplified or just inlined as it tests a key and then returns either the same value or null -- it could have just returned the packet.get() straight away.

russgold:
So the change lets the tests work for secrets? I don't see the new test.

lennyphan:
> So the change lets the tests work for secrets? I don't see the new test.

the change  to KubernetesTestSupport replicates the behavior in AsyncRequestStep which then caused the unit tests to fail (failure was masked by not removing the CONTINUE flag).  This replicates the failure that Rakshit was seeing.  The change(s) to DomainValidationSteps then resolves the issue and the current unit tests cover the feature for reading in multiple results, hence I didn't add a new unit test.

rjeberhard (approved):

doxiao (approved):

PR #2507, opened 2021-08-18
Owls 91273 - Create network access points for kubectl port-forward and console access

ankedia:
The notes containing use-cases and configurations for OWLS-91273 are listed in this wiki - https://confluence.oraclecorp.com/confluence/display/ASDevWLS/OWLS-91273+-+Solution+for+kubectl+port-forward+and+console+access+notes . 
I have tested various combinations using kubectl port-forwarding to access the admin console and connecting to the admin server using WLST. 

Integration test results are at - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6032/


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -115,6 +115,15 @@
           + "Defaults to true if `domainHomeSourceType` is PersistentVolume; false, otherwise.")
   private Boolean logHomeEnabled; // Boolean object, null if unspecified
 
+  /**
+   * Whether the port forwarding is enabled.
+   */
+  @Description(
+          "Specifies whether the port forwarding is enabled. "
+                  + "Defaults to true.")
+  private Boolean portForwardingEnabled = true;
```
rjeberhard, line 124: Need to add portForwardingEnabled to equals(), hashMap(), and toString()

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
```diff
@@ -115,6 +115,15 @@
           + "Defaults to true if `domainHomeSourceType` is PersistentVolume; false, otherwise.")
   private Boolean logHomeEnabled; // Boolean object, null if unspecified
 
+  /**
+   * Whether the port forwarding is enabled.
+   */
+  @Description(
+          "Specifies whether the port forwarding is enabled. "
```
rjeberhard, line 122: Most customers won't understand what this means. Perhaps something like, "When this flag is enabled, the operator enhances the domain configuration with a NetworkAccessPoint for the admin protocol that listens on localhost. This allows access to the Administration Console using the `kubectl port-forward` pattern, which is often more secure than exposing the console to the public Internet. Defaults to true."

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -1167,6 +1170,48 @@ def _writeIstioNAP(self, name, server, listen_address, listen_port, protocol, ht
...
+  def _getPortForwardNapConfigOverrideAction(self, svr, testname):
+    add_action = 'f:combine-mode="add"'
+    found = False
+    for nap in svr.getNetworkAccessPoints():
+      if nap.getName() == testname:
+        found = True
+        break
+
+    if found:
+      trace("SEVERE","Found NetWorkAccessPoint with name %s in the WebLogic Domain, this is an internal name used by the WebLogic Kubernetes Operator, please remove it from your domain and try again." % testname)
```
rjeberhard, line 1182: NetWorkAccessPoint -> NetworkAccessPoint

rjeberhard:
Two general questions:

1. If you have a domain running with operator 3.3.0 and then upgrade to your branch, do servers restart? Is the console accessible using port-forward? I'm guessing that the answer is mixed... For instance, for MII, I think that introspection will no-op because there has been no change in the model, etc., and so servers will not restart, but also that the port-forwarding won't be available until there is a domain restart or reason for introspection to rerun. For DII and DPV, introspection will eventually run, situational configuration will be updated and then port forwarding should work without a restart.

2. How does your flag interact with Istio support being enabled?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -323,6 +323,7 @@ private String getAsServiceName() {
       addEnvVar(vars, IntrospectorJobEnvVars.WDT_DOMAIN_TYPE, getWdtDomainType());
       addEnvVar(vars, IntrospectorJobEnvVars.DOMAIN_SOURCE_TYPE, getDomainHomeSourceType().toString());
       addEnvVar(vars, IntrospectorJobEnvVars.ISTIO_ENABLED, Boolean.toString(isIstioEnabled()));
+      addEnvVar(vars, IntrospectorJobEnvVars.PORT_FORWARDING_ENABLED, Boolean.toString(isPortForwardingEnabled()));
```
jshum2479, line 326: This will cause introspection runs and eventually roll the domain in upgrade situation.

ankedia: This environment variable is only added to the introspector job pod and not to the WLS server pods. So I think pods should not roll in case of an upgrade situation. I'll give it a try to confirm this.

jshum2479: Oh, ok.

tbarnes-us: Please specifically test this for Model-in-Image -- I suspect the MII introspector will detect the change in configuration and recommend a roll...

ankedia: I tested it for MII and am not seeing a roll. Since the env variable is added only in the job helper and the previous job will not be available for comparison, I think it'll not be able to detect the change in job configuration.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
ankedia: Done.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainSpec.java
ankedia: Fixed.

operator/src/main/resources/scripts/introspectDomain.py
ankedia: Fixed.

tbarnes-us:
General comment: I've started a private slack thread amongst the reviewers to discuss if we want to default to the new setting, the setting's name/location, and other issues.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: I was also thinking about the configuration change. How about when the MII online-update `domain.spec.configuration.model.onlineUpdate` is set to `enabled=true` and `onNonDynamicChanges=CommitUpdateAndRoll`?

jshum2479 (approved):

tbarnes-us (approved):

PR #2508, opened 2021-08-18
Implement new domain status approach

russgold:
This is a massive reworking of the domain status approach, as described in [OWLS-89964](https://jira.oraclecorp.com/jira/browse/OWLS-89964).

It includes a fair bit of refactoring.

It is still missing implementation of the ReplicasTooHigh failure, as I am trying to figure out the right way to do it. 

This change is absolutely not backwards-compatible, and functional tests that depend on the current approach to domain conditions are going to break, which is one reason I would like this available now. Those will need either to be disabled or modified before this is actually merged.


russgold:
The integration test is, as expected, failing a lot. There are 168 failures, and the ones I looked at including long and futile waits for the obsolete conditions to appear. https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6036/

rjeberhard:
Okay, I've looked at this change and everything "looks" good. I'll admit that it's really hard to validate if the proposed changes match the requirements listed in the JIRA; however, I think that it does. I'm going to review the test failures and give an initial assessment.

tbarnes-us:
General comment: There are multiple open issues in https://jira.oraclecorp.com/jira/browse/OWLS-89964 -- these should be resolved before 4.0 is shipped.  If the JIRA is closed, then new JIRAs should be created to track the pending impl & design work listed in the open issues.

russgold:
With that second commit, the number of failures is down to 21 https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6089/

russgold:
Replaced by https://github.com/oracle/weblogic-kubernetes-operator/pull/2529

PR #2512, opened 2021-08-23
OWLS-91273 - Script solution for kubectl port-forward and console access

ankedia:
This PR is for setting up the 'kubectl port-forward' using a script that can be run independently. The script creates a port-forward server pod in the domain namespace using an image containing 'socat' binary (default alpine/socat) and it forwards the requests to the administration server pod. The host and admin port of the admin server is derived from the domain topology stored in the introspector config map. The port can also be provided using the command line "-r" option.

### DIFF
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -899,6 +899,53 @@ function validateClusterName {
   fi
 }
 
+#
+# Function to get the host and admin port of the admin server for "kubectl port-forward".
+# 1) If the domain wide admin port is enabled, it returns the administration port.
+# 2) Otherwise if a custom admin channel is configured, it returns the
+#    listen port of the custom admin channel.
+# 3) Othewise if the default secure channel is configured, it returns the listen port
+#    of the default secure channel.
+# 4) If none of the above condition is true, it returns the listen port of the
+#    default channel.
+#
+# $1 - Domain unique id.
+# $2 - Domain namespace.
+# $3 - Return value containing host name of the admin server listen address.
+# $4 - Return value containing domain administration port.
+#
+function getAdminServerHostAdminPort {
+  local domainUid=$1
+  local domainNamespace=$2
+  local __host=$3
+  local __port=$4
+  local __remoteHost=""
+  local __remotePort=0
+
+  getTopology "${domainUid}" "${domainNamespace}" jsonTopology
+  adminServer=$(echo $jsonTopology | jq -r .domain.adminServerName)
+  adminServerCmd="(.domain.servers[] | select(.name == \"${adminServer}\"))"
+  server=$(echo $jsonTopology | jq "${adminServerCmd}")
+  customAdminChannel=$(echo $server | jq '.networkAccessPoints[] | select(.protocol=="admin")')
+  listenPort=$(echo $server | jq -r .listenPort)
+  sslListenPort=$(echo $server | jq -r .sslListenPort)
+  adminPort=$(echo $server | jq -r .adminPort)
+  customAdminChannelPort=$(echo $customAdminChannel | jq -r .listenPort)
+  __remoteHost=$(echo $server | jq -r .listenAddress)
+  if [ $adminPort != null ]; then
+    __remotePort=$adminPort
+  elif [[ ! -z $customAdminChannelPort && $customAdminChannelPort != null ]]; then
+    __remotePort=$customAdminChannelPort
+  elif [ $sslListenPort != null ]; then
+    __remotePort=$sslListenPort
+  else
+    __remotePort=$listenPort
+  fi
+  eval $__host="'${__remoteHost}'"
+  eval $__port="'${__remotePort}'"
+
+}
+
 function getTopology {
   local domainUid=$1
   local domainNamespace=$2
@@ -1047,3 +1094,13 @@ function toDNS1123Legal {
   eval $__result="'$val'"
 }
 
+#
+# Function to cleanup a port-forward server pod
+#
+function portForwardCleanup {
+  local domainNamespace=$1
+  local portForwardPodName=$2
+
+  echo ""
+  ${kubernetesCli} -n $domainNamespace delete pod/$portForwardPodName --grace-period 1 --wait=false
+}
--- a/kubernetes/samples/scripts/domain-lifecycle/helper.sh
+++ b/kubernetes/samples/scripts/domain-lifecycle/helper.sh
@@ -923,19 +923,18 @@ function getAdminServerHostAdminPort {
   local __remotePort=0
 
   getTopology "${domainUid}" "${domainNamespace}" jsonTopology
-  adminServer=$(echo $jsonTopology | jq -r .domain.adminServerName)
-  adminServerCmd="(.domain.servers[] | select(.name == \"${adminServer}\"))"
-  server=$(echo $jsonTopology | jq "${adminServerCmd}")
-  customAdminChannel=$(echo $server | jq '.networkAccessPoints[] | select(.protocol=="admin")')
-  listenPort=$(echo $server | jq -r .listenPort)
-  sslListenPort=$(echo $server | jq -r .sslListenPort)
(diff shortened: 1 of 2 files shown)


kubernetes/samples/scripts/domain-lifecycle/helper.sh
```diff
@@ -1047,3 +1093,15 @@ function toDNS1123Legal {
...
+#
+# Function to cleanup a port-forward server pod
+# $1 - Domain namespace
+# $2 - Name of the port-forward server pod
+#
+function portForwardCleanup {
+  local domainNamespace=$1
+  local portForwardPodName=$2
+
+  echo ""
```
jshum2479, line 1105: Shouldn't there be something inside the echo?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+portForwardPodName=port-forward-server
+localAddress=""
+localPort=""
+adminHost=""
+remotePort=""
+address=""
+imagePullPolicy=IfNotPresent
+imageName=alpine/socat
+waitTimeout=30
+podRunningTimeout=
```
jshum2479, line 78: should there be a default for podRunningTimeout ? 1m?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+
+  cat << EOF
+
+  This script sets up the 'kubectl port-forward' for a WebLogic domain to allow access to the
+  WebLogic Administration Console. The 'kubectl port-forward' pattern is often more secure than
+  exposing the console to the public Internet. The script creates a pod in the domain namespace
+  using the image containing 'socat' binary and it forwards the request to the WebLogic
+  Admistration Server pod. The script derives the address and the adminstration port of the
+  WebLogic Admistration Server using the domain topology. The port can also be provided on the
+  command line using '-r' option.
```
tbarnes-us, line 22: Usage could use more detail:
- Make it more clear that the console requests redirect from the port-forward, through a 'socat pod' on to the admin server. (something like "it forwards the request" --> "the script runs a port-forward command which forwards console requests to the socat pod which in turn forwards requests to the admin server pod'
- Explain that the script runs a port-forward command in foreground after launching the pod and a that a ^C not only terminates the port-forward command but also implicitly tears down the pod.

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+                   [-t pod_running_timeout]
+
+    -d <domain_uid>                : Domain unique-id. This parameter is required.
+
+    -n <namespace>                 : Domain namespace. This parameter is required.
+
+    -a <local_ip_address>          : Addresses to listen on (comma separated value for --address flag). 
+                                     Only accepts IP addresses or localhost as a value. Default is 'localhost'.
+
+    -l <local_port>                : A local port to listen on. If not specified, it listens on a random local port.
```
tbarnes-us, line 37: If its random, how do you find out the port number?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+    $(basename $0) -n namespace -d domain_uid [-a local_ip_address] [-l localPort] [-i image_name] 
+                   [-p image_pull_policy] [-w wait_timeout] [-p port_forward_server_name] [-m kubecli] 
+                   [-t pod_running_timeout]
+
+    -d <domain_uid>                : Domain unique-id. This parameter is required.
+
+    -n <namespace>                 : Domain namespace. This parameter is required.
+
+    -a <local_ip_address>          : Addresses to listen on (comma separated value for --address flag). 
+                                     Only accepts IP addresses or localhost as a value. Default is 'localhost'.
```
tbarnes-us, line 35: It'd help to make the purpose of this more clear:  Is this the externally available address on the Kubernetes node that redirects to the admin server within the k8s cluster?  What is `--address` ?  Some sort of parameter for the port-forward command?

tbarnes-us: Maybe a better name is 'external_ip_address' given the context?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
tbarnes-us: Maybe a better name is 'external port' given the context?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+
+    -l <local_port>                : A local port to listen on. If not specified, it listens on a random local port.
+
+    -r <remote_port>               : The port of the WebLogic Administartion Server to forward the request to.
+                                     If not specified, this script will derive the adminstration port from the
```
tbarnes-us, line 40: Maybe something like remote_port --> internal_admin_port?

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+                                     WebLogic domain topology stored in the introspector config map.
```
tbarnes-us, line 41: The existence and nature of the "introspector config map" is not commonly known to customers.  Maybe add. "The introspector config map is automatically created by the Operator for a domain after the domain's introspector job succeeds".

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+    -h                             : This help.
+   
+EOF
+exit $1
+}
+
+kubernetesCli=${KUBERNETES_CLI:-kubectl}
+domainUid=""
+domainNamespace=""
+portForwardPodName=port-forward-server
```
tbarnes-us, line 69: Can there be more than one domain active at one time?  If so I assume the portForwardPodName should be decorated with the DOMAIN_UID.

Can there be more than one port-forward active for a domain at one time? If so, I assume it should also be decorated with the local port number...

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
```diff
@@ -0,0 +1,181 @@
...
+
+trap "portForwardCleanup \"${domainNamespace}\" \"$portForwardPodName\"" EXIT
+
+getAdminServerHostAdminPort "${domainUid}" "${domainNamespace}" adminHost derivedAdminPort
+if [ -z $remotePort ]; then
+remotePort=$derivedAdminPort
+fi
+
+printInfo "Creating a pod with name '$portForwardPodName' in the namespace '$domainNamespace' using image '$imageName.'"
+${kubernetesCli} run -n $domainNamespace --image-pull-policy $imagePullPolicy --restart=Never --image=$imageName $portForwardPodName -- -d -d tcp-listen:$remotePort,fork,reuseaddr tcp-connect:$adminHost:$remotePort
```
tbarnes-us, line 152: It's possible for there to be a pod with the same name already running, perhaps because someone else is already running the script, or someone did a 'kill -9' on the script (bypassing the trap). In which case would the command fail and the script exit plus trap/delete someone else's pod?  Perhaps the script should exit early with a failure explanation if the pod is already there (without setting the trap first).

kubernetes/samples/scripts/domain-lifecycle/setupPortForwarding.sh
tbarnes-us: Typo: Administartion

tbarnes-us: Server --> Server Pod

PR #2518, opened 2021-08-27
Create testUntil() method in common test utilities

rjeberhard:
The logging from our use of the Awaitility until() method was not optimal, especially when the test failed due to a timeout. This change creates a shared set of methods and updates all of the tests to use it so that we have access to improve that logging.

You will now see clear indication of "Waiting for", "Completed", or "Timed out" status of each of these activities.


vanajamukkara (approved):
LG

PR #2529, opened 2021-09-10
Owls 89964

russgold:
This PR replaces https://github.com/oracle/weblogic-kubernetes-operator/pull/2508

This is a massive reworking of the domain status approach, as described in OWLS-89964.

It includes a fair bit of refactoring.

It is still missing implementation of the ReplicasTooHigh failure, as I am trying to figure out the right way to do it.

This change is absolutely not backwards-compatible, and functional tests that depend on the current approach to domain conditions are going to break, which is one reason I would like this available now. Those will need either to be disabled or modified before this is actually merged.

At present, there are still a number of failing integration tests, and I am going to need some help debugging them.

This is a full run, with 20 failures: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6330/ 

And here is a run of one suite after trying to fix the most obvious problems. It seems not to have improved. https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6333/


russgold:
Down to 9 failures https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6425/

russgold:
The reversion in the last commit was found by Sankar to make a difference; I admit that I don't understand why, but it makes sense that Available should suffice.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
```diff
@@ -208,7 +206,7 @@ void testDomainK8SEventsSuccess() {
     logger.info("verify the DomainCreated event is generated");
     checkEvent(opNamespace, domainNamespace1, domainUid, DOMAIN_CREATED, "Normal", timestamp);
     logger.info("verify the DomainProcessing Starting/Completed event is generated");
-    checkEvent(opNamespace, domainNamespace1, domainUid, DOMAIN_PROCESSING_STARTING, "Normal", timestamp);
+    // checkEvent(opNamespace, domainNamespace1, domainUid, DOMAIN_PROCESSING_STARTING, "Normal", timestamp);
```
rjeberhard, line 209: Do we know generate an event when the Complete status is achieved or is that a follow-on JIRA?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
```diff
@@ -570,7 +568,7 @@ void testLogHomeChangeEvents() {
 
 
   /**
-   * The test modifies the includeServerOutInPodLog property and verifies the domain roll starting events are logged.
+   * The test modifies the erOutInPodLog property and verifies the domain roll starting events are logged.
```
rjeberhard, line 571: I'm guessing that this was an inadvertent change.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItParameterizedDomain.java
```diff
@@ -758,14 +757,6 @@ void testMultiClustersRollingRestart() {
         "domain event {0} to be logged",
         DOMAIN_CHANGED);
 
-    // verify the DomainProcessing Starting/Completed event is generated
```
rjeberhard, line 761: Even if we delay testing for a Completed event to a later JIRA, I'm worried that if we just delete these checks then we will have trouble finding where to restore them. I guess we will have the source control. Still, it might be easier to put a TO-DO type marker.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
russgold: Yes, we do. I can try to see if the test can detect that rather than processing complete, within the time limit.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
russgold: yup, fixed.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItParameterizedDomain.java
russgold: We don't want the domain_processing_starting event any more. We do want domain-completed.

russgold:
Down to three failures

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6620/

russgold:
And all three failures appear to be intermittent, as they passed when run separately
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6637/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6635/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6634/

PR #2531, opened 2021-09-13
Changed the image names in MII sample script tests

hzhao-github:
Jenkins job

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6318 (parallel)


integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
```diff
@@ -78,6 +75,16 @@
     AUX
   }
 
+  private String getModelImageName(String suffix) {
+    String currentDateTime = getDateAndTimeStamp();
```
tbarnes-us, line 79: I thought that the test depends on the timestamp being the same throughout the entire run? If so, then currentDateTime should remain a global static final instead of being recalculated every time.

hzhao-github: changed

tbarnes-us (approved):

anpanigr (approved):
lgtm

PR #2533, opened 2021-09-13
Owls 91448 - Prevent insecure file system warnings by ensuring files are at a minimum of umask 027 and handle Openshift platform.

ankedia:
- Added a helm variable with name `cloudPlatform` to detect the Openshift platform. 

- For the default (not Openshift) platforms, this change makes sure that files in DOMAIN_HOME dir are at a minimum of umask 027. For this, it (a) restricts the execute permission for shell scripts to user-only, (b) removes group write and execute permissions on boot.properties and (c) preserves permissions when extracting the files from zip file.

- For the operator running on OpenShift, this change adds a system property to WLS startup to disable umask 027 checks (all domain styles). For Model-in-image domain on Openshift, it also uses `chmod -R g=u {domain home}` to give group the same permission as the user for files in DOMAIN_HOME dir.


operator/src/main/resources/scripts/startServer.sh
```diff
@@ -290,12 +290,27 @@ createFolder ${DOMAIN_HOME}/servers/${SERVER_NAME}/security
...
   trace "Copying situational configuration files from operator cm to ${DOMAIN_HOME}/optconfig directory"
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jms         'Sit-Cfg-JMS--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jdbc        'Sit-Cfg-JDBC--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/diagnostics 'Sit-Cfg-WLDF--'
+else
+  if [[ ! -z ${CLOUD_PLATFORM} && ${CLOUD_PLATFORM} == "Openshift" ]]; then
+    # Operator running on Openshift platform - change file permissions in the DOMAIN_HOME dir to give
+    # group same permissions as user .
+    chmod -R g=u ${DOMAIN_HOME} || return 1
```
tbarnes-us, line 307: A recursive file traversal of the entire domain-home seems like it could be prohibitively expensive, especially as it's done for every pod cycle.  For DII, can this occur solely when the image is created?  And for DiPV can it be limited to occur only in the introspector, and, even then, only when the first time the introspector sees the domain?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerEnvVars.java
```diff
@@ -62,10 +62,12 @@
   /** If present, pod scripts will watch for changes to override configurations and move them into place. */
   public static final String DYNAMIC_CONFIG_OVERRIDE = "DYNAMIC_CONFIG_OVERRIDE";
 
+  public static final String CLOUD_PLATFORM = "CLOUD_PLATFORM";
+
   private static final List<String> RESERVED_NAMES = Arrays.asList(
         DOMAIN_UID, DOMAIN_NAME, DOMAIN_HOME, NODEMGR_HOME, SERVER_NAME, SERVICE_NAME,
         ADMIN_NAME, AS_SERVICE_NAME, ADMIN_PORT, ADMIN_PORT_SECURE, ADMIN_SERVER_PORT_SECURE,
-        LOG_HOME, SERVER_OUT_IN_POD_LOG, DATA_HOME, ACCESS_LOG_IN_LOG_HOME, DYNAMIC_CONFIG_OVERRIDE);
+        LOG_HOME, SERVER_OUT_IN_POD_LOG, DATA_HOME, ACCESS_LOG_IN_LOG_HOME, DYNAMIC_CONFIG_OVERRIDE, CLOUD_PLATFORM);
```
jshum2479, line 70: Is this going to cause operator upgrade rolling side effect ?

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -290,12 +290,27 @@ createFolder ${DOMAIN_HOME}/servers/${SERVER_NAME}/security
...
+
+
 if [ ${DOMAIN_SOURCE_TYPE} != "FromModel" ]; then
   trace "Copying situational configuration files from operator cm to ${DOMAIN_HOME}/optconfig directory"
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jms         'Sit-Cfg-JMS--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jdbc        'Sit-Cfg-JDBC--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/diagnostics 'Sit-Cfg-WLDF--'
+else
+  if [[ ! -z ${CLOUD_PLATFORM} && ${CLOUD_PLATFORM} == "Openshift" ]]; then
```
tbarnes-us, line 304: To minimize user error, consider using a "toupper" equivalent - something like `${CLOUD_PLATFORM^^} == "OPENSHIFT"`.

operator/src/main/resources/scripts/startServer.sh
jshum2479, line 307: Should we also do it for non mii case? Or this is left as an exercise for the user when they create the domain?

operator/scripts/operator.sh
```diff
@@ -59,6 +59,10 @@ if [ "${MOCK_WLS}" == 'true' ]; then
   MOCKING_WLS="-DmockWLS=true"
 fi
 
+if [ ! -z "${CLOUD_PLATFORM}" ]; then
+  PLATFORM="-DcloudPlatform=$CLOUD_PLATFORM"
+fi
```
tbarnes-us, line 64: To minimize the need for null/empty string checks through the remainder of the code, maybe just set it to "unknown" by default?  `PLATFORM="-DcloudPlatform=${CLOUD_PLATFORM:-unknown}"`

operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -1079,6 +1080,10 @@ private boolean mockWls() {
     return Boolean.getBoolean("mockWLS");
   }
 
+  private String getCloudPlatform() {
+    return System.getProperty("cloudPlatform");
```
tbarnes-us, line 1084: `kubernetesPlatform`?

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -64,6 +64,8 @@ spec:
           value: "false"
         - name: "JAVA_LOGGING_LEVEL"
           value: {{ .javaLoggingLevel | quote }}
+        - name: "CLOUD_PLATFORM"
+          value: {{ .cloudPlatform | quote }}
```
tbarnes-us, line 68: There technically might not be a cloud involved. How about `.kubernetesPlatform` and `KUBERNETES_PLATFORM` throughout?

tbarnes-us: Also, this setting should be documented in `userguide/managing-operators/using-helm.md`, `security/openshift.md`, and` `userguide/platforms/environments.md#openshift`.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerEnvVars.java
tbarnes-us: +1

operator/src/main/resources/scripts/startServer.sh
tbarnes-us: @jshum2479 

I originally suggested that we handle the non-MII cases, but I have a similar concern.  In detail:

For one, a recursive file traversal of the entire domain-home seems like it could be prohibitively expensive, especially as it's done for every pod cycle. 

For DII in particular, maybe this can this occur solely when the image is created (and be left up to the customer)? 

And for DiPV, if we must do something automatic, can it be limited to occur only in the introspector, and, even then, only the first time the introspector sees the domain?

jshum2479: I agreed, doing it in the introspector (before zipping up the domain) is a better place.  Do we really need -R or is the check only for bin directory?

tbarnes-us: @rjeberhard FYI

jshum2479: Also, have we explored doing it in WDT?

jshum2479: Sorry, one more question.  Does the new was check provide any guidance on JKS keystore files?   Best practice is user read only.

ankedia: I'll set up a meeting to go over this. As per the requirements in JIRA OWLS-91448, we should NOT use `chmod -R g=u` for Model-in-image domain unless the environment is OpenShift. My understanding from JIRA was that, DII case will be handled in image tool and it's customers responsibility to do it for domain-on-pv.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerEnvVars.java
ankedia: I believe it will not cause the rolling side effect, I'll run the upgrade to confirm.

operator/src/main/resources/scripts/startServer.sh
tbarnes-us: Also, there's no need for the -z check if the second macro is encased in quotes, so then we have:
`if [[ ${CLOUD_PLATFORM^^} == "OPENSHIFT" ]]; then`

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ServerEnvVars.java
ankedia: I have verified that rolling doesn't happen during upgrade when `kubernetesPlatform` helm variable is not present. I add the `KUBERNETES_PLATFORM` environment variable to server pods only when `kubernetesPlatform` is set in the helm values.

operator/src/main/resources/scripts/startServer.sh
ankedia: Done.

operator/scripts/operator.sh
ankedia: I think this can result in rolling side effects unless we check for `unknown` and not set the environment variable. Additionally, if someone runs the `ps -ef` command to look at operator process command-line arguments, it might be confusing. I have made changes as per the above suggestion that minimizes the need for null/empty string checks.

tbarnes-us: ok

operator/src/main/resources/scripts/startServer.sh
```diff
@@ -290,12 +290,27 @@ createFolder ${DOMAIN_HOME}/servers/${SERVER_NAME}/security
...
+
+
 if [ ${DOMAIN_SOURCE_TYPE} != "FromModel" ]; then
   trace "Copying situational configuration files from operator cm to ${DOMAIN_HOME}/optconfig directory"
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig             'Sit-Cfg-CFG--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jms         'Sit-Cfg-JMS--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/jdbc        'Sit-Cfg-JDBC--'
   copySitCfgWhileBooting /weblogic-operator/introspector ${DOMAIN_HOME}/optconfig/diagnostics 'Sit-Cfg-WLDF--'
+else
+  if [[ ${KUBERNETES_PLATFORM^^} == "OPENSHIFT" ]]; then
```
tbarnes-us, line 304: I know KUBERNETES_PLATFORM is always set to something now, but it 'future proofs for safety' if you encase `${KUBERNETES_PLATFORM^^}` in quotes -- e.g. `"${KUBERNETES_PLATFORM^^}"`

russgold (changes requested):
>> I see changes to two Java files, but no unit tests for those changes.
[ANIL]
Added following new unit tests in JobHelperTest.java and PodHelperTestBase.java -
whenOperatorHasKubernetesPlatformConfigured_introspectorPodSpecHasKubernetesPlatformEnvVariable
whenNotConfigured_KubernetesPlatform_introspectorPodSpecHasNoKubernetesPlatformEnvVariable
whenOperatorHasKubernetesPlatformConfigured_createdPodSpecContainerHasKubernetesPlatformEnvVariable
whenNotConfigured_KubernetesPlatform_createdPodSpecContainerHasNoKubernetesPlatformEnvVariable

operator/src/main/resources/scripts/startServer.sh
ankedia: Done.

operator/src/main/resources/scripts/startServer.sh
ankedia: I have made the changes to perform the `chmod -R g=u ${DOMAIN_HOME}` in the introspector for model-in-image domain and to perform `chmod -R g=u ${DOMAIN_HOME}` in startServer.sh script for domain-home-in-image. My understanding is that we should not be doing `chmod -R` for the domain-on-pv domains since those changes will be persistent.

tbarnes-us: If the image tool is going to handle it  then why does startServer need to? (You wrote _"DII case will be handled in image tool "_ and _"I have made the changes to perform ...  chmod -R g=u ${DOMAIN_HOME} in startServer.sh script for domain-home-in-image."_)

ankedia: My comment "DII case will be handled in the image tool" on 9/13 was based on my previous understanding. However, in the standup meeting discussion on Tuesday 9/14, Derek and Robert Patrick suggested to also do `chmod -R u=g ${DOMAIN_HOME}` for DII in the operator.

documentation/staging/content/userguide/platforms/environments.md
```diff
@@ -65,6 +65,12 @@ Operator 2.5.0+ is certified for use on OpenShift Container Platform 4.3.0+ with
...
 
+Beginning with operator version 4.0, you should specify the `kubernetesPlatorm` Helm chart property and set its value to `Openshift` when installing the operator in Openshift. With this setting, the operator:
+- Configures the correct file permissions for WebLogic Server to work in Openshift for [Model in Image]({{< relref "/samples/domains/model-in-image/_index.md" >}}), and [Domain home in Image]({{< relref "/samples/domains/domain-home-in-image/_index.md" >}}) domains.
+- Sets `weblogic.SecureMode.WarnOnInsecureFileSystem` Java system property to `false` on each target WebLogic Server instance. This flag suppresses the insecure file system warnings in the WebLogic Server console in production mode. These warnings result from setting the correct file permissions to work with restricted security context constraints on Openshift.
+For more information about the Helm chart, see the
+[Operator Helm configuration values]({{<relref "/userguide/managing-operators/using-helm#operator-helm-configuration-values">}}).
```
tbarnes-us, line 72: Are extra steps needed for Domain in PV (I'm wondering if the user may need to run a chmod -R from within a container after creating the domain home...)?  If so, they should be documented here.

russgold (approved):

documentation/staging/content/userguide/platforms/environments.md
ankedia: I'm not sure, I can't run the Domain in PV test in Openshift environment. I checked with Bhavani and she's also having trouble running any of the Domain in PV tests on OKD.

ankedia: I was able to successfully run the Domain in PV in the PKD environment with and without `kubernetesPlatform=Openshift` flag. The test passed in both cases, it looks like no extra steps are needed for Domain in PV.

https://build.weblogick8s.org:8443/job/wko-okd-test/16/ (with `kubernetesPlatform=Openshift`)
https://build.weblogick8s.org:8443/job/wko-okd-test/14/ (without `kubernetesPlatform` flag)

documentation/staging/content/userguide/managing-operators/using-helm.md
```diff
@@ -96,6 +96,18 @@ Example:
 javaLoggingLevel:  "FINE"
 ```
 
+##### `kubernetesPlatform`
+Allows you to set the Kubernetes platform on which the operator is running. The value is case-insensitive.
```
tbarnes-us, line 100: Please consider replacing this with something like the following slight rewrite of the current wording in order to change its emphasis and include a little additional info: 

Specify the Kubernetes platform on which the operator is running. This setting has no default, the only valid value is `OpenShift`, and the setting should be left unset for other platforms. When set to `OpenShift`, the operator:

- Sets the domain home file permissions in each WebLogic Server pod to work correctly in OpenShift for [Model in Image]({{< relref "/samples/domains/model-in-image/_index.md" >}}), and [Domain home in Image]({{< relref "/samples/domains/domain-home-in-image/_index.md" >}}) domains. Specifically, it sets file group permissions so that they match file user permissions. 

- Sets the `weblogic.SecureMode.WarnOnInsecureFileSystem` Java system property to `false` on the command line of each WebLogic Server. This flag suppresses insecure file system warnings that are reported in the WebLogic Server console when the WebLogic Server is in production mode. These warnings result from setting the file permissions that are necessary to work with restricted security context constraints on OpenShift.

For more information about the security requirements for running WebLogic in OpenShift, see the [OpenShift chapter]({{<relref "/security/openshift.md">}}) in the Security section.

Example:
```yaml
kubernetesPlatform: Openshift
```

documentation/staging/content/security/openshift.md
```diff
@@ -116,3 +116,6 @@ see [OpenShift]({{<relref  "/userguide/platforms/environments#openshift">}}).
 #### Using a dedicated namespace
 
 When the user that installs an individual instance of the operator does not have the required privileges to create resources at the Kubernetes cluster level, a dedicated namespace can be used for the operator instance and all the WebLogic domains that it manages. For more details about the `dedicated` setting, please refer to [Operator Helm configuration values]({{< relref "/userguide/managing-operators/using-helm#operator-helm-configuration-values" >}}).
+
+#### Set the Helm chart property `kubernetesPlatorm` to `OpenShift`
+Beginning with operator version 4.0, set the operator `kubernetesPlatform` Helm chart property to `OpenShift`. This property accommodates OpenShift security requirements. For more information, see [Operator Helm configuration values]({{<relref "/userguide/managing-operators/using-helm#operator-helm-configuration-values">}}).
```
rjeberhard, line 121: I think that we should backport this change to 3.3.2. @mriccell, do you agree?

documentation/staging/content/userguide/platforms/environments.md
```diff
@@ -62,8 +62,9 @@ Operator 2.0.1+ is certified for use on OpenShift Container Platform 3.11.43+, w
 
 Operator 2.5.0+ is certified for use on OpenShift Container Platform 4.3.0+ with Kubernetes 1.16.2+.
 
-When using the operator in OpenShift, a security context constraint is required to ensure that WebLogic containers run with a UNIX UID that has the correct permissions on the domain file system.
-This could be either the `anyuid` SCC or a custom one that you define for user/group `1000`. For more information, see [OpenShift]({{<relref "/security/openshift.md">}}) in the Security section.
+To accommodate OpenShift security requirements:
+- For security requirements to run WebLogic in OpenShift, see the [OpenShift chapter]({{<relref "/security/openshift.md">}}) in the Security section.
+- Beginning with operator version 4.0, specify the `kubernetesPlatorm` Helm chart property with value `OpenShift`. For more information, see [Operator Helm configuration values]({{<relref "/userguide/managing-operators/using-helm#operator-helm-configuration-values">}}).
```
rjeberhard, line 67: As above, I think that this change should go in to 3.3.2.

ankedia: Changed to "Beginning with operator version 3.3.2".

documentation/staging/content/security/openshift.md
ankedia: Changed to "Beginning with operator version 3.3.2".

rjeberhard (approved):

PR #2538, opened 2021-09-17
Support for network changes starting with Istio v1.10 and greater

lennyphan:
Beginning with Istio v1.10, network changes were made by Istio where the proxy no longer redirects the traffic to the localhost interface, but instead forwards it to the network interface associated with the pod IP (e.g. eth0).  Since the operator has to support all versions of Istio, we are adding additional WebLogic network channels for binding to the localhost and pod IP network interfaces.  This will allow the WebLogic managed servers to continue to receive application traffic through either network interfaces.  During introspection, copies of the readiness probe and any custom WebLogic NAP's are created and listening addresses set to '127.0.0.1' so that the NAP's are bound to the localhost interface.  We continue to customize any existing defined NAPs and override the listening address by setting it to the defined service name (e.g. sample-domain1-managed-server${id}).

For DII and DomainOnPV domains, we generate additional NAP's in situational config files.  For example:

<d:network-access-point f:combine-mode="add">
      <d:name f:combine-mode="add">http-probe</d:name>
      <d:protocol f:combine-mode="add">http</d:protocol>
      <d:listen-address f:combine-mode="add">sample-domain1-admin-server.sample-domain1-ns</d:listen-address>
      <d:public-address f:combine-mode="add">sample-domain1-admin-server.sample-domain1-ns</d:public-address>
      <d:listen-port f:combine-mode="add">8888</d:listen-port>
      <d:http-enabled-for-this-protocol f:combine-mode="add">true</d:http-enabled-for-this-protocol>
      <d:outbound-enabled f:combine-mode="add">true</d:outbound-enabled>
      <d:enabled f:combine-mode="add">true</d:enabled>
    </d:network-access-point>
    <d:network-access-point f:combine-mode="add">
      <d:name f:combine-mode="add">http-probe-lh01</d:name>
      <d:protocol f:combine-mode="add">http</d:protocol>
      <d:listen-address f:combine-mode="add">127.0.0.1</d:listen-address>
      <d:public-address f:combine-mode="add">sample-domain1-admin-server.sample-domain1-ns</d:public-address>
      <d:listen-port f:combine-mode="add">8888</d:listen-port>
      <d:http-enabled-for-this-protocol f:combine-mode="add">true</d:http-enabled-for-this-protocol>
      <d:outbound-enabled f:combine-mode="add">true</d:outbound-enabled>
      <d:enabled f:combine-mode="add">true</d:enabled>
    </d:network-access-point>

Since container port names are limited to a maximum  of 15 characters, we truncate the names of any cloned network channel  to the first 10 characters and append the identifier '-lhNN'  (e.g 'http-probe-lh01')

Have successfully run all Istio integration tests on Kind cluster using Istio v1.7.3, v1.10.4, and v1.11.2

### DIFF
--- a/documentation/staging/content/userguide/istio/istio.md
+++ b/documentation/staging/content/userguide/istio/istio.md
@@ -146,9 +146,9 @@ ensures that the Istio sidecar is not injected into the introspector job's pods.
  
 Starting with Istio 1.10, the networking behavior was changed in that the proxy no longer redirects 
 the traffic to the localhost interface, but instead forwards it to the network interface associated
-to the pod's IP.  
+with the pod's IP.  
 
-To learn more about changes to Istio networking beginning with Istio 1.10, see [Upcoming networking changes in Istio 1.10](https://istio.io/latest/blog/2021/upcoming-networking-changes//).
+To learn more about changes to Istio networking beginning with Istio 1.10, see [Upcoming networking changes in Istio 1.10](https://istio.io/latest/blog/2021/upcoming-networking-changes/).
 
 In order to support Istio v1.10 and later, as well as previous releases, the
 operator will:
--- a/documentation/staging/content/userguide/istio/istio.md
+++ b/documentation/staging/content/userguide/istio/istio.md
@@ -142,6 +142,39 @@ If the WebLogic administration port is enabled on the Administration Server:
 Additionally, when Istio support is enabled for a domain, the operator
 ensures that the Istio sidecar is not injected into the introspector job's pods.
 
+#### Support for network changes in Istio v1.10 and later
+
+Prior to Istio release 1.10, the Istio Envoy proxy redirected all inbound traffic to the localhost
+network interface.  The network channels listed above, which the operator automatically added, 
+were bound to the localhost interface in order for WebLogic to receive the inbound traffic.  
+Starting with Istio 1.10, the networking behavior was changed in that the proxy no longer redirects 
+the traffic to the localhost interface, but instead forwards it to the network interface associated
+to the pod's IP.  In order to support Istio v1.10 and later, as well as pre-1.10 releases, the
+operator will:
+    
+* Add an additional WebLogic HTTP protocol network channel for the readiness probe that is bound to the localhost network interface.
+* Add additional WebLogic network channels, bound to the localhost network interface, for each defined custom network channel.  
+* Continue to automatically add the network channels described above in [How Istio-enabled domains differ from regular domains](#how-istio-enabled-domains-differ-from-regular-domains)
+
+When adding additional WebLogic network channels for the readiness probe and any defined custom channels,
+the name of the additional channel will be appended with '-lhNN', where NN represents
+a two digit value for uniqueness.
+
+For example, the additional WebLogic HTTP protocol network channel for the readiness probe would be
+defined as follows:
+
+|Name|Port|Listening address|Protocol|Exposed as a container port|
+|----|----|----|--------|-----|
+|`http-probe-lh01`|From configuration Istio `readinessPort` | `127.0.0.1` | `http`| No |
+
+As another example, for a custom WebLogic network channel defined as `T3Channel` with port `5556`
+and protocol `t3`, the additional channel would be defined as follows: 
+
+|Name|Port|Listening address|Protocol|Exposed as a container port|
+|----|----|----|--------|-----|
+|`T3Channel-lh01`| `5556` | `127.0.0.1` | `t3`| Yes |
+
+To learn more about changes to Istio networking beginning with Istio 1.10, see [Upcoming networking changes in Istio 1.10](https://istio.io/latest/blog/2021/upcoming-networking-changes//).
 
 ### Apply the Domain YAML file
 
--- a/documentation/staging/content/userguide/istio/istio.md
+++ b/documentation/staging/content/userguide/istio/istio.md
@@ -143,13 +143,14 @@ Additionally, when Istio support is enabled for a domain, the operator
 ensures that the Istio sidecar is not injected into the introspector job's pods.
 
 #### Support for network changes in Istio v1.10 and later
-
-Prior to Istio release 1.10, the Istio Envoy proxy redirected all inbound traffic to the localhost
-network interface.  The network channels listed above, which the operator automatically added, 
-were bound to the localhost interface in order for WebLogic to receive the inbound traffic.  
+ 
 Starting with Istio 1.10, the networking behavior was changed in that the proxy no longer redirects 
 the traffic to the localhost interface, but instead forwards it to the network interface associated
-to the pod's IP.  In order to support Istio v1.10 and later, as well as pre-1.10 releases, the
+to the pod's IP.  
+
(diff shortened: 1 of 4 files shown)


operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -99,6 +99,8 @@
 from utils import *
 from weblogic.management.configuration import LegalHelper
 
+ISTIO_NAP_NAMES = ['tcp-cbt', 'tcp-ldap', 'tcp-iiop', 'tcp-snmp', 'http-default', 'tcp-default', 'https-secure', 'tls-ldaps', 'tls-default', 'tls-cbts', 'tls-iiops', 'https-admin']
+WLS_LOCALHOST_IDENTIFIER = '-lh'
```
tbarnes-us, line 103: i'm not sure if the WLS_LOCALHOST_IDENTIFIER '-lh' value is unique enough for the `nameContainsLocalHostIdentifier` code below, which uses this value to distinguish localhost channels from other types of channels). 

just brain-storming, but maybe instead follow the pattern of other generated channels, where IIRC, we fail introspection if we detect a customer has used a reserved name/prefix?  we could start all localhost channels with 'lhs-' _and_ have a check if customer prefixed their own channel starting with 'lhs-'...

tbarnes-us: or maybe it'd be 'good enough' to advance the `nameContainsLocalHostIdentifier` to look for '-lhNN' instead of just '-lh'....

tbarnes-us: change LGTM

tbarnes-us (approved):

documentation/staging/content/userguide/istio/istio.md
```diff
@@ -142,6 +142,39 @@ If the WebLogic administration port is enabled on the Administration Server:
 Additionally, when Istio support is enabled for a domain, the operator
 ensures that the Istio sidecar is not injected into the introspector job's pods.
 
+#### Support for network changes in Istio v1.10 and later
+
+Prior to Istio release 1.10, the Istio Envoy proxy redirected all inbound traffic to the localhost
+network interface.  The network channels listed above, which the operator automatically added,
```
rjeberhard, line 148: Let's link to the article on this change.

documentation/staging/content/userguide/istio/istio.md
```diff
@@ -142,6 +142,39 @@ If the WebLogic administration port is enabled on the Administration Server:
 Additionally, when Istio support is enabled for a domain, the operator
 ensures that the Istio sidecar is not injected into the introspector job's pods.
 
+#### Support for network changes in Istio v1.10 and later
+
+Prior to Istio release 1.10, the Istio Envoy proxy redirected all inbound traffic to the localhost
```
rjeberhard, line 147: I don't think the "prior to" paragraph is necessary and you can just begin with "Starting with Istio 1.10". Also, let's link to that article describing the change.

documentation/staging/content/userguide/istio/istio.md
```diff
@@ -142,6 +142,39 @@ If the WebLogic administration port is enabled on the Administration Server:
...
+network interface.  The network channels listed above, which the operator automatically added, 
+were bound to the localhost interface in order for WebLogic to receive the inbound traffic.  
+Starting with Istio 1.10, the networking behavior was changed in that the proxy no longer redirects 
+the traffic to the localhost interface, but instead forwards it to the network interface associated
+to the pod's IP.  In order to support Istio v1.10 and later, as well as pre-1.10 releases, the
```
rjeberhard, line 152: "pre-1.10 releases" -> "previous releases"

documentation/staging/content/userguide/istio/istio.md
rjeberhard: I see that you did link the article later. Move the link up here :)

documentation/staging/content/userguide/istio/istio.md
```diff
@@ -142,6 +142,39 @@ If the WebLogic administration port is enabled on the Administration Server:
...
+|`http-probe-lh01`|From configuration Istio `readinessPort` | `127.0.0.1` | `http`| No |
+
+As another example, for a custom WebLogic network channel defined as `T3Channel` with port `5556`
+and protocol `t3`, the additional channel would be defined as follows: 
+
+|Name|Port|Listening address|Protocol|Exposed as a container port|
+|----|----|----|--------|-----|
+|`T3Channel-lh01`| `5556` | `127.0.0.1` | `t3`| Yes |
+
+To learn more about changes to Istio networking beginning with Istio 1.10, see [Upcoming networking changes in Istio 1.10](https://istio.io/latest/blog/2021/upcoming-networking-changes//).
```
jshum2479, line 177: Does the user need to do anything if upgrading istio from 1.9 to 1.10 or higher,  does he/she needs to re-run the introspector by updating the introspectVersion or something else?

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -1056,6 +1059,221 @@ def writeListenAddress(self, originalValue, newValue):
...
+  def writeProtocol(self, protocol):
+    self.writeln("<d:protocol f:combine-mode=\"add\">" + protocol + "</d:protocol>")
+
+  def writeListenPort(self, listen_port):
+    # offline WLST initializes int values to 0
+    if listen_port > 0:
+      self.writeln("<d:listen-port f:combine-mode=\"add\">" + str(listen_port) + "</d:listen-port>")
+
+  def writePublicAddress(self, public_address):
+    self.writeln("<d:public-address f:combine-mode=\"add\">" + public_address + "</d:public-address>")
```
jshum2479, line 1074: Will public_address be None and can sitconfig take None at this stanza (vs not set) ?

documentation/staging/content/userguide/istio/istio.md
lennyphan: This is a very good question.   With this change, we can support all versions of Istio and so  I don't believe re-running introspection would make much difference or is necessary.  Do we have a documented process for upgrading Istio in a cluster?  That is, do we recommend shutting down the domain before upgrading Istio or can you upgrade Istio while a domain is running?

documentation/staging/content/userguide/istio/istio.md
lennyphan: Changed.  Moved the link up.

documentation/staging/content/userguide/istio/istio.md
lennyphan: Changed.

operator/src/main/resources/scripts/introspectDomain.py
lennyphan: MBean method for setting public address can handle null when looking at the generated MBean method implementation but added a conditional check.  Please review.

documentation/staging/content/userguide/istio/istio.md
lennyphan: Also, removed the text describing the previous Istio network behavior.

rjeberhard (approved):

PR #2580, opened 2021-10-20
OWLS-91212 - Fix for the introspector retry behavior after the job times out. 

ankedia:
This change is to fix the introspector retry behavior following a introspection job time out. We changed the logic for incrementing the introspector job failure count. Additionally, we also changed the logic for when to create a new introspection job if a job already exists as part of OWLS-90180 as job might have been created before the operator started. 

I noticed that we currently rely on the failure count in domain status when introspector fails with SEVERE error in logs. However, when the introspector job times out, we were relying on the retryCount (failure count) stored in DPI. I have changed to use failure count in domain status for all types of failures. 

Fixed couple of NPEs - (1) when the domain resource had a missing secret and (2) when callResponse.getRequestParams() was null, the call to print the ASYNC_NO_RETRY log message was generating NPE.

Finally, this PR also has changes to capture the WDT log files to the LOG_HOME directory and a potential fix for intermittent failure in domain status integration test. Integration tests reesults are at -
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6899/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6884/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -1170,7 +1170,8 @@ public class DomainProcessorImpl implements DomainProcessor {
             bringAdminServerUp(info, delegate.getPodAwaiterStepFactory(info.getNamespace())),
             managedServerStrategy);
 
-    if (hasExceededRetryCount(info)) {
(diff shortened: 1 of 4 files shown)


ankedia:
Created https://jira.oraclecorp.com/jira/browse/OWLS-93380 to track the change to the `ItKubernetesEvents.testDomainK8SEventsFailed` integration test.

PR #2583, opened 2021-10-21
trimming the use cases that are covered in other tests

vanajamukkara:
Removing testMiiAddDynmicClusteriWithNoReplica and testMiiAddConfiguredCluster as they have the same coverage as testMiiAddDynmicCluster for MII offline update. Also we have tests in ItServerStartPolicy for configured cluster.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6814/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/6853/ (Full suite)


vanajamukkara:
> Remove the unused method private void checkPodNotCreated() Remove model.config.cluster.yaml from wdt-models ( if not being used by any other class )

Done, yaml is not used in any other classes, deleted. Thanks

PR #2599, opened 2021-11-01
Add unit tests for JobHelper changes in OWLS-92894

lennyphan:
Added unit tests for new Istio related environment variables for introspector job.

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
@@ -1049,9 +1049,8 @@ class JobHelperTest extends DomainValidationBaseTest {
   }
 
   @Test
-  void whenDomainIsIstioEnabled_localhostBindingsDisable_hasIstioUseLocalhostBindingsEnv() {
-    configureDomain()
-        .withIstioLocalhostBindingsEnabled(false);
+  void whenDomainIsIstioEnabled_localhostBindingsDisabled_hasIstioUseLocalhostBindingsEnv() {
+    configureDomain().withIstioLocalhostBindingsEnabled(false);
 
     V1JobSpec jobSpec = createJobSpec();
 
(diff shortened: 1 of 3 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -1045,6 +1048,87 @@ void whenIntrospectRequestSet_runIntrospector() {
     assertThat(job, notNullValue());
   }
 
+  @Test
+  void whenDomainIsIstioEnabled_localhostBindingsDisable_hasIstioUseLocalhostBindingsEnv() {
+    configureDomain()
```
russgold, line 1053: There is no need to break this on to a second line; that only makes sense when there are multiple calls that make it impossible to be a single line.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
```diff
@@ -1045,6 +1048,87 @@ void whenIntrospectRequestSet_runIntrospector() {
...
+
+    assertThat(
+        getMatchingContainerEnv(domainPresenceInfo, jobSpec),
+        not(hasEnvVar(ISTIO_USE_LOCALHOST_BINDINGS, "false"))
+    );
+  }
+
+  @Test
+  void whenDomainIsIstioEnabled_replicationChannelPort_hasIstioEnvVar() {
+    String port = "6789";
```
russgold, line 1079: Breaking out a constant is a good idea; follow through:

- declare the variable final
- write its name in all caps
- use it for all cases in the method rather than hard-coding the same value in the assertion

russgold (changes requested):
Overall good. There are some stylistic issues. It is generally easier to read single-line statements than multiple-line ones. The latter should be reserved for cases when there is not enough room on the line.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
lennyphan: Fixed to be on single line.

operator/src/test/java/oracle/kubernetes/operator/helpers/JobHelperTest.java
lennyphan: Fixed as a constant.

russgold (approved):
LGTM

rjeberhard (approved):

PR #2601, opened 2021-11-01
Owls 93146 - fix infinite introspector retries

russgold:
Correct integration job failure-count functionality.

Includes substantial refactoring of JobHelper.

[Integration test run](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7078/)

Note: this run has two failures:

- ItLivenessProbeCustomization.testCustomLivenessProbeFailureThresholdSuccessThreshold, which [passes when run again](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7084/), and
- ItProductionSecureMode.testVerifyProductionSecureMode, which [fails in main](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7086/).


ankedia (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -31,134 +30,79 @@
...
-
-  private static boolean checkIfIntrospectionRequestedAndReset(Packet packet) {
-    return packet.remove(DOMAIN_INTROSPECT_REQUESTED) != null;
-  }
-
-  private static boolean isIntrospectVersionChanged(Packet packet, DomainPresenceInfo info) {
-    return Optional.ofNullable(packet.get(INTROSPECTION_STATE_LABEL))
-            .map(introspectVersionLabel -> !introspectVersionLabel.equals(getIntrospectVersion(info))).orElse(false);
-  }
+  //----------- for unit testing only ---------
```
jshum2479, line 67: Is there a way to name these unit test methods so that it will be easier to recognize they are for unit test only?

tbarnes-us:
Does this include a merge that it shouldn't? For example, there are doc changes that look related to unrelated work.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -56,6 +56,9 @@
   @Range(minimum = 0)
   private Integer introspectJobFailureCount = 0;
 
+  @Description("Unique id of the last failed introspector job.")
```
rjeberhard, line 59: id -> ID
introspector -> introspection

rjeberhard (approved):

russgold:
> Does this include a merge that it shouldn't? For example, there are doc changes that look related to unrelated work.

I don't think so. I had to do a couple of merges from main, where there were doc changes, but that's it. Are there specific changes you're wondering about?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: I'm open to suggestions, but most of them make specific reference to the UnitTestAdapter.

rjeberhard:
> >    Does this include a merge that it shouldn't? For example, there are doc changes that look related to unrelated work.

> I don't think so. I had to do a couple of merges from main, where there were doc changes, but that's it. Are there specific changes you're wondering about?

We have a few files, such as the documentation/domains directory and the CRD, that are generated during the build. We likely shouldn't be doing this, but Russ doesn't have to fix that as part of this PR. If you can merge the latest from main the I believe that we can merge this PR.

rjeberhard:
@russgold, any idea how to resolve this conflict in JobHelperTest?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
jshum2479: I am concerned if someone trying to use these methods based on the names along later maybe confused until reading the actual code.  If the three methods cannot be encapsulated in UnitTestAdapter, maybe just rename with UnitTest appended to the end of the name.

russgold:
> @russgold, any idea how to resolve this conflict in JobHelperTest?

Yes, it's done. Somebody else added tests.

PR #2607, opened 2021-11-04
Enhance fiber breadcrumbs

russgold:
This adds some additional debugging capabilities for fibers:

1. When returning a `NextAction` it is now possible to add some static or computed information:

```
return doNext(getNext(), packet).withDebugComment(() -> "assume presence up-to-date");
```

indicates that the fiber has chosen a path in which the domain presence is considered to be up-to-date

```
return doNext(createIntrospectionSteps(), packet)
                    .withDebugComment(packet, this::introspectionNeededReason);
```
indicates that the fiber has decided that it needs to run an introspection, along with its reasoning, derived from the packet.

2. In each case, the comment is actually derived from a method, which is not invoked unless the feature is enabled for the current fiber (intended to avoid excess computation). That means that it may safely be left in the code.

3. To enable the feature, the code must add a string to the packet, using the key Fiber.DEBUG_FIBER

4. When the fiber terminates, it will be logged, along with the value of the Fiber.DEBUG_FIBER entry as a prefix.

Note that, at present, the feature is disabled unless specifically enabled in the code.

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
@@ -487,13 +487,18 @@ public class Kubernetes {
             null, // Timeout for the list/watch call.
             Boolean.FALSE // Watch for changes to the described resources.
         );
+    getLogger().info("REG-> found {0} pods in namespace {1}", v1PodList.getItems().size(), namespace);
     for (V1Pod item : v1PodList.getItems()) {
+      getLogger().info("REG-> comparing actual name {0} against specified pod name {1}",
+            item.getMetadata().getName(), podName);
       if (item.getMetadata().getName().contains(podName.trim())) {
         getLogger().info("Name: {0}, Namespace: {1}, Phase: {2}",
             item.getMetadata().getName(), namespace, item.getStatus().getPhase());
+        getLogger().info("REG-> found a match");
         return item;
       }
     }
+    getLogger().info("REG-> did not find a match");
     return null;
   }
 
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
@@ -487,18 +487,13 @@ public class Kubernetes {
             null, // Timeout for the list/watch call.
             Boolean.FALSE // Watch for changes to the described resources.
         );
-    getLogger().info("REG-> found {0} pods in namespace {1}", v1PodList.getItems().size(), namespace);
     for (V1Pod item : v1PodList.getItems()) {
-      getLogger().info("REG-> comparing actual name {0} against specified pod name {1}",
-            item.getMetadata().getName(), podName);
       if (item.getMetadata().getName().contains(podName.trim())) {
         getLogger().info("Name: {0}, Namespace: {1}, Phase: {2}",
             item.getMetadata().getName(), namespace, item.getStatus().getPhase());
-        getLogger().info("REG-> found a match");
         return item;
       }
     }
-    getLogger().info("REG-> did not find a match");
     return null;
   }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -72,6 +72,7 @@ import oracle.kubernetes.weblogic.domain.model.DomainSpec;
 import oracle.kubernetes.weblogic.domain.model.DomainStatus;
 import oracle.kubernetes.weblogic.domain.model.ServerHealth;
 import oracle.kubernetes.weblogic.domain.model.ServerStatus;
+import org.jetbrains.annotations.NotNull;
 
 import static oracle.kubernetes.operator.DomainStatusUpdater.createStatusUpdateStep;
 import static oracle.kubernetes.operator.LabelConstants.INTROSPECTION_STATE_LABEL;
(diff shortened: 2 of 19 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -262,6 +269,36 @@ private boolean isIntrospectionNeeded(Packet packet) {
...
+        } else if (isBringingUpNewDomain(packet)) {
+          sb.append("bringing up new domain");
+        } else {
+          sb.append("something else");
+        }
+        return sb.toString();
+      }
+
+      private String introspectionNotNeededReason(Packet packet) {
+        StringBuilder sb = new StringBuilder("introspection not needed because ");
```
rjeberhard, line 285: How do we use these strings? I'm fine with using non-internationalizable strings for debug messages, but not for anything used in INFO for greater logs.

rjeberhard:
@ankedia, I'll in particular be looking for your review on this since I'm concerned that this will overlap with the changes you were working on that we discussed the other day.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
rjeberhard: Ignore... I see where you are using these messages. I wouldn't have asked if ™️ the work "Debug" was in the method name.

rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: I built this to debug fiber actions, and currently the only way to enable it is to change the code to add a prefix, such as
```
    @Override
      packet.put(Fiber.DEBUG_FIBER, "REG-> (CRD)");
      return doNext(context.verifyCrd(getNext()), packet);
    }
```

which I had added to CrdHelper. There is not any current way for a user to enable it, and it would make little sense to somebody without access to the source code in any event. That said, I have a longer-term idea that might make it usable for certain end-users, and if we went in that direction, internationalization would absolutely make sense.

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
```diff
@@ -487,13 +487,18 @@ public static V1Pod getPod(String namespace, String labelSelector, String podNam
             null, // Timeout for the list/watch call.
             Boolean.FALSE // Watch for changes to the described resources.
         );
+    getLogger().info("REG-> found {0} pods in namespace {1}", v1PodList.getItems().size(), namespace);
```
ankedia, line 490: Should `REG` (seems like initials) be something like `DEBUG`?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -976,6 +977,116 @@ public Step createSteps() {
...
+      FiberGate gate = domainProcessor.getMakeRightFiberGate(getNamespace());
+      CompletionCallback cc =
+          new CompletionCallback() {
+            @Override
+            public void onCompletion(Packet packet) {
+              // no-op
+            }
+
+            @Override
+            public void onThrowable(Packet packet, Throwable throwable) {
```
ankedia, line 992: FYI - I have changes in the same area in my PR 2609 which might conflict with these changes.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
```diff
@@ -574,11 +574,12 @@ private void resetFailureCount() {
     retryCount.set(0);
   }
 
-  public int incrementAndGetFailureCount() {
-    return retryCount.incrementAndGet();
+  public void indicateFailure() {
```
ankedia, line 577: I have removed the retryCount from DPI in PR 2609 and use the failure count in the domain status instead.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -217,14 +218,20 @@ public NextAction onSuccess(Packet packet, CallResponse<T> callResponse) {
         if (isKnownFailedJob(job)) {
           return doNext(cleanUpAndReintrospect(), packet);
         } else if (job != null) {
-          return doNext(processIntrospectionResults(), packet);
+          return doNext(processIntrospectionResults(), packet).withDebugComment(job, this::foundJobDescription);
```
ankedia, line 221: The logic has changed in PR 2609 to account for failures due to image pull error and introspector job timeout.

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
russgold: No, they _are_ my initials, and are supposed to remind me to take that line out before putting it in a PR. Oops.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: Yes, and I'll have to deal with that.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainPresenceInfo.java
russgold: Understood.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: Understood.

russgold:
Replaced by #2643

PR #2612, opened 2021-11-05
fix olcne test failures

xiancao:
In olcne Jenkins job, the test ItMiiAuxiliaryImage fails when run it in the whole profile. However the test passes when run in single class. The issue is that the image name in ItMiiAuxiliaryImage is the same as in ItMiiAuxiliaryImageCluster. Update the image name to be unique. Also update the image path to be unique to the test class. 

Jenkins result:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7129/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7128/


anpanigr (approved):
LGTM

maggiehe00 (approved):

PR #2613, opened 2021-11-05
Owls 91212 - Merge fixes for the introspector retry behavior after the job times out and to capture WDT logs. 

ankedia:
This PR is for merging changes from PR 2580 to main. It has below changes -

- Fix for the introspector retry behavior following a introspection job time out.
- Fix for NPE when the domain resource has a missing secret
- Changes to capture the WDT log files to the LOG_HOME directory to facilitate in debugging introspector timeout.
- Improve error message when introspector job times out with suggestion to increase the timeout value.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -841,7 +841,7 @@ public class DomainProcessorImpl implements DomainProcessor {
         return true;
       } else if (shouldReportAbortedEvent()) {
         return true;
-      } else if (hasExceededRetryCount() && !isImgRestartIntrospectVerChanged(liveInfo, cachedInfo)) {
+      } else if (hasExceededRetryCount(liveInfo) && !isImgRestartIntrospectVerChanged(liveInfo, cachedInfo)) {
         LOGGER.severe(ProcessingConstants.EXCEEDED_INTROSPECTOR_MAX_RETRY_COUNT_ERROR_MSG);
         return false;
       } else if (isFatalIntrospectorError()) {
@@ -852,10 +852,7 @@ public class DomainProcessorImpl implements DomainProcessor {
         return false;  // we have already cached this
       } else if (shouldRecheck(cachedInfo)) {
 
-        if (hasExceededRetryCount()) {
-          resetIntrospectorJobFailureCount();
-        }
-        if (getCurrentIntrospectFailureRetryCount() > 0) {
+        if (getCurrentIntrospectFailureRetryCount(liveInfo) > 0) {
           logRetryCount(cachedInfo);
         }
         LOGGER.fine("Continue the make-right domain presence, explicitRecheck -> " + explicitRecheck);
@@ -869,29 +866,9 @@ public class DomainProcessorImpl implements DomainProcessor {
       return Optional.ofNullable(eventData).map(EventData::getItem).orElse(null) == DOMAIN_PROCESSING_ABORTED;
     }
 
-    private void resetIntrospectorJobFailureCount() {
-      Optional.ofNullable(liveInfo)
-          .map(DomainPresenceInfo::getDomain)
-          .map(Domain::getStatus)
-          .map(DomainStatus::resetIntrospectJobFailureCount);
-    }
-
-    private boolean hasExceededRetryCount() {
-      return getCurrentIntrospectFailureRetryCount()
-          >= DomainPresence.getDomainPresenceFailureRetryMaxCount();
-    }
-
-    private Integer getCurrentIntrospectFailureRetryCount() {
-      return Optional.ofNullable(liveInfo)
-          .map(DomainPresenceInfo::getDomain)
-          .map(Domain::getStatus)
-          .map(DomainStatus::getIntrospectJobFailureCount)
-          .orElse(0);
-    }
-
     private void logRetryCount(DomainPresenceInfo cachedInfo) {
       LOGGER.info(MessageKeys.INTROSPECT_JOB_FAILED_RETRY_COUNT, cachedInfo.getDomain().getDomainUid(),
-          getCurrentIntrospectFailureRetryCount(),
+          getCurrentIntrospectFailureRetryCount(liveInfo),
           DomainPresence.getDomainPresenceFailureRetryMaxCount());
     }
 
@@ -1020,6 +997,19 @@ public class DomainProcessorImpl implements DomainProcessor {
         .orElse(null);
   }
 
+  private Integer getCurrentIntrospectFailureRetryCount(DomainPresenceInfo info) {
+    return Optional.ofNullable(info)
+            .map(DomainPresenceInfo::getDomain)
+            .map(Domain::getStatus)
+            .map(DomainStatus::getIntrospectJobFailureCount)
+            .orElse(0);
+  }
+
+  private boolean hasExceededRetryCount(DomainPresenceInfo info) {
+    return getCurrentIntrospectFailureRetryCount(info)
+            >= DomainPresence.getDomainPresenceFailureRetryMaxCount();
+  }
+
   private static boolean isCachedInfoNewer(DomainPresenceInfo liveInfo, DomainPresenceInfo cachedInfo) {
     return liveInfo.getDomain() != null
         && KubernetesUtils.isFirstNewer(cachedInfo.getDomain().getMetadata(), liveInfo.getDomain().getMetadata());
@@ -1048,7 +1038,8 @@ public class DomainProcessorImpl implements DomainProcessor {
             gate.startFiberIfLastFiberMatches(
                 domainUid,
                 Fiber.getCurrentIfSet(),
-                DomainStatusUpdater.createFailureRelatedSteps(throwable),
+                Step.chain(DomainStatusUpdater.createFailureCountStep(null),
+                        DomainStatusUpdater.createFailureRelatedSteps(throwable)),
                 plan.packet,
                 new CompletionCallback() {
                   @Override
@@ -1071,7 +1062,7 @@ public class DomainProcessorImpl implements DomainProcessor {
                                  LoggingContext.setThreadContext().namespace(ns).domainUid(domainUid)) {
                           existing.setPopulated(false);
                           // proceed only if we have not already retried max number of times
-                          int retryCount = existing.incrementAndGetFailureCount();
+                          int retryCount = getCurrentIntrospectFailureRetryCount(existing);
                           LOGGER.fine(
                               "Failure count for DomainPresenceInfo: "
                                   + existing
@@ -1128,6 +1119,11 @@ public class DomainProcessorImpl implements DomainProcessor {
             bringAdminServerUp(info, delegate.getPodAwaiterStepFactory(info.getNamespace())),
             managedServerStrategy);
 
+    if (hasExceededRetryCount(info) && isImgRestartIntrospectVerChanged(info,
+            getExistingDomainPresenceInfo(info.getNamespace(), info.getDomainUid()))) {
+      domainUpStrategy = Step.chain(DomainStatusUpdater.createResetFailureCountStep(), domainUpStrategy);
+    }
+
     return Step.chain(
           createDomainUpInitialStep(info),
           ConfigMapHelper.readExistingIntrospectorConfigMap(info.getNamespace(), info.getDomainUid()),
@@ -1174,8 +1170,7 @@ public class DomainProcessorImpl implements DomainProcessor {
 
     @Override
     public NextAction apply(Packet packet) {
-      packet.getSpi(DomainPresenceInfo.class).complete();
-      return doNext(packet);
+      return doNext(DomainStatusUpdater.createResetFailureCountStep(), packet);
     }
   }
 
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -164,36 +164,33 @@ public class DomainStatusUpdater {
   static class FailureCountStep extends DomainStatusUpdaterStep {
 
     private final V1Job domainIntrospectorJob;
-    private final boolean resetFailureCount;
 
     public FailureCountStep(V1Job domainIntrospectorJob) {
-      this(domainIntrospectorJob, false);
-    }
-
-    public FailureCountStep(V1Job domainIntrospectorJob, boolean resetFailureCount) {
       super(null);
       this.domainIntrospectorJob = domainIntrospectorJob;
-      this.resetFailureCount = resetFailureCount;
     }
 
     @Override
     void modifyStatus(DomainStatus domainStatus) {
-      if (resetFailureCount) {
-        domainStatus.resetIntrospectJobFailureCount();
-      } else {
-        domainStatus.incrementIntrospectJobFailureCount(getJobUid());
-      }
+      domainStatus.incrementIntrospectJobFailureCount(getJobUid());
     }
 
     @Nullable
     private String getJobUid() {
-      return Optional.ofNullable(domainIntrospectorJob).map(V1Job::getMetadata)
-              .map(V1ObjectMeta::getUid).orElse(null);
+      return Optional.ofNullable(domainIntrospectorJob).map(V1Job::getMetadata).map(V1ObjectMeta::getUid).orElse(null);
+    }
+  }
+
+  static class ResetFailureCountStep extends DomainStatusUpdaterStep {
+
+    @Override
+    void modifyStatus(DomainStatus domainStatus) {
+      domainStatus.resetIntrospectJobFailureCount();
     }
   }
 
   public static Step createResetFailureCountStep() {
-    return new FailureCountStep(null, true);
+    return new ResetFailureCountStep();
   }
 
   private static String getEventMessage(@Nonnull DomainFailureReason reason, String message) {
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -164,21 +164,36 @@ public class DomainStatusUpdater {
   static class FailureCountStep extends DomainStatusUpdaterStep {
 
     private final V1Job domainIntrospectorJob;
+    private final boolean resetFailureCount;
 
-    public FailureCountStep(@Nonnull V1Job domainIntrospectorJob) {
+    public FailureCountStep(V1Job domainIntrospectorJob) {
+      this(domainIntrospectorJob, false);
+    }
+
+    public FailureCountStep(V1Job domainIntrospectorJob, boolean resetFailureCount) {
       super(null);
       this.domainIntrospectorJob = domainIntrospectorJob;
+      this.resetFailureCount = resetFailureCount;
     }
 
     @Override
     void modifyStatus(DomainStatus domainStatus) {
-      domainStatus.incrementIntrospectJobFailureCount(getJobUid());
+      if (resetFailureCount) {
+        domainStatus.resetIntrospectJobFailureCount();
+      } else {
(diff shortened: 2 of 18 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -164,23 +164,38 @@ public static Step createFailureCountStep(V1Job domainIntrospectorJob) {
   static class FailureCountStep extends DomainStatusUpdaterStep {
 
     private final V1Job domainIntrospectorJob;
+    private final boolean resetFailureCount;
```
russgold, line 167: You have changed this step to do one of two things, based on the value of a flag. That is generally poor practice. It would be better simply to add a new step to reset the failure count, if that is needed, and leave the original step alone.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -869,29 +866,9 @@ private boolean shouldReportAbortedEvent() {
       return Optional.ofNullable(eventData).map(EventData::getItem).orElse(null) == DOMAIN_PROCESSING_ABORTED;
     }
 
-    private void resetIntrospectorJobFailureCount() {
```
russgold, line 872: Main already has functionality for resetting the failure count. Why discard it in favor of the approach used in 3.3.3?

operator/src/main/java/oracle/kubernetes/operator/steps/HttpRequestProcessing.java
```diff
@@ -58,12 +58,14 @@ AuthorizationSource getAuthorizationSource() {
...
 
   final HttpRequest.Builder createRequestBuilder(String url) {
-    return HttpRequest.newBuilder()
+    HttpRequest.Builder builder = HttpRequest.newBuilder()
           .uri(URI.create(url))
-          .header("Authorization", getAuthorizationSource().createBasicAuthorizationString())
           .header("Accept", "application/json")
           .header("Content-Type", "application/json")
           .header("X-Requested-By", "WebLogic Operator");
+    Optional.ofNullable(getAuthorizationSource())
```
russgold, line 66: What problem led to this change? Is there a unit test for it?

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -214,21 +220,90 @@ public NextAction onSuccess(Packet packet, CallResponse<T> callResponse) {
...
+        public NextAction onSuccess(Packet packet, CallResponse<V1PodList> callResponse) {
+          V1Pod pod = getJobPod(callResponse);
+          V1Job job = (V1Job) packet.get(ProcessingConstants.DOMAIN_INTROSPECTOR_JOB);
+          boolean hasImagePullError = checkJobPodForImagePullError(pod);
+          if (hasImagePullError) {
+            packet.put(DOMAIN_INTROSPECT_REQUESTED, ReadPodLogResponseStep.INTROSPECTION_FAILED);
+          }
+
+          if (isKnownFailedJob(job) || hasImagePullError) {
+            return doContinueListOrNext(callResponse, packet, cleanUpAndReintrospect());
```
russgold, line 252: doContinueListOrNext is intended for list operations. Why are you using it here?

russgold (changes requested):
I have left a number of detailed comments; my biggest concern, though, is that this change seems to replace something that works in main with a very different implementation in 3.3, and I don't understand why. It maybe an artifact of the forward-porting process, which is why it is generally best to start by porting only the tests and then let them guide you as to which other changes are needed.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
```diff
@@ -1404,6 +1424,7 @@ private void forceExceptionDuringProcessing() {
   }
 
   @Test
+  @Disabled
```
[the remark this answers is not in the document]
ankedia: The test is not valid since it tries to force an exception by generating a NullPointerException by removing WebLogic  credentials secret. The `WebLogic  credentials secret` is a mandatory field and the domain validation should fail when this field is missing. We had a bug where missing secret would generate the NPE and I fixed that bug. I have another test where we force the exception when an introspector job times out. I wanted to discuss this with you and Ryan to see if we need another test for forcing exception or if this can be removed the test completely.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorDelegateStub.java
```diff
@@ -99,6 +101,9 @@ public Step waitForDelete(V1Pod pod, Step next) {
   private class PassthroughJobAwaiterStepFactory implements JobAwaiterStepFactory {
     @Override
     public Step waitForReady(V1Job job, Step next) {
+      if (isFailed(job) && "DeadlineExceeded".equals(getFailedReason(job))) {
```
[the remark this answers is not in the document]
ankedia: I wanted the unit test to replicate the behavior where an introspector job fails due to timeout (i.e. job failed reason is DeadlineExceeded) and the operator throws an exception. I'll change the class name or create another class to avoid the confusion.

operator/src/main/java/oracle/kubernetes/operator/steps/HttpRequestProcessing.java
ankedia: This change is also to avoid the NullPointerException when WebLogicCredentialsSecret is removed. I didn't know whether we need unit test when fixing the NPE. I'll look into adding unit test.

operator/src/main/java/oracle/kubernetes/operator/helpers/SecretHelper.java
```diff
@@ -62,10 +62,13 @@ public NextAction apply(Packet packet) {
         secretName = dpi.getDomain().getWebLogicCredentialsSecretName();
         namespace = dpi.getNamespace();
 
-        LOGGER.fine(MessageKeys.RETRIEVING_SECRET, secretName);
-        Step read = new CallBuilder().readSecretAsync(secretName, namespace, new SecretResponseStep(getNext()));
-
-        return doNext(read, packet);
+        if (secretName != null) {
```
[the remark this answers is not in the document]
ankedia: This change is to avoid the NullPointerException when WebLogicCredentialsSecret is removed. I didn't know whether we need unit test when fixing a NPE. I'll look into adding unit test.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
```diff
@@ -458,7 +458,7 @@ public Domain withStatus(DomainStatus status) {
    * @return the secret name
    */
   public String getWebLogicCredentialsSecretName() {
```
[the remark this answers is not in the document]
ankedia: This change is also to avoid the NullPointerException when WebLogicCredentialsSecret is removed. I didn't know whether we need unit test when fixing the NPE. I'll look into adding unit test and make the change use `V1SecretReference::getName`.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
ankedia: This method is resetting the failure count in the domain status. I believe this approach is same in latest 3.3.3 release. Previously we had 2 failure/retry counts (1) in the DomainStatus and (2) in-memory count in DomainPresenceInfo. After the fix for OWLS-90180, we can no longer rely on the in-memory state of the operator as the introspector job might have been created before the operator started. I have removed the in-memory retry count in DomainPresenceInfo and made changes to use the failure count in domain status instead.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
ankedia: ok. I had originally created two steps but later thought that I can reuse the same method and put additional logic in it. I'll change it back to add a new step for resetting failure count and keep the original step same.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
ankedia: The `onSuccess` method is for listing the introspector job pod. Since we don't know the exact name of the introspector job pod, we list the pods having `job-name` label. The list pods operation returns pods in multiple chunks and it's possible that introspector job pod is not returned in first chunk. My understanding is that `doContinueListOrNext` method will either return the next chunk or perform the next action.

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -137,7 +137,12 @@ public static boolean isComplete(V1Job job) {
     return false;
   }
 
-  static boolean isFailed(V1Job job) {
+  /**
```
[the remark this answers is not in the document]
ankedia: Fixed

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -295,7 +295,7 @@ public Integer getIntrospectJobFailureCount() {
    * @param uid the Kubernetes-assigned UID of the job which discovered the introspection failure
    */
   public void incrementIntrospectJobFailureCount(String uid) {
-    if (!Objects.equals(uid, failedIntrospectionUid)) {
+    if ((uid == null) || (!Objects.equals(uid, failedIntrospectionUid))) {
```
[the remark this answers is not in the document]
ankedia: This change is for the case when the introspector job times out or fiber encounters an exception during processing. I had added a new unit test `whenIntrospectionJobTimedout_failureCountIncremented` which covers this change. As you have suggested, I'll extract the `if` into it's own method.

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
```diff
@@ -591,6 +600,87 @@ void whenPreviousFailedJobExists_createNewJob() {
...
+    testSupport.defineResources(asTimedoutJob(createIntrospectorJob("TIMEDOUT_JOB")));
+  }
+
+  private V1Job asTimedoutJob(V1Job job) {
+    job.setStatus(new V1JobStatus().addConditionsItem(new V1JobCondition().status("True").type("Failed")
+            .reason("DeadlineExceeded")));
+    return job;
+  }
+
+  private long getIntrospectorJobActiveDeadlineSeconds() {
```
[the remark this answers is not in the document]
ankedia: Thanks, I'll delete it.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: That would be a concern if we expected to find hundreds of active pods with that name, which would shock me. Did you run into a problem that prompted this change?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: I hear the reasoning; I'm just surprised by the size of the change needed to do that.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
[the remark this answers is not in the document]
russgold: It is best to add a unit test when fixing any problem; it means that we didn't think of that case when the code was originally written, or else that case is new.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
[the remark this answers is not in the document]
russgold: If the system now prevents us from ever removing the secret, we should not need a test for it. If it doesn't, we still need to test that we handle that case.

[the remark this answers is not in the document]
ankedia: If the secret is missing during initial domain creation, the operator will generate a domain validation error. If the secret is removed after the domain is up and running, we no longer throw a NullPointerException. The REST API to read server health will fail due to missing credentials until the next make-right operation. The domain validation during the next make-right cycle will catch and report a validation error.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
ankedia: I was following the same approach that we use in the `PodListResponseStep` on line 637 of JobHelper which is called from `ReadDomainIntrospectorPodStep`. The `PodListResponseStep` records the introspector job pod-name in the packet and is called after the introspector job completes or fails. I thought it's possible that if the chunk size is very small and first chunk returns empty then the active pod might be returned in second chunk.

russgold: I suspect that it wasn't actually needed there, either.

Can you explain the problem you were solving? That is, what was the old code doing, why was it a problem, and what is the new approach? There is a basic assumption that is either presumably not valid, and yet when I look at the code, it appears to be make the same assumptions, and just getting the job pod with more steps.

ankedia: This particular change is for the case when the introspector job fails because the job pod container is waiting to start due to image pull error. In this case, we need to delete the failed introspector job and run the introspection again. The job status in case of image pull error is `BackoffLimitExceeded` and the job pod container waiting reason is `ErrImagePull`. The problem with the old code was that it was not deleting and recreating the failed introspector job when the `introspectVersion` or `restartVersion` is changed. You can run `whenPreviousFailedJobWithImagePullErrorExistsAndMakeRightContinued_createNewJob` and `whenPreviousFailedJobWithImagePullBackoffErrorExistsAndMakeRightContinued_createNewJob` unit tests in DomainIntrospectorJobTest.java  to reproduce the issue.

The new approach is to list the introspector job pod and look at the reason why container is waiting to start. It deletes and recreates the job when the domain specs have changed.

russgold: I reverted your changes to JobHelper, and all of the unit tests still pass.

ankedia: I had run the test on my previous branch since I was waiting for failure_counts branch to be merged to main. Let me rerun the unit tests and functional tests to see if those changes are still necessary.

ankedia: If I run with JobHelper from main branch, below 3 unit tests are failing for me. How did you revert the change locally?
whenPreviousFailedJobWithImagePullErrorExistsAndMakeRightContinued_createNewJob
whenPreviousTimedoutJobExists_createNewJob
whenPreviousFailedJobWithImagePullBackoffErrorExistsAndMakeRightContinued_createNewJob

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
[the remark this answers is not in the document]
ankedia: I have changed the test to assert the NPE and DomainProcessingAborted event is not generated when WebLogic credentials secret is removed in eaf7651 .

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Domain.java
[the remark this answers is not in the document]
ankedia: ok, I have changed the tets `whenWebLogicCredentialsSecretRemoved_NullPointerExceptionAndAbortedEventNotGenerated` in DomainProcessorTest to assert that NPE is not generated when WebLogicCredentialsSecret is removed.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorDelegateStub.java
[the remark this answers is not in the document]
ankedia: I have changed the class name to `TestJobAwaiterStepFactory` instead of PassthoughJobAwaiterStepFactory`. Please let me know if this is not clear and if you have other suggestion for the class name.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: Apparently not correctly. My apologies. I restored my directory and tried it again and now I see the failures.

russgold: I think that the issue can be fixed much more simply. By changing line 217 of JobHelper from main branch to
```
        if (isKnownFailedJob(job) || JobWatcher.isFailed(job)) {
```

those tests all pass, but whenNewFailedJobExists_readPodLogAndReportFailure fails. That suggests that all we need is to have the existing step that finds the job pod look for the relevant errors.

ankedia: ok. Above change means we're always going to delete an existing failed job and do the reintrospection. So it's expected that `whenNewFailedJobExists_readPodLogAndReportFailure` fails. I had initially looked into reusing the existing step that finds the job pod but was running into some issues since that step was invoked from different places. Let me see if the changed logic main makes it possible to reuse that step now.

ankedia: In current change, I have added a new step to read the `job pod` after reading the `job` and look for the relevant errors. I thought this was more clear than adding the logic into existing step to execute different set of next steps based on different errors. The introspector job pod may not always be there (for e.g. k8s might delete the pod once it times out) and it's not possible to read the logs of a job pod that has failed due to image pull error. It might be good to have a zoom call to see if it's worthwhile to spend time experimenting with that if current change is more clear.

russgold: OK, can we meet in the morning?

ankedia: ok, that works for me. Let me schedule a time in the calendar.

ankedia: I tried making the changes to use existing step but the logic seems to be get more complicated and some unit tests are hanging. Based on my current understanding, below is the psuedo code of the new logic. Can you confirm if this understanding is correct or if I'm missing something?

The `onSuccess` method of exsiting `findIntrospectorPodName` step will do following. 
1. Fetch the job from the packet
2. Read the job pod
3. Get the job pod name

If (known failed job or image pull error) then 
 Execute deleteIntrospectorJob -> readExistingIntrospectorConfigMap -> createNewJob -> waitForIntrospectionToComplete -> findIntrospectorPodName -> readNamedPodLog -> deleteIntrospectorJob -> createIntrospectorConfigMap -> next action
else if (job is timed-out) then 
 Execute deleteDomainIntrospectorJobStep -> createNewJob -> waitForIntrospectionToComplete -> findIntrospectorPodName -> readNamedPodLog -> deleteIntrospectorJob -> createIntrospectorConfigMap -> next action 
else if (job is not null)
 Execute waitForIntrospectionToComplete -> findIntrospectorPodName -> readNamedPodLog -> deleteIntrospectorJob -> createIntrospectorConfigMap -> next action
else if (introspection is needed)  
 Execute readExistingIntrospectorConfigMap -> createNewJob -> waitForIntrospectionToComplete -> findIntrospectorPodName -> readNamedPodLog -> deleteIntrospectorJob -> createIntrospectorConfigMap -> next action
else 
 Execute next action
fi

russgold (approved):

ankedia:
Integration test run results look good, the three failures look unrelated to latest changes - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7172/

jshum2479 (approved):

PR #2615, opened 2021-11-05
Upgrade to apiVersion v9 and product version 4.0.0

rjeberhard:
@russgold, @ankedia, some context... I spent time looking at the various options of how we could support upgrading to Domain resource apiVersion v9 with 4.0.0. You can see the basic documentation here: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview. The default is the `None` strategy where Kubernetes will *only* upgrade the apiVersion field. This is what we've been using since the beginning and, long story short, is what I'm still using in this PR. The alternative is to provide a webhook endpoint that K8s would call when it needs to upgrade an older version of the resource. This seems really good and it does work, but it occurred to me that this introduces a single point of failure based on this webhook (if the endpoint is down then you can't create or change Domain resources). So, I didn't think that this was worth it. Instead, I've put code in to the CallBuilder and other classes to do the conversion when we read a domain or list domains.

Next, I made a small change to the SchemaGenerator to support obsolete enum-values and made Progressing obsolete. This means that the v9 schema does not list the Progressing value as a valid choice, but the v8 schema does. The Java code still accepts the value as it deserializes.  However, if we read a document that has such a Progressing condition then the upgrade logic will remove that condition. 

Finally, I've done the normal blocking and tacking to upgrade the product version.


integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
```diff
@@ -14,7 +14,7 @@
 
   // domain constants
   public static final String DOMAIN_VERSION = Optional.ofNullable(System.getenv("DOMAIN_VERSION"))
-      .orElse("v8");
+      .orElse("v9");
```
russgold, line 17: why not use the constant in integration-tests/src/test/java/oracle/weblogic/kubernetes/actions/impl/primitive/Kubernetes.java rather than changing this to keep in sync?

json-schema-generator/src/main/java/oracle/kubernetes/json/ObsoleteCheck.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.json;
+
+public interface ObsoleteCheck {
```
russgold, line 6: A more Java-standard name for this interface would be something like `Obsoleteable`. A check is a process, whereas this describes an object that can have a particular property or behavior.

operator/src/main/java/oracle/kubernetes/operator/helpers/CallBuilder.java
```diff
@@ -304,12 +304,63 @@
 
   private final String resourceVersion = "";
```
russgold, line 306: I'm a bit troubled by adding this functionality into CallBuilder, as it seems to be something of a violation of the [Single Responsibility Principal](https://stackify.com/solid-design-principles/). Our convention has generally been to use CallBuilder simply as an interface to the Kubernetes API and put additional resource-specific behavior into a xxxHelper class. The calls to listDomain and readDomain would then go through the new DomainHelper.

operator/src/test/sh/scaling/scalingActionTest.sh
```diff
@@ -31,7 +31,7 @@ test_get_domain_api_version() {
 
   result=$(get_domain_api_version)
 
-  assertEquals "Did not return expected api version" 'v8' "${result}"  
+  assertEquals "Did not return expected api version" 'v9' "${result}"
```
russgold, line 34: I would suggest using a constant, here, to minimize the number of changes we'll need to make each time we upgrade the domain resource version.

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
rjeberhard: I can look at this, but my thought process was that any place that was previously defaulting to "v8" would likely want to default now to "v9" unless it was explicitly an upgrade test.

integration-tests/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>oracle.kubernetes</groupId>
         <artifactId>operator-parent</artifactId>
-        <version>3.3.4</version>
+        <version>4.0.0</version>
```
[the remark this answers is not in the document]
rjeberhard: As above, I'll look at this too... Part of the reason that I've not used the Maven plugin is that we have a bunch of references that aren't easily driven from the Maven version, such as documentation, Helm chart default values, and samples, etc. Therefore, I'd just found it easier to do a grep and replace. I did find an easier way to update the references in the documentation with a change to just one file, but that is still separate from the Maven version.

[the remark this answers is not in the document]
russgold: I'm thinking that now that we are running a process to publish the docs, we could incorporate the maven version at that time. It would mean, of course, that we would only update them when we publish the release, rather than on every push to main.

russgold (approved):

russgold:
Oops - I didn't mean to approve this yet.

integration-tests/pom.xml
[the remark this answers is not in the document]
russgold: The maven release plugin runs steps in the project poms based on profiles. We could certainly incorporate grep-type replacements as part of that. We could also include the Hugo generation, and move all of the web page updating into that step. It would be much easier to test than a GitHub action.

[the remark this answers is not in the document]
rjeberhard: I'm open to using the release plugin but that needs to be more thought through and, therefore, outside the scope of this PR. For instance, the Hugo generation needs to be separate or at least separable from the release process since we often correct errors or update the documentation outside of the release process. I do strongly support automating as much of the release process as possible.

rjeberhard:
@russgold,
> Do you have a replacement test for the new approach of using Upgradable?

The original test added to CallBuilderTest goes through this codepath. I'm happy to find some place to add a simple test just on a Domain or DomainList.

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
rjeberhard: I made the other codepoint depend on this one.

russgold (approved):
LGTM

ankedia (approved):

PR #2616, opened 2021-11-08
OWLS-93783: WLSVERSIONEARLIERTHAN METHOD IS MISSING IN 3.3.3

lennyphan:
Add missing wlsVersionEarlierThan method in model filtering for MII and add unit tests to verify isSecureModeEnabled and missing import.

### DIFF
--- a/operator/src/main/resources/scripts/model_wdt_mii_filter.py
+++ b/operator/src/main/resources/scripts/model_wdt_mii_filter.py
@@ -139,6 +139,13 @@ class OfflineWlstEnv(object):
   def getModel(self):
     return self.model
 
+  def wlsVersionEarlierThan(self, version):
+    # unconventional import within function definition for unit testing
+    from weblogic.management.configuration import LegalHelper
+    # WLS Domain versions supported by operator are 12.2.1.3 + patches, 12.2.1.4
+    # and 14.1.1.0 so current version will only be one of these that are listed.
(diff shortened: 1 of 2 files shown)


operator/src/main/resources/scripts/model_wdt_mii_filter.py
```diff
@@ -139,6 +139,13 @@ def toDNS1123Legal(self, address):
   def getModel(self):
     return self.model
 
+  def wlsVersionEarlierThan(self, version):
+    # unconventional import within function definition for unit testing
+    from weblogic.management.configuration import LegalHelper
+    # WLS Domain versions supported by operator are 12.2.1.3 + patches, 12.2.1.4
+    # and 14.1.1.0 so current version will only be one of these that are listed.
+    return LegalHelper.versionEarlierThan("14.1.1.0", version)
```
doxiao, line 147: The code that was removed by PR #2582 checks against 14.1.2.0. Just want to make sure that changing from 14.1.2.0 to 14.1.1.0 here is intentional.

lennyphan: yes, it turns out from talking to Quang that secure mode will be enabled by default in 14.1.2.0 so had to change it.  This is intentional.

doxiao (approved):
LGTM.

rjeberhard (approved):

PR #2622, opened 2021-11-10
Merge changes for OWLS-930732 to include retry stage and fatal condition in domain status message 

ankedia:
Merge changes for OWLS-930732 from PR 2571 for following domain conditions -

- When retries exceed the max limit, add "introspect failure retry count exceeded error" to the domain status message.
- In case of a fatal error, include the fatal status in the domain status message.
- For non-fatal retries, indicate the current retry stage i.e. "Introspection failed on try 1 of x".


operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -54,7 +54,13 @@
       + "exceeded configured domainPresenceFailureRetryMaxCount: "
       + DomainPresence.getDomainPresenceFailureRetryMaxCount()
       + " The domainPresenceFailureRetryMaxCount is an operator tuning parameter and can be controlled"
-      + " by adding it to the weblogic-operator-cm configmap.";
+      + " by adding it to the weblogic-operator-cm configmap."
+      + " To force the introspector to start retrying again, update 'domain.spec.introspectVersion'.";
+
+  String INTROSPECTION_ERROR = "Introspection Error: ";
+
+  String FATAL_ERROR_DOMAIN_STATUS_MESSAGE = "Introspection encountered a fatal error and will not retry automatically."
```
rjeberhard, line 62: Can we internationalize this by pulling from the resource bundle?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -173,12 +176,42 @@ public FailureCountStep(V1Job domainIntrospectorJob) {
...
+      return domainStatus.getIntrospectJobFailureCount() >= DomainPresence.getDomainPresenceFailureRetryMaxCount();
+    }
+
+    private String createStatusMessage(DomainStatus domainStatus, String retryStatusMessage) {
+      return onSeparateLines(retryStatusMessage, INTROSPECTION_ERROR, domainStatus.getMessage());
+    }
+
+    private String getNonFatalRetryStatusMessage(DomainStatus domainStatus) {
+      return "Introspection failed on try " + domainStatus.getIntrospectJobFailureCount()
+              + " of " + DomainPresence.getDomainPresenceFailureRetryMaxCount() + ".";
```
doxiao, line 208: Shall we also provide the information about when the next retry will happen?

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
ankedia: Fixed in ea644db .

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
ankedia: @rjeberhard @tbarnes-us @jshum2479 - Any thoughts about this and where do we get this information from? Thanks.

doxiao: Would DomainPresence.getDomainPresenceFailureRetrySeconds() work for this?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -173,12 +175,51 @@ public FailureCountStep(V1Job domainIntrospectorJob) {
...
     private String getJobUid() {
       return Optional.ofNullable(domainIntrospectorJob).map(V1Job::getMetadata).map(V1ObjectMeta::getUid).orElse(null);
     }
+
+    private void addRetryInfoToStatusMessage(DomainStatus domainStatus) {
+
+      if (hasExceededMaxRetryCount(domainStatus)) {
+        domainStatus.setMessage(createStatusMessage(domainStatus, exceededMaxRetryCountErrorMessage()));
+      } else if (isFatalError(domainStatus)) {
+        domainStatus.setMessage(
```
russgold, line 191: The name FATAL_ERROR_DOMAIN_STATUS_MESSAGE doesn't seem to be consistent with the naming we use in MessageKeys. Maybe DOMAIN_FATAL_ERROR would be better, and allow you to dispense with the line break (If not, use a static import on the constant to shorten the line).

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -173,12 +175,51 @@ public FailureCountStep(V1Job domainIntrospectorJob) {
...
+              domainStatus.getIntrospectJobFailureCount(), getFailureRetryMaxCount());
+    }
+
+    private boolean isFatalError(DomainStatus domainStatus) {
+      return Optional.ofNullable(domainStatus.getMessage())
+              .map(m -> m.contains(FATAL_INTROSPECTOR_ERROR)).orElse(false);
+    }
+
+    private String exceededMaxRetryCountErrorMessage() {
+      return LOGGER.formatMessage(EXCEEDED_INTROSPECTOR_MAX_RETRY_COUNT_ERROR_MSG, getFailureRetryMaxCount());
```
russgold, line 217: A more consistent name might be INTROSPECTOR_MAX_ERRORS_EXCEEDED

operator/src/main/java/oracle/kubernetes/utils/OperatorUtils.java
```diff
@@ -176,4 +176,14 @@ private static StringBuilder toStringBuilder(Reader from) throws IOException {
     return sb;
   }
 
+  /**
+   * Returns a new String composed of copies of the CharSequence elements joined together with a copy
```
russgold, line 180: The comment says CharSequence, but the method takes String... And there is no need to mention "copy" as that is how string concatenation works.

russgold (approved):
Overall looks very good; I have requested some minor changes.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
jshum2479: Do we need exact timing?  or can we just tell the user it will retry approximately in next make right interval (2 mins or the configured time).

ankedia: In some cases such as when an introspector job times out, we retry after `DomainPresence.getDomainPresenceFailureRetrySeconds` but in other cases we retry during next make right interval.

ankedia:
> Overall looks very good; I have requested some minor changes.

Thanks, I have made the suggested changes in ec9d534.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
ankedia: We will track this new requirement with a separate JIRA and address in a new pull request.

tbarnes-us: please let me know the retry JIRA# when filed

note there's an existing umbrella retry JIRA OWLS-84438 that's meant for brain-storming purposes.  

IMO, it's confusing to have two different retry settings - as this makes it hard for customers to figure out which setting they need to tune

ankedia: @tbarnes-us new JIRA is https://jira.oraclecorp.com/jira/browse/OWLS-93899 .

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -408,16 +408,17 @@ private void processIntrospectionResult(Packet packet, String result) {
...
-              createFailureRelatedSteps(Introspection, onSeparateLines(severeStatuses)),
-              getNextStep(packet, domainIntrospectorJob));
-
         if (!severeStatuses.isEmpty()) {
           nextSteps = Step.chain(createFailureCountStep(domainIntrospectorJob), nextSteps);
         }
 
+        nextSteps = Step.chain(
+              createFailureRelatedSteps(Introspection, onSeparateLines(severeStatuses)),
+              getNextStep(packet, domainIntrospectorJob), nextSteps);
```
doxiao, line 420: I believe getNextSteop() method can return null. When it is null, will we have a null step in the middle of the chain here?

ankedia: My understanding was that Step.chain can handle null elements in the middle of the chain. The reason I made the above change is because we want the `FailureCountStep` to be executed last (i.e. after the `FailedStep`) to retain the correct message in domain status. If `FailedStep` is executed last, it overwrites the domain status message created by the `FailureCount` step. I didn't see any issues in the unit or the integration test with this approach. Please let me know if my understanding is incorrect, I can try to come up with some other approach.

ankedia: From the Javadoc of `Step.chain`, the parameter stepGroups is multiple groups of steps. Any ideas if the group of steps have to be non-null?
```
  /**
   * Chain the specified step groups into a single chain of steps.
   *
   * @param stepGroups multiple groups of steps
   * @return the first step of the resultant chain
   */
```

russgold: Look at StepChainTest.ignoreNullMiddleSteps. It actually verifies that this case works - null steps are ignored.

doxiao: Thanks, Russ. We are good then.

doxiao: Ideally, the FailureCount step should be after the FailedStep and before createEventStep so that the event can be populated with the domain status message. This will help to match the event message with the domain status message. If this requires too much changes, we can consider to make this change in a future PR.

operator/src/main/resources/Operator.properties
```diff
@@ -144,6 +144,12 @@ WLSKO-0191=Rolling restart of domain {0} completed
 WLSKO-0192=Executing make right domain operation, recheck count for server {0} is {1}.
 WLSKO-0193=Waiting for server {0} to start, recheck count is {1}.
 WLSKO-0194=Internal identity initialization step failed with exception {0}.
+WLSKO-0195=Introspection encountered a fatal error and will not retry automatically. \
+  Please resolve the error and then update 'domain.spec.introspectVersion' to force a retry.
+WLSKO-0196=Stop introspection retry - exceeded configured domainPresenceFailureRetryMaxCount: {0}. The \
+  domainPresenceFailureRetryMaxCount is an operator tuning parameter and can be controlled by adding it to the \
+  weblogic-operator-cm configmap. To force the introspector to start retrying again, update 'domain.spec.introspectVersion.
```
doxiao, line 151: Missing a "'" at the end.

operator/src/main/resources/Operator.properties
```diff
@@ -144,6 +144,12 @@ WLSKO-0191=Rolling restart of domain {0} completed
 WLSKO-0192=Executing make right domain operation, recheck count for server {0} is {1}.
 WLSKO-0193=Waiting for server {0} to start, recheck count is {1}.
 WLSKO-0194=Internal identity initialization step failed with exception {0}.
+WLSKO-0195=Introspection encountered a fatal error and will not retry automatically. \
+  Please resolve the error and then update 'domain.spec.introspectVersion' to force a retry.
+WLSKO-0196=Stop introspection retry - exceeded configured domainPresenceFailureRetryMaxCount: {0}. The \
+  domainPresenceFailureRetryMaxCount is an operator tuning parameter and can be controlled by adding it to the \
```
doxiao, line 150: Can domainPresenceFailureRetryMaxCount also be controlled via helm upgrade command?

ankedia: AFAIK, this can't be done using `helm upgrade` command.

operator/src/main/resources/Operator.properties
ankedia: Fixed in ea2d103 .

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
ankedia: Changed in ea2d103.

doxiao (approved):
LGTM.

PR #2627, opened 2021-11-16
OWLS-93641 - Fix for MII auxiliary image negative test case failure on OKD

ankedia:
The permission denied error was because of missing "&" when redirecting the standard error to standard output during `ls -1` command execution. This didn't fail on other platforms but fails on OKD due to different permissions. Maggie has verified that the fix resolves the issue on OKD and entire MIIAuxiliaryImage test suite is passing with this change on OKD.

### DIFF
--- a/operator/src/main/resources/scripts/utils.sh
+++ b/operator/src/main/resources/scripts/utils.sh
@@ -715,7 +715,7 @@ function checkAuxiliaryImage() {
     rm -f ${AUXILIARY_IMAGE_PATH}/testaccess.tmp || return 1
 
     # The container .out files embed their container name, the names will sort in the same order in which the containers ran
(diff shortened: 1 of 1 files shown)


maggiehe00 (approved):

tbarnes-us (approved):
Approved.  This makes a lot of sense.  That said, I have a small kernel of worry:  I could be mistaken but I vaguely recall the '&' was there at one point and now seems. to have been taken out - perhaps by @jshum2479 ? Not sure how to work back through the history?

ankedia:
> Approved. This makes a lot of sense. That said, I have a small kernel of worry: I could be mistaken but I vaguely recall the '&' was there at one point and now seems. to have been taken out - perhaps by @jshum2479 ? Not sure how to work back through the history?
Thanks. In looking at the git history, it seems '&' was missing in the original pull and I don't think it was taken out.

rjeberhard (approved):

rjeberhard:
I'll cherry pick this commit to release/3.3 just in case.

PR #2628, opened 2021-11-16
Correct some status update oversights

russgold:
This addresses multiple issues:

1. It brings the life-cycle of the Available condition in line with that of Completed. Both are now created false when the domain is recognized by the operator. (OWLS-93193, OWLS-94048)
2. It defines all servers shut down (such as start policy = NEVER) as Completed/true (OWLS-93493)
3. It corrects a bug in the detection of ReplicasTooHigh (OWLS-93692), but does not completely solve the issue; thus the JIRA remains open.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -4,6 +4,7 @@
 package oracle.kubernetes.operator;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -14,6 +15,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
@@ -84,6 +86,7 @@ import static oracle.kubernetes.operator.logging.MessageKeys.DOMAIN_FATAL_ERROR;
 import static oracle.kubernetes.operator.logging.MessageKeys.INTROSPECTOR_MAX_ERRORS_EXCEEDED;
 import static oracle.kubernetes.operator.logging.MessageKeys.TOO_MANY_REPLICAS_FAILURE;
 import static oracle.kubernetes.utils.OperatorUtils.onSeparateLines;
+import static oracle.kubernetes.weblogic.domain.model.DomainCondition.TRUE;
 import static oracle.kubernetes.weblogic.domain.model.DomainConditionType.Available;
 import static oracle.kubernetes.weblogic.domain.model.DomainConditionType.Completed;
 import static oracle.kubernetes.weblogic.domain.model.DomainConditionType.ConfigChangesPendingRestart;
@@ -98,8 +101,6 @@ import static oracle.kubernetes.weblogic.domain.model.DomainConditionType.Failed
 public class DomainStatusUpdater {
 
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
-  private static final String TRUE = "True";
-  private static final String FALSE = "False";
 
   private DomainStatusUpdater() {
   }
@@ -473,8 +474,8 @@ public class DomainStatusUpdater {
     @Override
     void modifyStatus(DomainStatus status) {
       if (status.getConditions().isEmpty()) {
-        status.addCondition(new DomainCondition(Completed).withStatus("False"));
-        status.addCondition(new DomainCondition(Available).withStatus("False"));
+        status.addCondition(new DomainCondition(Completed).withStatus(false));
+        status.addCondition(new DomainCondition(Available).withStatus(false));
       }
     }
   }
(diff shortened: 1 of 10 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -535,20 +535,10 @@ private boolean domainJustAvailable() {
...
-          status.addCondition(new DomainCondition(Completed).withStatus(FALSE));
-          if (sufficientServersRunning()) {
-            status.addCondition(new DomainCondition(Available).withStatus(TRUE));
-          } else if (status.hasConditionWithType(Available)) {
-            status.addCondition(new DomainCondition(Available).withStatus(FALSE));
-          }
-
-          addTooManyReplicasFailures(status);
-        }
+        status.addCondition(new DomainCondition(Completed).withStatus(isProcessingCompleted() ? TRUE : FALSE));
```
ankedia, line 538: It looks like status for domain condition `Completed` is set to `TRUE`  when `isProcessingCompleted()` returns true. However, DOMAIN_COMPLETE event is generated when `processingJustCompleted()`  returns true. Should the DOMAIN_COMPLETE event be generated only when the status for domain condition `Completed` is set to `TRUE` ?

russgold: `processingJustCompleted()` indicates a transition from Completed/false to Completed/true. The logic in this method computes the new overall conditions, which may be the same as the previous ones, in which case no such event is generated.

xiancao:
I verified the OWLS-93493 with this branch. When the serverStartPolicy is set to NEVER in **DOMAIN** level, the conditions have the expected result. However when the serverStartPolicy is set to NEVER in **CLUSTER** level, the condition type Completed is still false.

russgold:
> I verified the OWLS-93493 with this branch. When the serverStartPolicy is set to NEVER in **DOMAIN** level, the conditions have the expected result. However when the serverStartPolicy is set to NEVER in **CLUSTER** level, the condition type Completed is still false.

Thanks; I will try that case.

russgold:
@xiancao can you try your case again?

xiancao:
> @xiancao can you try your case again?

It's working now.

russgold:
Converting to draft, as I have found a couple of cases that I missed.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -564,47 +563,143 @@ private void setStatusConditions(DomainStatus status) {
...
+        }
+
+        List<DomainCondition> getNewConditions(DomainStatus status) {
+          return conditions.stream()
+                .filter(c -> "True".equals(c.getStatus()))
+                .filter(c -> status == null || !status.hasConditionWith(matchFor(c)))
+                .collect(Collectors.toList());
+        }
+
+        Predicate<DomainCondition> matchFor(DomainCondition condition) {
```
doxiao, line 587: Shall we take the failure reason into consideration as well in the logic here when the condition type is Failed?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -564,47 +563,143 @@ private void setStatusConditions(DomainStatus status) {
...
-      private List<WlsClusterConfig> getConfiguredClusters() {
-        return Optional.ofNullable(config).map(WlsDomainConfig::getConfiguredClusters).orElse(Collections.emptyList());
+      class Conditions {
+
+        private final ClusterCheck[] clusterChecks = createClusterChecks();
+        private final List<DomainCondition> conditions = new ArrayList<>();
+
+        void apply(DomainStatus status) {
+          status.removeConditionsMatching(c -> c.hasType(Failed) && ReplicasTooHigh.name().equals(c.getReason()));
+          conditions.forEach(status::addCondition);
```
doxiao, line 577: Shall we remove all Failed conditions in "status" if we have Completed condition with "True" status in "conditions"?

russgold: Yes, that makes sense.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: Can you explain? Take it into consideration how?

doxiao: When matchFor method checks if one condition matches another condition, currently it only take the condition type into consideration. When the type is "Failed",  we have many different failure reasons, such as DomainInvalid, Kubernetes, Internal, ReplicasTooHigh and etc. Should a "Failed" condition with one failure reason be considered different from another "Failed" condition with a different failure reason?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: We create the Completed condition, set to true or false, on every make right and status update. The case you describe is not relevant.

doxiao: What is the expected behavior for the case that I described? What is the expected status of the Completed condition after a fully running domain is updated with a weblogic credential setting that points to a non-existing secret?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: The purpose of this method is to return the new conditions for which we might generate events. Currently, that means only Available and Completed, as we have other places that generate Failure events. If we want to change that so this is also the source of the Failure events, we will indeed need to make that change.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: In such a case, we will have a failure, and that will result in completed/false.

russgold:
Fixing the ReplicasTooHigh case is going to require a fair bit of work, as we now need to do some processing in the scheduled status thread that was previously not needed. For that reason, and since Dongbo needs some of these changes, I would like to merge this now and defer the rest of the work to a different PR.

ankedia (approved):
LGTM

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
doxiao: Thanks. 

Is the answer to my original question still true at this point? If so, do we need to address it in this PR?

russgold: I think I didn't give you a good answer, originally. The statuses are recomputed every time we do the status update. If there are failures, we do not reach completed/true, so there is no need to remove them if we do reach that state - as it means that they have already been cleared.

russgold:
> > Russ: Fixing the ReplicasTooHigh case is going to require a fair bit of work, as we now need to do some processing in the scheduled status thread that was previously not needed. For that reason, and since Dongbo needs some of these changes, I would like to merge this now and defer the rest of the work to a different PR.
> 
> Tom: Sounds good to me, assuming the outage is tracked via JIRA & "Known Issue". I'm passing my sign-off off to Dongbo & Anil, as they're more familiar with Operator internals.

Yes, OWLS-93692 tracks the issue and is not closed by this PR.

doxiao (approved):
LGTM.

PR #2629, opened 2021-11-18
Temporarily disable trap in diff_model as logFileRotate or testLogFileRotate may return non-zero code.

ankedia:
The wdt shell script or logFileRotate may return non-zero code and it will go to trap if the trap is on. This change temporarily disables the trap and then restores it in the end.

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -1357,13 +1357,8 @@ function logSevereAndExit() {
 #   1 - Name of the log file to rotate and copy to WDT output directory.
 function wdtRotateAndCopyLogFile() {
   local logFileName=$1
-  # temporarily disable the trap as logFileRotate may return non-zero code
-  stop_trap
 
   logFileRotate "${WDT_OUTPUT_DIR}/${logFileName}" "${WDT_LOG_FILE_MAX:-11}"
 
   cp ${WDT_ROOT}/logs/${logFileName} ${WDT_OUTPUT_DIR}/
-
-  # restore trap
-  start_trap
 }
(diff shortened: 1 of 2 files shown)


jshum2479 (approved):

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -1354,7 +1360,7 @@ function wdtRotateAndCopyLogFile() {
   testLogFileRotate "${WDT_OUTPUT_DIR}/${logFileName}"
```
jshum2479, line 1360: Should this be removed since logFileRotate already called it?

ankedia: Removed.

tbarnes-us:
I delegate to Johnny - who approved yesterday but hasn't seen today's diff.  
Minor comment: stop/start in the rotate function itself looks unhelpful since all of the callers will have already stopped, it implicitly starts a trap that's about to be started by all of the callers, and if a caller actually deliberately had no trap it'd confusingly start one...

ankedia:
> I delegate to @jshum2479 - who approved yesterday but hasn't seen today's diff.
> 
> Minor comment: stop/start in the rotate function itself looks unhelpful since all of the callers will have already stopped, it implicitly starts a trap that's about to be started by all of the callers, and if a caller actually deliberately had no trap it'd confusingly start one...

Currently the rotate function is only called in the modelInImage.sh script which had trap enabled. I’ll either remove or check if the trap is enabled before start/stop.

jshum2479 (approved):

tbarnes-us (approved):
Welp since Johny is happy I'm happy :)

PR #2635, opened 2021-11-22
OWLS-91213: Shutdown servers using REST operations prior to deleting pods

lennyphan:
When the operator is going to shutdown a pod then it will first connect to the server instance and try to shutdown the server using REST shutdown request prior to deleting the pod.  This is a best effort attempt by the operator to gracefully shutdown the managed server before forcefully deleting the pod. The changes:

1. Issues a synchronous REST shutdown request to the managed server instance.
2. Added 'waitForAllSessions' parameter to Shutdown configuration.
3. Added FINE level logging messages for responses.


documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -227,6 +230,8 @@ the operator will not override the environment variable based on the shutdown co
 | `SHUTDOWN_TYPE` | `Graceful` | `Graceful` or `Forced` |
 | `SHUTDOWN_TIMEOUT` | 30 | Whole number in seconds where 0 means no timeout |
 | `SHUTDOWN_IGNORE_SESSIONS` | `false` | Boolean indicating if active sessions should be ignored; only applicable if shutdown is graceful |
+| `SHUTDOWN_WAIT_FOR_ALL_SESSIONS` | `false` | `true` to wait for all HTTP sessions during inflight work handling; `false` to wait for non-persisted HTTP sessions only ; only applicable if shutdown is graceful |
```
tbarnes-us, line 233: How does SHUTDOWN_WAIT_FOR_ALL_SESSIONS interact with SHUTDOWN_IGNORE_SESSIONS?

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -227,6 +230,8 @@ the operator will not override the environment variable based on the shutdown co
...
+
```
tbarnes-us, line 234: What happens when a customer configures both serverPod.waitForAllSessions and SHUTDOWN_WAIT_FOR_ALL_SESSIONS?

tbarnes-us, line 234: Why is there a serverPod.waitForAllSessions domain resource configurable when all of the other shutdown configurable's use env vars?  For example, it is confusing to say `For graceful shutdown only` in theserverPod.waitForAllSessions resources without having a sister configuration at same scope for configuring shutdown. IMO, they either all should have a `serverPod` setting or none should...

tbarnes-us: Ignore this. I didn't notice that there are already sister settings.

tbarnes-us: In other words, the documentation states `When a user-configured environment variable is present,
the operator will not override the environment variable based on the shutdown configuration.` ; so does this hold true for the new setting?

operator/src/main/java/oracle/kubernetes/operator/steps/ShutdownManagedServerStep.java
```diff
@@ -0,0 +1,249 @@
...
+      Boolean waitForAllSessions = getWaitForAllSessions(shutdown);
+
+      HttpRequest request = createRequestBuilder(getRequestUrl(isGracefulShutdown))
+            .POST(HttpRequest.BodyPublishers.ofString(getManagedServerShutdownPayload(
+                isGracefulShutdown, ignoreSessions, timeout, waitForAllSessions))).build();
+      return request;
+    }
+
+    private Boolean isGracefulShutdown(Shutdown shutdown) {
+      String shutdownType = Optional.ofNullable(shutdown).map(Shutdown::getShutdownType)
```
tbarnes-us, line 130: I could be wrong, but this doesn't appear to check if the env var is set for the pod, or if graceful shutdown is configured on WebLogic server mbean itself...

operator/src/main/resources/Operator.properties
```diff
@@ -142,6 +142,8 @@ WLSKO-0191=Rolling restart of domain {0} completed
 WLSKO-0192=Executing make right domain operation, recheck count for server {0} is {1}.
 WLSKO-0193=Waiting for server {0} to start, recheck count is {1}.
 WLSKO-0194=Internal identity initialization step failed with exception {0}.
+WLSKO-0195=Successfully shutdown managed server {0} via REST interface.
+WLSKO-0196=Failed to shutdown managed server {0} via REST interface with response: {1}.
```
tbarnes-us, line 146: I assume there's a pre-existing FINE that indicates start of shutdown for a managed server?
Perhaps all fines related to shutdown could be prefixed with something like "WL pod shutdown:" or some-such.
"managed server" --> "WebLogic server" (because admin server is technically not a managed server)

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: The 'ignoreSessions' property takes precedence.  If 'ignoreSessions=true' then 'waitForAllSessions' is ignored.  If 'ignoreSessions=false' and  'waitForAllSessions=true' then shutdown will wait till all sessions are complete before shutdown.  If 'ignoreSessions=false' and  'waitForAllSessions=false' then shutdown will only wait for non-persisted sessions to complete.  I'll try and update the description here to try and make it more clear.  I don't even think out docs point this out as I cut-pasted the descriptions I've given from our docs.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: For the new property, I just followed the current pattern as the other shutdown properties (type, timeout and ignoreSessions) that already existed.  We only add the environment variable if it doesn't already exist.  In PodStepContext.updateEnvForShutdown():

Shutdown shutdown = getShutdownSpec();
addDefaultEnvVarIfMissing(env, "SHUTDOWN_TYPE", shutdown.getShutdownType());
addDefaultEnvVarIfMissing(env, "SHUTDOWN_TIMEOUT", String.valueOf(shutdown.getTimeoutSeconds()));
addDefaultEnvVarIfMissing(env, "SHUTDOWN_IGNORE_SESSIONS", String.valueOf(shutdown.getIgnoreSessions()));
addDefaultEnvVarIfMissing(env, "SHUTDOWN_WAIT_FOR_ALL_SESSIONS",String.valueOf(shutdown.getWaitForAllSessions()));

operator/src/main/java/oracle/kubernetes/operator/steps/ShutdownManagedServerStep.java
lennyphan: Yes, you're right.  The configuration in the step is just looking at the domain resource shutdown properties and  not checking the environment.  Let me see how I can get to the env vars.

operator/src/main/resources/Operator.properties
lennyphan: When looking at all the messages, I didn't see any pre-existing FINE message that indicates the start of a shutdown of a managed server as we were basically deleting the pod which then called the stop scripts.  The only messages I found were in the stop-server.py script and look like the following:

print('Shutdown: Calling server shutdown with force = ' + force)
print('Shutdown: Successfully shutdown the server')
print('Shutdown: Exception in connect or shutdown')

I'm happy to change the message to have the prefix you suggested, as well as remove 'managed server'.

lennyphan: Thanks for the review Tom.  I'll make the changes.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
tbarnes-us: Then this (and the other env vars), should mention that they override the corresponding domain resource field when explicitly set...

tbarnes-us: In short, this confirms that the precedence is following the already documented convention for all env vars.  Thanks.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: I've added a note to try and describe the parameter precedence, please review and let me what you think.

operator/src/main/java/oracle/kubernetes/operator/steps/ShutdownManagedServerStep.java
lennyphan: Changed so that we take the environment variable as precedence.  By the time we are deleting a server pod, the environment variable would already be populated with the configuration from the shutdown configuration of the domain resource but the code will still check the shutdown configuration if the environment variable doesn't exist.

operator/src/main/resources/Operator.properties
lennyphan: Added another FINE log message for when we initiate shutdown of a server via REST.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
tbarnes-us: Looks good!

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/Shutdown.java
```diff
@@ -35,6 +36,13 @@
           + " Defaults to false.")
   private Boolean ignoreSessions;
 
+  @Description(
+      "For graceful shutdown only, Set to true to wait for all HTTP sessions"
```
tbarnes-us, line 40: Minor: Set --> set

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
tbarnes-us: Looks good.

operator/src/main/resources/Operator.properties
tbarnes-us: Looks good.

tbarnes-us:
Holding off approval until review with @rjeberhard in light of the original goal of this pull. The main goal seems to be preventing K8S from destroying the pod prematurely and, my guess, is thereby abandoning NFS v3 locks because the "NFS lock manager" is destroyed before it can clean up any locks held by destroyed processes that own locks. 

Using REST, IMO, doesn't quite address the goal - it just changes the timing.  If the REST call fails to return within X seconds, then the pod is deleted as it was before, which in turn waits up to X before pod destruction.  In effect, we've moved the goal-posts from X to 2X, and CAGBU can thereby get a similar benefit to the current version of this pull simply by doubling their shutdown timeout. This is a rough approximation:  a side benefit of using REST is that it doesn't rely on WLST, which should give us a few extra seconds within X (the cost of spinning up WLST is no longer included in X).

If we are trying to avoid abandoned locks, I think pod delete needs extra logic. For example, shutdown script could start an 'time bomb' script that does a kill -9 of teh NM and WLS at X-5seconds or some-such...  This would give the normal shutdown a chance to finish, but, if it takes to long, would have the affect of removing any left behind locks (the kill -9 kills the process which causes the NFS lock manager to remove any locks it hold).

russgold:
All of our interactions with Kubernetes are via REST calls (that's what underlies CallBuilder). Why is this different?

lennyphan:
The CallBuilder is used for calling the K8s API server. Here, the operator is calling shutdown directly on the managed server using WebLogic Management REST API, similar to reading the managed servers health state.

russgold (commented):
Looks good; don't forget to update the copyright dates.

lennyphan:
Still waiting for results from integration tests run on kind cluster:  https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8138/

Results from previous run had some failures but did not look related as there were issues in test setup/initialization.

operator/src/main/java/oracle/kubernetes/operator/steps/HttpRequestProcessing.java
```diff
@@ -35,8 +35,13 @@
...
 
   static HttpAsyncRequestStep createRequestStep(HttpRequest request, HttpResponseStep responseStep) {
+    return createRequestStep(request, responseStep, HTTP_TIMEOUT_SECONDS);
+  }
+
+  static HttpAsyncRequestStep createRequestStep(HttpRequest request, HttpResponseStep responseStep,
+      Long httpTimeoutSeconds) {
     return HttpAsyncRequestStep.create(request, responseStep)
-          .withTimeoutSeconds(HTTP_TIMEOUT_SECONDS);
+          .withTimeoutSeconds(httpTimeoutSeconds);
```
tbarnes-us, line 44: Should this be wl-shutdown-secs instead of the http-timeout? plus maybe a few seconds as a "buffer"?

It's normal not to get a response if shutdown is taking a long time...

tbarnes-us: @rjeberhard Any thoughts?

lennyphan: The `httpTimeoutSeconds` value passed in comes from the Shutdown.timeoutSeconds value specified and I believe this is the same as the default weblogic graceful shutdown value.  I can add the same fudge factor (10 seconds) used when we delete the pod:

PodHelper.DEFAULT_ADDITIONAL_DELETE_TIME = 10;

tbarnes-us: Thanks. It's good that the http timeout in this case is based on the shutdown timeout.  OTOH the variable name is confusing if that's the case. Why not just directly use the shutdown timeout variable (assuming there is one somewhere)? 

Adding a fudge factor sounds good to me. It'd be great if @russgold and/or @rjeberhard could weigh in here too.

lennyphan: The original method specified the constant HTTP_TIMEOUT_SECONDS and I created an overloaded method to be able pass the timeout as a parameter and just ended up keeping the name, hence httpTimeoutSeconds.  I can add the same fudge factor that we use in PodHelper.DeletePodStep if we all think  an extra 10 seconds is reasonable, even though it's used for allowing extra time when connecting via WLST.

lennyphan: I added the additional fudge factor to the request timeout, see ShutdownManagedServerStep.getRequestTimeoutSeconds().  We can make modify it if we need to based on everyone's thoughts.  @russgold and @rjeberhard could weigh in here too.

tbarnes-us (approved):

operator/src/main/java/oracle/kubernetes/operator/steps/HttpRequestProcessing.java
```diff
@@ -35,8 +35,13 @@
   }
 
   static HttpAsyncRequestStep createRequestStep(HttpRequest request, HttpResponseStep responseStep) {
+    return createRequestStep(request, responseStep, HTTP_TIMEOUT_SECONDS);
+  }
+
+  static HttpAsyncRequestStep createRequestStep(HttpRequest request, HttpResponseStep responseStep,
```
russgold, line 41: Since the `withTimeoutSeconds` method is public, have the caller use that rather than creating an additional method that does the same thing.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTestSetup.java
```diff
@@ -34,7 +34,9 @@ public static void defineRequiredResources(KubernetesTestSupport testSupport) {
   }
 
   private static V1Secret createSecret() {
-    return new V1Secret().metadata(new V1ObjectMeta().name(SECRET_NAME).namespace(NS));
+    return new V1Secret().metadata(new V1ObjectMeta().name(SECRET_NAME).namespace(NS))
```
russgold, line 37: The varied indentation of the keys in the data map makes this harder to read. Either put them on the same line or line them up vertically.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/BaseConfiguration.java
```diff
@@ -170,6 +170,11 @@ Shutdown getShutdown() {
     return serverPod.getShutdown();
   }
 
+  public void setShutdown(String shutdownType, Long timeoutSeconds, Boolean ignoreSessions,
```
russgold, line 173: Is this used anywhere?

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -207,15 +207,27 @@ The server will count toward the cluster's `replicas` count.  Also, if you confi
 The Domain YAML file includes the field `serverPod` that is available under `spec`, `adminServer`, and each entry of
 `clusters` and `managedServers`. The `serverPod` field controls many details of how Pods are generated for WebLogic Server instances.
 
-The `shutdown` field of `serverPod` controls how servers will be shut down and has three fields:
-`shutdownType`, `timeoutSeconds`, and `ignoreSessions`.  The `shutdownType` field can be set to either `Graceful`, the default,
+The `shutdown` field of `serverPod` controls how servers will be shut down and has four fields:
```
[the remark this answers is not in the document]
rosemarymarano: @lennyphan Yes, please make me a reviewer this PR after the technical edits are completed. Thank you.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -207,15 +207,27 @@ The server will count toward the cluster's `replicas` count.  Also, if you confi
...
-The operator runtime monitors this property but will not restart any server pods solely to adjust the shutdown options.
+The `shutdown` field of `serverPod` controls how managed servers will be shut down and has the following four properties:
+`shutdownType`, `timeoutSeconds`, `ignoreSessions` and `waitForAllSessions`. The operator runtime monitors these properties but will not restart any server pods solely to adjust the shutdown options.
 Instead, server pods created or restarted because of another property change will be configured to shutdown, at the appropriate
 time, using the shutdown options set when the WebLogic Server instance Pod is created.
 
+| Field| Default Value | Supported Values | Description |
+| --- | --- | --- | --- |
+| `shutdownType` | `Graceful` | `Graceful` or `Forced` | Specifies how the operator will shut down server instances. |
+| `timeoutSeconds` | 30 | Whole number in seconds where 0 means no timeout | For graceful shutdown only, number of seconds to wait before aborting in-flight work and shutting down the server. |
```
rosemarymarano, line 218: no timeout. (period at the end of the sentence)

documentation/domains/Domain.md
```diff
@@ -257,6 +257,7 @@ The current status of the operation of the WebLogic domain. Updated automaticall
 | `ignoreSessions` | Boolean | For graceful shutdown only, indicates to ignore pending HTTP sessions during in-flight work handling. Defaults to false. |
 | `shutdownType` | string | Specifies how the operator will shut down server instances. Defaults to graceful shutdown. |
 | `timeoutSeconds` | number | For graceful shutdown only, number of seconds to wait before aborting in-flight work and shutting down the server. Defaults to 30 seconds. |
+| `waitForAllSessions` | Boolean | For graceful shutdown only, set to true to wait for all HTTP sessions during inflight work handling; false to wait for non-persisted HTTP sessions only. Defaults to false. |
```
rosemarymarano, line 260: inflight  -> in-flight

documentation/domains/index.html
```diff
@@ -1891,6 +1891,10 @@
         "timeoutSeconds": {
           "description": "For graceful shutdown only, number of seconds to wait before aborting in-flight work and shutting down the server. Defaults to 30 seconds.",
           "type": "number"
+        },
+        "waitForAllSessions": {
+          "description": "For graceful shutdown only, set to true to wait for all HTTP sessions during inflight work handling; false to wait for non-persisted HTTP sessions only. Defaults to false.",
```
rosemarymarano, line 1896: inflight -> in-flight

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -207,15 +207,27 @@ The server will count toward the cluster's `replicas` count.  Also, if you confi
...
+| `ignoreSessions` | `false` | `true` or `false` | Boolean indicating if active sessions should be ignored; only applicable if shutdown is graceful. |
+| `waitForAllSessions` | `false` | `true` or `false` | For graceful shutdown only, set to `true` to wait for all HTTP sessions during inflight work handling; `false` to wait for non-persisted HTTP sessions only during inflight work handling. |
```
rosemarymarano, line 220: inflight -> in-flight (please change this globally)

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -207,15 +207,27 @@ The server will count toward the cluster's `replicas` count.  Also, if you confi
...
+
+{{% notice note %}}
+The `waitForAllSessions` property does not apply when the `ignoreSessions` property is `true`. When the
+`ignoreSessions` property is `false` then `waitForAllSessions` property is taken into account during 
+the WebLogic graceful shutdown process. When the`waitForAllSessions` is `true`, the graceful shutdown
+process will wait for all HTTP session to complete or be invalidated before proceeding. When `waitForAllSessions`
```
rosemarymarano, line 226: HTTP session -> HTTP sessions

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
```diff
@@ -207,15 +207,27 @@ The server will count toward the cluster's `replicas` count.  Also, if you confi
...
+is `false`, the graceful shutdown process will only wait for non-persisted HTTP sessions to complete
```
rosemarymarano, line 227: will only wait -> will wait only

rosemarymarano (commented):
Few minor edits

operator/src/test/java/oracle/kubernetes/operator/steps/ShutdownManagedServerStepTest.java
```diff
@@ -0,0 +1,404 @@
...
+      CONFIGURED_CLUSTER_NAME);
+  private final V1Service standaloneServerService = createServerService(MANAGED_SERVER1, null);
+  private final V1Service dynamicServerService = createServerService(DYNAMIC_MANAGED_SERVER1,
+      DYNAMIC_CLUSTER_NAME);
+  private final List<LogRecord> logRecords = new ArrayList<>();
+  private final List<Memento> mementos = new ArrayList<>();
+  private final KubernetesTestSupport testSupport = new KubernetesTestSupport();
+  private final HttpAsyncTestSupport httpSupport = new HttpAsyncTestSupport();
+  private final TerminalStep terminalStep = new TerminalStep();
+  private Step shutdownConfiguredManagedServer = ShutdownManagedServerStep
```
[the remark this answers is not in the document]
lennyphan: fields are final

operator/src/main/java/oracle/kubernetes/operator/steps/ShutdownManagedServerStep.java
```diff
@@ -0,0 +1,390 @@
+// Copyright (c) 2018, 2022, Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
lennyphan: Eliminated as many of the warnings as possible.  Thanks for the advice.

operator/src/main/java/oracle/kubernetes/operator/steps/ServerDownStep.java
```diff
@@ -43,6 +43,7 @@ public NextAction apply(Packet packet) {
       next = pw.waitForDelete(oldPod, new DomainPresenceInfoUpdateStep(serverName, next));
     }
 
-    return doNext(oldPod != null ? PodHelper.deletePodStep(serverName, next) : next, packet);
+    return doNext(oldPod != null ? ShutdownManagedServerStep
```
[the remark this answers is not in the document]
lennyphan: extracted to its  own method for easier readability

operator/src/main/java/oracle/kubernetes/operator/steps/HttpRequestProcessing.java
lennyphan: removed the additional method and had caller directly use the call that includes `withTimeoutSeconds` method

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTestSetup.java
lennyphan: put on the same line.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/BaseConfiguration.java
lennyphan: remove unused method

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: added period

documentation/domains/Domain.md
lennyphan: made global change for inflight -> in-flight

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: Fixed.

documentation/staging/content/userguide/managing-domains/domain-lifecycle/startup.md
lennyphan: Fixed

rosemarymarano (approved):

PR #2639, opened 2021-11-24
Support hostAliases

rjeberhard:
A customer asked us to support the `hostAliases` field as part of our `serverPod`. This field allows you to add elements to the `/etc/hosts` file of the containers. 

We had previously not supported most of the networking related fields that could be on a pod because of concern that they could easily be misused; however, the customer found that other workarounds (such as using ExternalName Services) were insufficient.


ankedia (approved):
LGTM

PR #2643, opened 2021-11-29
Enhance fiber breadcrumbs

russgold:
This adds some additional debugging capabilities for fibers:

When returning a NextAction it is now possible to add some static or computed information:
return doNext(getNext(), packet).withDebugComment(() -> "assume presence up-to-date");
indicates that the fiber has chosen a path in which the domain presence is considered to be up-to-date

return doNext(createIntrospectionSteps(), packet)
                    .withDebugComment(packet, this::introspectionNeededReason);
indicates that the fiber has decided that it needs to run an introspection, along with its reasoning, derived from the packet.

In each case, the comment is actually derived from a method, which is not invoked unless the feature is enabled for the current fiber (intended to avoid excess computation). That means that it may safely be left in the code.

To enable the feature, the code must add a string to the packet, using the key Fiber.DEBUG_FIBER

When the fiber terminates, it will be logged, along with the value of the Fiber.DEBUG_FIBER entry as a prefix.

Note that, at present, the feature is disabled unless specifically enabled in the code.

Replaces #2607

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
@@ -79,7 +79,7 @@ public class DomainValidationSteps {
     }
 
     static List<V1Secret> getSecrets(Packet packet) {
-      return (List<V1Secret>) Optional.ofNullable(packet.get(SECRETS)).orElse(new ArrayList<>());
+      return Optional.ofNullable(packet.<List<V1Secret>>getValue(SECRETS)).orElse(new ArrayList<>());
     }
   }
 
@@ -99,7 +99,7 @@ public class DomainValidationSteps {
     }
 
     static List<V1ConfigMap> getConfigMaps(Packet packet) {
-      return (List<V1ConfigMap>) Optional.ofNullable(packet.get(CONFIGMAPS)).orElse(new ArrayList<>());
+      return Optional.ofNullable(packet.<List<V1ConfigMap>>getValue(CONFIGMAPS)).orElse(new ArrayList<>());
     }
   }
 
@@ -116,7 +116,7 @@ public class DomainValidationSteps {
       List<String> validationFailures = domain.getValidationFailures(new KubernetesResourceLookupImpl(packet));
 
       if (validationFailures.isEmpty()) {
-        return doNext(packet);
+        return doNext(packet).withDebugComment(packet, this::domainValidated);
       }
 
       LOGGER.severe(DOMAIN_VALIDATION_FAILED, domain.getDomainUid(), perLine(validationFailures));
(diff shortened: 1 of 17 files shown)


rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -127,6 +127,10 @@ public NextAction apply(Packet packet) {
     private String perLine(List<String> validationFailures) {
       return String.join(lineSeparator(), validationFailures);
     }
+
+    private String domainValidated(Packet packet) {
+      return "Validated " + DomainPresenceInfo.fromPacket(packet).orElse(null);
```
ankedia, line 132: It seems debug message is printing the contents of DomainPresenceInfo. Should this be just the domain name?

lennyphan (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -252,6 +259,36 @@ private boolean isIntrospectionNeeded(Packet packet) {
...
+          sb.append("bringing up new domain");
+        } else {
+          sb.append("something else");
+        }
+        return sb.toString();
+      }
+
+      private String introspectionNotNeededReason(Packet packet) {
+        StringBuilder sb = new StringBuilder("introspection not needed because ");
+        if (getNumRunningServers() != 0) {
```
ankedia, line 276: The introspection may not be needed for different reasons even when there are running servers. For e.g. if the `domain.spec.introspectVersion` or  `domain.spec.image` value didn't change. Would the debug message be incorrect in such scenarios?

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
russgold: Yes, I agree. It would be better just to print the domain name.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
russgold: Well, it's intended for developers to understand what is happening in the code, so I probably didn't handle every possible case, nor is that the goal. It should be easy enough to add cases as needed. If this were intended to end-users, I'd agree that we need to be more careful about not printing debug messages.

ankedia: ok. I was thinking that the debug message can just print the values for different conditions rather than checking for those conditions. For e.g.

```
StringBuilder sb = new StringBuilder("introspection not needed : ");
sb.append("Number of running servers: " + getNumRunningServers() + ",");
sb.append("Creating new servers: " + creatingServers(info) + ",");
sb.append("domain generation = ").append(getDomainGeneration())
                .append(" and packet last generation is ").append(packet.get(INTROSPECTION_DOMAIN_SPEC_GENERATION));
return sb.toString(); 
```
Anyways, I'm fine with refining and adding those cases later.

ankedia (approved):

PR #2648, opened 2021-12-01
OWLS-93891 - Update operator 4.0 docker pre-reqs and change the 'auxilliary image' docker file

ankedia:
Updated docker pre-reqs for Operator 4.0 and changed the docker file for 'auxilliary image' to use a single statement for COPY and chown.


ddsharpe (approved):

tbarnes-us:
@rjeberhard @rosemarymarano Note that this includes a 4.0 only doc change. We've been making changes  throughout staging (and have more changes coming) assuming it'd be back-ported to 3, so this change seems tricky to me...

ankedia:
> @rjeberhard @rosemarymarano Note that this includes a 4.0 only doc change. We've been making changes throughout staging (and have more changes coming) assuming it'd be back-ported to 3, so this change seems tricky to me...

I'm wondering whether we need to backport this change. Since the pre-requisites for the operator 3.3.x is Docker 18, the change to the docker file won't work in 3.3.x as it requires docker 19.03.

rosemarymarano:
The text is fine. Why would you assume that updating a prerequisite for operator 4.0 should be backported to the 3.x doc line?

robertpatrick (approved):

PR #2660, opened 2021-12-09
Fix intermittent events issue

sankarpn:
It looks like the event was logged couple of seconds before the timestamp was obtained. The call to now() should be before createNewCluster(), the test should pass without any timing issue.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7417/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesEvents.java
@@ -389,8 +389,8 @@ class ItKubernetesEvents {
   @Order(5)
   @Test
   void testK8SEventsMultiClusterEvents() {
-    createNewCluster();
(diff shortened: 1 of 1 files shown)


anpanigr (commented):
If this is the reason, it should fail always.  Is it is intermittent issue ? Please post the jenkin result.

sankarpn:
> If this is the reason, it should fail always. Is it is intermittent issue ? Please post the jenkin result.

The logging of events doesn't happen immediately. Some times it takes a while to log it and other times it happens immediately. When it happens immediately the test fails. The reason it was failing some times is because the DomainAvailable event logged immediately after the domain resource was patched. But we were capturing the before timestamp after the event was logged which causes the test to fail.

Basically we should capture the timestamp before a domain resource is created or patched.

anpanigr (approved):
lgtm

maggiehe00 (approved):
LGTM

PR #2664, opened 2021-12-14
OWLS-94812 - Fix for auxiliary image error after the main container crashes and restarts.

ankedia:
- Retain the `$AUXILIARY_IMAGE_PATH/auxiliaryImageLogs` dir after the check in `checkAuxiliaryImage` is finished.
- Change to redirect the output of `ls -1` command to `/dev/null`.

### DIFF
--- a/operator/src/main/resources/scripts/utils.sh
+++ b/operator/src/main/resources/scripts/utils.sh
@@ -722,9 +722,11 @@ function checkAuxiliaryImage() {
     rm -f ${AUXILIARY_IMAGE_PATH}/testaccess.tmp || return 1
 
     # The container .out files embed their container name, the names will sort in the same order in which the containers ran
-    out_files=$(set -o pipefail ; ls -1 $AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out > /dev/null 2>&1 | sort --version-sort) \
-      || (trace SEVERE "Auxiliary Image: Assertion failure. No files found in '$AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out" \
-      && return 1)
+    out_files=$(ls -1 $AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out 2>/dev/null | sort --version-sort)
+    if [ -z ${out_files} ]; then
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/utils.sh
```diff
@@ -722,9 +722,11 @@ function checkAuxiliaryImage() {
     rm -f ${AUXILIARY_IMAGE_PATH}/testaccess.tmp || return 1
 
     # The container .out files embed their container name, the names will sort in the same order in which the containers ran
-    out_files=$(set -o pipefail ; ls -1 $AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out 2>&1 | sort --version-sort) \
-      || (trace SEVERE "Auxiliary Image: Assertion failure. No files found in '$AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out" \
-      && return 1)
+    out_files=$(ls -1 $AUXILIARY_IMAGE_PATH/auxiliaryImageLogs/*.out 2>/dev/null | sort --version-sort)
+    if [ -z ${out_files} ]; then
```
tbarnes-us, line 726: I think ${out_files} should be quoted "${out_files}". The -z check expects a single parameter...

operator/src/main/resources/scripts/utils.sh
```diff
@@ -742,7 +744,6 @@ function checkAuxiliaryImage() {
       trace "Auxiliary Image: End of '${out_file}' contents"
     done
     [ "${severe_found}" = "true" ] && return 1
-    rm -fr $AUXILIARY_IMAGE_PATH/auxiliaryImageLogs
     [ -z "$(ls -A $AUXILIARY_IMAGE_PATH)" ] \
```
tbarnes-us, line 747: This check is no longer valid now that auxiliaryImageLogs is always going to be there.

tbarnes-us: Maybe have `"$(ls -A $AUXILIARY_IMAGE_PATH 2>/dev/null | grep -v auxiliaryImageLogs)"`

tbarnes-us (approved):
LGTM

Note: Advanced users may be surprised to now see the "auxiliaryImageLogs" directory in their aux mount, but I can't think of a way to fix that off the top of my head, and it seems doubtful it'll cause an insurmountable problem in the rare case its a problem (if any)...

jshum2479 (approved):

PR #2678, opened 2021-12-17
OWLS-94652 - Enable AUXILIARY_IMAGE_PATH validation when command is not customized.

ankedia:
OWLS-94652 - Enable the AUXILIARY_IMAGE_PATH check when custom command is not specified and the default command is used.

Integration test run started - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7537/console

### DIFF
--- a/operator/src/main/resources/scripts/utils_base.sh
+++ b/operator/src/main/resources/scripts/utils_base.sh
@@ -234,6 +234,17 @@ function initAuxiliaryImage() {
     return
   fi
 
+  if [[ "$AUXILIARY_IMAGE_COMMAND" == "cp -R \$AUXILIARY_IMAGE_PATH/* \$AUXILIARY_IMAGE_TARGET_PATH" ]]; then
+    # Auxiliary image command is default, validate that AUXILIARY_IMAGE_PATH exists and is not empty.
+    if [ ! -d ${AUXILIARY_IMAGE_PATH} ]; then
+      trace SEVERE "Auxiliary Image: Dir '${AUXILIARY_IMAGE_PATH}' doesn't exist. Exiting."
+      return
+    elif [ -z "$(ls -A ${AUXILIARY_IMAGE_PATH})" ]; then
+      trace SEVERE "Auxiliary Image: Dir '${AUXILIARY_IMAGE_PATH}' is empty. Exiting."
+      return
+    fi
+  fi
+
   trace FINE "Auxiliary Image: About to execute command '$AUXILIARY_IMAGE_COMMAND' in container image='$AUXILIARY_IMAGE_CONTAINER_IMAGE'. " \
              "AUXILIARY_IMAGE_PATH is '$AUXILIARY_IMAGE_PATH' and AUXILIARY_IMAGE_TARGET_PATH is '${AUXILIARY_IMAGE_TARGET_PATH}'."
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -234,6 +234,17 @@ function initAuxiliaryImage() {
     return
   fi
 
+  if [[ "$AUXILIARY_IMAGE_COMMAND" == "cp -R \$AUXILIARY_IMAGE_PATH/* \$AUXILIARY_IMAGE_TARGET_PATH" ]]; then
```
tbarnes-us, line 237: This looks to me like a somewhat fragile way to check if COMMAND is at the default. Any small change in the default at the operator level would forever more skip this check.

Instead:

-Could the operator pass some sort of flag that the COMMAND is at the default? 
 
-Or could you add a unit test in the operator that checks the default for COMMAND exactly matches this string, and have a notation in the unit test that the unit test exists because there's a corresponding equals check in `utils_base.sh`.

Something like that.

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -234,6 +234,17 @@ function initAuxiliaryImage() {
...
+    # Auxiliary image command is default, validate that AUXILIARY_IMAGE_PATH exists and is not empty.
+    if [ ! -d ${AUXILIARY_IMAGE_PATH} ]; then
```
tbarnes-us, line 239: minor: `${AUXILIARY_IMAGE_PATH}` should ideally be `"${AUXILIARY_IMAGE_PATH}"` in case the path has spaces in it.  i'm sure we have forgotten to do this in many places, but might as well do it here regardless

operator/src/main/resources/scripts/utils_base.sh
ankedia: - I thought about introducing another environment variable for the default command. However, adding env variable will cause the pods to roll since the pod specs will change due to that. 
- I will check and add the unit test if one doesn't exist already.  Thanks.

robertpatrick (commented):
Two comments:

1.) Generating an error or warning if the directory doesn't exist is misleading.  This directory won't exist if WDT and the models are in a custom location.  The user should not be misled by the logs implying that something is wrong.
[ANIL] Ok. My understanding was that we discussed this and decided to generate an error only in cases when the auxiliary image command is not specified and default command is used. In such cases, if WDT and the modeles are in a custom location, then the default command will fail.

2.) I don't understand the value of the check.  If we want to check the non-empty directory, then we should make the check meaningful.  For example, does the weblogic-deploy/bin/createDomain.sh file exist and is the models subdirectory present and contain something. In my opinion, if you cannot do anything more than check if the top-level directory exists and isn't empty, then this check doesn't really buy us much.
[ANIL] ok, yes. As you suggested, I will make the check meaningful by checking for the existence of `weblogic-deploy/bin/createDomain.sh` file and/or the models subdirectory contents. I'll setup  a meeting to discuss this and see whether we want to support other use-cases for auxiliary image which doesn't involve model-in-image or WDT. Thanks.

ankedia:
Closing this PR since we decided to leave the check removed.

PR #2679, opened 2021-12-17
OWLS-94707, fixed regression due model file change

marinakog:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7538

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMonitoringExporterSideCar.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMonitoringExporterSideCar.java
@@ -214,7 +214,7 @@ class ItMonitoringExporterSideCar {
     try {
       // create and verify one cluster mii domain
       logger.info("Create domain and verify that it's running");
-      String modelFile = SessionMigrationUtil.generateSessionMigrYaml("ItMonitoringExporterSide", domain3Uid);
+      String modelFile = SessionMigrationUtil.generateSessionMigrYaml("ItMonitoringExporterSideCar", domain3Uid);
       String miiImage1 = createAndVerifyMiiImage(modelFile);
       String yaml = RESOURCE_DIR + "/exporter/rest_webapp.yaml";
       createAndVerifyDomain(miiImage1, domain3Uid, domain3Namespace, "FromModel", 2, false, yaml, exporterImage);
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMonitoringExporterSideCar.java
```diff
@@ -23,6 +23,7 @@
 import oracle.weblogic.kubernetes.annotations.IntegrationTest;
 import oracle.weblogic.kubernetes.annotations.Namespaces;
 import oracle.weblogic.kubernetes.logging.LoggingFacade;
+import oracle.weblogic.kubernetes.utils.SessionMigrationUtil;
```
anpanigr, line 26: We should create a new Utility Class MonitoringExporterUtils instead of using SessionMigrationUtil in MonitoringExporter usecase.  OR if this is a generic, we can add the method to CommonTestUtil class.

marinakog: I use this method from SessionMigrationUtil because it generates model file with unique domain name String modelFile = SessionMigrationUtil.generateSessionMigrYaml("ItMonitoringExporterSideCar", domain3Uid); Method is implemented to use specific file called model.sessmigr.yaml so I don't think it is generic, this is why I just reuse that method. Why I can't call utils I need in the test?

anpanigr: Do not you think it can be made more generic, which can take any model (template) file not  only model.sessmigr.yaml ( hard-coded) to create a usable model file by replacing the domain id on the fly

anpanigr: As I discussed with Marina, we can make the method generateSessionMigrYaml()  a generic utility  ( in a another tracking JIRA) which will take a template file and generate a usable model file based on the properties to be replaced

anpanigr (commented):
lgtm

russgold (approved):

anpanigr (approved):
lgtm

hzhao-github (approved):

hzhao-github:
LGTM

PR #2689, opened 2022-01-04
OWLS-94849: update create-domain.sh script to handle null or empty pod name due t…

lennyphan:
ENEL uses the built-in WDT scripts (in the samples) for creating the WebLogic domain with Domain-on-PV model.  They are experiencing a timing issue where the pod associated to the create domain job has not been started by the time the script attempts to get the pod name and hence the script execution hangs.  The 'workaround' proposed by ENEL was to add a sleep statement to allow the pod to start.  This pull request resolves the issue by:

1. Retrieving the pod name by continuously checking for the pod, and returning it's name when the pod is started, up to 120 seconds (2 minutes) instead of using a 'sleep'.
2. Fix 'kubectl wait' command which is missing the namespace where the create domain job is running.


kubernetes/samples/scripts/common/utility.sh
```diff
@@ -926,3 +926,27 @@ function checkService(){
...
+ pod=$1
+ ns=$2
+
+ pname=`kubectl get po -n ${ns} | grep -w ${pod} | awk '{print $1}'`
+ while [ -z ${pname} -a $count -le $max ] ; do
+   sleep 1
+   count=`expr $count + 1`
+   pname=`kubectl get po -n ${ns} | grep -w ${pod} | awk '{print $1}'`
+ done
+
```
tbarnes-us, line 945: A slightly more elegant timeout pattern would be: 
max=$((SECONDS + 120))
...
while [ -z ${pname} -a $SECONDS -le $max ] ; do
...
echo "[ERROR] Could not find Pod [$pod] after $max seconds";

(The shell automatically populates SECONDS with the number of seconds since the script started.)

tbarnes-us (approved):
Approved. Just had one minor comment - feel free to address it if you like.

kubernetes/samples/scripts/common/utility.sh
tbarnes-us: Or even:
```
 pname=""
 while [ $count -le $max ] ; do
   pname=`kubectl get po -n ${ns} | grep -w ${pod} | awk '{print $1}'`
   [ -z "${pname}" ] || break
   sleep 1
 done
```

kubernetes/samples/scripts/common/utility.sh
```diff
@@ -926,3 +926,27 @@ function checkService(){
...
+function getPodName(){
+
+ max=120
+ count=1
+
+ pod=$1
+ ns=$2
+
+ pname=`kubectl get po -n ${ns} | grep -w ${pod} | awk '{print $1}'`
+ while [ -z ${pname} -a $count -le $max ] ; do
```
tbarnes-us, line 940: I think ${pname} must be encased in quotes. This will not parse correctly if pname is empty.

tbarnes-us (changes requested):

bhavaniravichandran (approved):

ddsharpe (changes requested):
Copyrights need to be updated to 2022

kubernetes/samples/scripts/common/utility.sh
tbarnes-us: Looks good.

kubernetes/samples/scripts/common/utility.sh
```diff
@@ -926,3 +926,26 @@ function checkService(){
...
+# Get pod name when pod available in a given namespace
+function getPodName(){
+
+ max=$((SECONDS + 120))
+
+ pod=$1
+ ns=$2
+
+ pname=""
+ while [ $SECONDS -le $max ] ; do
```
tbarnes-us, line 939: Minor, but worth it: Please add "local " before each of max, pod, ns, and pname - in case they are globally used somewhere else.

tbarnes-us (approved):
Pre-approving (taking rest of today off). I have added one new minor comment that you probably should address, but it's not a show stopper.

ddsharpe (approved):

robertpatrick:
Does this need to be backported?

lennyphan:
Yes, I will be creating the backport PR to 3.3.  Thanks.

PR #2699, opened 2022-01-06
Owls 95448 Handle WDT archive domainBin expansion in MII

jshum2479:
Currently the domainBin is expanded by WDT during domain creation.  If the domain is updated with a new archive with only files under archives's wlsdeploy/domainBin changes,  the files are not updated in the domain, changes are to expand the bin before starting the domain

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright (c) 2018, 2021, Oracle and/or its affiliates.
+# Copyright (c) 2018, 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 #
 # This script contains the all the function of model in image
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -1054,12 +1054,22 @@ function wdtHandleOnlineUpdate() {
   createFolder "${DOMAIN_HOME}/lib" "This is the './lib' directory within directory 'domain.spec.domainHome'." || exitOrLoop
   for file in $(sort_files ${IMG_ARCHIVES_ROOTDIR} "*.zip")
     do
-        # expand the archive domain libraries to the domain lib
+        # expand the archive domain libraries to the domain lib, 11 is caution when zip entry doesn't exists
         cd ${DOMAIN_HOME}/lib || exitOrLoop
-        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/
+        unzip -jo ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainLibraries/*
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -1054,12 +1054,22 @@ function wdtHandleOnlineUpdate() {
...
 
-        if [ $? -ne 0 ] ; then
-          trace SEVERE  "Error extracting domain lib '${IMG_ARCHIVES_ROOTDIR}/${file}'"
+        # expand the domain bin, in update case user may only update a file in the domainBin archive, 11 is caution when
+        # zip entry doesn't exists
+        cd ${DOMAIN_HOME}/bin || exitOrLoop
+        unzip -jo ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainBin/*
+        if [ $? -ne 0 && $? -ne 11 ] ; then
+          trace SEVERE  "Domain Source Type is FromModel, error in extracting domainBin
+          ${IMG_ARCHIVES_ROOTDIR}/${file}"
```
tbarnes-us, line 1072: I didn't know that you could break a string across multiple lines like this, but don't have access to a shell at the moment to try it out.  Cool. Will it ignore the indent spaces?

OTOH It seems a little subtle and easy to break.  Alternatively, I think trace supports any number of arguments, so you could do this instead:
```
 trace SEVERE  "Domain Source Type is FromModel, error in extracting domainBin" \
          " ${IMG_ARCHIVES_ROOTDIR}/${file}"
```

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -1286,21 +1296,25 @@ function prepareMIIServer() {
...
         fi
 
-        # Flatten the jars to the domain lib root
-
-        if [ -d ${WLSDEPLOY_DOMAINLIB} ] && [ "$(ls -A ${WLSDEPLOY_DOMAINLIB})" ] ; then
-          mv ${WLSDEPLOY_DOMAINLIB}/* .
+        # expand the domain bin, in update case user may only update a file in the domainBin archive, 11 is caution when
+        # zip entry doesn't exists
+        cd ${DOMAIN_HOME}/bin || exitOrLoop
+        unzip -jo ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/domainBin/*
```
tbarnes-us, line 1312: For each unzip, please consider capturing stdout & stderr for each unzip to some unique file (or env var), and then 'cat' the capture just after the SEVERE.  This will provide more diagnostic info in the SEVERE. (When the operator parses out a log message and reports it, it appends everything found up until the next log message).

robertpatrick (approved):

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: There are many lines we simple exit, this kind is rare, I don't think we need every line to be so verbose.

operator/src/main/resources/scripts/modelInImage.sh
jshum2479: Yes, it will, shell knows it's continuation since the double quote has not been ended. I can change it for the easier to read convention

tbarnes-us (approved):

PR #2708, opened 2022-01-10
fix ItMiiSampleWlsAux and ItMiiSampleWlsMain image name structure

xiancao:
Currently the image name uses format mii-<timestamp>-<namespace>:wls-v1. When we push the image to phx.ocir.io, each image will create a new repository. When we hit the repository limit in phx.ocir.io, we can not push the image anymore.

Change the image name use format mii-<itclassname>-wlsv1:<timestamp>-wls-v1. When we push the image, each test will create a repository with different image tag.

Jenkins test result:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7920/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/7919/


kubernetes/samples/scripts/create-weblogic-domain/model-in-image/ai-docker-file/Dockerfile
```diff
@@ -25,5 +25,6 @@ ENV AUXILIARY_IMAGE_PATH=${AUXILIARY_IMAGE_PATH}
 RUN adduser -D -u ${USERID} -G $GROUP $USER
 # ARG expansion in COPY command's --chown is available in docker version 19.03.1+.
 # For older docker versions, change the Dockerfile to use separate COPY and 'RUN chown' commands.
-COPY --chown=$USER:$GROUP ./ ${AUXILIARY_IMAGE_PATH}/
+COPY ./ ${AUXILIARY_IMAGE_PATH}/
+RUN chown -R $USER:$GROUP ${AUXILIARY_IMAGE_PATH}
```
tbarnes-us, line 29: This looks like it reverts a 4.0 change/optimization by @ankedia ?  Did you use a recent enough of version of docker  so the docker file would work?

operator/integration-tests/model-in-image/run-test.sh
```diff
@@ -54,6 +54,7 @@ function usage() {
     WORKDIR               : /tmp/\$USER/mii-sample-work-dir
     DOMAIN_NAMESPACE      : sample-domain1-ns
     MODEL_IMAGE_NAME      : model-in-image 
+    MODEL_IMAGE_TAG       : ${WDT_DOMAIN_TYPE}-v2
```
tbarnes-us, line 57: Defaults to JRF-v1, JRF-v2, WLS-v1, or WLS-v2 depending on use-case and domain type.

tbarnes-us: (` ${WDT_DOMAIN_TYPE}-v2` is incorrect because WDT_DOMAIN_TYPE is hard coded to WLS when , and v2/v1 varies based on the situation).

operator/integration-tests/model-in-image/run-test.sh
```diff
@@ -110,7 +111,7 @@ function usage() {
                     checked into the mii sample git location.
 
     -initial-image: Build image required for initial use case.
-                    Image is named '\$MODEL_IMAGE_NAME:WLS-v1' or '...:JRF-v1'
+                    Image is named '\$MODEL_IMAGE_NAME:\$MODEL_IMAGE_TAG' or '...:\$MODEL_IMAGE_TAG'
```
tbarnes-us, line 114:
```
Image is named '\$MODEL_IMAGE_NAME:WLS-v1' or '...:JRF-v1' if MODEL_IMAGE_TAG not specified.
Image is named '\$MODEL_IMAGE_NAME:\$MODEL_IMAGE_TAG' if MODEL_IMAGE_TAG is specified.
```

operator/integration-tests/model-in-image/run-test.sh
tbarnes-us:
```
      MODEL_IMAGE_TAG       : Defaults to JRF-v1, JRF-v2, WLS-v1, or WLS-v2 
                                                  depending on use-case and domain type.
                                                  IMPORTANT: If setting this env var, do not
                                                  run multiple use cases on the same command line,
                                                  as then all use cases will use the same image.
```

operator/integration-tests/model-in-image/run-test.sh
```diff
@@ -127,7 +128,7 @@ function usage() {
                     Depends on '-initial-main' (calls its app).
 
     -update3-image: Build image required for update3 use case.
-                    Image is named '\$MODEL_IMAGE_NAME:WLS-v2' or '...:JRF-v2'
+                    Image is named '\$MODEL_IMAGE_NAME:\$MODEL_IMAGE_TAG' or '...:JRF-v2'
```
tbarnes-us, line 131:
```
Image is named '\$MODEL_IMAGE_NAME:WLS-v2' or '...:JRF-v2' if MODEL_IMAGE_TAG is not set.
And is named '\$MODEL_IMAGE_NAME:\$MODEL_IMAGE_TAG` if MODEL_IMAGE_TAG is set.'
```

operator/integration-tests/model-in-image/run-test.sh
tbarnes-us, line 57: Also I think it'd be more clear if this was called MODEL_IMAGE_TAG_OVERRIDE and used to override MODEL_IMAGE_TAG where-ever it was set internally. (E.g. leave MODEL_IMAGE_TAG as an internal variable.)

tbarnes-us, line 57: Finally, none of the java IT tests are setting this variable. So how is this change to "run-test.sh" changing IT test behavior?

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/ai-docker-file/Dockerfile
robertpatrick: I agree with Tom, I don't understand why we would split the COPY w/ chown command into two separate commands.

operator/integration-tests/model-in-image/run-test.sh
xiancao: The MODEL_IMAGE_TAG is set in ItMiiSampleHelper.java which is used by ItMiiSampleWlsAux.java, ItMiiSampleWlsMain.java, ItMiiSampleFmwAux.java and ItMiiSampleFmwMain.java. The reason for this change is that in the old way, the tests set image name as format mii-<timestamp>-<namespace>:wls-v1. When we push the image to phx.ocir.io, each image will create a new repository. When we hit the repository limit in phx.ocir.io, we can not push the image anymore.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/ai-docker-file/Dockerfile
xiancao: fixed. I split the command since in my local box, the docker version is lower than 19.03.1.

operator/integration-tests/model-in-image/run-test.sh
xiancao: fixed

operator/integration-tests/model-in-image/run-test.sh
xiancao: fixed

operator/integration-tests/model-in-image/run-test.sh
xiancao: fixed

operator/integration-tests/model-in-image/run-test.sh
xiancao: If MODEL_IMAGE_TAG is not set, it will use the default value.

tbarnes-us: How about instead of replacing the entire generated tag, instead _prepend_ extra information to the tag? There's much descriptive value in the tag that's useful for debugging/tracing/image-tracking:  the generated tag indicates version (v1 or v2), whether or not image is aux, and whether image is WLS or JRF.  Furthermore, the generated tag is intended to differentiate the different use cases run by the test, so when you eliminate the differentiation, I assume the command line is no longer safe to run multiple use cases, since each use case will have the same tag (luckily the IT tests only run one use case at a time, so I assume that helps)...

tbarnes-us: The point is that the usage help is incorrect.  The default changes based on use case (version, whether or not aux is involved, and whether WLS or JRF are involved).  The default is _not_ ${WDT_DOMAIN_TYPE}-v2 except for specific use cases (specific combinations of command-line and env var settings).

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
```diff
@@ -75,9 +76,6 @@
   private String getModelImageName(String suffix) {
     return new StringBuffer(DOMAIN_IMAGES_REPO)
         .append("mii-")
-        .append(currentDateTime)
-        .append("-")
-        .append(domainNamespace)
```
tbarnes-us, line 80: I assume it's this fix that is preventing the explosion of image names, but I thought this was needed to allow parallel runs?  Unless you have removed deletion/cleanup of these objects?

tbarnes-us: (Also, if this is the main cause of the explosion of image names, is it truly necessary to alter the tag logic?)

xiancao: The image name was mii-<timestamp>-<namespace>-wlsv1:WLS-v1. It is changed to mii-<testClassName>-wlsv1:<timestamp>-WLS-v1. To prevent the explosion of the image name and allow parallel runs, to be consistent with other tests, the image name should be fixed and the image tag contains the timestamp.

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
```diff
@@ -288,13 +291,19 @@ public void callInitialUseCase() {
...
-                                       String errString) {
+  public void callUpdateUseCase(String args, String errString, String testClassName) {
+
     if (args.contains("update3")) {
       String imageName = (domainType.equals(DomainType.WLS))
-          ? getModelImageName("-wlsv2") : getModelImageName("-jrfv2");
+          ? getModelImageName(testClassName + "-wlsv2") : getModelImageName(testClassName + "-jrfv2");
       envMap.put("MODEL_IMAGE_NAME", imageName);
+      String decoration = (envMap.get("DO_AI") != null && envMap.get("DO_AI").equalsIgnoreCase("true"))  ? "AI-" : "";
+      envMap.put("MODEL_IMAGE_TAG", MII_BASIC_IMAGE_TAG + "-" + domainType + "-" + decoration + "v2");
```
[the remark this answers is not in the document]
xiancao: MII_BASIC_IMAGE_TAG is set to:
public static final String MII_BASIC_IMAGE_TAG =
      Boolean.valueOf(SKIP_BUILD_IMAGES_IF_EXISTS) ? "local" : getDateAndTimeStamp();

It is actually a current timestamp. So each test run will generate a new image.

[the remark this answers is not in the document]
tbarnes-us: Oh - I missed that. Thanks.

tbarnes-us (approved):
LGTM - Thank you for your patience.

hzhao-github (approved):

hzhao-github:
LGTM

PR #2710, opened 2022-01-12
Ensure all pods have rolled, if needed, before declaring complete/true

russgold:
This addresses issue OWLS-95265.

The problem was that only the RUNNING state was checked, and not the changes that would require a roll. We now mark pods needing a roll during the verifyPod step, by patching it to add a label before adding it to the list of pods to roll.

Jenkins run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8120/


tbarnes-us:
General comment 1: My naive read of the code sees checks in this change for running, start version, and introspect version. The internal specification we have for declaring completed has "All servers expected to be running are ready and at their target image, auxiliary images (formerly known as common mount images), restart version, and introspect version".  So is ready checked? Images? Would it be better to check ready and everything in https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/domain-lifecycle/startup/#fields-that-cause-servers-to-be-restarted ?

General comment 2: IIRC, there's a related integration test that is disabled.  Does this test now pass? Can this change be updated to enable the test?  If there's no test, then we need a JIRA to add one.

russgold:
@tbarnes-us The logic about what to check is found in the new class, RollStateComputation. It tests the image and restart versions as well as the introspect version, as specified in the document WKO Status Conditions 4.0. The only thing that is not present pertains to auxiliary images, which was just redefined. Once that is working, we can add that. 

There is a conceptual problem with expanding this past that: In theory, we should know when any make-right processing is needed, but we also update status during the scheduled status update fiber, which has no information as to which pods are still supposed to roll. A more thorough implementation might run make-right each time, in which case we could check the list of pods that still need to roll.

You mention a disabled integration test. I was not aware of one. Do you know which one it is? The JIRA does not mention it.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -1029,7 +1029,15 @@ private boolean isNonClusteredServer(String serverName) {
       }
 
       private boolean isRunning(@Nonnull String serverName) {
-        return RUNNING_STATE.equals(getRunningState(serverName));
+        return RUNNING_STATE.equals(getRunningState(serverName)) && isRollCompleteFor(serverName);
+      }
+
+      private boolean isRollCompleteFor(@Nonnull String serverName) {
+        return isAdminServer(serverName) || rollStateComputation.isRollCompleteFor(serverName);
```
jshum2479, line 1036: Why we don't need to calculate the rolling state for admin server?

tbarnes-us:
@russgold I can't find an IT either. I've filed two internal JIRAs to track the follow on work. OWLS-95683 which is the conditions dev work that awaits the planned auxiliary image updates, and OWLS-95682 which is a request to create an integration test for the problem fixed in this pull.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: We don't roll the admin server. 

Rolling is something we do with the managed servers to prevent too many from being down at the same time. That means that it is very possible that we know that some managed servers still need to be replaced, but we're not going to do that yet, until others have been. But there is only one admin server, so we replace it immediately when we know we need to.

jshum2479: Really?  If I changed the restartVersion,  I saw all servers rolled including the admin server.  Am I missing something about the use case? @doxiao?

russgold: I expect that you saw all of the server pods _restarted_. We using _rolling_ to mean that the restarts are spread out over time. If the admin server restart doesn't happen for a minute or so after some managed servers are restarted, that would be a problem.

doxiao: @jshum2479 , Yes, when restartVersion is changed for the domain, all servers will be restarted one by one, and the domain can only be marked as Completed when all servers reach the expected state. I think the confusion is about terminology. I noticed that Russ mentioned that "we replace it (admin server) immediately when we know we need to". In implementation, the restart happened in two different ways: for the admin server, the pod is replaced right away (as Russ mentioned in his response), and for the managed servers, the restart is deferred until, all managed server pods are looked at, and then restarted in a rolling fashion.

doxiao: The question is if we need to verify if the admin server has reached the expected state here? I think we probably should.

russgold: Under what conditions would the admin server be running at this point, but _not_ yet in the expected state? Remember that we don't even get to the status update until we've created the replacement admin server, if it needs to be replaced.

doxiao: I was thinking about the periodical status update code path as well.

russgold: And in that code path, the admin server is either not running, or is up-to-date (pending a future make-right).

robertpatrick (approved):

jshum2479 (approved):

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -1029,7 +1028,24 @@ private boolean isNonClusteredServer(String serverName) {
       }
 
       private boolean isRunning(@Nonnull String serverName) {
-        return RUNNING_STATE.equals(getRunningState(serverName));
+        return RUNNING_STATE.equals(getRunningState(serverName)) && isRollCompleteFor(serverName);
```
tbarnes-us, line 1031: Why is running tied to whether or not roll is complete? (I assume "isRunning" feeds into the running status reporting, which makes this change suspect. If the change isn't suspect, can this be changed to "isRunningAndRollComplete" or some-such.)

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -1029,7 +1028,24 @@ private boolean isNonClusteredServer(String serverName) {
...
+      }
+
+      private boolean isRollCompleteFor(@Nonnull String serverName) {
+        return isAdminServer(serverName) || isNotMarkedForRoll(serverName);
```
tbarnes-us, line 1035: Why is roll always complete for admin server?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: If the roll is not complete, the pod is still running an obsolete configuration, and therefore does not count as completed. The name is likely unclear, though, and should be changed.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: We don't roll the admin server. If it needs to be replaced, we do so immediately. We roll the managed server pods so as never to have too many offline at once. Since you are the second person to ask, it seems that a comment or better names are needed.

operator/src/main/java/oracle/kubernetes/operator/helpers/PodHelper.java
```diff
@@ -452,16 +455,34 @@ String getServerName() {
...
+
+    private boolean isPodAlreadyLabeledToRoll(V1Pod pod) {
+      return pod.getMetadata().getLabels().containsKey(LabelConstants.TO_BE_ROLLED_LABEL);
+    }
+
+    private Step createPatchStep() {
+      JsonPatchBuilder patchBuilder = Json.createPatchBuilder();
+      patchBuilder.add("/metadata/labels/" + LabelConstants.TO_BE_ROLLED_LABEL, "true");
+      return new CallBuilder()
+          .patchPodAsync(getPodName(), getNamespace(), getDomainUid(),
```
doxiao, line 481: I am wondering if we should use a sync call here to avoid a possible race condition.

russgold: What kind of race condition are you thinking of?

russgold: I see; if we are also going to roll that particular pod, it could be bad to add this label afterwards. Looking into making it a synchronous call.

russgold: @doxiao Patching the pod synchronously works, and should avoid the problem. Please take a look.

doxiao (approved):
LGTM.

PR #2720, opened 2022-01-19
Owls93995 fix a race condition when DomainNamespaceSelectionStrategy is changed from List to LabelSelector

doxiao:
Problem: If the DomainNamespaceSelectionStrategy is LabelSelector, the operator relies on specifying the LabelSelector on the listNamespaceAsync call to filter the namespaces. When the strategy is changed from List to LabelSelector when the operator is running, listNamespaceAsync call may return the full list of namespaces instead of the namespaces that match the label selector, because the strategy is still List when the call is made. 

Fix: The changes in this PR removes the label selectors from the listNamespaceAsync call so that it always returns the full list, and modifies the isDomainNamespace method, which is used to filter the namespaces that are returned from the listNamespaceAsync call; instead of always returning true, it now actually evaluates the selector in the LabelSelector case.  This approach makes the LabelSelector handling consistent with all other strategies. 

This PR also fixes a NPE that I noticed during testing/debugging in DomainProcessorImpl (in the apply method of DomainPresenceInfoStep).

Integration test results (no unknown failures):
Main branch: 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8066/ 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8080/

This branch:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8067/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8081/

New results to be posted once available.
Main branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8187/
This branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8188/


operator/src/main/java/oracle/kubernetes/operator/Namespaces.java
```diff
@@ -121,15 +125,40 @@ private String getConfiguredNamespaceList() {
       }
 
       @Override
-      public boolean isDomainNamespace(@Nonnull String namespaceName) {
-        return true;  // filtering is done by Kubernetes list call
+      public boolean isDomainNamespace(@Nonnull V1ObjectMeta nsMetadata) {
+        // although filtering is done by Kubernetes list call, there is a rice condition where readExistingNamespaces
```
russgold, line 129: I'm trying to imagine what a "rice" condition could be. :)

Can you explain what this race condition is? Why would K8s give us a non-matching namespace?

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
```diff
@@ -475,6 +475,19 @@ void withLabelSelector_onCreateReadNamespaces_startsNamespaces() {
     assertThat(getStartingNamespaces(), contains(NS_WEBLOGIC1, NS_WEBLOGIC3, NS_WEBLOGIC5));
   }
 
+  @Test
+  void withLabelSelector_onCreateReadNamespaces_ignoreSelectorOnList_startsNamespaces() {
```
russgold, line 479: This test name is unclear. What is it trying to verify?

I'm very concerned about the `ignoreSelectorOnListOperation` call. What K8s behavior is being simulated or tracked?

operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesTestSupport.java
```diff
@@ -801,7 +812,9 @@ Object listResources(String namespace, Integer limit, String cont, String fieldS
     }
 
     private Predicate<Object> withLabels(String[] labelSelectors) {
-      return o -> labelSelectors == null || hasLabels(getMetadata(o), labelSelectors);
+      return o -> (labelSelectors == null
```
russgold, line 815: To check for a null or empty array, use ArrayUtils.isEmpty()

russgold (changes requested):

operator/src/main/java/oracle/kubernetes/operator/Namespaces.java
```diff
@@ -47,11 +49,11 @@
   private static final String ALL_DOMAIN_NAMESPACES = "ALL_DOMAIN_NAMESPACES";
 
   /**
```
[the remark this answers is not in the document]
doxiao: We need the name and labels of the namespace to complete the check, and the metadata has both. We could pass the name and labels as two separate parameters if that reduce the confusion.

operator/src/main/java/oracle/kubernetes/operator/Namespaces.java
doxiao: The intermittent issue only happens when the selection strategy is changed from List to LabelSelector. When the readExistingNamespaceAysnc is called, the strategy may still be List, so the K8S returns all namespaces.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
doxiao: The behavior/condition the change is trying to simulate is when the readExistingNamespace async call is issued, the strategy is List, so the results is as if the selector does not exist although the strategy is LabelSelector when the returned valued are processed.

russgold: I'm not following that; if the strategy has been changed to LabelSelector, why is K8s acting as though it is List? Are we sending the wrong call? Do we have an in-flight list call going when the strategy is changed? If so, maybe we should address that in the list response?

doxiao: yes, the in-flight call uses List. We simulate this race condition by ignoring the selector.

doxiao: In the product code, we do address the issue in the list response as the changes in DomainRecheck.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
```diff
@@ -171,7 +171,7 @@ private boolean useBackupStrategy(CallResponse<V1NamespaceList> callResponse) {
 
     @Override
     public NextAction onSuccess(Packet packet, CallResponse<V1NamespaceList> callResponse) {
-      final Set<String> domainNamespaces = getNamespacesToStart(getNames(callResponse.getResult()));
+      final Set<String> domainNamespaces = getNames(getNamespacesToStart(getNSMetadata(callResponse.getResult())));
```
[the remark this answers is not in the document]
doxiao: I'll try to simplify the logic here.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
doxiao: As to the test method name, what about withLabelSelector_returnAllNamespacesOnCreateReadNamespaces_startsExpectedNamespaces?

operator/src/main/java/oracle/kubernetes/operator/Namespaces.java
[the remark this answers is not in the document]
doxiao: Having thought more about passing the name and labels as two separate parameters, I feel it is probably cleaner to pass namespace metadata as what we have right now.

russgold:
I believe that the current approach is flawed, and requires too many changes, plus is not guaranteed to work in all cases.

The basic problem starts with the reality that our list namespace call is asynchronous, meaning that the strategy used to create the call might not match the strategy when we process the response. This has been identified in a specific case: changing from List to LabelSelector, as the latter does no post-processing, assuming that the filtering is being done by Kubernetes. It should also, then, logically be a problem in other combinations. For example, changing from LabelSelector to any other strategy means that the namespaces we start handling will exclude those without the label, even though we have explicitly said that we want to see them all. In general, any attempt to compensate for the mismatch runs into a problem of this kind. I therefore suggest that we have two simpler approaches, depending on how we want the operator to react. Both require creating the list response step with the current strategy

1. We could simply use the specified strategy for all strategy-dependent computations, thus ensuring that the request and processing are consistent. This would mean that the user’s change of strategies would take effect with the next list operation, not the in-flight one.

2. We could declare a mismatch to be an error, and restart the request. This would cause the change to take effect even with an in-flight request. The difficulty is whether we have a way to restart a multi-step list (used when there are a large number of namespaces), when the change happens partway through. If there is not, I would recommend against this approach.

doxiao:
> I believe that the current approach is flawed, and requires too many changes, plus is not guaranteed to work in all cases.
> 
The PR contains fixes for other existing and unrelated issues that I noticed during testing, such as NPE. The only change for this is to filter the namespaces that the operator get from the listNamesapces call, which matches the handling of all other strategies.

> The basic problem starts with the reality that our list namespace call is asynchronous, meaning that the strategy used to create the call might not match the strategy when we process the response. This has been identified in a specific case: changing from List to LabelSelector, as the latter does no post-processing, assuming that the filtering is being done by Kubernetes. It should also, then, logically be a problem in other combinations. For example, changing from LabelSelector to any other strategy means that the namespaces we start handling will exclude those without the label, even though we have explicitly said that we want to see them all. In general, any attempt to compensate for the mismatch runs into a problem of this kind. I therefore suggest that we have two simpler approaches, depending on how we want the operator to react. Both require creating the list response step with the current strategy
> 
>     1. We could simply use the specified strategy for all strategy-dependent computations, thus ensuring that the request and processing are consistent. This would mean that the user’s change of strategies would take effect with the next list operation, not the in-flight one.
> 
This is error-prune because the operator uses the current selection strategy in many other places.  If the operator continues with the in-flight one here, we need to make sure all other places use the in-flight strategy as well, which is hard to do and error-prune. 

>     2. We could declare a mismatch to be an error, and restart the request. This would cause the change to take effect even with an in-flight request. The difficulty is whether we have a way to restart a multi-step list (used when there are a large number of namespaces), when the change happens partway through. If there is not, I would recommend against this approach.

russgold:
> This is error-prune because the operator uses the current selection strategy in many other places. If the operator continues with the in-flight one here, we need to make sure all other places use the in-flight strategy as well, which is hard to do and error-prune.

In what other cases do you think a problem will occur? It's a problem in the list case because the request and post-processing need to use the same strategy. In most cases, it should not be an issue, as far as I can tell. And if it is indeed a problem, how do the changes here address it?

doxiao:
Having thought more about the problem, I think we should not have used the label selector in the list call for LabelSelector case. We should always list all namespaces (without a selector), and filter the returned list with the selector. All other strategies are handled this way.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
russgold: > Having thought more about the problem, I think we should not have used the label selector in the list call for LabelSelector case. We should always list all namespaces (without a selector), and filter the returned list with the selector. All other strategies are handled this way.

That's fine, and lends itself to a simple solution: NamespaceListResponseStep.onSuccess() can call a new method, `getMetadatas` rather `getNames` and the Namespaces.isDomainNamespace() method can take a @NonNull V1ObjectMeta rather than a strong.

russgold: As to the test, I would suggest that it should now test the change of strategy between the list call and the response. In that case, the thing to add to KubernetesTestSupport would be a `doOnList` call, similar to `doOnCreate`, etc. The test would then set the strategy to, say, List and have the `doOnList` change it to LabelSelection. That could be shown to fail without any other changes.

I would name such a test, `whenSelectionStrategyChangesDuringList_startDesiredNamespaces`.

As a rule, it is better to have the test name describe the desired behavior rather than the implementation of the test.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -1376,7 +1376,7 @@ private Packet createPacket() {
     private class DomainPresenceInfoStep extends Step {
       @Override
       public NextAction apply(Packet packet) {
-        packet.with(DOMAINS.get(getNamespace()).get(getDomainUid()));
+        Optional.ofNullable(DOMAINS.get(getNamespace()).get(getDomainUid())).map(i -> packet.with(i));
```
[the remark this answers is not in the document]
doxiao: Modified.

operator/src/main/java/oracle/kubernetes/operator/DomainRecheck.java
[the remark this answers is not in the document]
doxiao: done

operator/src/main/java/oracle/kubernetes/operator/Namespaces.java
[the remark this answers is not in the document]
doxiao: The method takes a V1Namespace object now.

operator/src/test/java/oracle/kubernetes/operator/MainTest.java
doxiao: we don't need the new test case any more. One of the existing test case covers it already now that we changed the listNamespaces for LabelSelector case.

operator/src/test/java/oracle/kubernetes/operator/helpers/KubernetesTestSupport.java
doxiao: done

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -526,21 +523,21 @@ private NamespaceWatcher createNamespaceWatcher(String initialResourceVersion) {
     return NamespaceWatcher.create(
         threadFactory,
         initialResourceVersion,
-        Namespaces.getLabelSelectors(),
         TuningParameters.getInstance().getWatchTuning(),
         this::dispatchNamespaceWatch,
         new AtomicBoolean(false));
   }
 
   void dispatchNamespaceWatch(Watch.Response<V1Namespace> item) {
```
russgold, line 531: The `metadata` variable is not used after this check, meaning that you can simplify it to a single expression that returns the namespace. The computation is now complex enough to warrant being moved into its own method.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -1376,7 +1376,11 @@ private Packet createPacket() {
     private class DomainPresenceInfoStep extends Step {
       @Override
       public NextAction apply(Packet packet) {
-        packet.with(DOMAINS.get(getNamespace()).get(getDomainUid()));
+        DomainPresenceInfo info =
+            Optional.ofNullable(DOMAINS.get(getNamespace())).map(n -> n.get(getDomainUid())).orElse(null);
+        if (info != null) {
+          info.addToPacket(packet);
+        }
```
russgold, line 1383: Since you already have an Optional, here, there is no need separately to test the resultant info. Instead, you can use `ifPresent` to control the addition of the info to the packet.

russgold (approved):
Looks very good. I have left a couple of stylistic recommendations.

ankedia (approved):
LGTM. I have a minor comment about copyright.

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/K8sEvents.java
```diff
@@ -293,7 +293,7 @@ public static CoreV1Event getOpGeneratedEvent(
             logger.info("Verifying domain event type {0}", type);
             assertTrue(event.getType().equals(type));
             int countAfter = getDomainEventCount(domainNamespace, domainUid, reason, "Normal");
-            return (countAfter == countBefore + 1);
+            return (countAfter >= countBefore + 1);
```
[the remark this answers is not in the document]
doxiao: good catch. fixed

PR #2725, opened 2022-01-20
Owls 93559 enable and fix the disabled test case (no product code change needed)

doxiao:
Enable the disabled test case that verifies when the replicas are set higher that the max number of servers in the cluster, the domain completed condition is false, and Failed condition is true.
Note that the max number of servers is configured as 5 on WebLogic cluster (not 2 as the test code originally assumed), I changed the test code to align with this.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8130/


tbarnes-us (approved):

xiancao:
LGTM

xiancao (approved):

PR #2728, opened 2022-01-24
Status update tuning

russgold:
Addresses:
- OWLS-93144: fix conditions for admin-only
- OWLS-95965: domain Completed condition must check Ready (not Running)
- NOJIRA: domain Available condition must check Ready (not Running)

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -795,7 +795,7 @@ public class DomainStatusUpdater {
         }
 
         boolean isAvailable() {
-          return isClusterIntentionallyShutDown() || sufficientServersRunning();
+          return isClusterIntentionallyShutDown() || sufficientServersReady();
         }
 
         boolean hasTooManyReplicas() {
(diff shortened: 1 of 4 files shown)


tbarnes-us (approved):

tbarnes-us:
**Note:** I see nothing in this change that seems directly related to "Ensure that the 4.0 available condition is false if domain.spec.serverStartPolicy is ADMIN_ONLY" (that's from OWLS-93144).  If this is incorrect, then OWLS-93144 is not completely resolved by this change.

russgold:
> **Note:** I see nothing in this change that seems directly related to "Ensure that the 4.0 available condition is false if domain.spec.serverStartPolicy is ADMIN_ONLY" (that's from OWLS-93144). If this is incorrect, then OWLS-93144 is not completely resolved by this change.

I added unit tests to confirm that behavior; as far as I can tell, it was already working, since the status computation depends on what servers are supposed to be running, which is computed from the start policy.

doxiao:
Can you post Jenkins integration test results?

russgold:
Jenkins run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8207

doxiao (approved):
LGTM

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -828,19 +828,14 @@ private int minReplicas() {
 
         private long numServersRunning() {
```
tbarnes-us, line 829: numServersRunning is misnamed - it's now checking ready

PR #2748, opened 2022-01-28
Certify DB operator tests in Openshift OKD cluster

sankarpn:
In the Openshift OKD k8s cluster we are using it is required to create a storageclass with NFS share provisioning instead of hostPath. This PR adds the necessary storageclass and changes needed to run the tests in OKD.


sankarpn:
will resume work later

PR #2749, opened 2022-01-28
OWLS-96076 - improvements to validateScripts.sh

alai8:
Update validateScripts.sh script that checks for usages of function keyword in shell scripts during maven build:

1. ignore docker-images directory since it comes from different project that WKO itself
2. list problematic file with absolute path, not relative path
3. Report all violations instead of just the first one it finds

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8308/


lennyphan (approved):
LGTM

tbarnes-us (approved):
LGTM2

PR #2771, opened 2022-02-09
Rework domain cycle start/complete events

russgold:
This change involves the following:

1. the domain status now includes a boolean field, `rolling` which is true while the domain is rolling
2. the cycle start event is emitted when we actually begin a pod cycle step while the domain is not already rolling
3. when the cycle start event is emitted, we also set the `rolling` status to true
4. the cycle complete event is emitted when the domain reaches the complete/true state while the domain is rolling
5. reaching the complete/true state clears the `rolling` field.
6. the domain cycle start event no longer attempts to describe the reason for the roll, since we do not have a reliable way to compute that

Some questions here:

1. Should we also remove the computed reason from the pod cycle event, given that we cannot be certain about getting that right, either?
2. The ItPodsRestart test has variations for multiple restart reasons; do we actually need them all?

Test run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8696

Rerun for failures in:
[ItKubernetesEvents](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8708/)
[ItIntrospectVersion](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8709/)


kubernetes/crd/domain-crd.yaml
```diff
@@ -5,7 +5,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
```
robertpatrick, line 5: Please update the copyright date

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServerUpAfterStep.java
```diff
@@ -19,8 +19,6 @@
 import oracle.kubernetes.operator.work.Packet;
```
robertpatrick, line 19: Please update the copyright date

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -49,6 +49,9 @@
       "A brief CamelCase message indicating details about why the domain is in this state.")
```
robertpatrick, line 49: Please update the copyright date

operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdaterTest.java
```diff
@@ -4,8 +4,10 @@
 package oracle.kubernetes.operator;
```
robertpatrick, line 4: Please update the copyright date

operator/src/test/java/oracle/kubernetes/operator/helpers/PodHelperTest.java
```diff
@@ -114,7 +114,7 @@ void afterDeletePodStepRun_runSpecifiedNextStep() {
```
robertpatrick, line 114: Please update the copyright date

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainStatusTest.java
```diff
@@ -232,6 +232,18 @@ void whenConditionRemovedAndNoOtherHasMessage_setDomainStatusMessageNull() {
     assertThat(domainStatus.getReason(), nullValue());
```
robertpatrick, line 232: Please update the copyright date

robertpatrick (approved):

documentation/domains/Domain.json
```diff
@@ -496,6 +496,10 @@
           "description": "A brief CamelCase message indicating details about why the domain is in this state.",
           "type": "string"
         },
+        "rolling": {
+          "description": "True while pods have been intentionally deleted and have not yet returned to the ready state",
```
rjeberhard, line 500: We need a statement about the default value, which I assume is false, for when the field is not populated.

tbarnes-us:
General comments:

A) Why not make rolling a condition?
B) Are there any follow up JIRAs for integration testing this new behavior?
C) As for question #1, "Should we also remove the computed reason from the pod cycle event".  Since you say it can be incorrect, I'm fine with removing it -- administrators should already know why the roll started in the first place...

operator/src/main/java/oracle/kubernetes/operator/helpers/EventHelper.java
```diff
@@ -355,6 +360,11 @@ public String getReason() {
         return DOMAIN_COMPLETED_EVENT;
       }
 
+      @Override
```
doxiao, line 363: Wondering if it is important, certainly helpful, to make sure that all the pod cycling events that are associated with a roll are between the domain roll start event and domain roll complete event, say domain roll start, pod1 cycling, pod2 cycling, ..., domain roll complete.

doxiao: I saw that you only have an explicit ordering value for three event types. Not sure if you need it for roll start and pod cycling as well.

russgold: Oh, I see. The events so ordered were generated simultaneously, and therefore we need to add them to Kubernetes in the appropriate order. Those other events are generated in separate steps and don't therefore need to be sorted.

doxiao (approved):

PR #2772, opened 2022-02-09
Skip validating domain list ability if operator has already read, created, or updated CRD

rjeberhard:
We have functionality in Main that trails the behavior in CRDHelper and additionally does a list of domains to validate the the custom type is available. Oddly, I could consistently reproduce getting a 404 trying to list domains immediately after creating the CRD. This meant that most customers would see a SEVERE message in the operator's log when the operator was started pointing at a fresh cluster. 

Looking at the requirement, we added the domain list as a CRD "presence check" in response to use cases where the operator did not have privilege to read the CRD. We don't actually need this extra check if the operator was able to read, create, or update the CRD.

This PR skips the presence check if the operator was able to read, create, or update the CRD.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Main.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Main.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nonnull;
 
 import io.kubernetes.client.openapi.models.CoreV1EventList;
+import io.kubernetes.client.openapi.models.V1CustomResourceDefinition;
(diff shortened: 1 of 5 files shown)


ankedia:
> Oddly, I could consistently reproduce getting a 404 trying to list domains immediately after creating the CRD. This meant that > most customers would see a SEVERE message in the operator's log when the operator was started pointing at a fresh cluster.
My understanding is that `listDomainAsync` call will return an empty list if CRD is installed but no domains are created. If son, can we have an explict check for the 404 (HTTP_NOT_FOUND) response code and not log the SEVERE error in that case?

ankedia (approved):
LGTM

lennyphan (approved):

PR #2774, opened 2022-02-10
WIP Check server Running state and server pod ready too

doxiao:
Add additional checks that report Failed true when any WebLogic server is not in Running state and any server pod not ready.

Integration test results: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8680/

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
@@ -293,6 +293,40 @@ abstract class DomainStatusUpdateTestBase {
   }
 
   @Test
+  void whenServerIntentionallyNotStarted_reportDomainFailedCondition() {
+    defineScenario()
+        .withCluster("clusterB", "server1", "server2")
+        .withServersReachingState(RUNNING_STATE, "server1", "server2")
+        .build();
+    deactivateServerAndPod(0);
+
+    updateDomainStatus();
+
+    assertThat(getRecordedDomain(), hasCondition(Failed));
+  }
+
+  @Test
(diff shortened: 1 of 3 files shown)


operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
```diff
@@ -292,6 +292,40 @@ void whenServerIntentionallyNotStarted_reportItsStateAsShutdown() {
           hasStatusForServer("server1").withState(SHUTDOWN_STATE).withDesiredState(SHUTDOWN_STATE));
   }
 
+  @Test
```
russgold, line 295: I don't understand the purpose of this test. If a server is _intentionally_ not started, that would not be a failure. Reading the JIRA, I think what you want is to test that the server did not come up within some predefined time, but that's not what you're checking, here - and it's not what this test (or the next) claim to be testing.

russgold (changes requested):

operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
doxiao: See the scenario in https://jira.oraclecorp.com/jira/browse/OWLS-94702 or ItDiagnosticsFailedCondition#testMSBootFailureStatus.

russgold: The JIRA specifies that it should be a failure if they don't start within some (presumably configurable) time. There is nothing time-based in either the tests or the implementation. It is perfectly reasonable for the servers not to have started by the time we next update the status; that should not be considered a failure. At the very least, we need to verify that any started servers which are not yet ready have had enough time.

PR #2777, opened 2022-02-11
OWLS-94847 remove failed conditions only when the introspector job completes

doxiao:
This is to fix an intermittent failure in ItDiagnosticsFailedCondition.testIntrospectorTimeoutFailure. Depends on timing, the introspector timeout condition can be reported in the job pod list response, or in the waitfor-job-ready logic.  In the former case, the operator sets Failed:true condition, deletes and restarts an introspector job immediately. The PodScheduled/Ready notifications of the new job remove the Failed condition almost right after it is set. As a result, the integration test never detects the Failed:true condition.

The fix is to only remove the Failed conditions after the introspector job completes.

Integration test results (no unknown failures reported): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8733/


rjeberhard:
> What do you mean by "the race window"?

> The intent has generally been to remove the failure conditions when we run the introspector, ...

@russgold, the design wiki says that we remove the introspection failure after introspection next succeeds, not just when the next introspection attempt is started. I believe that the code was previously removing the failed condition when introspection was next started, which often gives too little time for the customer to read the failure message.

doxiao:
> What do you mean by "the race window"?
> 
I updated the description to clarify this.

doxiao:
> The intent has generally been to remove the failure conditions when we run the introspector, so that we can detect new failures rather than those in the previous introspection cycle. How are we ensuring that we clear those, now? Maybe we should only remove an _introspection_ failure when we restart it?

We currently remove the Failed conditions too soon. With the change, we clear those in WatchDomainIntrospectorJobReadyStep when we know the introspector job has completed.

russgold (approved):

PR #2782, opened 2022-02-14
OWLS-88984 - add consensus leasing validation to introspector

alai8:
Consensus leasing and whole server migration features are not supported in WKO. This PR adds checks and flags validation errors if whole server migration or feature that requires cluster leasing is configured in a cluster, but cluster is configured to use consensus leasing, or if a leasing datasource is not configured for database leasing.

These validations can be skipped by setting the SKIP_LEASING_VALIDATIONS environment variable.

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8774/


operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -554,6 +608,108 @@ def validateServerTemplateNapListenPortIsSet(self, server_or_template):
...
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(jdbcStore, migratableTarget, "JDBCStore")
+    for fileStore in domain.getFileStores():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(fileStore, migratableTarget, "FileStore")
+    return leasingRequired
+
+  def validateJMSResourceMigrationPolicy(self, jmsResource, migratableTarget, typeStr):
+    leasingRequired = False
+    # valid migration policy values are "Off", "On-Failure", and "Always".
+    # Cluster leasing must be configured for "On-Failure" and "Always".
+    if jmsResource.getMigrationPolicy() != None and jmsResource.getMigrationPolicy() != "Off":
```
tbarnes-us, line 699: I'm not sure the logic is quite right here.

- The MigrationPolicy attribute directly on stores, path services, and bridges cluster targeted, and leasing, should only be checked when these resource are cluster targeted.  Leasing is needed when when this is not "Off".

- The MigrationPolicy on MTs for the JMS case only apply to jms servers, saf agents, stores, and bridges, that are migratable targeted, and only need consensus leasing when the MT migration policy is not "Manual".

The two cases should be treated distinctly - not as an overlapping union.

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -554,6 +608,108 @@ def validateServerTemplateNapListenPortIsSet(self, server_or_template):
...
+    domain = self.env.getDomain()
+    for messagingBridge in domain.getMessagingBridges():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(messagingBridge, migratableTarget, "Messaging bridge")
+    for jdbcStore in domain.getJDBCStores():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(jdbcStore, migratableTarget, "JDBCStore")
+    for fileStore in domain.getFileStores():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(fileStore, migratableTarget, "FileStore")
+    return leasingRequired
+
+  def validateJMSResourceMigrationPolicy(self, jmsResource, migratableTarget, typeStr):
```
tbarnes-us, line 695: The migratableTarget parameter isn't being used. It can be removed.

jshum2479 (approved):

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -554,6 +608,108 @@ def validateServerTemplateNapListenPortIsSet(self, server_or_template):
...
+
+  def validateJMSResourcesLeasing(self, migratableTarget):
+    leasingRequired = False;
+    domain = self.env.getDomain()
+    for messagingBridge in domain.getMessagingBridges():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(messagingBridge, migratableTarget, "Messaging bridge")
+    for jdbcStore in domain.getJDBCStores():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(jdbcStore, migratableTarget, "JDBCStore")
+    for fileStore in domain.getFileStores():
+      leasingRequired |= self.validateJMSResourceMigrationPolicy(fileStore, migratableTarget, "FileStore")
```
tbarnes-us, line 692: Add JMS Servers, SAF Agents, and Path Services to this list.  Technically this isn't needed since their respective stores will have the same target, but it wouldn't hurt just in case it's possible to configure one without a store.

operator/src/main/resources/scripts/introspectDomain.py
alai8: Updated and added check for cluster targeted

operator/src/main/resources/scripts/introspectDomain.py
alai8: Sure. added validations for JMS servers, SAF agents, and path services

operator/src/main/resources/scripts/introspectDomain.py
```diff
@@ -519,11 +566,23 @@ def validateNonDynamicClusterServerHaveSameCustomChannels(self, cluster):
                  self.addError("The WebLogic configured cluster " + self.name(cluster) + " has mismatched network access point " + self.name(nap) + " in servers " + self.name(firstServer) + " and " + self.name(server) + ". All network access points in a cluster must be the same.")
                  return
 
+  def validateNonDynamicClusterLeasing(self, cluster):
+    leasingRequired = self.validateSingletonServices(cluster)
+    for server in self.env.getDomain().getServers():
+      if cluster is self.env.getClusterOrNone(server):
+        leasingRequired |= self.validateJTAMigrationPolicy(server, cluster)
+        self.validateAutoMigrationDisabled(server, cluster)
```
tbarnes-us, line 574: Minor. Consider changing "Auto" to "WholeServer".

tbarnes-us (approved):
Looks good! I especially like the way you abstracted out the error message handling.

Just one minor comment left: up to you if you want to address it.  Approved!

operator/src/main/resources/scripts/introspectDomain.py
alai8: updated as suggested. thanks

PR #2783, opened 2022-02-15
Not throw exception when listJobs fails

sankarpn:
In ItInrospectversion class, test tries to create a domain with creating a Deployment job and waits for it finish to completion. Some times the listJobs api call timesout and test fails when a api exception is thrown. The fix handles the issue by not re throwing the exception and wait until the test timeout.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/8786/


anpanigr (approved):
lgtm

xiancao (approved):
LGTM

PR #2810, opened 2022-03-01
Fix false failures in ItIntrospectVersion tests due to k8s delay in updating the metadata

sankarpn:
The testUpdateImageName test in ItIntrospectVersion class has 2 clusters and when the test updates the image used by the domain resource it rolls all of the pods in the domain. But the test only waits for one of the cluster pods to roll and it causes failures in subsequent test when the subsequent test checks whether the second cluster has rolled.

The fix is to wait for all pods in the domain to roll and conclude the tests.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9012

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
@@ -944,6 +944,8 @@ class ItIntrospectVersion {
             domainUid, domainNamespace));
     assertNotNull(domain1, "Got null domain resource");
     assertNotNull(domain1.getSpec(), domain1 + " /spec is null");
+
+
     logger.info("Getting timestamps for the following pods");
     for (int i = 1; i <= 3; i++) {
       String managedServerPodName = cl2managedServerPodNamePrefix + i;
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
```diff
@@ -1039,11 +1067,11 @@ void testDedicatedModeSameNamespaceScale() {
   /**
    * Update the introspectVersion of the domain resource using lifecycle script.
    * Refer to kubernetes/samples/scripts/domain-lifecycle/introspectDomain.sh
-   * The usecase update the introspectVersion by passing differnt value to -i 
-   * option (non-numeic, non-numeric with space, no value) and make sure that 
-   * the introspectVersion is updated accrodingly in both domain sepc level 
-   * and server pod level. 
-   * It also verifies the intospector job is started/stoped and none of the 
+   * The usecase update the introspectVersion by passing differnt value to -i
+   * option (non-numeic, non-numeric with space, no value) and make sure that
```
anpanigr, line 1071: non-numeic -> non-numeric

anpanigr (commented):
Start a complete suite run.

vanajamukkara (approved):
LG

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItIntrospectVersion.java
sankarpn: fixed typos.

anpanigr (approved):

PR #2811, opened 2022-03-02
Owls94702 handle server pod not ready within a pre-defined period of time

doxiao:
Operator needs to report a Failed condition when expected servers are not all ready within a given period of time. 

The changes include: 
1) Add an additional check for the situation where one or more server pods that are supposed to be up do not get into the ready status within a given period of time. Once the situation is detected, operator a Failed condition to the domain status to report the failure. The clock starts at the later between the creation time of the pod and the lastTransitionTime of the Ready condition to a non-true status (Failed or Unknown). The check is during a make right or a scheduled domain status update.
2) Add  a new configurable, maxReadyWaitTimeSecodns in ServerPod, to specify the maximum time in seconds that the operator waits for a WebLogic server pod to get into ready state before it considers the server pod failed. Default is 30 minutes. This can be specified at the domain spec level, a cluster level or a server level.
3) Add two status fields in ServerStatus to report the phase and the status of the Ready condition of the corresponding server pod to help customers identify the servers that have problem.

Jenkins run with this branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9076/
Jenkins run with main branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9073/


documentation/domains/Domain.md
```diff
@@ -136,6 +136,7 @@ The current status of the operation of the WebLogic domain. Updated automaticall
 | `initContainers` | Array of [Container](k8s1.13.5.md#container) | Initialization containers to be included in the server Pod. See `kubectl explain pods.spec.initContainers`. |
 | `labels` | Map | The labels to be added to generated resources. The label names must not start with "weblogic.". |
 | `livenessProbe` | [Probe Tuning](#probe-tuning) | Settings for the liveness probe associated with a WebLogic Server instance. |
+| `maxReadyWaitTimeSeconds` | number | The maximum time in seconds that the operator waits for a WebLogic server pod to get into ready state before it considers the server pod failed. Defaults to 1800 seconds. |
```
rjeberhard, line 139: "WebLogic server" -> "WebLogic Server"
"to get into" -> "to reach the"
last "server pod" -> "pod"

rjeberhard:
Everything looks good except for a few comments on descriptions and messages. @rosemarymarano, would you mind please taking a look to see if I've missed anything?

documentation/domains/Domain.md
```diff
@@ -190,6 +191,8 @@ The current status of the operation of the WebLogic domain. Updated automaticall
 | `desiredState` | string | Desired state of this WebLogic Server instance. Values are RUNNING, ADMIN, or SHUTDOWN. |
 | `health` | [Server Health](#server-health) | Current status and health of a specific WebLogic Server instance. |
 | `nodeName` | string | Name of Node that is hosting the Pod containing this WebLogic Server instance. |
+| `podPhase` | string | Phase of server pod. Possible values are: Pending, Succeeded, Failed, Running, or Unknown. |
```
rosemarymarano, line 194: Phase of server pod. -> Phase of the server pod.

tbarnes-us:
General comments per our slack:  

- A Update the pull description to reflect that the timeout is actually 30 minutes, not 30 seconds. (Already done.)

- B Please also update the pull description to describe the details about "when the clock starts" and "when the clock is checked".

- C I don't know if we want to do anything about this, but there will be wide variation in the time it takes to detect if the timeout has passed, which would be especially noticeable if the timeout was set to a lower value. This is because some of the clock checks occur during make-right. Perhaps the doc should state that it might take up to two minutes for the operator to act on a timeout? Or perhaps we can ignore this detail?

- D It seems like it could be problematic that the "clock start" can start if ready is 'unknown'. Since 'unknown' might be normal.  Is 'unknown' common? What is the advantage of checking 'unknown'?

documentation/domains/Domain.md
```diff
@@ -190,6 +191,8 @@ The current status of the operation of the WebLogic domain. Updated automaticall
...
+| `podReady` | string | Status of server pod's condition type Ready if the pod is in Running phase, otherwise Unknown. Possible values are: True, False or Unknown. |
```
rosemarymarano, line 195: Status of server pod's condition type Ready if the pod is in Running phase, otherwise Unknown. -> (I'm having a hard time interpreting this description, perhaps, due to a lack of punctuation.) -> Do you mean, "Status of the server pod's condition type; Ready if the pod is in the Running phase, otherwise Unknown."

documentation/domains/index.html
```diff
@@ -1805,6 +1765,50 @@
           "description": "If specified, the Pod will be dispatched by the specified scheduler. If not specified, the Pod will be dispatched by the default scheduler. See `kubectl explain pods.spec.schedulerName`.",
           "type": "string"
         },
+        "maxReadyWaitTimeSeconds": {
+          "description": "The maximum time in seconds that the operator waits for a WebLogic server pod to get into ready state before it considers the server pod failed. Defaults to 1800 seconds.",
```
rosemarymarano, line 1769: WebLogic server -> WebLogic Server (globally)

documentation/domains/index.html
```diff
@@ -1852,10 +1856,18 @@
...
+        "podReady": {
+          "description": "Status of server pod\u0027s condition type Ready if the pod is in Running phase, otherwise Unknown. Possible values are: True, False or Unknown.",
+          "type": "string"
+        },
         "desiredState": {
           "description": "Desired state of this WebLogic Server instance. Values are RUNNING, ADMIN, or SHUTDOWN.",
           "type": "string"
         },
+        "podPhase": {
+          "description": "Phase of server pod. Possible values are: Pending, Succeeded, Failed, Running, or Unknown.",
```
rosemarymarano, line 1868: Phase of server pod -> Phase of the server pod

rosemarymarano (commented):
Few edits, one clarification needed.

integration-tests/src/test/java/oracle/weblogic/domain/ServerStatus.java
```diff
@@ -27,9 +28,19 @@
   @ApiModelProperty("Name of node that is hosting the Pod containing this WebLogic Server.")
   private String nodeName;
 
+  @ApiModelProperty("Phase of server pod. Possible values are: Pending, Succeeded, Failed, Running, or Unknown.")
+  @Expose
+  private String podPhase;
+
```
tbarnes-us, line 34: server -> WebLogic Server 

ditto for podReady

(follows the pattern of the the other status field descriptions)

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
```diff
@@ -542,6 +543,50 @@ public static boolean verifyRollingRestartOccurred(Map<String, OffsetDateTime> p
...
+                                                                             String podReadyStatus) {
+    LoggingFacade logger = getLogger();
+
+    return () -> {
+      oracle.weblogic.domain.Domain domain =
+          assertDoesNotThrow(() -> getDomainCustomResource(domainUid, domainNamespace));
+
+      if (domain != null && domain.getStatus() != null) {
+        List<ServerStatus> serverStatusList = domain.getStatus().getServers();
+        logger.info(Yaml.dump(serverStatusList));
```
tbarnes-us, line 569: (1) This, and the other logger.info should be changed to something like 'fine'/'finer'/'finest' I assume? Or even removed?
(2) If keeping them, I think they should have more context in their messages...

documentation/domains/Domain.md
doxiao: Ready is a pod condition type.

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
doxiao: Note that this is a test logger. This has followed other methods. It is helpful in debugging a test failure.

documentation/domains/Domain.md
doxiao: how about "Status of server pod's Ready condition if the pod is in Running phase, otherwise Unknown."?

rosemarymarano: @doxiao So, then does it make sense to write it as I suggested? "Status of the server pod's condition type; (semi-colon)  Ready if the pod is in the Running phase, otherwise Unknown."

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/TestAssertions.java
tbarnes-us: Oops! I thought this was in the main code base.

tbarnes-us:
One more general comment:

E: IMO, it'd be helpful to document the new feature in the sameplace we discuss readiness probe customization: 
  ./content/userguide/managing-domains/domain-lifecycle/liveness-readiness-probe-customization.md

documentation/domains/Domain.md
rosemarymarano: @doxiao oh, now I understand. Yes, your suggestion is ok. "Status of the server pod's Ready condition if the pod is in Running phase, otherwise Unknown."

doxiao: fixed. thanks

doxiao:
> One more general comment:
> 
> E: IMO, it'd be helpful to document the new feature in the sameplace we discuss readiness probe customization: ./content/userguide/managing-domains/domain-lifecycle/liveness-readiness-probe-customization.md

Hmm, I don't quite see what to put there. This is not directly related to what is in that doc. Can you propose what you would like to add there?

doxiao:
> General comments per our slack:
> 
>     * A Update the pull description to reflect that the timeout is actually 30 minutes, not 30 seconds. (Already done.)
> 
>     * B Please also update the pull description to describe the details about "when the clock starts" and "when the clock is checked".
Done.
> 
>     * C I don't know if we want to do anything about this, but there will be wide variation in the time it takes to detect if the timeout has passed, which would be especially noticeable if the timeout was set to a lower value. This is because some of the clock checks occur during make-right. Perhaps the doc should state that it might take up to two minutes for the operator to act on a timeout? Or perhaps we can ignore this detail?
Hmm, I am not sure if we could do anything that is really helpful for this. 
> 
>     * D It seems like it could be problematic that the "clock start" can start if ready is 'unknown'. Since 'unknown' might be normal.  Is 'unknown' common? What is the advantage of checking 'unknown'?

We are not sure what the pod's state really is when its Ready status is Unknown. We consider the pod is ready when it is in Running phase and the Ready condition is "True". Anything else (Ready:False or Ready:Unknown) is considered not ready.

rosemarymarano (approved):
LGTM

tbarnes-us:
>> Tom: D It seems like it could be problematic that the "clock start" can start if ready is 'unknown'. Since 'unknown' might be normal.  Is 'unknown' common? What is the advantage of checking 'unknown'?
> Dongbo: We are not sure what the pod's state really is when its Ready status is Unknown. We consider the pod is ready when it is in Running phase and the Ready condition is "True". Anything else (Ready:False or Ready:Unknown) is considered not ready.

I am worried that 'Unknown' might simply reflect a problem in k8s rather than a problem with WL itself - we don't want to accidentally kill pods that are actually just running fine.

tbarnes-us:
>> 
>> E: IMO, it'd be helpful to document the new feature in the same place we discuss readiness probe customization: ./content/userguide/managing-domains/domain-lifecycle/liveness-readiness-probe-customization.md
>
> Hmm, I don't quite see what to put there. This is not directly related to what is in that doc. Can you > propose what you would like to add there?

I think the new feature should be mentioned there because one of the reasons for "probe customization" is to handle the use case that the new configurable now covers. If you like, I can make a suggested change in a branch on your branch, and then you can choose whether to merge.

doxiao:
> > > Tom: D It seems like it could be problematic that the "clock start" can start if ready is 'unknown'. Since 'unknown' might be normal.  Is 'unknown' common? What is the advantage of checking 'unknown'?
> 
> > Dongbo: We are not sure what the pod's state really is when its Ready status is Unknown. We consider the pod is ready when it is in Running phase and the Ready condition is "True". Anything else (Ready:False or Ready:Unknown) is considered not ready.
> 
> I am worried that 'Unknown' might simply reflect a problem in k8s rather than a problem with WL itself - we don't want to accidentally kill pods that are actually just running fine.
I understand your concern. But when there is a k8s problem for the operator to get the pod status, we don't know if it is more likely that the pod is ready or not ready. There are 50% chance that we may get it wrong. In addition, we don't usually immediately delete a server pod after a failure is detected. In most of the cases, the operator will retry, and the condition may resolve on subsequent retries.

doxiao:
> > > E: IMO, it'd be helpful to document the new feature in the same place we discuss readiness probe customization: ./content/userguide/managing-domains/domain-lifecycle/liveness-readiness-probe-customization.md
> > 
> > 
> > Hmm, I don't quite see what to put there. This is not directly related to what is in that doc. Can you > propose what you would like to add there?
> 
> I think the new feature should be mentioned there because one of the reasons for "probe customization" is to handle the use case that the new configurable now covers. If you like, I can make a suggested change in a branch on your branch, and then you can choose whether to merge.

Sure, if you don't mind. thanks

documentation/domains/Domain.md
```diff
@@ -190,6 +191,8 @@ The current status of the operation of the WebLogic domain. Updated automaticall
 | `desiredState` | string | Desired state of this WebLogic Server instance. Values are RUNNING, ADMIN, or SHUTDOWN. |
 | `health` | [Server Health](#server-health) | Current status and health of a specific WebLogic Server instance. |
 | `nodeName` | string | Name of Node that is hosting the Pod containing this WebLogic Server instance. |
+| `podPhase` | string | Phase of the WebLogic Server pod. Possible values are: Pending, Succeeded, Failed, Running, or Unknown. |
+| `podReady` | string | Status of the WebLogic Server pod's Ready condition if the pod is in Running phase, otherwise Unknown. Possible values are: True, False or Unknown. |
```
rosemarymarano, line 195: pod is in Running phase -> pod is in the Running phase

documentation/domains/index.html
```diff
@@ -1805,6 +1765,50 @@
...
+        },
+        "env": {
+          "description": "A list of environment variables to set in the container running a WebLogic Server instance. More info: https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/domain-resource/#jvm-memory-and-java-option-environment-variables. See `kubectl explain pods.spec.containers.env`.",
+          "type": "array",
+          "items": {
+            "$ref": "https://github.com/garethr/kubernetes-json-schema/blob/master/v1.13.5/_definitions.json#/definitions/io.k8s.api.core.v1.EnvVar"
+          }
+        },
+        "restartPolicy": {
+          "description": "Restart policy for all containers within the Pod. One of Always, OnFailure, Never. Default to Always. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy. See `kubectl explain pods.spec.restartPolicy`.",
```
rosemarymarano, line 1795: One of Always, OnFailure, Never -> Possible values are: Always, OnFailure, or Never

documentation/domains/index.html
```diff
@@ -1852,10 +1856,18 @@
           "description": "Name of Node that is hosting the Pod containing this WebLogic Server instance.",
           "type": "string"
         },
+        "podReady": {
+          "description": "Status of the WebLogic Server pod\u0027s Ready condition if the pod is in Running phase, otherwise Unknown. Possible values are: True, False or Unknown.",
```
rosemarymarano, line 1860: is in Running phase -> is in the Running phase

documentation/staging/content/userguide/managing-domains/domain-lifecycle/liveness-readiness-probe-customization.md
```diff
@@ -83,15 +83,26 @@ The following operator-populated environment variables are available for use in
...
-```yaml
-  serverPod:
-    readinessProbe:
-      periodSeconds: 10
-      timeoutSeconds: 10
-      failureThreshold: 3
-```
+- By default, the readiness probe is configured to check readiness every 5 seconds, to timeout after 5 seconds, and to perform the first check after 30 seconds. The default success and failure thresholds values are 1. You can customize the readiness probe initial delay, interval, timeout, success and failure thresholds using the `readinessProbe` attribute under the `serverPod` element of the domain resource.
+
+  Following is an example configuration to change readiness probe interval, timeout and failure threshold value.
```
rosemarymarano, line 92: timeout  -> timeout, (comma)

rosemarymarano (commented):
Minor edits

tbarnes-us (approved):

PR #2816, opened 2022-03-04
Owls 95632 - Automated upgrade of WKO 3.x with legacy auxiliary images config to auxiliary images 4.0  

ankedia:
Changes to support both the runtime and manual conversion of 3.x auxiliary image configuration to "raw" serverPod init-container/volume configuration in a 4.0/v9-schema. 
- Uses conversion webhook to automatically upgrade 3.x/v8-schema auxiliary image to 4.0/v9-schema aux images.
   - Implements a new Webhook REST Server and the REST end-point conversion webhook.
   - Creates a new webhook deployment using REST endpoint of the conversion webhook as the entry point.
   - The conversion webhook deployment generates a self-signed CA certificate to be used in CRD.
-  Auxiliary image scripts modified to support both 3.x and 4.0 auxiliary images.
- Changes to prevent pods from rolling when upgrading the Operator from 3.x to 4.0
- Initial implementation of a standalone tool to manually convert v8 domain resource to v9 domain resource.
Latest Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9284/ , https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9287/ and https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9288/.


operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionResponse.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    ConversionResponse that = (ConversionResponse) o;
+    return Objects.equals(uid, that.uid)
+            && Objects.equals(convertedObjects, that.convertedObjects)
+            && Objects.equals(result, that.result);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(uid, convertedObjects, result);
```
tbarnes-us, line 85: Minor Nit: I'm not familiar with REST programming - but it seems odd that the request's and response's hash and equals methods depend on values that can change via 'setters'. The usual practice is to make all such values final and set them in a constructor (to eliminate the chance of mutations after putting such objects in a hash map).

operator/src/main/resources/Operator.properties
```diff
@@ -157,6 +157,14 @@ WLSKO-0201=WL pod shutdown: Failed to shutdown WebLogic server {0} via REST inte
 WLSKO-0202=WL pod shutdown: Failed to shutdown WebLogic server {0} via REST interface due to HTTP request timeout after {1} seconds.
 WLSKO-0203=WL pod shutdown: Failed to shutdown WebLogic server {0} via REST interface due to exception: {1}.
 WLSKO-0204=WL pod shutdown: Retry shutdown of WebLogic server {0} via REST interface.
+WLSKO-0205=Conversion Webhook For Oracle WebLogic Kubernetes Operator, version: {0}, implementation: {1}, build time: {2}
```
tbarnes-us, line 160: I wonder if give the webhook a precise name in all public references. This not only future proofs the feature in case additional webhooks are added later, but it helps make its purpose more clear, and sets a consistent name for documentation purposes. If so, then using "WebLogic Conversion Webhook" (WCW) or "Domain Conversion Webhook" (DCW) (or similar) might be adequate, and should be used in place of just plain 'webhook' in the other log messages and anywhere else that is public.

operator/src/main/resources/scripts/utils_base.sh
```diff
@@ -357,6 +357,37 @@ checkSourceWDTInstallDirExistsAndNotEmpty() {
   fi
 }
 
+#
```
tbarnes-us, line 360: The legacy "initAuxiliaryImage" function was not reverted back to the original? It looks like it has been repurposed and is now used for the new approach. I think this risks pods that started before the operator was converted -- they will have their scripts changed out from underneath them, leading to unpredictable behavior. The old functions should all be left in place and unchanged - plus hopefully leveraged for the 'converted' domains. New functions should be added for handling the new approach -- instead of re-using the function names of the old approach.

operator/src/main/resources/scripts/auxImage.sh
```diff
@@ -32,8 +32,17 @@ UNKNOWN_SHELL=true
...
+  cat /tmp/auxiliaryImage.out
+
+  mkdir -p ${AUXILIARY_IMAGE_TARGET_PATH}/auxiliaryImageLogs
+  cp /tmp/auxiliaryImage.out ${AUXILIARY_IMAGE_TARGET_PATH}/auxiliaryImageLogs/${AUXILIARY_IMAGE_CONTAINER_NAME}.out
+elif [[ "$AUXILIARY_IMAGE_CONTAINER_NAME" == "compatibility-mode-operator-aux-container"* ]]; then
+  initCompatibilityModeInitContainersWithLegacyAuxImages > /tmp/compatibilityModeInitContainers.out 2>&1
+  cat /tmp/compatibilityModeInitContainers.out
+
+  mkdir -p ${AUXILIARY_IMAGE_TARGET_PATH}/compatibilityModeInitContainerLogs
+  cp /tmp/compatibilityModeInitContainers.out ${AUXILIARY_IMAGE_TARGET_PATH}/compatibilityModeInitContainerLogs/${AUXILIARY_IMAGE_CONTAINER_NAME}.out
```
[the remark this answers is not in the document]
tbarnes-us: For code clarity, maybe have a different file, 'auxImageV2.sh' for new approach, and leave 'auxImage.sh' unchanged.  For safety, IMO, it should remain unchanged regardless for 'old' pods...

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
```diff
@@ -0,0 +1,386 @@
...
+      return domain;
+    }
+    LOGGER.fine("Converting domain " + domain + " to " + desiredAPIVersion + " apiVersion.");
+
+    String apiVersion = (String)domain.get("apiVersion");
+    Map<String, Object> adminServerSpec = (Map<String, Object>) spec.get("adminServer");
+    Boolean adminChannelPortForwardingEnabled = (Boolean)Optional.ofNullable(adminServerSpec)
+            .map(as -> as.get("adminChannelPortForwardingEnabled")).orElse(null);
+    if ((adminChannelPortForwardingEnabled == null) && (API_VERSION_V8.equals(apiVersion))) {
+      Optional.ofNullable(adminServerSpec).map(as -> as.put("adminChannelPortForwardingEnabled", false));
```
tbarnes-us, line 99: Why does new code for converting aux volumes need to handle 'admin channel port forwarding enabled'?

operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionResponse.java
ankedia: Not sure if I'm understanding this correctly. Most of the model classes (e.g. ProbeTuning or ServerPod etc.) has hash and equals method that depend on values that can change via 'setters'. In fluent style of coding, we don't set values using constructor but have methods such as below to construct the object. 
```
    new ProbeTuning()
        .initialDelaySeconds(initialDelay)
        .timeoutSeconds(timeout)
        .periodSeconds(period);
```

operator/src/main/resources/Operator.properties
ankedia: Since the conversion webhook is used only in the context of the WebLogic Operator, how about "WebLogic Operator Conversion Webhook" or "WKO Conversion Webhook"? I'm fine with other names as well. @rjeberhard ?

operator/src/main/resources/scripts/auxImage.sh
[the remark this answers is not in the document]
ankedia: I have tested the current scripts and they work with both the old and new pods. I can create a different 'auxImageV2.sh' for the code clarity as you suggested.

operator/src/main/resources/scripts/utils.sh
```diff
@@ -735,3 +735,64 @@ checkAuxiliaryImage() {
...
+# checkCompatibilityModeInitContainersWithLegacyAuxImages
+#   purpose: If the AUXILIARY_IMAGE_PATH directory exists, it echoes the contents of output files
+#            in ${AUXILIARY_IMAGE_PATH}/compatibilityModeInitContainerLogs dir. It returns 1 if a SEVERE message
+#            is found in any of the output files in ${AUXILIARY_IMAGE_PATH}/compatibilityModeInitContainerLogs
+#            dirs. It also returns 1 if 'successfully' message is not found in the output files
+#            or if the AUXILIARY_IMAGE_PATH directory is empty. Otherwise it returns 0 (success).
+#            See also 'auxImage.sh'.
+#            See also initCompatibilityModeInitContainersWithLegacyAuxImages in 'utils_base.sh'.
+#
+checkCompatibilityModeInitContainersWithLegacyAuxImages() {
```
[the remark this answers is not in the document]
ankedia: We don't populate certain environment variables such as `AUXILIARY_IMAGE_PATHS` and `AUXILIARY_IMAGE_PATH` for the init containers created for auxiliary images using new approach. Also previously, we were checking if there are no files in `$AUXILIARY_IMAGE_PATH` location then script was failing with a SEVERE error. That concept doesn't exist in the new auxiliary image approach.

operator/src/main/resources/scripts/utils_base.sh
ankedia: I'm not sure if I understand this correctly. I have tested the scripts with both the 3.x auxiliary images and new 4.0 auxiliary images and it works fine in both cases. The 3.x auxiliary image pods will always be converted to "raw" init containers using special names (starting with compatibility-mode-operator-aux-container) before running in the 4.0 operators. I don't see a need to keep the old functions in place and unchanged.

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
ankedia: This is for disabling the port forwarding by default in 4.0 if it was not explicitly enabled in 3.x. This is to prevent pods from rolling during operator upgrade scenario.

tbarnes-us: ok

operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionResponse.java
tbarnes-us: ok.

operator/src/main/resources/scripts/auxImage.sh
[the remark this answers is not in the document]
ankedia: As suggested, I have added an 'else' to fail if the container name is unexpeted.

operator/src/main/resources/scripts/utils_base.sh
tbarnes-us: "The 3.x auxiliary image pods will always be converted to "raw" init containers using special names"

The old "already running" pods could theoretically still be there, right? And it's not clear to me if a subsequent external pod-restart will cause them to run through the webhook, or simply restart the pod without operator/webhook intervention. So it seems risky to change the scripts out from underneath these pods...

ankedia: If the pod is externally deleted by the customer, it'll be recreated by the Operator based on converted domain resource yaml i.e. using the specs containing "raw" init containers with special names.

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestServer.java
```diff
@@ -0,0 +1,319 @@
...
+        // if we didn't get a chance to start all of the ports because an exception
+        // was thrown, then stop the ones we did manage to start
+        stop();
+      }
+    }
+    LOGGER.exiting();
+  }
+
+  /**
+   * Stops WebLogic operator's REST api.
```
jshum2479, line 193: minor,  it is stopping the webhook rest server not the operator's rest API?

operator/src/main/resources/Operator.properties
rjeberhard: Kubernetes calls this a "custom resource conversion webhook." The full name could be something like, "WebLogic Domain Custom Resource Conversion Webhook" and then in the rest of the document we can just call it the "webhook" or "conversion webhook."

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestServer.java
rjeberhard: These should be two different REST servers.

rjeberhard:
@ankedia, just about to go through in detail... Here are some items that I'll be looking for (likely these aren't all addressed yet, but listing them now so that we can discuss):

1. The webhook deployment can handle having replicas > 1 (I would probably default to having 2 replicas)

Once you have 2 or more replicas then multiple webhook instances may be starting in parallel. They may each create a cert and try to update the CRD. One of them will "win" and update the CRD first. The other will get a HTTP 409 / Conflict. The "loser" should respond by reading the cert & key from whatever Secret where the winner stored it and using those values rather than the one that it had generated. Similarly, webhook instances started later should read the existing key/cert from the Secret. (The conflict might be on creating/updating that Secret rather than the CRD.)

2. The "uninstall" problem

After I uninstall the webhook, I expect that the CRD will be stranded so that Domain resources don't get deleted. There needs to be a way to reinstall the webhook, perhaps in a different namespace, and have this new installation of the webhook "know" that it is responsible for updating the CRD.

3. v8 fidelity

A 3.x operator (presumably 3.4) is running in a cluster where a 4.0 webhook has been installed. This operator will be reading and storing Domain resources as v8. Since the storage version is v9, the webhook will be converting Domains back and forth between v8 and v9 as the 3.x operator works. Are there any unexpected changes to the Domain when the 3.x operator reads it? There may not be an easy way to "restore" a Progressing status condition that existed.

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -165,3 +167,170 @@ spec:
...
+    # webhook does not exist or chart version is newer, create a new webhook
+    apiVersion: "v1"
+    kind: "ConfigMap"
+    metadata:
+      labels:
+        weblogic.webhookName: {{ .Release.Namespace | quote }}
+      name: "weblogic-webhook-cm"
+      namespace: {{ .Release.Namespace | quote }}
+    data:
+      serviceaccount: {{ .serviceAccount | quote }}
```
rjeberhard, line 183: I'm assuming that this is the ConfigMap where the webhook will store its certificate and keys, correct? If so, consider preserving the values from any earlier version of the map, if present, just like the operator does for its internal certificate:

```
  {{- $configmap := (lookup "v1" "ConfigMap" .Release.Namespace "weblogic-operator-cm") }}
  {{- if (and $configmap $configmap.data) }}
  {{- $internalOperatorCert := index $configmap.data "internalOperatorCert" }}
  {{- if $internalOperatorCert }}
  internalOperatorCert: {{ $internalOperatorCert }}
  {{- end }}
  {{- end }}
```

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
```diff
@@ -165,3 +167,170 @@ spec:
...
+        "helm.sh/resource-policy": keep
+        "helm.sh/hook-delete-policy": "before-hook-creation"
+      {{- end }}
+    spec:
+      strategy:
+        type: Recreate
+      selector:
+        matchLabels:
+          weblogic.webhookName: {{ .Release.Namespace | quote }}
+      replicas: 1
```
rjeberhard, line 206: As described in my review note, I think that this should be "2" and we should resolve startup related concurrency issues.

kubernetes/charts/weblogic-operator/templates/_operator-internal-svc.tpl
```diff
@@ -17,4 +17,19 @@ spec:
   ports:
     - port: 8082
       name: "rest"
+---
+apiVersion: "v1"
+kind: "Service"
+metadata:
+  name: "internal-weblogic-operator-webhook-svc"
```
rjeberhard, line 24: Why "internal"? The operator has two different services with the external one meant to possibly be accessed from out side the cluster, but the webhook will only ever be accessed from the k8s master. I think you can just remove "internal-"

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
rjeberhard: Ignore this... You already used the suggestion in the webhook's secret.

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -159,6 +159,12 @@ javaLoggingFileCount: 10
 # the operator's deployment. See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector
 # for more information on node selectors.
 #nodeSelector:
+ 
+# webhookOnly - Only install the webhook and skip deploying the operator.
```
rjeberhard, line 163: Change the description of these two fields to be similar to how the descriptions of other fields are worded, e.g. "webhoookOnly specifies that ..."

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -192,6 +198,20 @@ internalDebugHttpPort: 30999
 # Otherwise, it is ignored.
 externalDebugHttpPort: 30999
 
+# internalWebhookDebugHttpPort specifies the port number inside the Kubernetes cluster for the operator's Java
```
rjeberhard, line 201: "operator's" -> "webhook's"

operator/pom.xml
```diff
@@ -108,7 +108,8 @@
             <configuration>
               <includeScope>runtime</includeScope>
               <outputDirectory>${project.build.directory}/lib</outputDirectory>
-              <excludeGroupIds>io.prometheus,com.google.protobuf</excludeGroupIds>
+              <!-- excludeGroupIds>io.prometheus,com.google.protobuf</excludeGroupIds -->
+              <excludeGroupIds>io.prometheus</excludeGroupIds>
```
rjeberhard, line 112: Why this change? Are you using protobuf somewhere? Also, if you really are then we don't need the commented out line.

operator/pom.xml
```diff
@@ -414,6 +415,11 @@
       <artifactId>simplestub</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>commons-cli</groupId>
```
rjeberhard, line 419: What is this dependency? Do we need it? There is, sadly, a complicated process with CoreArch and Legal to get approval to have new dependencies. If you really need it then we can go through the process.  

If we do add it, then this needs to be added to the top-level pom following the pattern of other dependencies and have the version as a property there.

operator/src/main/java/oracle/kubernetes/operator/DomainResourceConverter.java
```diff
@@ -0,0 +1,112 @@
...
+    this.outputFileName = Optional.ofNullable(outputFileName)
+            .orElse("Converted_" + new File(inputFileName).getName());
+    this.inputFileName = inputFileName;
+  }
+
+  private static DomainResourceConverter parseCommandLine(String[] args) {
+    CommandLineParser parser = new DefaultParser();
+    Options options = new Options();
+
+    Option helpOpt = new Option("h", "help", false, "Print help message");
```
rjeberhard, line 81: Can these descriptions come from a resource bundle?

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -467,36 +467,64 @@ Step createCRDPresenceCheck() {
...
-  class CrdPresenceResponseStep extends DefaultResponseStep<DomainList> {
+  // on failure, waits for the CRD to be installed.
+  private class CrdPresenceResponseStep extends DefaultResponseStep<DomainList> {
 
     CrdPresenceResponseStep(Step next) {
       super(next);
     }
 
     @Override
     public NextAction onSuccess(Packet packet, CallResponse<DomainList> callResponse) {
```
rjeberhard, line 520: With the removal of warnedOfCrdAbsence then you don't need to define onSuccess() at all as it just calls the super method.

operator/src/main/java/oracle/kubernetes/operator/WebhookLiveness.java
```diff
@@ -0,0 +1,34 @@
...
+
+import oracle.kubernetes.operator.logging.LoggingFacade;
+import oracle.kubernetes.operator.logging.LoggingFactory;
+import oracle.kubernetes.operator.logging.MessageKeys;
+
+/**
+ * This task maintains the "liveness" indicator so that Kubernetes knows the Webhook is still
+ * alive.
+ */
+public class WebhookLiveness implements Runnable {
```
rjeberhard, line 18: This is another copy... Similarly, there could be one Liveness class that takes constructor arguments for the differences.

operator/src/main/java/oracle/kubernetes/operator/WebhookReady.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator;
+
+import java.io.File;
+import java.io.IOException;
+
+/** This task creates the "readiness" indicator so that Kubernetes knows the Operator is ready. */
+public class WebhookReady {
```
rjeberhard, line 10: Copied comment for copied code 😄

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestConfig.java
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.rest;
+
+/**
+ * The RestWebhookConfig interface is used to pass the WebLogic Operator's WebHook REST configuration to the
+ * RestWebhookServer.
+ */
+public interface WebhookRestConfig {
```
rjeberhard, line 10: This seems like a copy too, but not sure if a copy is necessary.

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestServer.java
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * @param restConfig the operator REST configuration
+   * @return a resource configuration
+   */
+  static ResourceConfig createResourceConfig(WebhookRestConfig restConfig) {
+    ResourceConfig rc =
+        new ResourceConfig()
+            .register(JacksonFeature.class)
+            .register(ErrorFilter.class)
+            .register(RequestDebugLoggingFilter.class)
```
rjeberhard, line 125: I now know why your change is so big 🤣... This class is just the same as the original except that the set of filters is different (there are probably a few other differences), but I really do think the next "big thing" for this change should be removing unnecessary copies.

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
ankedia: I'm not clear why we need to default to "2" copies since webhook is going to be rarely used (only when converting from v8 to v9 or vice-versa). Not sure if it's worth creating 2 pods for just this reason. The customers can always change the replica if they feel it's necessary.

operator/src/main/java/oracle/kubernetes/operator/rest/resource/ConversionWebhookResource.java
```diff
@@ -0,0 +1,51 @@
...
+   * @return a String representation of JSON document describing the ConversionReview with conversion response.
+   */
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  public String post(String body) {
+    LOGGER.entering(href());
+    ConversionReviewModel conversionReview = conversionUtils.readConversionReview(body);
+    ConversionResponse conversionResponse = conversionUtils.createConversionResponse(conversionReview.getRequest());
+    LOGGER.exiting(conversionResponse);
+    return conversionUtils.writeConversionReview(new ConversionReviewModel()
```
rjeberhard, line 46: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#response

```
If conversion fails, a webhook should return a response stanza containing the following fields:

uid, copied from the request.uid sent to the webhook
result, set to {"status":"Failed"}
```

You need a guard so that this method never throws and returns an appropriate status when there is an exception.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
```diff
@@ -0,0 +1,368 @@
...
+import oracle.kubernetes.operator.work.Packet;
+import oracle.kubernetes.operator.work.Step;
+import oracle.kubernetes.operator.work.ThreadFactorySingleton;
+import oracle.kubernetes.weblogic.domain.model.DomainList;
+
+import static oracle.kubernetes.operator.helpers.NamespaceHelper.getWebhookNamespace;
+
+/** A Conversion Webhook for WebLogic Kubernetes Operator. */
+public class ConversionWebhookMain {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Webhook", "Operator");
```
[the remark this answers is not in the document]
ankedia: Well, this project started as a POC for the Webhook and I started out using Main as a template. I think the existing implementation of Main is very tightly coupled with CRD lifecyle and few other things and refactoring may not be straight-forward without understanding the complete design. I'll take a look at refactoring, is it critical to be done as part of this PR?

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
```diff
@@ -0,0 +1,187 @@
...
+  private NextAction createIdentity(Packet packet) throws Exception {
+    KeyPair keyPair = createKeyPair();
+    String key = convertToPEM(keyPair.getPrivate());
+    writeToFile(key, new File(WEBHOOK_CERTIFICATE_KEY));
+    X509Certificate cert = generateCertificate(WEBHOOK_CERTIFICATE, keyPair, SHA_256_WITH_RSA, COMMON_NAME,
+            CERTIFICATE_VALIDITY_DAYS);
+    writeToFile(getBase64Encoded(cert), new File(WEBHOOK_CERTIFICATE));
+    // put the new certificate in the webhook's config map so that it will be available
+    // the next time the webhook is started
+    return doNext(recordWebhookCert(WEBHOOK_CERTIFICATE, cert,
```
rjeberhard, line 95: This is the other potential concurrency point. The webhook is about to record its certificate and key, but the create/replace of the Secret may fail with a conflict, in which case the webhook should read the existing data and reuse the identity set by the other webhook instance that was faster.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  private static String getBase64Encoded(X509Certificate cert) throws IOException {
+    return Base64.getEncoder().encodeToString(convertToPEM(cert).getBytes());
+  }
+
+  private static Step recordWebhookCert(String certName, X509Certificate cert, Step next) throws IOException {
+    JsonPatchBuilder patchBuilder = Json.createPatchBuilder();
+    patchBuilder.add("/data/webhookCert", getBase64Encoded(cert));
```
rjeberhard, line 122: Thought... writing the cert to a CM and the key to a Secret just makes the concurrency problem harder... Why not just write both to the Secret? Then there will be one consistent update with both pieces of data.

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
```diff
@@ -0,0 +1,394 @@
...
+   */
+  public Object convertDomain(Map<String, Object> domain, String desiredAPIVersion) {
+    Map<String, Object> spec = getSpec(domain);
+    if (spec == null) {
+      return domain;
+    }
+    LOGGER.fine("Converting domain " + domain + " to " + desiredAPIVersion + " apiVersion.");
+
+    String apiVersion = (String)domain.get("apiVersion");
+    adjustAdminPortForwardingDefault(spec, apiVersion);
```
rjeberhard, line 95: I see transformations for the port-forwarding change and for auxiliary images... where is the change for removing the Progressing condition?

rjeberhard: I'll read below, but wondering if these methods will each handle both the v8 -> v9 and the v9 -> v8 conversion. I suspect those code flows will be quite different and it might read more intuitively if we made the conversion decision up front and then had specific methods.

webhook/scripts/webhook.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+export PATH=$PATH:/operator
+
+echo "Launching the Webhook for Oracle WebLogic Server Kubernetes Operator..."
```
rjeberhard, line 7: "Webhook" -> "Domain Custom Resource Conversion Webhook"

webhook/scripts/webhookLivenessProbe.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+# if the livenessProbeSuccessOverride file is available, treat failures as success
+#
+RETVAL=$(test -f /webhook/debug-config/livenessProbeSuccessOverride ; echo $?)
```
rjeberhard, line 7: I'm wondering if these files need to be different... even if installed together, the operator and the webhook will be running in their own containers so they could use the same names for the .alive and other files.

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
ankedia: My understanding was that "Progressing condition" is removed during upgrade in DomainStatus.java (line 576)

```
  public DomainStatus upgrade() {
    Optional.ofNullable(conditions).ifPresent(x -> x.removeIf(cond -> cond.hasType(Progressing)));
    return this;
  }
```

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
```diff
@@ -330,6 +365,12 @@ Step updateExistingCrd(Step next, V1CustomResourceDefinition existingCrd) {
           existingCrd.getMetadata().getName(), existingCrd, createReplaceResponseStep(next));
     }
 
+    Step updateExistingCrdWithConversion(Step next, V1CustomResourceDefinition existingCrd) {
+      existingCrd.getSpec().conversion(createConversionWebhook());
+      return new CallBuilder().replaceCustomResourceDefinitionAsync(
```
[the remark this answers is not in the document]
ankedia: Not sure if I'm missing something here. The `onFailure` method  of the `ReplaceResponseStep` calls the onFailure of the parent with the `conflictStep` defined in the context. If the "two webhook instances" are part of the same deployment, then they'll share the same identity (certificate/key data). The identity data is generated only once and stored in the Webhook's configmap and secret. For subsequent calls, it reuses the same identity instead of creating a new identity.

operator/src/main/java/oracle/kubernetes/operator/rest/model/ConvertedObject.java
```diff
@@ -0,0 +1,79 @@
...
+public class ConvertedObject {
+
+  @Expose
+  private String kind;
+  @Expose
+  private String apiVersion;
+  @Expose
+  private V1ObjectMeta metadata;
+  @Expose
+  private String host;
```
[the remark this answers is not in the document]
ankedia: I used the Gson schema to POJO conversion tool to generate these classes. There's a host and port defined in the example schema provided in conversion webhook kubernetes documentation (see below). Probably it's not needed in our case, I'll remove them.

```
    "convertedObjects": [
      {
        "kind": "CronTab",
        "apiVersion": "example.com/v1",
        "metadata": {
          "creationTimestamp": "2019-09-04T14:03:02Z",
          "name": "local-crontab",
          "namespace": "default",
          "resourceVersion": "143",
          "uid": "3415a7fc-162b-4300-b5da-fd6083580d66"
        },
        "host": "localhost",
        "port": "1234"
      },
```

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
ankedia: I was following the pattern that's similar to Operator's identity. In operator's case also, we write the cert to CM and key to a secret. Do you know why it's done that way in the operator code? Thanks.

webhook/scripts/webhookLivenessProbe.sh
ankedia: I changed the location of webhook liveness file to `/webhook/.alive` and the operator's liveness file is in `/operator/.alive` to avoid the confusion. If we decide to go with the same file, then both liveness files will need to be in the `/operator` directory.

operator/src/main/resources/scripts/utils_base.sh
tbarnes-us: How about if a pod restarts for other reasons? Like out-of-memory, etc.

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestServer.java
ankedia: Fixed.

kubernetes/charts/weblogic-operator/templates/_operator-internal-svc.tpl
ankedia: Removed "internal-".

kubernetes/charts/weblogic-operator/values.yaml
ankedia: Changed.

kubernetes/charts/weblogic-operator/values.yaml
ankedia: Removed external debug port.

operator/pom.xml
ankedia: Yes. IIRC, it's needed by `GsonOffsetDateTime.java` that's used when converting the JSON payload to ConversionReviewModel. At the time, I kept the line commented to be sure that it's not causing any other issues. I have removed the commented out line.

operator/pom.xml
ankedia: Thanks, that's good to know. It's used by the standalone command-line utility class `DomainResourceConverter`.

ankedia: I'll add it to the top-level pom and have the version as a property in this file.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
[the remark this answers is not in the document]
russgold: If it's not done now, there is a good chance that it will never be done, so I would strongly recommend that it be done as part of this PR.

kubernetes/charts/weblogic-operator/values.yaml
ankedia: Changed to have only one tuning value.

operator/src/main/java/oracle/kubernetes/operator/DomainResourceConverter.java
ankedia: Done.

operator/src/main/java/oracle/kubernetes/operator/Main.java
ankedia: Removed.

operator/src/main/java/oracle/kubernetes/operator/WebhookReady.java
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/operator/rest/resource/ConversionWebhookResource.java
ankedia: Fixed.

webhook/scripts/webhook.sh
ankedia: Fixed.

operator/src/main/java/oracle/kubernetes/operator/logging/MessageKeys.java
```diff
@@ -153,6 +153,20 @@
...
+  public static final String WEBHOOK_IDENTITY_INITIALIZATION_FAILED = "WLSKO-0212";
+  public static final String INPUT_FILE_NON_EXISTENT = "WLSKO-0213";
+  public static final String OUTPUT_FILE_NON_EXISTENT = "WLSKO-0214";
+  public static final String PRINT_HELP = "WLSKO-0215";
+  public static final String OUTPUT_DIRECTORY = "WLSKO-0216";
+  public static final String OUTPUT_FILE_NAME = "WLSKO-0217";
+  public static final String DOMAIN_CONVERSION_FAILED = "WLSKO-0218";
 
   // domain status messages
   public static final String DUPLICATE_SERVER_NAME_FOUND = "WLSDO-0001";
```
russgold, line 172: We already have a number of messages that wind up in the domain status; why is this different enough to require a new number sequence?

ankedia: This message is part of the conversion webhook which is in a separate deployment and is used while converting the domain custom resource from `v8` to `v9`.

operator/src/main/resources/scripts/utils_base.sh
ankedia: The pod restart due to out-of-memory should be same as customer deleting the pod externally. The pod will be recreated with new specs  with the "raw" init containers. If the container is restarted (for e.g. due to liveness probe failure), then the init container will not be run again.

operator/src/main/java/oracle/kubernetes/operator/rest/WebhookRestServer.java
ankedia: Refactored to use the existing RestServer and RestConfig clases in b0ef168 .

operator/src/main/java/oracle/kubernetes/operator/WebhookLiveness.java
ankedia: The `OperatorLiveness` class takes MainDelegate as a parameter and creates the liveness file in `OperatorHome` returned by the delegate which defaults to `/operator`. Do we want to create the webhook liveness file under '/operator' or '/webhook' or should we rename the `OperatorHome` to something like `DeploymentHome` which can be used by both the operator and the webhook and defaults to something like `/deployments`?

kubernetes/charts/weblogic-operator/templates/_operator-dep.tpl
rjeberhard: Okay, that's fair. I was thinking that if we defaulted to 2 replicas then we would actually make sure that the concurrency issues were resolved ;)

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -159,6 +159,20 @@ javaLoggingFileCount: 10
 # the operator's deployment. See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector
 # for more information on node selectors.
 #nodeSelector:
+ 
+# webhookOnly - webhoookOnly specifies whether only the webhook deployment named weblogic-operator-webhook
```
rjeberhard, line 163: "webhookOnly - webhoookOnly" -> "webhookOnly"

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -159,6 +159,20 @@ javaLoggingFileCount: 10
...
+# should be created during helm install and if the creation of the operator deployment should be skipped.
+# By default, the helm install will create both the webhook deployment and the operator deployment.
+# If set to true, the helm install will only create the webhook deployment.
+# The default value is false.
+webhookOnly: false
+ 
+# preserveWebhook - preserveWebhook specifies whether the previous webhook deployment should be preserved
```
rjeberhard, line 170: "preserveWebhook - preserveWebhook" -> "preserveWebhook"

operator/pom.xml
```diff
@@ -414,6 +414,11 @@
...
+        <artifactId>commons-cli</artifactId>
+        <version>${commons-cli-version}</version>
```
rjeberhard, line 420: You don't need <version> here if the dependency is at top-level.

operator/src/main/java/oracle/kubernetes/operator/WebhookLiveness.java
rjeberhard: Yes, some common path. /probes ?

operator/src/main/java/oracle/kubernetes/operator/WebhookReady.java
rjeberhard: I don't see that it is fixed. I was suggesting that there could be just one class, shared between the operator and webhook, that checked a .ready path at a common path, something like /probes/.ready.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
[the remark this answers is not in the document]
rjeberhard: Once I've got the tests working on release/3.4 then I'd be interested to work on this problem :).  I think that you are missing that if two webhook replicas start concurrently then they may each try to create a certificate/key and then they will need to negotiate when one of them wins -- that is that the other will have to rewind and read the data shared by the other.

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -191,31 +212,39 @@ String getInternalHttpsUri() {
...
   public void start(Container container) throws Exception {
     LOGGER.entering();
-    if (externalHttpsServer != null || internalHttpsServer != null) {
+    if ((externalHttpsServer != null || internalHttpsServer != null) && (!isWebhook())) {
       throw new AssertionError("Already started");
     }
     boolean fullyStarted = false;
     try {
-      if (isExternalSslConfigured()) {
+      if (!isWebhook() && isExternalSslConfigured()) {
```
rjeberhard, line 220: Can you please make this more OO :) ?

Instead of lots of  `if (!isWebhook())` statements, please have a method that is called and then that the operator REST or webhook implementations have content or a no-op.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
rjeberhard: I believe that Lenny's scaling script reads the certificate from the CM.

pom.xml
```diff
@@ -596,6 +596,7 @@
     <jms-api-version>2.0.1</jms-api-version>
     <kjetland-mbknor-jsonschema-version>1.0.39</kjetland-mbknor-jsonschema-version>
     <networknt-jsonschmema-validator-version>1.0.66</networknt-jsonschmema-validator-version>
+    <commons-cli-version>1.5.0</commons-cli-version>
```
rjeberhard, line 599: You will see that all of the dependencies are defined in the dependencyManagement section of this file and then can be used without the version in the module-specific pom.

kubernetes/charts/weblogic-operator/values.yaml
ankedia: Changed.

operator/pom.xml
ankedia: Moved the dependency  to top level.

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
ankedia: I refactored the code as suggested, please let me know if I missed something. Thanks.

pom.xml
ankedia: Changed.

operator/src/main/java/oracle/kubernetes/operator/rest/model/ConvertedObject.java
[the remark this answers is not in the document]
ankedia: Removed host and port values.

operator/src/main/java/oracle/kubernetes/operator/WebhookLiveness.java
ankedia: Changed.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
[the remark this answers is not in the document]
ankedia: I have done the refactoring to have the shared code into BaseMain with the abstract methods as suggested. Please let me know if I missed something. Thanks.

operator/pom.xml
ankedia: Fixed. I have reverted this change as it's no longer necessary.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMainDelegate.java
```diff
@@ -0,0 +1,30 @@
...
+
+import oracle.kubernetes.operator.helpers.KubernetesVersion;
+import oracle.kubernetes.operator.helpers.SemanticVersion;
+import oracle.kubernetes.operator.work.Packet;
+import oracle.kubernetes.operator.work.Step;
+
+/**
+ * Definition of an interface that returns values that the Webhook Main class requires.
+ */
+interface ConversionWebhookMainDelegate {
```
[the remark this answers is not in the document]
ankedia: Refactored to extend from the CoreDelegate class and have only webhook specific code here.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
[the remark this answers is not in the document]
ankedia: ok, sure :) I think this is related to the discussion on slack thread. If two replicas are started concurrently, there'll be a race and winner will create certificate/key and write them to secret. However, the other instance (loser) might overwrite the certificate/key stored by the winner. I think we need some way to synchornize so that loser doesn't overwrite the certificate/key stored by winner.

deployment/scripts/initialize-external-operator-identity.sh
```diff
@@ -12,11 +12,11 @@ fi
...
 EXTERNAL_KEY="externalOperatorKey"
-OPERATOR_DIR="/operator"
-EXTERNAL_IDENTITY_DIR="${OPERATOR_DIR}/external-identity"
+DEPLOYMENT_DIR="/deployment"
+EXTERNAL_IDENTITY_DIR="${DEPLOYMENT_DIR}/external-identity"
 NAMESPACE=`cat /var/run/secrets/kubernetes.io/serviceaccount/namespace`
-OPERATOR_CONFIG_DIR=${OPERATOR_DIR}/config
-OPERATOR_SECRETS_DIR=${OPERATOR_DIR}/secrets
+OPERATOR_CONFIG_DIR=${DEPLOYMENT_DIR}/config
+OPERATOR_SECRETS_DIR=${DEPLOYMENT_DIR}/secrets
```
rjeberhard, line 19: This is more of a "just curious". I very much understand using "/deployment" to align the operator and the webhook... but why change the mount location for the CM and Secret?

deployment/scripts/webhook.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+export PATH=$PATH:/operator
```
rjeberhard, line 5: Why change the PATH? Isn't everything fully qualified below?

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
rjeberhard: > My understanding was that "Progressing condition" is removed during upgrade in DomainStatus.java (line 576)

  public DomainStatus upgrade() {
    Optional.ofNullable(conditions).ifPresent(x -> x.removeIf(cond -> cond.hasType(Progressing)));
    return this;
  }

If your webhook removes Progressing then we don't need the upgrade() method any more. Our intention was to not have Progressing in v9; however, we were blocked by not yet having the webhook.

deployment/scripts/initialize-external-operator-identity.sh
ankedia: The location of cert and key file is using the `getDeploymentHome` method of the `CoreDelegate` in `InitializeInternalIdentityStep` . I thought it was consistent with keeping all the identity files included config and secret mounted cert and key under `/deployment` dir. 

```
    this.configInternalCertFile = new File(delegate.getDeploymentHome(), "/config/internalOperatorCert");
    this.secretsInternalKeyFile = new File(delegate.getDeploymentHome(), "/secrets/internalOperatorKey");  
```

deployment/scripts/webhook.sh
ankedia: ok. This is similar to the PATH settings used in `operator.sh` and I wasn't sure how the PATH is used. I'll go ahead and remove it from `webhook.sh`. I assume we can also similarly remove it from the `operator.sh`.

rjeberhard: Sounds good... I must have missed it in the other script. Either way, let's remove it if we can.

operator/src/main/java/oracle/kubernetes/operator/utils/DomainUpgradeUtils.java
ankedia: ok. I will make the change in the webhook to remove the Progressing condition.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
ankedia: Fixed in 6b88d5b.

operator/src/main/java/oracle/kubernetes/operator/helpers/CrdHelper.java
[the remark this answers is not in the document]
ankedia: ok. I think this is likely resolved by 6b88d5b. I did some tests by setting 2 replicas for the webhook deployment and it seem to work fine.

operator/src/main/java/oracle/kubernetes/operator/steps/InitializeWebhookIdentityStep.java
ankedia: I changed to use the secret for both webhook cert and key.

operator/src/main/resources/Operator.properties
ankedia: Changed the message to use "WebLogic Domain custom resource conversion webhook".

rjeberhard (approved):

deployment/scripts/operator.sh
```diff
@@ -15,7 +13,7 @@ relay_SIGTERM() {
 
 trap relay_SIGTERM SIGTERM
 
-/operator/initialize-external-operator-identity.sh
+/deployment/initialize-external-operator-identity.sh
```
jshum2479, line 16: maybe easier to use a variable for `/deployment`  similar to initialize-external-operator-identity.sh

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
```diff
@@ -0,0 +1,162 @@
...
+
+    try {
+      main.startDeployment(main::completeBegin);
+
+      // now we just wait until the pod is terminated
+      main.waitForDeath();
+
+      // stop the webhook REST server
+      stopWebhookRestServer();
+    } finally {
```
jshum2479, line 59: What happened if it crashed?  Does k8s automatically restart it and how will the user know something is wrong?

rjeberhard: Yes, k8s will restart. We are making sure that the webhook can support replicas > 1 so that there isn't a single point of failure if one webhook crashes.

deployment/scripts/operator.sh
rjeberhard: I don't think that we need a variable -- this set has to match between Dockerfile and script. Since this script can now be run from the webhook, it made sense to not use the "/operator" path.

ankedia: Added a `DEPLOYMENT_DIR` variable since it was used in two places in 938f839 .

jshum2479 (approved):

ankedia:
@tbarnes-us , @russgold - I think I addressed your review comments. Please let me know if you have follow up comments or if the comments are resolved. Thanks.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
ankedia: Created JIRA OWLS-97471 to track the event generation for conversion webhook unhappy paths.

operator/src/main/resources/scripts/auxImage.sh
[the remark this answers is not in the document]
tbarnes-us: I continue to prefer having two versions of this script.  One for "normal" internal use, and one for customer use. This way a customer that is using the custom path will not need to care about the container name, and cannot get it wrong...  Or, if the container name must have a certain format, then this way it will always (properly) expect the "compat" format instead of allowing for the (possibly incorrect) normal format...

probes/scripts/livenessProbe.sh
```diff
@@ -6,7 +6,7 @@
 #
 RETVAL=$(test -f /operator/debug-config/livenessProbeSuccessOverride ; echo $?)
 
-FILE=/operator/.alive
+FILE=/probes/.alive
```
tbarnes-us, line 9: I assume this file is now used by both operator and webhook? If so, then the error message for a missing liveness probe does not match the name of this file - at least in the webhook case - not sure about the operator case.

kubernetes/charts/weblogic-operator/values.yaml
```diff
@@ -159,6 +159,20 @@ javaLoggingFileCount: 10
...
+# If set to true, the helm install will only create the webhook deployment.
+# The default value is false.
+webhookOnly: false
+ 
+# preserveWebhook specifies whether the previous webhook deployment should be preserved
+# when the chart is uninstalled using helm uninstall.
+# By default, the helm uninstall will remove both the webhook deployment and the operator deployment.
+# If set to true, the helm uninstall command will skip removing the webhook deployment.
+# The default value is false.
+preserveWebhook: false
```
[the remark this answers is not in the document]
ankedia: The CRD behavior is still the same and operator uninstall would leave the CRD behind. The `preserveWebhook` option is if the webhook pod(s) needs to be left behind. I think majority of use-cases will have single operator and they will be surprised to see a running pod after they uninstall the operator. For multiple operators use-cases, we can recommend customers to install the webhook in a namespace that is separate from the operator namespace (using `webhookOnly` option). This way webhook will not be uninstalled when the operator is uninstalled. I would vote to keep this default to `false`.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
```diff
@@ -0,0 +1,162 @@
...
+
+  static class ConversionWebhookMainDelegateImpl extends CoreDelegateImpl {
+    public ConversionWebhookMainDelegateImpl(Properties buildProps, ScheduledExecutorService scheduledExecutorService) {
+      super(buildProps, scheduledExecutorService);
+    }
+
+    private void logStartup() {
+      ConversionWebhookMain.LOGGER.info(MessageKeys.CONVERSION_WEBHOOK_STARTED, buildVersion,
+              deploymentImpl, deploymentBuildTime);
+      ConversionWebhookMain.LOGGER.info(MessageKeys.WEBHOOK_CONFIG_NAMESPACE, getWebhookNamespace());
```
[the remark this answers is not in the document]
ankedia: I have created OWLS-97471 for generating the events for unhappy paths. My understanding was that we'll generate events only for the unhappy path and generating events for happy path is a lower priority. If it's a high priority, we can update existing JIRA.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
```diff
@@ -91,6 +93,30 @@ protected V1Container createPrimaryContainer(TuningParameters tuningParameters)
...
+
+    addAuxImagePathsEnvVarIfExists(tuningParameters, vars);
+
+    hideAdminUserCredentials(vars);
+    return doDeepSubstitution(varsToSubVariables(vars), vars);
+  }
+
   protected V1Volume createEmptyDirVolume() {
     V1EmptyDirVolumeSource emptyDirVolumeSource = new V1EmptyDirVolumeSource();
     Optional.ofNullable(getMedium()).ifPresent(emptyDirVolumeSource::medium);
```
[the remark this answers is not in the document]
ankedia: The BasePodStepContext is shared by `JobStepContext` (used for introspector job pod) and the `PodStepContext` (used for WLS pods) and they both need the emptydir for auxiliary images. This is existing code and not used in conversion webhook.

kubernetes/charts/weblogic-operator/values.yaml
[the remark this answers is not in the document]
tbarnes-us: I will leave it up to you. I think it's more intuitive and safer to have it match the CRD life-cycle.  

But if you do leave it 'false', please make absolutely sure the documentation for operator cleanup (in quickstart, etc, etc) is extremely clear about the potential "silent" side effects... 

To me, this is like having a light switch in every office that surprisingly can shut down the entire building unless you jiggle it in exactly the right way when switching it off.  Works fine in a single office building, but not otherwise.

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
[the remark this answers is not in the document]
tbarnes-us: I think happy path should be there too. Just do it all at the same time - a "happy" event gives helpful context when an unhappy event occurs later (makes it easier to interpret the unhappy event).

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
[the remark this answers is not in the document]
tbarnes-us: This implies that another level of refactoring would make things more clear in some ways - but perhaps less clear because we already have a bunch of classes.  I'll leave it up to you.

operator/src/main/java/oracle/kubernetes/operator/rest/RestConfigImpl.java
```diff
@@ -12,10 +12,20 @@
 /** RestConfigImpl provides the WebLogic Operator REST api configuration. */
```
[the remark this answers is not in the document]
ankedia: This same class is also used for webhook REST server config, the same class is shared by both operator and webhook.

probes/scripts/livenessProbe.sh
ankedia: I'm not sure which error message are referring to, can you please clarify?

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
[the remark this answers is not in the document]
ankedia: Updated JIRA to cover happy path as well.

operator/src/main/java/oracle/kubernetes/operator/DomainCustomResourceConverter.java
```diff
@@ -0,0 +1,121 @@
...
+   * @param outputFileName Name of the output file.
+   * @param inputFileName Name of the input file.
+   */
+  public DomainCustomResourceConverter(
+          String outputDir,
+          String outputFileName,
+          String inputFileName) {
+    this.outputDir = Optional.ofNullable(outputDir).orElse(new File(inputFileName).getParent());
+    this.outputFileName = Optional.ofNullable(outputFileName)
+            .orElse("Converted_" + new File(inputFileName).getName());
```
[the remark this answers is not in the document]
ankedia: Changed in [48e8591](https://github.com/oracle/weblogic-kubernetes-operator/pull/2816/commits/48e85910ce48ee2a68fe1705b2008dde434670fb).

operator/src/main/java/oracle/kubernetes/operator/DomainCustomResourceConverter.java
```diff
@@ -0,0 +1,121 @@
...
+   * @param outputDir Name of the output directory.
+   * @param outputFileName Name of the output file.
+   * @param inputFileName Name of the input file.
+   */
+  public DomainCustomResourceConverter(
+          String outputDir,
+          String outputFileName,
+          String inputFileName) {
+    this.outputDir = Optional.ofNullable(outputDir).orElse(new File(inputFileName).getParent());
+    this.outputFileName = Optional.ofNullable(outputFileName)
```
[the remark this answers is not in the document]
ankedia: Changed in [48e8591](https://github.com/oracle/weblogic-kubernetes-operator/pull/2816/commits/48e85910ce48ee2a68fe1705b2008dde434670fb).

operator/src/main/java/oracle/kubernetes/operator/rest/RestServer.java
```diff
@@ -45,16 +25,7 @@
...
-  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
-  private static final int CORE_POOL_SIZE = 3;
-  private static final String SSL_PROTOCOL = "TLSv1.2";
-  private static final String[] SSL_PROTOCOLS = {
-    SSL_PROTOCOL
-  }; // ONLY support TLSv1.2 (by default, we would get TLSv1 and TLSv1.1 too)
-  private static RestServer INSTANCE = null;
-  private final RestConfig config;
-  // private String baseHttpUri;
+public class RestServer extends BaseRestServer {
```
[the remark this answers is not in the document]
ankedia: Changed in [48e8591](https://github.com/oracle/weblogic-kubernetes-operator/pull/2816/commits/48e85910ce48ee2a68fe1705b2008dde434670fb).

operator/src/main/java/oracle/kubernetes/operator/Main.java
```diff
@@ -30,138 +22,74 @@
...
-  private static final ThreadFactory threadFactory = new WrappedThreadFactory();
-  private static final ScheduledExecutorService wrappedExecutorService =
-      Engine.wrappedExecutorService("operator", container);
-  private static final AtomicReference<OffsetDateTime> lastFullRecheck =
-      new AtomicReference<>(SystemClock.now());
-  private static final Semaphore shutdownSignal = new Semaphore(0);
-  private static final int DEFAULT_STUCK_POD_RECHECK_SECONDS = 30;
-
-  private static final File operatorHome;
+public class Main extends BaseMain {
```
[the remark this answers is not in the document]
ankedia: Changed in [48e8591](https://github.com/oracle/weblogic-kubernetes-operator/pull/2816/commits/48e85910ce48ee2a68fe1705b2008dde434670fb).

kubernetes/charts/weblogic-operator/values.yaml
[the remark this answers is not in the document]
ankedia: ok. I believe we discussed this previously and my understanding is that we decided to keep it simple for ~ 90% of the customers who have a single operator. I agree on the documentation part. I'll leave the default to 'false' unless @rjeberhard and @robertpatrick disagrees and we can change the default later if we decide to do it.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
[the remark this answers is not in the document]
ankedia: ok. The changes to `BasePodStepContext.java` in this PR is only to avoid the pod roll during operator upgrade. When the auxiliary image is created using old approach, AUXILIARY_IMAGE_PATHS env variable is added after all other user specified env variables and some other env variables. When the v8 domain resource is converted to v9, AUXILIARY_IMAGE_PATHS  env variable is now part of the user specified env variable.

operator/src/main/java/oracle/kubernetes/operator/DomainCustomResourceConverter.java
```diff
@@ -0,0 +1,121 @@
...
+    } catch (ParseException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static void printHelpAndExit(Options options) {
+    HelpFormatter help = new HelpFormatter();
+    help.printHelp(120, "java -cp <operator-jar> "
+                    + "oracle.kubernetes.operator.DomainCustomResourceConverter "
+                    + "<input-file> [-d <output_dir>] [-f <output_file_name>] [-h --help]",
```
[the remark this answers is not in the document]
ankedia: Added in [ae35805](https://github.com/oracle/weblogic-kubernetes-operator/pull/2816/commits/ae358053adb02f293fbe591617fe5005211bf986).

operator/src/main/resources/Operator.properties
```diff
@@ -159,6 +159,24 @@ WLSKO-0201=WL pod shutdown: Failed to shutdown WebLogic server {0} via REST inte
...
+WLSKO-0213=Domain custom resource converter: InputFile: {0} does not exist or could not be read.
+WLSKO-0214=Domain custom resource converter: The output dir {0} specified by ''-d'' does not exist or could not be read.
+WLSKO-0215=Domain custom resource converter: The output file {0} already exists, remove it and try again. \
+  Use ''-o'' option to overwrite the existing file.
+WLSKO-0216=Domain custom resource converter: Print help message.
+WLSKO-0217=Domain custom resource converter: The directory where the tool will place the converted file. \
+  If not specified, it defaults to the directory of the input file.
+WLSKO-0218=Domain custom resource converter: Name of the converted file. If not specified, it generates the \
+  file by appending "__converted." and the input file extension to the base name of the input file name.
+WLSKO-0219=Domain custom resource converter: Overwrites the existing output file.
```
tbarnes-us, line 179: Enable overwriting the existing output file, if any.

tbarnes-us: Also 0216 through 0219 should not have "Domain custom resource converter: " since the string probably looks a little confusing in the generated command line help.

operator/src/main/resources/Operator.properties
```diff
@@ -159,6 +159,24 @@ WLSKO-0201=WL pod shutdown: Failed to shutdown WebLogic server {0} via REST inte
...
+WLSKO-0208=WebLogic Domain custom resource conversion webhook namespace is: {0}
+WLSKO-0209=The Custom Resource Definition for 'domains.weblogic.oracle' is not installed, waiting for {0} seconds for the CRD to be installed.
+WLSKO-0210=The WebLogic Domain custom resource conversion webhook is shutting down
+WLSKO-0211=Starting Webhook liveness Thread
+WLSKO-0212=WebLogic Domain custom resource conversion webhook identity initialization step failed with exception {0}.
+WLSKO-0213=Domain custom resource converter: InputFile: {0} does not exist or could not be read.
+WLSKO-0214=Domain custom resource converter: The output dir {0} specified by ''-d'' does not exist or could not be read.
+WLSKO-0215=Domain custom resource converter: The output file {0} already exists, remove it and try again. \
+  Use ''-o'' option to overwrite the existing file.
+WLSKO-0216=Domain custom resource converter: Print help message.
```
tbarnes-us, line 174: Print this help message.

tbarnes-us:
I may have lost the comment - but I was thinking of error message WLSKO-0035, which seems to be shared by the operator and webhook cases - but refers to the operator location of the file?

operator/src/main/resources/Operator.properties
ankedia: Fixed in 16f5f6e.

operator/src/main/resources/Operator.properties
ankedia: Changed in 16f5f6e.

tbarnes-us (approved):

operator/src/main/java/oracle/kubernetes/operator/ConversionWebhookMain.java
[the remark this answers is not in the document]
rjeberhard: I disagree with this one. There will be a lot of webhook calls resulting in successful conversions (e.g. every time the 3.4 operator updates a domain for status). If we generate events for the happy paths then there will be a large number of events that are unimportant. The customer should really only care about the details of conversions that fail.

PR #2821, opened 2022-03-07
Replaced the generation of different image names with one name and different imagetags to avoid docker push issues

marinakog:


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiAuxiliaryImage.java
```diff
@@ -428,7 +444,7 @@ void testWithMultipleAIsHavingWDTInstallers() {
     final String managedServerPrefix = domainUid + "-managed-server";
     // using the first image created in initAll, creating second image with different WDT version here
 
-    createPushAuxiliaryImageWithWDTInstallOnly(miiAuxiliaryImage4, "1.9.19");
+    createPushAuxiliaryImageWithWDTInstallOnly(MII_AUXILIARY_IMAGE_NAME,miiAuxiliaryImage4Tag, "1.9.19");
```
anpanigr, line 447: Can we replace the hard-coded value ?

anpanigr (commented):
In each test method in ItMiiCreateAuxImageWithImageTool class we are creating required aux image. Can we use the same image name with tag. This way we can avoid creating a image per test method.  Delete the image in each test and create a new one with same name but with different image toll option.

marinakog:
addressed the comments

anpanigr (approved):
lgtm

vanajamukkara (approved):
LG

PR #2824, opened 2022-03-09
first pass at install and upgrade

rosemarymarano:
PR to update the Installation and Upgrade doc as per [OWLS-95907](https://jira.oraclecorp.com/jira/browse/OWLS-95907) Operator User Guide Feedback and Updates for 4.0. 

**NOTE**: The upgrade section is preliminarily modified _only_: 

- I created JIRA [OWLS-97331](https://jira.oraclecorp.com/jira/browse/OWLS-97331) for development input. 
- In this PR, I removed the section "Upgrading a 2.6 operator to a 3.x operator" as explained in [OWLS-95907](https://jira.oraclecorp.com/jira/browse/OWLS-95907).
- 
@rjeberhard @tbarnes-us 
**ALSO**: Please answer whether this text in the Uninstall section is needed:
After removing the operator deployment,
you should also remove the Domain custom resource definition (CRD) if it is no longer needed:
```text
$ kubectl delete customresourcedefinition domains.weblogic.oracle
```
Note that the Domain custom resource definition is shared.
Do not delete the CRD if there are other operators in the same cluster
or you have running domain resources.

### DIFF
--- a/documentation/staging/content/userguide/managing-operators/installation.md
+++ b/documentation/staging/content/userguide/managing-operators/installation.md
@@ -41,7 +41,6 @@ a Helm "release" name for the operator,
 the Helm chart location,
 the operator namespace and service account,
 the platform (if required),
-the operator image name,
 and name space selection settings.
 
 A typical Helm release name is `weblogic-operator`.
@@ -75,10 +74,9 @@ For example, using Helm 3.x, with the following settings:
 
 |Setting|Value and Notes|
 |-|-|
-|Helm release|`sample-weblogic-operator` (you may choose any name)|
-|Helm chart|`kubernetes/charts/weblogic-operator` in local operator source downloaded to `/tmp/weblogic-kubernetes-operator`|
+|Helm release name|`sample-weblogic-operator` (you may choose any name)|
+|Helm chart repo location|`https://oracle.github.io/weblogic-kubernetes-operator/charts`|
 |`namespace`|`sample-weblogic-operator-ns`|
-|`image`|`ghcr.io/oracle/weblogic-kubernetes-operator:{{< latestVersion >}}`|
 |`serviceAccount`|`sample-weblogic-operator-sa`|
 |`enableClusterRoleBinding`|`true` (gives operator permission to automatically install the Domain CRD and to manage domain resources in any namespace)|
 |`domainNamespaceSelectionStrategy`|`LabelSelector` (limits operator to managing namespaces that match the specified label selector)|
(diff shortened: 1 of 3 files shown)


documentation/staging/content/userguide/managing-operators/installation.md
```diff
@@ -91,13 +89,24 @@ $ kubectl create namespace sample-weblogic-operator-ns
...
+```text
+$ helm repo add weblogic-operator https://oracle.github.io/weblogic-kubernetes-operator/charts --force-update  
+```
+- To list the versions of the operator that you can install from the Helm chart repository:
+
+  ```text
+  $ helm search repo weblogic-operator/weblogic-operator --versions
+  ```
 
+- For a specified version of the Helm chart, use the `--version <value>` option with `helm install`
```
tbarnes-us, line 102: Add: This will also set the operator version (which defaults to latest).

documentation/staging/content/userguide/managing-operators/installation.md
```diff
@@ -91,13 +89,24 @@ $ kubectl create namespace sample-weblogic-operator-ns
...
+  to choose the version that you want, with the `latest` value being the default.
```
tbarnes-us, line 103: For a specified version of the Helm chart and operator,

documentation/staging/content/userguide/managing-operators/installation.md
```diff
@@ -171,29 +177,7 @@ Example updates:
...
-To delete the 2.6.0 operator:
-
-```text
-$ helm delete weblogic-operator -n weblogic-operator-namespace
-```
-
-Then install the 3.x operator using the [install the operator](#install-the-operator) instructions.
+You can upgrade a 3.x operator while the operator's domain resources are deployed and running.
 
 #### Upgrading a 3.x operator
```
tbarnes-us, line 182: In my opinion, this also needs a discussion of choosing and setting the helm chart version because we are choosing to "hide" the existence of the operator image as a simplification, which is what is being used to set the version today.

It might also be nice to teach folks how to quickly determine the current running version since we are "hiding" discussion of the operator image, which is the current way folks figure that out.

tbarnes-us: @rjeberhard The new simplification ties helm chart version to operator version, right? I assume this is OK?  Or are there cases when customers want to use the latest helm chart but an older version of the operator image?

rosemarymarano: I can make specific edits, but if sections need more info or elaboration, please add the description to the JIRA [OWLS-97331](https://jira.oraclecorp.com/jira/browse/OWLS-97331).

rjeberhard: Oh, I think I understand... the upgrade use case is a special case because we've said "--reuse-values" which would preserve the image name from the original install. The customer could alternatively not use "--reuse-values" but instead provide the arguments from the original install or that the customer now wants.

robertpatrick: I am confused.  We can combine `—reuse-values` with a set of the image to point to the new version.  Is there some reason that this isn’t desirable or won’t work?  We definitely want to make things easy but for the upgrade use case, I feel like the combination of `—reuse-values` and `—set` to specify the new image name is the most understandable mechanism.

rjeberhard: Retracting my comment based on our discussion in Slack.

rjeberhard (approved):

PR #2831, opened 2022-03-10
Added more log and output files collection for scaling Via script testcase

marinakog:
1. Added operator log collection before operator pod is deleted
2. Switched to FINE for operator logging 
3. Added collection for scalingAction output and log if it fails
4. Switched test case to use unique namespace.
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9252


integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/OperatorUtils.java
```diff
@@ -280,6 +280,7 @@ public static OperatorParams installAndVerifyOperator(String opNamespace,
     OperatorParams opParams = new OperatorParams()
         .helmParams(opHelmParams)
         .imagePullSecrets(secretNameMap)
+        .javaLoggingLevel("FINE")
```
anpanigr, line 283: Will this not cause a ton of information for all the tests. Can we create the OperatorParam on the tests with the FINE logging level.

rjeberhard: I agree... What is it that you're looking for that you would need the FINE level?

marinakog: changed code to use FINE logging only for one test

rjeberhard: Okay, but why would we commit this change? Why not use it only when trying to diagnose this problem?

marinakog: it is intermittent and happen only in OKE run, I have run OKE nightly against this branch for 2 days and can't get it failed anymore :).

rjeberhard (approved):

anpanigr (approved):

PR #2851, opened 2022-03-21
Simply computation of events from conditions

russgold:
Wherever possible, events are now computed directly from added or removed conditions. Failures are now created in a single place, which will allow us to log them at the appropriate severity level once we add that functionality.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
@@ -4,9 +4,18 @@
 package oracle.kubernetes.weblogic.domain.model;
 
 import oracle.kubernetes.json.Obsoleteable;
+import oracle.kubernetes.operator.helpers.EventHelper;
+
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_AVAILABLE;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_COMPLETE;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_FAILURE_RESOLVED;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_INCOMPLETE;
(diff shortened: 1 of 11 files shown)


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainConditionType.java
```diff
@@ -4,9 +4,18 @@
...
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_COMPLETE;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_FAILURE_RESOLVED;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_INCOMPLETE;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_ROLL_COMPLETED;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_ROLL_STARTING;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_UNAVAILABLE;
 
 public enum DomainConditionType implements Obsoleteable {
-  Failed {
+  Failed(null, DOMAIN_FAILURE_RESOLVED) {
```
doxiao, line 18: It would be more consistent/clean if the addedEvent for the Failed condition can be handled the same as others. I understand that Domain_Failed event has the message and reason fields that need to be set, but the constructor of the event can take the condition object where the values can be retrieved.

russgold: Addition of a Failed condition is handled differently in a number of ways, including frequently coming from an explicit failure step and requiring logging (not yet implemented). That is in contrast to the common behavior of other conditions, which are computed and changed. In addition, while for the other conditions, we typically only want to add an event on a transition (such as not rolling to rolling), for failures, we generally want to log them and add events whenever they happen - that is, most failures are not state-changed, but an error during processing. For that reason, we cannot infer them from a change in the domain status.

For example. If we get two kubernetes failures for the same call in subsequent make-rights, that counts a two failures. It should be logged twice and we should report two events. But the domain status should not change: the first Failed condition will be preserved.

rjeberhard (approved):

PR #2853, opened 2022-03-22
Jenkinsfile-based Kind IT tests

robertpatrick:


Jenkinsfile
```diff
@@ -0,0 +1,576 @@
+// Copyright (c) 2017, 2022, Oracle and/or its affiliates.
```
rjeberhard, line 1: Very minor point, but as a new file this should only have the 2022 date.

Jenkinsfile
```diff
@@ -0,0 +1,576 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+//
+def kind_k8s_map = [
+     '0.8': [
```
rjeberhard, line 5: I would be okay with eliminating all of the data for pre-0.11.1 Kind. We're not going to ever go back to those releases.

Jenkinsfile
```diff
@@ -0,0 +1,576 @@
...
+
+        kind_name = "kind"
+        kind_network = "kind"
+        registry_name = "kind-registry"
+        registry_host = "${registry_name}"
+        registry_port = "5000"
+
+        start_time = sh(script: 'date +"%Y-%m-%d %H:%M:%S"', returnStdout: true).trim()
+        wle_download_url="https://github.com/oracle/weblogic-logging-exporter/releases/latest"
+        wko_build_email='robert.patrick@oracle.com'
```
rjeberhard, line 118: Can we use the ecnj_sonar account? This seems like the kind of thing we don't want CorpArch to find if they do an audit.

Jenkinsfile
```diff
@@ -0,0 +1,576 @@
...
+               description: 'kubectl version',
+               defaultValue: '1.21.5'
+        )
+        string(name: 'HELM_VERSION',
+               description: 'Helm version',
+               defaultValue: '3.7.2'
+        )
+        string(name: 'ISTIO_VERSION',
+               description: 'Other Possible Values 1.7.3, 1.8.1, 1.7.6',
+               defaultValue: '1.10.4'
```
rjeberhard, line 181: I thought that we were defaulting to 1.11 now

sankarpn:
Since ocir is internal do we rename it to something generic?

Jenkinsfile
robertpatrick: Of course.  As it turns out this Jenkins server's email configuration is still pointing at stbeehive so email doesn't work.  Unless we fix this, we don't need at email address anyway.  One downside of using this generic account's email address is that someone will have to be checking that user's email to see failure emails.  Might be better to use a mailing list instead.

Jenkinsfile
robertpatrick: I just copied the default from the current job.  We can change this, if desired.

Jenkinsfile
robertpatrick: Obviously, we can.  The question is does the 3.3 or 3.4 branches need earlier versions?

rjeberhard: No, they don't. We've been running all branches with the modern Kind binary for a while.

Jenkinsfile
rjeberhard: Derek had me sign-up for that account, so I get the reports. Now that I think about it, the ecnj_sonar account doesn't make any sense for this other than that it isn't a personal account. I'm really okay not getting or sending e-mail about build failures.

robertpatrick:
> Since ocir is internal do we rename it to something generic?

OCIR is not internal, it is an Oracle Cloud offering so I don't think there should be any need to worry about this...

Jenkinsfile
rjeberhard: Okay, I don't think this should hold up your PR. It will be an ongoing activity to update defaults.

sankarpn (approved):
lgtm

PR #2862, opened 2022-03-25
Avoid adding two copies of subsequent steps

russgold:
Addresses bug OWLS-97689

The problem was caused by the DomainStatusUpdater adding two copies of the `next` step chain to the fiber: one as part of creating the status update step, and one after the event steps. In production, that chain ends with the DomainProcessorImpl.TailStep, which terminates the fiber. As a result, the event steps were never being run (nor the duplicate domain processing).

The unit tests did not catch it because the unit test equivalent, _TerminalStep_ was invoking the default `doNext` method, which does run the subsequent steps. Changing that explicitly to select `null` as the next step, reproduces the error and confirms that avoiding the duplication solves the problem.

Jenkins jobs:
[ItKubernetesEvents](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9416/)
[ItPodsRestart](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9420/)
[ItIntrospectVersion](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9418/)

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -318,8 +318,9 @@ public class DomainStatusUpdater {
       return doNext(createContext(packet).createUpdateSteps(getNext()), packet);
     }
 
+    // Note: this step is created with no next step, as that is added via a call to Step.chain, later.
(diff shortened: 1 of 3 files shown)


rjeberhard (approved):

doxiao (approved):
LGTM.

PR #2864, opened 2022-03-25
Clear topology failures when no longer needed

russgold:
Rather than clearing all failures at the start of a make-right, which can cause them to flicker, this commit ensures that the topology validation preserves any current `TopologyMismatch` failures and only deletes them if they are no longer applicable. The step which formerly cleared all failures now specifically excludes this type. Future changes will address other cases.

I see one failure; I don't know if it is an intermittent one.

Jenkins run 1 https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9423/
Jenkins run 2 https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9432/ (updated with merge from Main)

I note that we get different failures this time, so I am presuming that they are intermittent.


rjeberhard:
You get the first 100% coverage award!

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainCondition.java
```diff
@@ -55,6 +56,9 @@
   @NotNull
   private String status = "True";
 
+  // internal: used to select failure conditions for deletion
+  private volatile boolean marked;
```
doxiao, line 60: The comment helps, but the field name could be more informative, IMO.

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainStatusConditionMatcher.java
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2019, 2022, Oracle and/or its affiliates.
```
doxiao, line 1: Is this a new class? if so, the copyright is not correct.

doxiao (commented):
Look good overall. Had a couple of minor comments.

Do we need to add unit test cases for the newly added createRemoveSelectedFailuresStep and createRemoveUnSelectedFailuresStep?

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainStatusConditionMatcher.java
russgold: I renamed it. It used to be called DomainConditionMatcher. Apparently git doesn't preserve history in such cases.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainCondition.java
russgold: will change

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -183,11 +183,27 @@ public static Step createStatusInitializationStep() {
...
 
+  /**
+   * Asynchronous step to remove selected failure conditions.
+   */
+  public static Step createRemoveSelectedFailuresStep(DomainFailureReason... selectedReasons) {
+    return new RemoveSelectedFailuresStep(Arrays.asList(selectedReasons));
+  }
+
+  /**
+   * Asynchronous step to remove selected failure conditions.
```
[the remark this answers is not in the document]
russgold: oops. Fixing

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainStatusConditionMatcher.java
doxiao: I see it now.

doxiao (approved):

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainStatusConditionMatcher.java
robertpatrick: @russgold `git add -A` will allow git to preserve the history across file renames…

russgold: Ah, thanks. I had thought it would do so automatically. I'll take advantage of this in the future.

rjeberhard (approved):

PR #2878, opened 2022-03-29
Kubernetes Client Java 15.0.0 and update Enum pattern to match

rjeberhard:


bryantidd (approved):

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainTestBase.java
```diff
@@ -263,11 +263,11 @@ void whenOnlyAsStateSpecified_managedServerDesiredStateIsRunning() {
 
   @Test
   void whenClusterStateSpecified_managedServerDesiredStateIsAsSpecified() {
-    configureCluster(CLUSTER_NAME).withDesiredState("NEVER");
+    configureCluster(CLUSTER_NAME).withDesiredState(ServerStartState.ADMIN);
 
     ServerSpec spec = domain.getServer(SERVER1, CLUSTER_NAME);
 
-    assertThat(spec.getDesiredState(), equalTo("NEVER"));
+    assertThat(spec.getDesiredState(), equalTo("ADMIN"));
```
ankedia, line 270: Can this use ServerStartState.ADMIN instead of "ADMIN"?

rjeberhard: I found that `getDesiredState()` was being used for status and would be set to values, such as "SHUTDOWN", that weren't legal values for the enum, so I had to leave it as String. This value isn't customer-visible.

rjeberhard:
Anil approved in Slack :)

PR #2891, opened 2022-03-31
owls-96791 - Rerun introspector immediately after changing auxiliary image or image

alai8:
jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9513


ankedia:
Does this also cover the cases when the auxiliary image is deleted or other attributes such as the `image-pull-policy` of the aux image is changed? I'm really not sure if we need to re-run the introspector if the aux image attributes such as `image-pull-policy` changes.

alai8:
> Does this also cover the cases when the auxiliary image is deleted or other attributes such as the image-pull-policy of the aux image is changed? I'm really not sure if we need to re-run the introspector if the aux image attributes such as image-pull-policy changes.

Only the `image` attribute is checked. Changes to other attributes such as `image-pull-policy` would not trigger a rerun.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
```diff
@@ -311,4 +314,19 @@ protected boolean isK8sContainer(V1Container c) {
...
+
+  private boolean isAuxiliaryContainerName(@Nonnull String name) {
+    return name.startsWith(AUXILIARY_IMAGE_INIT_CONTAINER_NAME_PREFIX);
+  }
+
+  protected Optional<V1Container> getAuxiliaryContainer(V1PodSpec v1PodSpec) {
+    return Stream.ofNullable(v1PodSpec.getInitContainers())
+        .flatMap(Collection::stream)
+        .filter(this::isAuxiliaryContainer)
+        .findFirst();
```
ankedia, line 330: It looks like this is only checking for the first auxiliary image. There can be multiple auxiliary images such as one aux image for WDT install and one or more images containing WDT models etc. It'll be good to handle changes to all the init container images.

alai8: Thanks for catching this. Updated code to check for auxiliary images in multiple initcontainers.

rjeberhard (approved):

rjeberhard:
I've started another integration test run (pr-merge) to see if it will be clean.

PR #2898, opened 2022-04-01
Port in use debugging

sankarpn:
Created a PortInuseEventWatcher class to start collecting services and lsof logs when the integration tests encounters "port already in use" error. It runs as a thread scanning the events across all namespaces(default and starting with "ns-") every 10 seconds. When it encounters a port in use error it dumps the list of services and events across all namespaces and lsof logs.

The thought behind this is to get more information on what is using the port in question.
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9673/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9683/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9684/


robertpatrick (approved):

vanajamukkara (approved):
LGTM

PR #2902, opened 2022-04-04
Owls 97471 - Generate events for the failure paths in the conversion webhook

ankedia:
OWLS-97471 - Generate events in the conversion webhook for the unhappy paths to improve the diagnosability of the conversion webhook. 

Integration test run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9682/


tbarnes-us:
General comment: IMO, it'd be helpful to have a positive event for Conversion WebHook deployment success, and another for undeployment (if practical). This will give WH sufficient event logging to determine not only if it has failed, but if it is properly running at all...

common/src/main/resources/Operator.properties
```diff
@@ -178,6 +178,7 @@ WLSKO-0218=Domain custom resource converter: Name of the converted file. If not
   file name by appending "__converted." and the input file extension to the base name of the input file name.
 WLSKO-0219=Enable overwriting the existing output file, if any.
 WLSKO-0220=Successfully generated upgraded domain custom resource file ''{0}''.
+WLSKO-0221=Domain custom resource conversion failed due to ''{0}''.
```
tbarnes-us, line 181: I'm not sure - but it seems to me that all of the webhook related messages -- the old ones and this new one - should ideally be updated to clearly call out their unique origin "WebLogic Domain custom resource conversion webhook" or similar. Not just "conversion" or just "webhook", but perhaps minimally "domain resource conversion webhook"...

robertpatrick (changes requested):
The main build job is failing with compiler errors.  You need to resolve these errors and get the tests passing at https://build.weblogick8s.org:8443/job/WebLogic%20Kubernetes%20Operator%20Kind/view/change-requests/job/PR-2902/

ankedia:
> The main build job is failing with compiler errors. You need to resolve these errors and get the tests passing at https://build.weblogick8s.org:8443/job/WebLogic%20Kubernetes%20Operator%20Kind/view/change-requests/job/PR-2902/

[ANIL] The build passes in jenkins run and in my local env. I'll take a look but am curious what's the difference between external jenkins run and this build, any ideas? 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9797/console

tbarnes-us:
General comment: Take this with a grain of salt as I am told that this was already discussed in design meetings, but it is a bit surprising IMO that failed domains always get "evented" in the NS of the WH instead of logged in the same NS as the domain, unlike almost all other domain related failures, so...  Could we "best effort" put domain failures in the same namespace as the domain? For example, on a parse failure, use a simple reg-ex to best effort try and get the namespace... OTOH, such failures are already communicated back by failing the a domain creation/application so perhaps we're already being sufficiently helpful for interactive use cases if not automated use cases...

robertpatrick:
The new build job is our new gate for merging PRs into `main` (and soon, into `release/3.4`).  It only runs a subset of the tests and is defined in the `Jenkinsfile` in the project root directory.  One difference is that it is using JDK 17 and Maven 3.8.5...

ankedia:
> General comment: IMO, it'd be helpful to also have a positive event for Conversion WebHook deployment success, domain success in the namespace of the domain (perhaps not this one? the operator already events this one too, so an event here might distract newbies from the operator event), and another for undeployment (if practical). This will give WH sufficient event logs sufficient context to reveal not only if it has failed (as this pull implements), but also if it is properly running at all... Making it easier to debug when the WH can be ignored, or to compare success to failure, etc, etc.

[ANIL] There are events generated by the kubernetes during webhook deployment and undeployment. For e.g. below `Scaled up replica set` is generated during WH deployment. In addition, the deployment READY status shows `1/1` and `Running` after deployment is ready that can be used by interactive and automated use-cases. This is similar to how the operator deployment works currently and I'm not sure if generating events for  `Conversion WebHook deployment success` is going be very helpful.

```
71s         Normal    ScalingReplicaSet        deployment/weblogic-operator-webhook             Scaled up replica set weblogic-operator-webhook-c5856ff5b to 1
```
```
2s          Normal    Killing                  pod/weblogic-operator-webhook-c5856ff5b-pltjp    Stopping container weblogic-operator-webhook
```

ankedia:
> General comment: Take this with a grain of salt as I am told that this was already discussed in design meetings, but it is a bit surprising IMO that failed domains always get "evented" in the NS of the WH instead of logged in the same NS as the domain, unlike almost all other domain related failures, so... Could we "best effort" put domain failures in the same namespace as the domain? For example, on a parse failure, use a simple reg-ex to best effort try and get the namespace... OTOH, such failures are already communicated back by failing the a domain creation/application so perhaps we're already being sufficiently helpful for interactive use cases if not automated use cases...

[ANIL] Yes, the `kubectl apply -f` command will fail when the conversion webhook fails and domain will not get created. It's possible that the failure happens even before the domain namespace is created and/or the request payload doesn't have domain namespace during failure. In such cases, it won't be possible to create the event in the domain namespace.

common/src/main/resources/Operator.properties
ankedia: Changed in [c11a13e](https://github.com/oracle/weblogic-kubernetes-operator/pull/2902/commits/c11a13e7113f1e3c533b7fac583440da62021087).

operator/src/main/java/oracle/kubernetes/operator/EventConstants.java
```diff
@@ -78,4 +80,5 @@
   String SERVER_POD_ERROR_SUGGESTION = "";
   String REPLICAS_TOO_HIGH_ERROR_SUGGESTION = "Lower replicas in the domain resource, or increase the number "
       + "of WebLogic servers in the WebLogic domain configuration for the cluster.";
+  String DOMAIN_CONVERSION_FAILED_PATTERN = "WebLogic Domain custom resource conversion webhook failed due to '%s' %s";
```
rjeberhard, line 83: Why is this here? You can access the string from the resource bundle and format against it without having a copy in the code.

operator/src/main/java/oracle/kubernetes/operator/helpers/EventHelper.java
```diff
@@ -615,6 +622,42 @@ public V1ObjectReference createInvolvedObject(EventData eventData) {
...
+      }
+
+      @Override
+      public String getReason() {
+        return CONVERSION_WEBHOOK_FAILED_EVENT;
+      }
+
+      @Override
+      public String getPattern() {
+        return DOMAIN_CONVERSION_FAILED_PATTERN;
```
rjeberhard, line 639: Okay, the constant with the English text is used here and is used in a user visible string... This needs to come from the resource bundle.

From DomainValidationMessages.java (you can do this):
```
  private static String getBundleString(String key) {
    return ResourceBundle.getBundle("Operator").getString(key);
  }
```

common/src/main/resources/Operator.properties
```diff
@@ -178,6 +178,7 @@ WLSKO-0218=Domain custom resource converter: Name of the converted file. If not
   file name by appending "__converted." and the input file extension to the base name of the input file name.
 WLSKO-0219=Enable overwriting the existing output file, if any.
 WLSKO-0220=Successfully generated upgraded domain custom resource file ''{0}''.
+WLSKO-0221=WebLogic Domain custom resource conversion webhook failed due to ''{0}''.
```
tbarnes-us, line 181: A) Is there additional information available? Name of the domain? (anything to help identify the source when available)

B) I still see missing decorations on other WH text ^^^, for example 0211, 0206, 0207, and possibly others

common/src/main/resources/Operator.properties
tbarnes-us: Looks like 0221 is fixed now - left a new comment for the others.

operator/src/main/java/oracle/kubernetes/operator/EventConstants.java
ankedia: I was following same pattern as other messages such as `REPLICAS_TOO_HIGH_ERROR_SUGGESTION `, I have removed it.

operator/src/main/java/oracle/kubernetes/operator/helpers/EventHelper.java
ankedia: Changed to get it from resource bundle and use LOGGER.formatMessage().

common/src/main/resources/Operator.properties
ankedia: (a) I added conversion request details to the message. 
(b) Since 0211, 0206 and 0207 were added in previous PR, I missed it earlier. I have added decoration now.

rjeberhard (approved):

robertpatrick (approved):

tbarnes-us (approved):

PR #2908, opened 2022-04-05
update and divide WebLogic Images doc

rosemarymarano:
Divided the WebLogic Images into two separate documents. Updated them according to our meeting discussion and per Robert's JIRA [OWLS-95907](https://jira.oraclecorp.com/jira/browse/OWLS-95907), where applicable.

### DIFF
--- a/documentation/staging/content/userguide/base-images/_index.md
+++ b/documentation/staging/content/userguide/base-images/_index.md
@@ -2,132 +2,85 @@
 title: "WebLogic images"
 date: 2019-02-23T16:45:55-05:00
 weight: 6
-description: "Create, obtain, or inspect images for WebLogic Server or Fusion Middleware Infrastructure deployments."
+description: "Obtain and inspect images for WebLogic Server or Fusion Middleware Infrastructure deployments."
 ---
+
 
 ### Contents
 
 - [Overview](#overview)
+- [Understand Oracle Container Registry images](#understand-oracle-container-registry-images)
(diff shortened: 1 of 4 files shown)


tbarnes-us:
I have a few general comments/recommendations, as the "user guidance" has been somewhat lost with the re-org:

rjeberhard (approved):

rosemarymarano:
@tbarnes-us In response to your comments/suggestions:
A. Yes, I agree that both documents are/should be at the same level. I'll create an index page and doing the re-naming as appropriate.
B.  "The deleted overview material that ties the chapters together by domain home source type" I found that this information was basically repeated in the "Apply patched images to a running domain" subsections (Domain in PV, Model in Image with auxiliary images, Model in Image without auxiliary images, and Domain in Image) and definitely does **not** warrant/nor would it make sense to put it on the upper level index page. 
C. Because the document is about creating custom images with WIT, I am quite comfortable with the first para being, "Use the WebLogic Image Tool to create custom images." IMO, creating custom images using WIT **is** the overview, despite it containing detailed information with regard to the commands. 
D. Yes, I agree with you here: "The "dynamic update" chapter does not quite fit in the 'custom image' chapter with the new org. Perhaps give it its own chapter on same level as the other two chapters." I will make "Apply patched images to a running domain" a separate document, at the same level as the other two.
E.  "In short, propose: top level _index chapter "WebLogic Images", and three sub chapters "OCR Images", "Custom Images", and "Patch Running Domains" - YES. Very good. As always, thank you for your input and suggestions!

robertpatrick (approved):

PR #2910, opened 2022-04-05
Tag tests for gate and samples-gate

vanajamukkara:
Jenkins results - https://build.weblogick8s.org:8443/job/WebLogic%20Kubernetes%20Operator%20Kind/view/change-requests/job/PR-2910/16/
https://build.weblogick8s.org:8443/job/WebLogic%20Kubernetes%20Operator%20Kind/view/change-requests/job/PR-2910/15/

https://build.weblogick8s.org:8443/job/WebLogic%20Kubernetes%20Operator%20Kind/view/change-requests/job/PR-2910/


Jenkinsfile
```diff
@@ -584,7 +584,7 @@ EOF
                                 export NO_PROXY
                             fi
 
-                            if ! time mvn -pl integration-tests -P ${MAVEN_PROFILE_NAME} verify 2>&1 | tee "${result_root}/kindtest.log"; then
+                            if ! time mvn -pl integration-tests -Dgroups="gate,samples" -DskipITs=false verify 2>&1 | tee "${result_root}/kindtest.log"; then
```
robertpatrick, line 587: I thought we were going to change the Maven profiles to specify the groups instead of the test classes?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItAuxV8DomainImplicitUpgrade.java
```diff
@@ -151,6 +152,7 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
    */
   @Test
   @DisplayName("Test implicit upgrade of v8 version of Auxiliary Domain")
+  @Tag("gate")
```
robertpatrick, line 155: This seems to be a new gating test.  Are we confident that it is stable?

robertpatrick (commented):
You seem to have added a bunch of new gate tests that we were not running before.  My concerns with this are:

- Are these tests stable?
- How much time are we adding to the test execution?

Also, you have used the samples tag for a lot more samples than what the gating build was previously running (and that we discussed yesterday in the meeting).  I suspect that the gating build is going to take 4 or 5 hours now--I think that is too long.  My suggestion on the samples is to use two tags, `samples-gate` and `samples`, where `samples-gate` are the tests we want to run in the gating build and the `samples` tags represents the rest of the samples not labeled with `samples-gate`.

vanajamukkara:
> You seem to have added a bunch of new gate tests that we were not running before. My concerns with this are:
> 
>     * Are these tests stable?
> 
>     * How much time are we adding to the test execution?
> 
> 
> Also, you have used the samples tag for a lot more samples than what the gating build was previously running (and that we discussed yesterday in the meeting). I suspect that the gating build is going to take 4 or 5 hours now--I think that is too long. My suggestion on the samples is to use two tags, `samples-gate` and `samples`, where `samples-gate` are the tests we want to run in the gating build and the `samples` tags represents the rest of the samples not labeled with `samples-gate`.

I picked the tests to cover all the features(at least one in each feature) as we discussed for gate. This is just to start, I need to see how long these tests take and how stable are they by running these tests on Jenkins multiple times, open for any changes based on that. I can use the two tags `samples-gate` and `samples`.

Jenkinsfile
vanajamukkara: For that I would have to tag all the classes and change all the profiles to use the groups. @anpanigr suggested to first do the `gate` and then go for the rest.

robertpatrick: Right, but now you have a Jenkinsfile job with a dead parameter (MAVEN_PROFILE_NAME) that isn't being used anywhere.

robertpatrick: If we just changed the `wls-srg` profile to use groups, wouldn't that allow you to do just the gate/samples-gate tags first and not have to convert everything else?

vanajamukkara: I think I tried that and did not work. Let me try again and see.

vanajamukkara: modified `wls-srg` profile to use groups

robertpatrick (approved):

rjeberhard (approved):

sankarpn (approved):
lgtm

PR #2911, opened 2022-04-05
OWLS-98083 - Create port-forward NAP when admin server name not defined in WDT model

ankedia:
OWLS-98083 - This change handles the case where admin server name is not defined in WDT model file. It uses the default admin server name `AdminServer` when creating the network channel in admin server for port-forwarding. 

Jenkins run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9822/consoleFull


jshum2479 (approved):

operator/src/main/resources/scripts/model_wdt_mii_filter.py
```diff
@@ -318,14 +318,19 @@ def customizeServers(model):
 
 def customizeServer(model, server, name):
   listen_address=env.toDNS1123Legal(env.getDomainUID() + "-" + name)
-  adminServer = model['topology']['AdminServerName']
+  try:
+    adminServer = model['topology']['AdminServerName']
+  except KeyError:
+    adminServer = None
```
robertpatrick, line 324: Why not just use an if test instead of relying on an exception?

```
adminServer = None
if 'AdminServerName' in model['topology'] and len(model['topology']['AdminServerName']) > 0:
    adminServer = model['topology']['AdminServerName']
```

ankedia: Changed to use the `if` test as suggested. thanks.

robertpatrick (approved):

PR #2924, opened 2022-04-07
check rolling condition at test end

russgold:
Some code simplification (mostly removal of duplicate code) and the addition of a validation that the `Rolling` condition is not present at the end of the run.


rjeberhard (approved):

sankarpn (approved):
lgtm

Issue #2819, opened 2022-03-07
Introspector pod fails initialisation


th3n3rd:
## Context
I am using the Model In Image deployment strategy, with auxiliary images, in order to deploy a sample application on k8s via the Oracle WebLogic Operator (v3.3.7).
The auxiliary image currently uses ubuntu bionic as a base image, which default shell is dash as in `/bin/sh -> dash`.
The latter does only support POSIX compliant instructions.

## Problem
When the introspector pod kicks in, the init containers fails to initialise the pod with the following error:

```shell
./auxImage.sh: 25: [: 1: unexpected operator
./auxImage.sh: 27: ./auxImage.sh: source: not found
[SEVERE] Missing file /weblogic-operator/scripts/utils_base.sh
```

The source of the problem seems to be related to the [auxImage](https://github.com/oracle/weblogic-kubernetes-operator/blob/main/operator/src/main/resources/scripts/auxImage.sh#L1) script shebang setup, currently equal to `#!/bin/sh`.
The script currently uses a non portable set of instructions, i.e. non-POSIX compliant, hence if the default shell in the OS is strictly supporting only POSIX instructions (like `dash`) the script does not work as expected.

## Potential solutions
1. Replace the shebang with `#!/bin/bash`, leaving the code as is
2. Replace script instructions with portable POSIX compliant ones, e.g. use `.` instead of `source`, etc

robertpatrick:
This issue should be addressed and available in the upcoming 3.4.0 release.

PR #2925, opened 2022-04-07
changing script to require bash

robertpatrick:
Fixing #2819 in main

### DIFF
--- a/operator/src/main/resources/scripts/auxImage.sh
+++ b/operator/src/main/resources/scripts/auxImage.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # Copyright (c) 2021, 2022, Oracle and/or its affiliates.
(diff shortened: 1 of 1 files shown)


rjeberhard (dismissed):

ankedia:
AFAIK, busybox doesn't have bash shell. Will this work in environments that doesn't have bash shell?

rjeberhard:
Oh, right... I forgot about that.

robertpatrick:
@rjeberhard @ankedia OK, I didn't think about that.  Then the right solution is to use `.` instead of `source` (which is not a POSIX-compliant command)

rjeberhard (approved):

ankedia:
> @rjeberhard @ankedia OK, I didn't think about that. Then the right solution is to use `.` instead of `source` (which is not a POSIX-compliant command)
[ANIL] Ok. I think `source` is also used in other scripts such as `startServer.sh` and `utils.sh`. However those scripts run in the main container which I assume is based on OEL. If this is the only non-POSIX compliant command, then the change looks good to me.

PR #2948, opened 2022-04-12
Merge topology checks

russgold:
The commit merges the two sets of topology validations into a single set of code and tests. All are now handled similarly, with regard to replacing old failures so that they don't flicker, generating events and logging. 

Note that currently, they are logged as warnings and do not halt the make-right, which which be changed in a future PR.

Jenkins run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/9982/


rjeberhard (approved):

ankedia (approved):

rjeberhard:
@russgold, I've run the gating tests twice with the same (or at least similar -- I didn't look carefully) failures.

russgold:
> @russgold, I've run the gating tests twice with the same (or at least similar -- I didn't look carefully) failures.

I'm seeing failures to POST to WLS instances, and NPEs in ResponseStep. Investigating now.

russgold:
The operator log now contains messages such as:

```
{"timestamp":"2022-04-12T23:27:48.757980941Z","thread":28,"fiber":"engine-operator-thread-9-fiber-556 NOT_COMPLETE","namespace":"ns-ewvvlt","domainUID":"myintrodomain","level":"WARNING","class":"oracle.kubernetes.operator.helpers.WlsConfigValidator","method":"reportFailure","timeInMillis":1649806068757,"message":"Cluster 'cl2' specified in the domain resource does not exist in the WebLogic domain home configuration.","exception":"","code":"","headers":{},"body":""}
```

which is a direct result of this change - it means that our tests are configuring topology elements that don't exist. I think we'll need help from the testing group to fix this.

robertpatrick (approved):

common/src/main/resources/Operator.properties
```diff
@@ -201,16 +201,16 @@ WLSDO-0014=The domain {0} introspect job failed, it will automatically retry in
   retries = {2}
 WLSDO-0015="Retrying failed domain {0} introspect job, retry {1} of a maximum {2}."
 WLSDO-0016=DomainUID ''{0}'' introspector job name ''{1}'' exceeds maximum allowed length ''{2}''.
-WLSDO-0017=DomainUID ''{0}'' and cluster name ''{1}'' combination ''{2}'' exceeds maximum allowed length ''{3}''.
```
doxiao, line 204: The message is probably clearer with the combination in it because the combination includes additional characters, such as "-", besides the domainUID and cluster name.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -266,8 +266,9 @@ public static Step createAbortedFailureSteps() {
    * Asynchronous steps to set Domain condition to Failed and to generate DOMAIN_FAILED event.
    *
    * @param message a fuller description of the problem
+   * @param next the next step to run. May be null.
    */
-  public static Step createTopologyMismatchFailureSteps(String message) {
+  public static Step createTopologyMismatchFailureSteps(String message, Step next) {
```
doxiao, line 271: I don't see newly added "next" is used.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -173,103 +165,19 @@ static class ValidateDomainTopologyStep extends Step {
...
-      boolean isExplicitRecheck = MakeRightDomainOperation.isExplicitRecheck(packet);
-      WlsDomainConfig wlsDomainConfig = (WlsDomainConfig) packet.get(ProcessingConstants.DOMAIN_TOPOLOGY);
-      validate(info, wlsDomainConfig);
-
-      return doNext(getNextStep(info.getValidationWarningsAsString(), isExplicitRecheck, getNext()), packet);
+      List<String> failures = new WlsConfigValidator(packet).loggingTo(LOGGER).getFailures();
+      if (failures.isEmpty()) {
+        return doNext(createRemoveSelectedFailuresStep(getNext(), TOPOLOGY_MISMATCH), packet);
+      } else {
+        return doNext(createTopologyMismatchFailureSteps(getNext(), failures), packet);
```
doxiao, line 174: Is it possible that the domain already has a TopologyMistach failure condition from the previous make-right operation? If so, we may need to remove the old one before adding a new one.

doxiao (changes requested):

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
russgold: Yes, it was supposed to be passed to the FailureStep. Thanks for catching it.

common/src/main/resources/Operator.properties
russgold: I spoke with Rosemary, who felt that it made the message actually more difficult to understand.

doxiao: we could reword it to make it easier to understand, but the key is that we need to show the generated name in the message so that customer knows why it exceeds the limit.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
russgold: In the test, we update the topology and the domain and then compare the new domain with the old topology, which are expected to be incompatible. We're not supposed to be comparing them. It is possible that failing to call the correct FailureStep constructor is actually causing the test to fail, but the logic needs to be fixed in any case.

common/src/main/resources/Operator.properties
russgold: Why don't we discuss with Rosemary; I think that showing the generated message doesn't provide any more information than the number of total characters allowed.

doxiao: An example may help here. 

Let us say we have a domain with UID mydomain1, and a cluster mycluster1. The generated service name for this cluster is mydomain1-cluster-mycluster1, which has 28 characters. If the limit is 25 characters (for the sake of the discussion), the customer would see an error message like

The combination of UID "mydomain1" and cluster name ''mycluster1" exceeds maximum allowed length 25. 

But the total number of characters in domainUID and cluster name is only 19.

doxiao: What about something like the following?

WLSDO-0017=The generated service name "{3}" for domain ''{0}'' and cluster name ''{1}'' exceeds maximum allowed length ''{2}''.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
doxiao: My comment is general, not related to any test cases.  
If we have a TopologyMismatch failure in the previous make-right, what should we do?

common/src/main/resources/Operator.properties
russgold: In all cases where the generated name is too long, the number displayed in the error message is the number allowed for the combination of uid and name. So in your case, it would say 

The combination of UID "mydomain1" and cluster name ''mycluster1" exceeds maximum allowed length 16. The number displayed is not the maximum size of the _generated_ name. That would indeed not be reasonable. It is the maximum characters that the user is allowed to specify.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
russgold: TopologyValidationStepTest has two tests that address this issue:

``removeOldToplogyFailures`` verifies that when the validation step is run, if there were topology failures from the previous make-right and they have since been corrected, they are removed.

``preserveTopologyFailuresThatStillExist`` verifies that if the problems have not been corrected, the old failures are not removed, and the original failure time is preserved.

This is the pattern that will be used for all failures, to avoid the flickering issue you found previously. Rather than clearing all failures at the start of the make-right, it is the responsibility of steps which check for failures to remove those that no longer exist. I'm not sure yet how to handle Kubernetes failures, but I think this can be made to work for ServerPod failures - and it certainly works for state-based failures such as DomainInvalid, TopologyMismatch, Introspection and ReplicasTooHIgh.

russgold:
The integration test failure, it turns out, was not related to the comparison of the new domain with the old topology, although that is a problem. The actual failures were happening because of the oversight that Dongbo noticed - that the `next` parameter was unused. There was no relevant unit test, so I didn't notice when I didn't complete my refactoring.

The tests are now clean: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10048/

common/src/main/resources/Operator.properties
robertpatrick: I think that showing the user the generated name makes it clearer to the user.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
doxiao: I am not suggesting to remove all failures at the start of the make-right. I was thinking about a situation where the domain failed with some topology mismatch errors, and then the user changed the domain to fix the problem, but still has some errors remaining, or introduced new errors.  The remaining and new errors will be reported. The question is what we need to do with the original/previous failure condition. I think it is confusing to leave the stale failure condition in the domain status because errors in the original failure condition do not apply to the current version of the domain spec any more.

common/src/main/resources/Operator.properties
russgold: @doxiao Please show me where you think that is happening. The unit test explicitly asserts that the maximum length allowed is the length of the user-specified names. 

My objection to including both the user-specified names and the resultant combined name is that the message has a lot of duplicate information and is difficult to read, given that it is close to two hundred chars long, and wrapped several times in all the places we display it.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
doxiao: Are you saying that the old Failed/TopologyMismatch condition is removed when the new Failed/TopologyMismatch condition is added? I don't see where it is done.

common/src/main/resources/Operator.properties
doxiao: Since your change has deducted the number of additional characters that the operator adds to the generated names. The information in the message is correct. The only issue is usability. I am okay with the current change since you have run it with Rosemary.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
russgold: Yes, look at DomainStatusUpdater.createTopologyMismatchSteps(). It appends a call to ``removingOldFailures(TOPOLOGY_MISMATCH)`` which is then used to select existing failures for removal.

doxiao: I see. thanks.

doxiao (approved):
LGTM.

PR #2953, opened 2022-04-14
Download Istio from object storage and update default Istio version

rjeberhard:


anpanigr (commented):
When we run the test outside Oracle Network for example e.g. Thirdparty Contractor of OpenShift certification or some other certification purpose, the objectstorage may not be accessible.

rjeberhard:
@anpanigr, yes, object storage is accessible to third parties because this object storage bucket is public. It's the same bucket that we're using for Helm, Kind, and kubectl downloads.

robertpatrick (approved):

PR #2955, opened 2022-04-15
separated doc

rosemarymarano:
Moved "Set up Kubernetes to access domain images" into a separate document, as per Tom B.


documentation/staging/content/userguide/base-images/access-images.md
```diff
@@ -0,0 +1,115 @@
...
+Alternatively, it may be preferable to manually place an image
+on each Kubernetes worker node in your Kubernetes cluster in advance.
+
+For example, if the desired image is located in a Docker registry,
+then you can manually call `docker login` and `docker pull` on each
+worker node. For the steps with Oracle Container Registry images,
+see [Obtain images from the Oracle Container Registry](#obtain-images-from-the-oracle-container-registry).
+
+As another example,
+if the Docker image is located in a local Docker cache,
```
rjeberhard, line 107: We've been slowly moving away from the term "Docker" except when referring specifically to that product. I think that "Docker image" is a place where we can globally move to "container image" or just "image" regardless if the context is about the Docker product. The same thing applies to "Docker registry", which can globally be "container registry".

robertpatrick: In my opinion, `image registry` is a more accurate term...

documentation/staging/content/quickstart/install.md
```diff
@@ -60,7 +60,7 @@ $ helm install traefik-operator traefik/traefik \
...
 
-    **Note:** Prior to version 3.1.0, the operator's Helm chart supported configuring only the namespaces that the operator would manage using a list of namespaces. The chart now supports specifying namespaces using a label selector, regular expression, or list. Review the available [Helm configuration values]({{< relref "/userguide/managing-operators/using-helm#domainnamespaceselectionstrategy" >}}).
+    **Note:** Prior to version 3.1.0, the operator's Helm chart supported configuring only the namespaces that the operator would manage using a list of namespaces. The chart now supports specifying namespaces using a label selector, regular expression, or list. Review the choices for selecting a domain namespace strategy in the [Namespace management]({{< relref "/userguide/managing-operators/namespace-management.md" >}}) document.
```
robertpatrick, line 64: How long do we need to keep this historical note around?

documentation/staging/content/userguide/base-images/access-images.md
rosemarymarano: I agree with both suggestions and will edit the doc accordingly.

documentation/staging/content/userguide/base-images/access-images.md
```diff
@@ -0,0 +1,115 @@
...
+then you can get an inventory of the cache by calling `docker images`:
+
+- Save the image to a TAR file: `docker save -o myimage.tar myimagerepo:myimagetag`.
+- Then, copy the TAR file to each node.
+- Call `docker load -o myimage.tar` on each node.
+
+If you choose this approach, then a Kubernetes Secret is not required
+and your domain resource `domain.spec.imagePullPolicy` must be set to `Never` or `IfNotPresent`.
```
robertpatrick, line 115: Hmm...  do we really need to even mention this manual distribution of images to node hack in this day and age?

documentation/staging/content/userguide/base-images/access-images.md
rjeberhard: Yeah, I agree about "image registry" except that my Google search suggests that most others have chosen "container registry", e.g. Azure Container Registry, Oracle Container Registry.

documentation/staging/content/userguide/base-images/access-images.md
rjeberhard: I suspect that customers who do this know why they selected it and don't need us to tell them. We've become more verbose over time because a surprising number of customers can't figure out why they are getting image pull backoff errors. I would lean toward agreeing that the comment about image pull secrets and registries is sufficient.

rosemarymarano: @rjeberhard Need clarification here: should I remove this sentence, "If you choose this approach, then a Kubernetes Secret is not required and your domain resource `domain.spec.imagePullPolicy` must be set to `Never` or `IfNotPresent`"?

documentation/staging/content/userguide/base-images/access-images.md
rosemarymarano: @rjeberhard Need clarification here: do you want me to search and replace all instances in the documentation of Docker image with "container image" or just "image"?

documentation/staging/content/userguide/base-images/access-images.md
rjeberhard: I think that @robertpatrick is suggesting that the entire option 3 is unnecessary. The reason is that an extremely small number of customers do this and, for those that do, they are already needing to do this for all of their images and so don't need our help to understand how to do this specifically for the operator's image.

rosemarymarano: okey doke. Got it. I will remove Option 3.

documentation/staging/content/userguide/base-images/access-images.md
rjeberhard: It looks like you are already using "image" in most places, so I think that it is unnecessary that these become "container image". The full "container image" only seems necessary if there is a possible ambiguity with the use of just "image". That is likely not the case in this document, so I'd leave them all as "image". We could generally have a first use of "container image" followed after by uses of "image", but I don't think that's necessary in this document as the context is clear.

robertpatrick: yeah, ok...

rosemarymarano: @rjeberhard Yes, for this document. However, we use the term, Docker registry, throughout the doc set. Do you want me to change them accordingly?

rjeberhard: Yes, please.

rosemarymarano: @rjeberhard FYI. Done, except for this AKS user guide document, https://oracle.github.io/weblogic-kubernetes-operator/userguide/aks/#configure-aks-cluster, which, as seen in the second table, **Field** column, has this field name, "Use a pre-existing WebLogic Server Docker image from Oracle Container Registry (OCR)?" I did not change anything that was related to this field.

robertpatrick (approved):

rjeberhard (approved):

documentation/staging/content/quickstart/install.md
tbarnes-us: +1 I think it can go - the fact that the Helm command is using the 'domainNameSelectionStrategy' option is enough of a hint that there are other name space selection options.

documentation/staging/content/userguide/base-images/access-images.md
```diff
@@ -0,0 +1,92 @@
...
+    imagePullSecrets:
+    - name: SECRET_NAME
+  ...
+  ```
+
+- If you are using the Oracle Container Registry, then
+  you must use the web interface to accept the Oracle Standard Terms and Restrictions
+  for the Oracle software images that you intend to deploy.
+  You need to do this only once for a particular image.
+  See [Obtain images from the Oracle Container Registry](#obtain-images-from-the-oracle-container-registry).
```
tbarnes-us, line 60: Broken link.

documentation/staging/content/userguide/base-images/access-images.md
```diff
@@ -0,0 +1,92 @@
...
+  ```
+
+  Note that this patch command entirely replaces the current list of
+  image pull secrets (if any). To include multiple secrets, use
+  the following format:
+  `-p '{"imagePullSecrets": [{"name": "my-registry-pull-secret"}, {"name": "my-registry-pull-secret2"}]}'`.
+
+For more information about updating a Kubernetes `ServiceAccount`
+for accessing the registry, see the Kubernetes documentation,
+[Configure service accounts for pods](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secrets-to-a-service-account).
```
tbarnes-us, line 92: This should be a relref - not a full link

documentation/staging/content/quickstart/install.md
rosemarymarano: Done.

documentation/staging/content/userguide/base-images/access-images.md
rosemarymarano: Fixed.

documentation/staging/content/userguide/base-images/access-images.md
rosemarymarano: Not sure why a link to the Kubernetes docs needs to be a relref and I cannot figure out how to do it. So, for now, I'm leaving it as is.

tbarnes-us: Oops - sorry - didn't look closely - I mistakenly thought this was a ref into our docs. Resolving.

PR #2979, opened 2022-04-21
OWLS-96896 - Restart Evicted Pods

### LINKED ISSUE #2794: Restart Evicted Pods
Hey,

I'm currently using the operator in version 3.3.6 with the Domain-Home-In-Image mode.
I'm running into an issue, where a Pod might get evicted from a Node due to not enough ephemeral storage being available for logging.

The pods are assigned to qosclass guaranteed (so requests == limits for both, cpu and memory).
The issue here is that once pods are evicted by kubernetes, the operator does not seem to care about them anymore, so they stay in an Evicted state, thus the application might get unavailable.

Right now I'm working around this using a serviceaccount and a simple batch-script to manually remove those Pods, so the Operator can take over control again, but I think this is something the Operator should take care of.

Thanks!

alai8:
resolves #2794 

jenkins https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10119/


common/src/main/resources/Operator.properties
```diff
@@ -185,6 +185,8 @@ WLSKO-0223=When fluentdSpecification is specified in the domain spec, a secret c
   must be specified in {0}
 WKSKO-0224=Fluentd configmap created.
 WKSKO-0225=Fluentd configmap replaced.
+WLSKO-0226=Pod {0} was evicted due to {1}; validating domain
+WLSKO-0227=Pod {0} was evicted due to {1} but the operator is configured to not restart it.
```
jshum2479, line 189: to not -> not to

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -405,7 +406,15 @@ private void processServerPodWatch(V1Pod pod, String watchType) {
         info.setServerPodBeingDeleted(serverName, Boolean.FALSE);
         // fall through
       case MODIFIED:
-        info.setServerPodFromEvent(serverName, pod);
+        boolean podAlreadyEvicted = info.setServerPodFromEvent(serverName, pod, PodHelper::isEvicted);
+        if (PodHelper.isEvicted(pod) && !podAlreadyEvicted) {
```
jshum2479, line 410: What's the difference between PodHelper.isEvicted()  vs podAlreadyEvicted ?   When will podAlreadyEvicted be true?

alai8: PodHelper.isEvicted(pod) is true if the new pod from the PodWatch event is evicted.
podAlreadyEvicted is true if the original pod from getSko(serverName).getPod() was already evicted.

common/src/main/resources/Operator.properties
alai8: fixed. thanks

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
alai8: renamed to podPreviouslyEvicted. Thanks

jshum2479 (approved):

operator/src/main/java/oracle/kubernetes/operator/TuningParameters.java
```diff
@@ -258,12 +258,13 @@ class PodTuning {
     public final int livenessProbeFailureThreshold;
     public final long introspectorJobActiveDeadlineSeconds;
     public final long maxReadyWaitTimeSeconds;
+    public final boolean restartEvictedPods;
```
rjeberhard, line 261: Is the default false? It seems like the default behavior should be to restart evicted pods. I'll remove this comment if I find that it's initialized to "true" below.

rjeberhard (approved):

PR #2980, opened 2022-04-21
OWLS-96978 fix intermittent failures related to Rolling:true condition remains after rolling restart completes

doxiao:
This is to fix an intermittent issue that is related to missing Rolling Start events. The operator keeps all managed servers that need to roll in a map, and roll them after all managed servers are evaluated. The servers in the map remain after a rolling restart completes, which causes the Rolling condition remains in the domain status because the server-to-roll map is not empty. The presence of Rolling condition prevents the Rolling Start event from being generated.

The changes in this PR removes a server from the map after the server is restarted.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10150/ (still running)


alai8 (commented):
The change LGTM. How confident that this fix would address the intermittent failures? 
Please also take a look at the code smell and see if you want to address it. Thanks.

doxiao:
> The change LGTM. How confident that this fix would address the intermittent failures? 
Once I commented out the code in the test to wait for the Rolling condition to be cleared, I was able to reproduce the issue within 5 runs. With the change (and the same test change), the test has not been failing in 10 runs.

doxiao:
> What do you mean by "The servers in the map remain after a rolling restart completes"? The map should be generated anew on each make-right IMO, and not preserved between them. This would seem to be a design error.

Yes, the map will be initialized with an empty map in the next domain presence retry cycle (make-right operation with explicit=true). But a server that is successfully restarted should be removed from the map right away. This has also prevent the domain from becoming Complete:true at the end of the current inflight make-right operation, but the domain may get into the Completed:true condition in the next explicit retry cycle, which by default is 2 minutes.

rjeberhard (approved):

alai8 (approved):

russgold:
I'm still not seeing why this is a problem, and why the server should be removed from the map once we start it. Can you explain why this problem is happening, and why it is intermittent, and how this change fixes it?

That is, what, exactly, is the sequence of events that causes the problem?

doxiao:
> I'm still not seeing why this is a problem, and why the server should be removed from the map once we start it. Can you explain why this problem is happening, and why it is intermittent, and how this change fixes it?
> 
> That is, what, exactly, is the sequence of events that causes the problem?

When a make-right operation detects that some of the managed server pods need to roll, it adds the server in a map. The map is initialized as an empty map at the beginning of an explicit make-right operation. DomainStatusUpdater checks if the map is empty (and other things) when it needs to determine if the roll is completed (and the domain processing is completed), and if the Rolling condition needs to be cleared.  Without the change, the servers that are rolled already stay in the map until next the recheck cycle, as a result, the domain does not reach Completed:true and Rolling:false until the next explicit recheck cycle. This change remove a server from the map right after it is successfully restarted, so that when a rolling restart completes, the map becomes empty right away, and, therefore all conditions are updated right away.

russgold:
> Maybe I did not fully understand the purpose and expected behavior of this map. Can you explain the expected
> functionality of this map in the original design?

The intent of the map is simply to identify the servers which need to be restarted, in order to select the appropriate subset of them for a replace-pod operation. The map is _not_ supposed to be used to determine whether we have reached rolling-completed, which is why it surprises me that you believe that this is a fix.

I think what you need here is to demonstrate that you understand the intermittent failure by creating a unit test with the right set of conditions so that it will fail. Then you make a production change that causes the test to pass, by allowing the operator to handle those conditions.

Intermittent failures are nasty because the mere fact that you've had runs with no failures doesn't prove that you've fixed the problem. The unit test you've written merely demonstrates that you've changed some behavior - it does not show that the changed behavior is significant to the problem.

doxiao:
> > Maybe I did not fully understand the purpose and expected behavior of this map. Can you explain the expected
> > functionality of this map in the original design?
> 
> The intent of the map is simply to identify the servers which need to be restarted, in order to select the appropriate subset of them for a replace-pod operation. The map is _not_ supposed to be used to determine whether we have reached rolling-completed, which is why it surprises me that you believe that this is a fix.

This is a surprise to me. We have code to check if the map is empty to determine if the Rolling condition needs to be removed. See allIntendedServersReady() method in DomainStatusUpdater.java and the its callers. 

 private boolean allIntendedServersReady() {
          return haveServerData()
              && allStartedServersAreComplete()
              && allNonStartedServersAreShutdown()
              && serversMarkedForRoll().isEmpty();
        }

  if (allIntendedServersReady()) {
            this.status.removeConditionsWithType(ROLLING);
          }


> 
> I think what you need here is to demonstrate that you understand the intermittent failure by creating a unit test with the right set of conditions so that it will fail. Then you make a production change that causes the test to pass, by allowing the operator to handle those conditions.

The intermittent failures in the integration test is due to the timing when the next recheck happens. 

> 
> Intermittent failures are nasty because the mere fact that you've had runs with no failures doesn't prove that you've fixed the problem. The unit test you've written merely demonstrates that you've changed some behavior - it does not show that the changed behavior is significant to the problem.

The unit test I added showed that the map is not empty after rolling restart completes before the change, and is empty after the change.

russgold:
> We have code to check if the map is empty to determine if the Rolling condition needs to be removed. 
> See allIntendedServersReady() method in DomainStatusUpdater.java and the its callers.

Ah I see. I had forgotten about that. That is intended to verify that we don't have other servers marked as pending a roll in a subsequent make-right. That means that removing servers from the map once they are restarted should not cause any problems. That doesn't explain, though, why it should accomplish anything.

> The intermittent failures in the integration test are due to the timing when the next recheck happens. 

The map should be initialized on each make-right, not just the explicit ones - but it's actually not, and that is what is wrong. The map is initialized in ManagedServerUpIteratorStep.initialServersToRoll - but that only happens if ManagedServersUpStep.ServersUpStepFactory.createNextStep() determines that we have something to roll in the current make-right. If not, we could have the map cached in the DomainPresenceInfo, which would be out-of-date. I would bet that is what is actually happening.

That means that we could demonstrate the problem in a unit test by running the make right with one server left to start and then running it a second time after marking that server ready. The old map would still be in the DPI, and rolling complete would not be generated. The correct fix then is to ensure that we always clear that map in the DPI in DomainProcessorImpl.PopulatePacketServerMapsStep. If you do that, you'll convert a failing unit test to a passing one, and thus demonstrate that the problem is actually fixed.

russgold:
> I think it does accomplish something very important; the domain status (Rolling condition and Completed condition) 
> is up-to-date right away, not until the next recheck.

Only if the new pod can reach the ready state within milliseconds of being started (the time between initiating the actual pod roll and the time we reach the status update step). That's not going to happen, in practice. And the bug turns out to be the lack of initialization of the map at the start of the make-right, so that's what we should fix.

russgold (changes requested):
This turns out to be the incorrect fix and lacks a unit test that reproduces the failure in the absence of the correct fix.

Please change this to demonstrate the problem and fix it as simply as possible.

doxiao:
> > I think it does accomplish something very important; the domain status (Rolling condition and Completed condition)
> > is up-to-date right away, not until the next recheck.
> 
> Only if the new pod can reach the ready state within milliseconds of being started (the time between initiating the actual pod roll and the time we reach the status update step). That's not going to happen, in practice. And the bug turns out to be the lack of initialization of the map at the start of the make-right, so that's what we should fix.

I don't think you understand the fix. The newly introduced Step is reached until the pod is ready.  I don't think the domain status update should wait until the next make-right operation starts.

russgold:
> I don't think you understand the fix. The newly introduced Step is reached only after the pod is ready.
> I don't think the domain status update should wait until the next make-right operation starts.

Interesting. I don't understand why we would wait for ready during a make-right. @rjeberhard isn't that a bit of an issue, to count on the make-right lasting so long, especially when we can interrupt it? I'd have expected us to start up the new pod, or even just replace it - and then have the make-right terminate. If we don't do that, why do we bother with asynchronous status updates?

I still think that the correct fix is to fix the actual error, not add code to compensate for it.

I'm going off-line now until Saturday night, and can pick up this discussion then.

robertpatrick:
@russgold @doxiao I would suggest you meet and discuss live rather than trading comment on this PR.  It is critical to get the intermittent failures in the tests addressed by whatever means necessary.  If there is a better way to accomplish that and we know exactly how to do that now, then so be it.  If not, then I would suggest we consider the interim solution being proposed followed by a JIRA to design and test the preferred way.

russgold:
We can meet on Monday, certainly.

The biggest problem with this fix is that it assumes that the make-right will always be allowed to finish. But a make-right can be interrupted by events, and if that were to happen before the new pod were to come up and be ready, the flag would never be cleared, resulting in the current issue.

The simplest fix, as I said, is always to clear the collection at the start of the make-right, and it is simple to write a unit test that demonstrates both the problem and the efficacy of the fix. As an improvement, you can have a more complex unit test that involves running two make-rights, and achieve your goal by clearing the flag at the start of the cycle step, rather than adding a new step that waits for the pod to be ready. If the make-right is interrupted, the next one will detect the end of the roll, and if it is not interrupted, the status step at the end will detect it.

doxiao:
> We can meet on Monday, certainly.
Let's sync up in the slack today.

> 
> The biggest problem with this fix is that it assumes that the make-right will always be allowed to finish. But a make-right can be interrupted by events, and if that were to happen before the new pod were to come up and be ready, the flag would never be cleared, resulting in the current issue.

The purpose of the fix is to make sure that when a server is **_successfully restarted_**, the server is removed from the servers-marked-to-roll map, and when all servers are successfully restarted, the map becomes empty and the Rolling:true condition is removed. This is what the test cases in ItPodsRestart covers, and this is the use case where the intermittent failures in OWLS-96978 happen. 

It is reasonable and expected that a server remains in the map until the next make-right starts if the server’s restart is somehow interrupted, and the change in this PR does not (and does not need to) alter the behavior in these cases. If you think the operator has problems in the use cases that are not tested in the ItPodsRestart class, we can investigate those use cases in a separate JIRA.

russgold (approved):

robertpatrick (changes requested):
Please resolve the merge conflict and then we can merge, provided that the build passes the tests.

robertpatrick (approved):

PR #2990, opened 2022-04-25
Fix NPE when no domainConfig is defined in unit test

jshum2479:
The NPE is preventing new code merges https://github.com/oracle/weblogic-kubernetes-operator/pull/2948

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
@@ -592,7 +592,7 @@ class DomainIntrospectorJobTest extends DomainTestUtils {
   }
 
   @Test
-  void whenJobCreatedWithoutFluentdTerminatedDuringIntropsection() {
+  void whenJobCreatedWithFluentdAndSuccessIntrospection_JobIsTerminatedAndJobTerminatedMarkerInserted() {
     String jobName = UID + "-introspector";
     DomainConfiguratorFactory.forDomain(domain)
         .withFluentdConfiguration(true, "elastic-cred", null);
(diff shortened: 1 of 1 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
```diff
@@ -601,12 +601,13 @@ void whenJobCreatedWithoutFluentdTerminatedDuringIntropsection() {
...
     testSupport.defineResources(jobPod);
     defineNormalFluentdContainerInIntrospection();
+    testSupport.addToPacket(DOMAIN_TOPOLOGY, createDomainConfig("cluster-1"));
 
     Packet packet = testSupport.runSteps(JobHelper.createIntrospectionStartStep(terminalStep));
     logRecords.clear();
     assertThat(packet.get(JOB_POD_INTROSPECT_CONTAINER_TERMINATED),
             equalTo(JOB_POD_INTROSPECT_CONTAINER_TERMINATED_MARKER));
-    assertThat(terminalStep.wasRun(), is(false));
+    assertThat(terminalStep.wasRun(), is(true));
```
russgold, line 610: Can you clarify the purpose of the test in its name? I think, after speaking with you, that the intent is that JOB_POD_INTROSPECT_CONTAINER_TERMINATED_MARKER is added to the packet, and has nothing to do with the terminalStep. If so, you can remove that last assertion, and change the name to something like 'whenJobCreatedWithoutFluentd_recordTerminatedMarkerInPacket` (if that is what is going on here. I am still not sure of the purpose of this test).

russgold (changes requested):

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
jshum2479: I changed the name to `whenJobCreatedWithFluentdAndSuccessIntropsection_ensureJobIsTerminatedAndJobTerminatedMarkerInserted` but leave the last assert to make sure the job will run to it's end

russgold: The purpose of the terminal step is to verify that the make-right runs to the end. Is that the intent?

russgold (approved):

doxiao (approved):
LGTM.

PR #3000, opened 2022-04-28
Add and initialize severity field

russgold:
Adds a severity field to the DomainCondition, uses it compute condition ordering, and initializes the value based on the reason and message fields.

Addresses OWLS-97603


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainFailureSeverity.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.weblogic.domain.model;
+
+public enum DomainFailureSeverity {
+  FATAL("Fatal"), SEVERE("Severe"), WARNING("Warning");
```
rjeberhard, line 7: This will probably fail to deserialize from JSON properly without the `@SerializedName` annotation on each enum value similarly to what we have in DomainConditionType.

operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
```diff
@@ -947,7 +944,7 @@ void whenNoDynamicClusters_doNotAddReplicasTooHighFailure() {
   }
 
   @Test
-  void whenReplicaCountDoesNotExceedMaxReplicasForDynamicCluster_removeOldReplicasTooHighFailure() {
+  void whenReplicaCountDoesNotExceedMaxReplicasForDynamicCluster_removeOldReplicasTooHighFailure() {   // todo here
```
rjeberhard, line 947: What is the todo?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainFailureSeverity.java
russgold: That's possible; it could explain the failures. I will fix that, thanks.

rjeberhard (approved):

doxiao (approved):
LGTM.

PR #3028, opened 2022-05-03
OWLS-98864 - Set the default resource requests/limits for the aux image init containers

ankedia:
This change sets the default resource requests/limits for the aux image init containers based on values provided in `spec.adminServer.serverPod.resources`.


operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
```diff
@@ -160,6 +162,16 @@ protected List<V1EnvVar> createEnv(AuxiliaryImage auxiliaryImage, String name) {
     return vars;
   }
 
+  protected V1ResourceRequirements createResources() {
+    V1ResourceRequirements resources = getServerSpec().getResources();
+    V1ResourceRequirements resourceRequirements = null;
+    if (!resources.getLimits().isEmpty() || !resources.getRequests().isEmpty()) {
```
rjeberhard, line 168: Why do they both have to be non-empty? Isn't it a valid use case to have one but not the other?

ankedia: The check is to see if any one of them is non-empty (i.e. either limits is non-empty or requests is non-empty). Are you suggesting to only set the limits or requests value based on which is one is non-empty?

alai8:
Do you think we should document that we are now also setting the resources for aux image init containers based on values in `spec.adminServer.serverPod.resources`, in [Pod memory and CPU resources page](https://oracle.github.io/weblogic-kubernetes-operator/faq/resource-settings/#setting-resource-requests-and-limits-in-a-domain-resource)?

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
alai8, line 168: could getLImits() or getRequests() return null?

rjeberhard: Yeah, I think that I misread it, so just want to make sure. I think that it's valid to have a resources section that only configures limits or that only configures requests. Anthony's question is probably the right one... Would this fail if one of the values was configured and then where the other was null.

ankedia: ok. I have pushed a change to separate the two checks in [b8ae10d](https://github.com/oracle/weblogic-kubernetes-operator/pull/3028/commits/b8ae10deb3a77aaa0144000f69b3dfdc73ebae5e). My understanding is that setting the limits/requests value to `null` is same as not setting it.

ankedia: Changed in [b8ae10d](https://github.com/oracle/weblogic-kubernetes-operator/pull/3028/commits/b8ae10deb3a77aaa0144000f69b3dfdc73ebae5e).

ankedia:
> Do you think we should document that we are now also setting the resources for aux image init containers based on values in `spec.adminServer.serverPod.resources`, in [Pod memory and CPU resources page](https://oracle.github.io/weblogic-kubernetes-operator/faq/resource-settings/#setting-resource-requests-and-limits-in-a-domain-resource)?
[ANIL] We document the fact that the introspector job pod uses the same CPU and memory settings as the domain’s WebLogic Administration Server pod. The aux init container is part of the introspector job pod and not sure if we need to mention separately for each container. Similarly for WebLogic server pods, the resource requests are set at the pod level and not sure if it needs to be mentioned separately for each container.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
```diff
@@ -160,6 +162,22 @@ protected List<V1EnvVar> createEnv(AuxiliaryImage auxiliaryImage, String name) {
     return vars;
   }
 
+  protected V1ResourceRequirements createResources() {
+    V1ResourceRequirements resources = getServerSpec().getResources();
+    V1ResourceRequirements resourceRequirements = null;
+    if (!resources.getLimits().isEmpty()) {
```
alai8, line 168: Can V1ResourceReqirements::getLimits() return null and cause an NPE here

rjeberhard:
> Do you think we should document that ...

Yes, that's a good idea. I think just a sentence or updating the existing sentence is sufficient.

operator/src/main/java/oracle/kubernetes/operator/helpers/BasePodStepContext.java
ankedia: I don't this so. We, by default, initialize the resource requests and limits with empty hash maps. 
```
  private final V1ResourceRequirements resources =
      new V1ResourceRequirements().limits(new HashMap<>()).requests(new HashMap<>());
```

rjeberhard (approved):

alai8 (approved):

ankedia:
> > Do you think we should document that ...
> 
> Yes, that's a good idea. I think just a sentence or updating the existing sentence is sufficient.
[ANIL] Added a sentence in [f16c942](https://github.com/oracle/weblogic-kubernetes-operator/pull/3028/commits/f16c942918d4ad4882f844df2439881c78f1c693).

PR #3033, opened 2022-05-03
Update domain classes for new severe retry functionality

russgold:
As part of change OWLS-97642 this adds new fields to the domain and domain status and uses them to compute whether to retry. This change also includes some refactoring in DomainStatusUpdater to prepare for the transition from the old retry behavior to the new.

Some cleanup is included: three test configuration classes turned out to be in the main source directory and have been moved, and the tests included in DomainTestBase were moved to a new test to allow the test base to be used for new tests. It would be useful in the future to organize the domain tests; they are getting a bit unwieldy.


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -88,6 +92,12 @@ public class DomainStatus {
           + "or clusters are available.")
   private OffsetDateTime startTime = SystemClock.now();
 
+  @Description("RFC 3339 date and time at which a currently failing domain started automatic retries.")
```
tbarnes-us, line 95: A quibble: I am no expert on time standards by any stretch of the imagination but replacing 'RFC 3339' with 'ISO 8601' may be a little more appropriate. The javadoc for the OffsetDateTime method that's used here directly states that it uses the ISO 8601 standard, and, according to [this article](https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a), the main difference is that it requires the 'T' where-as RFC 3339 also allows using a space instead. We use the same 'T' format for the operator logs and introspector scripts.

russgold: Hmm... RFC 3339 is what we've been saying with the other time values I found, including DomainStatus.startTime and ServerHealth.activationTime, and [this article](https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats) disputes the article you cited.

@rjeberhard your call on this.

rjeberhard: Yes, Kubernetes time-based data types are RFC 3339: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata

rjeberhard: Here is how the creationTimestamp in metadata is described:
"CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC. Populated by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
Returns:
creationTimestamp"

tbarnes-us: I was looking at the javadoc for the API, which is pretty clear  `A date-time with an offset from UTC/Greenwich in the ISO-8601 calendar system, such as 2007-12-03T10:15:30+01:00.`.  No worries - the two standards are only slightly different - RFC being a superset.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainCondition.java
```diff
@@ -48,7 +48,7 @@ public class DomainCondition implements Comparable<DomainCondition>, PatchableCo
   @Description("Unique, one-word, CamelCase reason for the condition's last transition.")
   @SerializedName("reason")
   @Expose
-  private String reason;
+  private DomainFailureReason reason;
```
rjeberhard, line 51: In Slack, you correctly diagnosed that v8 Domains might have reason values that were not valid DomainFailureReason enums and therefore you would need to upgrade those reason values as part of the v8 to v9 conversion code in SchemaConversionUtils. Are you planning on doing that as part of another PR?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
rjeberhard: Yes, I think that you're right, Tom. The implementation is using ISO-8601 behavior because it aligns with RFC 3339 when UTC is enforced.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainCondition.java
russgold: Yes. This is not the full change needed for the functionality; I will be creating a second PR once this is merged. If preferred, I can put that change into this PR; I was concerned, after  a discussion with Derek, about making this one larger (I hope he is going to explain at today's status meeting).

doxiao (commented):
LGMT overall. Just had 2 comments.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainCondition.java
rjeberhard: That's fine. I'm okay with follow-on work coming in additional PR's. I just wanted to make sure that this requirement wasn't forgotten.

operator/src/test/java/oracle/kubernetes/weblogic/domain/model/DomainRetryTest.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Test
+  void whenDomainHasSevereFailure_retryIsNeeded() {
+    addFailureCondition(INTERNAL);
+
+    assertThat(domain.shouldRetry(), is(true));
+  }
+
+  private void addFailureCondition(DomainFailureReason internal) {
```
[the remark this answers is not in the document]
russgold: Yes, oops. That was left over from the original form of the test. I will fix it.

doxiao (approved):

rjeberhard (approved):

PR #3036, opened 2022-05-05
re-org doc content

rosemarymarano:
Re-organized the content in the Model in Image, Runtime Updates doc as per [OWLS-95907](https://jira.oraclecorp.com/jira/browse/OWLS-95907): Operator User Guide Feedback and Updates for 4.0:
⦁	The information provided is really hard to consume due to what feels like lots details about supported, unsupported, online, and offline updates.  What I am not getting is a good sense of what happens if my update includes something I didn’t expect (e.g., an offline-only update or an unsupported update when I thought I was going to apply an online update).  
⦁	Maybe reordering the page to discuss how to do updates and what happens if you mistakenly include an update that doesn’t fit the type of update you think you are doing first (since this is the most important content) and then have reference sections that categorize what updates fall into which categories?

### DIFF
--- a/documentation/staging/content/userguide/managing-domains/model-in-image/runtime-updates.md
+++ b/documentation/staging/content/userguide/managing-domains/model-in-image/runtime-updates.md
@@ -41,7 +41,7 @@ without first shutting down the domain:
 - _Offline updates_: [Offline updates](#offline-updates) are propagated to WebLogic pods by updating your model
   and then initiating a domain roll, which generates a new domain configuration,
   restarts the domain's WebLogic Server Administration Server with the updated configuration,
-  and then restarts the other pods in the cluster.
+  and then restarts other running servers.
 
  - _Online updates_: If model changes are configured to fully dynamic configuration MBean attributes,
    then you can optionally propagate changes to WebLogic pods without a roll using an [online update](#online-updates).
--- a/documentation/staging/content/userguide/managing-domains/model-in-image/runtime-updates.md
+++ b/documentation/staging/content/userguide/managing-domains/model-in-image/runtime-updates.md
@@ -9,20 +9,18 @@ description = "Updating a running Model in Image domain's images and model files
 ### Contents
 
  - [Overview](#overview)
- - [Supported updates](#supported-updates)
- - [Unsupported updates](#unsupported-updates)
  - [Updating an existing model](#updating-an-existing-model)
  - [Offline updates](#offline-updates)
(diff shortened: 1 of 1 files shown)


documentation/staging/content/userguide/managing-domains/model-in-image/runtime-updates.md
```diff
@@ -40,32 +38,29 @@ This is because a Model in Image domain home is regenerated from the model on ev
 There are two approaches for propagating model updates to a running Model in Image domain
 without first shutting down the domain:
 
+- _Offline updates_: [Offline updates](#offline-updates) are propagated to WebLogic pods by updating your model
+  and then initiating a domain roll, which generates a new domain configuration,
+  restarts the domain's WebLogic Server Administration Server with the updated configuration,
+  and then restarts the other pods in the cluster.
```
rjeberhard, line 44: I see that this was the existing text, but saying, "and then restarts the other pods in the cluster." is confusing as the Administration Server is not in a cluster and this also implies that there is exactly one cluster and that all other servers are clustered servers. This is true for many of our samples, but isn't universally true. Can this say something like, "... and then restarts other running servers."? I don't think that we need to even mention here that clustered servers will be restarted in a roll as to preserve cluster availability.

rosemarymarano: @rjeberhard Not sure exactly which sentence should be removed or modified according to this feedback, "I don't think that we need to even mention here that clustered servers will be restarted in a roll as to preserve cluster availability." 

Do you want this sentence changed to, "[Offline updates](#offline-updates) are propagated to WebLogic pods by updating your model and then restarts other running servers."

rjeberhard: You have the right sentence, but I was commenting on the content on line 44 that presently reads, "and then restarts the other pods in the cluster." I was pointing out that the Administration Server isn't in a cluster and so it doesn't make sense to say the "other pods in the cluster" unless the sentence is implying that there is always exactly one cluster and that all other servers are members of that one cluster -- that is true for many of our samples, but isn't generally true.

rosemarymarano: Yes, that part I got. Thank you.

rjeberhard (approved):

robertpatrick (approved):

PR #3037, opened 2022-05-05
Add domain condition reasons upgrade

russgold:
The primary intent of this change is to add support to the domain upgrade for the new behavior of condition reasons: since only values from DomainFailureReason are supported, the field was changed from a string to an enum. 

I have also done a bit of code cleanup, especially in the unit tests. This addresses warnings in the build and improves the readability of the domain conversion unit test.

I note some concerns that I did not address:
1.  The class `SchemaConversionUtils` is misleadingly named. It is not, as the name suggests, a set of utilities for doing schema conversion. It is, rather, the set of rules for upgrading an old domain. It should really be renamed and moved to a more appropriate package.
2. There are still some behaviors of the conversion that are not unit-tested. I added a test for removing the `Progressing` condition, but it is not the only gap.
3. I do not see a unit test for loading a domain resource into the `ConversionRequest` class. I modified the base type of the `objects` field because of a build warning. That should be safe because of the way Java handles generics, but the presence of a unit test would be reassuring, at the least.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionRequest.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionRequest.java
@@ -14,9 +14,12 @@ public class ConversionRequest {
 
   @Expose
   private String uid;
+
   @Expose
   private String desiredAPIVersion;
+
   @Expose
+  // The domains to be converted. Note that the field name 'objects' is required by the conversion API.
   private List<Map<String,Object>> objects = new ArrayList<>();
 
   public String getUid() {
@@ -31,16 +34,11 @@ public class ConversionRequest {
     return desiredAPIVersion;
   }
 
-  public void setDesiredAPIVersion(String desiredAPIVersion) {
-    this.desiredAPIVersion = desiredAPIVersion;
-  }
-
-  public List<Map<String,Object>> getObjects() {
+  /**
+   * Returns a list domains to be converted by the webhook.
+   */
+  public List<Map<String,Object>> getDomains() {
     return objects;
-  }
-
-  public void setObjects(List<Map<String,Object>> objects) {
-    this.objects = objects;
   }
 
   @Override
--- a/operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionRequest.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionRequest.java
@@ -5,6 +5,7 @@ package oracle.kubernetes.operator.rest.model;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import com.google.gson.annotations.Expose;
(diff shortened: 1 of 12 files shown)


ankedia (approved):

rjeberhard (approved):

ankedia:
> > I do not see a unit test for loading a domain resource into the ConversionRequest class.
> 
> Russ, just FYI... The K8s API server calls our webhook request, so the creation of the ConversionRequest with a domain to check is done by Kubernetes.

Yes, I missed this. The conversion request is sent by the k8s API server as part of the `ConversionReview` when conversion is needed.

ankedia:
> > > I do not see a unit test for loading a domain resource into the ConversionRequest class.
> > 
> > 
> > Russ, just FYI... The K8s API server calls our webhook request, so the creation of the ConversionRequest with a domain to check is done by Kubernetes.
> 
> Yes, I missed this. The conversion request is sent by the k8s API server as part of the `ConversionReview` when conversion is needed.
Also, there's existing unit test `whenConversionWebhookRequestSent_hasExpectedResponse` that reads the conversion review and sends to webhook REST service.

operator/src/main/java/oracle/kubernetes/operator/rest/model/ConversionRequest.java
```diff
@@ -34,11 +35,11 @@ public void setDesiredAPIVersion(String desiredAPIVersion) {
     this.desiredAPIVersion = desiredAPIVersion;
   }
 
-  public List<Object> getObjects() {
+  public List<Map<String,Object>> getObjects() {
```
ankedia, line 38: I think this method name is bit misleading since it's returning a list of map rather than list of objects.

rjeberhard: Isn't this returning the "objects" field from the conversion request? Each of those objects (i.e. Domains) is represented by a Map.

rjeberhard: This was confusing to me too because those maps can have entries that are not maps, but I believe that the top-level instance is always a Map. That would be consistent with the code we had previously where we always cast.

ankedia: ok, I was wondering if the method name can be renamed to something like `getDomains` or `getDomainMaps` to clarify it's intent ?

russgold: Good point. As I understand it, the value was always a list of maps (the domains to convert, rendered as maps). I changed the type, but didn't change the name.

russgold: And I see that the name "objects" is in the sample conversion request. So the next question is, how is the conversion request generated? Is "objects" specified by Kubernetes, or is it in the webhook definition?

rjeberhard: Kubernetes makes this request and defines the REST operation. The operator has to implement the operation that K8s has defined; we only configure the location of the endpoint.

ankedia: I believe we can leave the field name as `objects` but change the method names for getter and setter.

ankedia (approved):

PR #3039, opened 2022-05-06
Provide operatorOnly helm config value to only install the operator and skip the webhook installation

ankedia:
This change provides `operatorOnly` helm configuration value to only install the operator and skip the webhook installation. This might be needed in multiple operators use-case where cluster-level permissions can't be granted or all operators are deployed concurrently. This also includes a minor change to ignore the `preserveWebhook`  helm config value when `webhookOnly` is set to true.

### DIFF
--- a/documentation/staging/content/userguide/managing-operators/using-helm.md
+++ b/documentation/staging/content/userguide/managing-operators/using-helm.md
@@ -689,8 +689,14 @@ If set to `true`, the `helm install` will install _only_ the conversion webhook 
 
 Defaults to `false`.
 
+##### `operatorOnly`
+Specifies whether only the operator should be installed during the `helm install` and that the conversion webhook installation should be skipped. By default, the `helm install` command installs both the operator and the conversion webhook.
+If set to `true`, the `helm install` will install _only_ the operator (and not the conversion webhook).
+
+Defaults to `false`.
+
 ##### `preserveWebhook`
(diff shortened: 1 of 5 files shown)


documentation/staging/content/userguide/managing-operators/using-helm.md
```diff
@@ -689,8 +689,14 @@ If set to `true`, the `helm install` will install _only_ the conversion webhook
 
 Defaults to `false`.
 
+##### `operatorOnly`
```
tbarnes-us, line 692: Two comments:

- The table of contents does not have this, and it looks like like preserveWebhook is missing too - you might want to check all of them. To generate a sample TOC, run `./getlinks.sh userguide/managing-operators/using-helm.md` from within the staging directory.

- Also, IMO, this section would fit better if moved just after "Creating the operator pod" at the same level (or even under it), so that all operations related to basic install/uninstall are near each-other.

ankedia: I have made the suggested changes in [6a08484](https://github.com/oracle/weblogic-kubernetes-operator/pull/3039/commits/6a0848471b62626db8d4eaa46871fb54eb630604).

tbarnes-us (approved):

jshum2479 (approved):

rjeberhard (approved):

PR #3041, opened 2022-05-06
add WDT to sample script info

rosemarymarano:
Clarify in FMW domain home in PV sample (doc) that both WLST or WDT can be used to create the domain as per [OWLS-99137](https://jira.oraclecorp.com/jira/browse/OWLS-99137).


mriccell (approved):

bhavaniravichandran (approved):
LGTM

PR #3043, opened 2022-05-06
Use time rather than number of retries to determine timeouts

russgold:
Next part of the changes to retries for severe errors.

This includes using time to determine both the retry limit and the increases in the instrospector job timeout.

Currently, two values (the maximum number of increases and size of each increase) are hard-coded. The next change will make them tuning parameters. It will also remove the rest of the retry count functions, and add removing those fields to the domain upgrade.

### DIFF
--- a/common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
+++ b/common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
@@ -145,6 +145,7 @@ public class MessageKeys {
   public static final String LOG_WAITING_COUNT = "WLSKO-0193";
   public static final String INTERNAL_IDENTITY_INITIALIZATION_FAILED = "WLSKO-0194";
   public static final String DOMAIN_FATAL_ERROR = "WLSKO-0195";
+  public static final String INTROSPECTOR_FATAL_ERROR = "WLSKO-0196";
   public static final String INTROSPECTOR_MAX_ERRORS_EXCEEDED = "WLSKO-0196";
   public static final String NON_FATAL_INTROSPECTOR_ERROR = "WLSKO-0197";
   public static final String DUMP_BREADCRUMBS = "WLSKO-0198";
@@ -193,7 +194,6 @@ public class MessageKeys {
   public static final String NO_MANAGED_SERVER_IN_DOMAIN = "WLSDO-0012";
   public static final String CANNOT_EXPOSE_DEFAULT_CHANNEL_ISTIO = "WLSDO-0013";
   public static final String INTROSPECT_JOB_FAILED = "WLSDO-0014";
-  public static final String INTROSPECT_JOB_FAILED_RETRY_COUNT = "WLSDO-0015";
   public static final String ILLEGAL_INTROSPECTOR_JOB_NAME_LENGTH = "WLSDO-0016";
   public static final String ILLEGAL_CLUSTER_SERVICE_NAME_LENGTH = "WLSDO-0017";
   public static final String ILLEGAL_SERVER_SERVICE_NAME_LENGTH = "WLSDO-0018";
--- a/common/src/main/resources/Operator.properties
+++ b/common/src/main/resources/Operator.properties
@@ -145,12 +145,11 @@ WLSKO-0191=Rolling restart of WebLogic domain {0} completed
 WLSKO-0192=Executing make right domain operation, recheck count for server {0} is {1}.
 WLSKO-0193=Waiting for server {0} to start, recheck count is {1}.
 WLSKO-0194=Internal identity initialization step failed with exception {0}.
-WLSKO-0195=The operator encountered a fatal error: ''{0}'' and will not retry automatically. \
+WLSKO-0195=The operator failed after retrying for {0,number,#} minutes. \
+  This time limit may be specified in spec.failureRetryLimitMinutes. \
+  Please resolve the error and then update 'domain.spec.introspectVersion' to force another retry.
+WLSKO-0196=Introspection encountered a fatal error: ''{0}'' and will not retry automatically. \
   Please resolve the error and then update 'domain.spec.introspectVersion' to force a retry.
-WLSKO-0196=Introspection failed: ''{0}'' after {1} attempt(s) and will not retry. \
-  Please correct the error and force the operator to try again by updating 'domain.spec.introspectVersion'. \
-  The number of retries is controlled by the operator tuning parameter 'domainPresenceFailureRetryMaxCount', \
-  which may be specified in the weblogic-operator-cm configmap.
 WLSKO-0197=Introspection failed: ''{0}'' and will retry. Attempt {1} of {2}.
 WLSKO-0198={0} Fiber {1}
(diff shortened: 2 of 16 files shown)


common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
```diff
@@ -145,6 +145,7 @@ public class MessageKeys {
   public static final String LOG_WAITING_COUNT = "WLSKO-0193";
   public static final String INTERNAL_IDENTITY_INITIALIZATION_FAILED = "WLSKO-0194";
   public static final String DOMAIN_FATAL_ERROR = "WLSKO-0195";
+  public static final String INTROSPECTOR_FATAL_ERROR = "WLSKO-0196";
```
rjeberhard, line 148: Something is off here because you have "WLSKO-0196" duplicated and Operator.properties is using "WLSKO-0195" for the new message.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -71,10 +73,10 @@ public class DomainStatus {
           + "You cannot configure a limit for other types of failures, such as a Domain resource reference "
           + "to an unknown secret name; in which case, the retries are unlimited.")
   @Range(minimum = 0)
-  private Integer introspectJobFailureCount = 0;
+  private Integer introspectJobFailureCount = 0;  //todo remove this field
```
rjeberhard, line 76: I'm okay that this is a todo for a later PR, but just making sure that this is your intention.

common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
russgold: The new retry strategy requires different messages, and I was trying to reuse the numbers. Maybe that's an error.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
russgold: Yes, as noted in the description, I will be removing those fields in a subsequent PR.

common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
rjeberhard: The String value, "WLSKO-0196", is the lookup key in to Operator.properties. If these two use cases need to generate the same message then this could work, but in that case I think it would clearer to use the same constant in MessageKeys.java.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
russgold: Since this is the second PR on which you've made such a comment, should I assume that putting TODOs in the code for a future PR is not a practice you endorse? I should keep that information in a separate place?

rjeberhard: It's not that -- todo's are okay... I needed more in the comment following the todo so that I don't have to look elsewhere for a statement of what needs to be done (at least simply) and if this is near-term work or for someday.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainStatus.java
```diff
@@ -587,10 +589,41 @@ public OffsetDateTime getInitialFailureTime() {
...
+  /**
+   * Return the number of times that failure timeouts should be increased as a result of failures.
+   * Will be approximately the number of times a failure has been reported, assuming one failure per retry.
+   * @param retrySeconds the number of seconds between retries.
+   */
+  public int getNumDeadlineIncreases(long retrySeconds) {
+    if (initialFailureTime == null || lastFailureTime == null) {
+      return 0;
+    } else {
+      return (int) (1 + divideRoundingUp(getSecondsFromInitialToLastFailure(), retrySeconds));
```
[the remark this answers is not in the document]
doxiao: The code seems to calculate how many times the failure has been retried based on the initialFailureTime, lastFailureTime and retry interval. But the retries do not happen back-to-back; each retry attempt takes sometime to fail especially here we are handling the situation where the job/pod timed out because the active deadline has exceeded.  The current calculation will yield a higher number than the actual number of times the retry happened.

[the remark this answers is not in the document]
russgold: I think I see what you mean; since the intent is to come up with a reasonable approximation of the number of retries for the purpose of computing the new active deadline, we could presumably come up with a better algorithm that allows for the possibility that the failure was caused by an introspector timeout rather than one of the other failure modes. Can you think of a better algorithm that is not too complex? Ryan said that he wasn't worried about getting it precise, given that we are no longer tracking the actual number of failures.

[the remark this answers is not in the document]
rjeberhard: As Russ said here, I agree that the focus is on simplicity and transparency. I don't mind thinking about what the algorithm could be; however, if the operator is consistently communicating what the failure is and when the next retry will occur then I think this will be perfect for most customers -- put another way, very few if any customers will care that the next retry time is some exact value based on the number of retries if the time is communicated and is approximately what would be given by an algorithm that did have the exact number of retires.

doxiao (approved):

common/src/main/resources/Operator.properties
```diff
@@ -145,12 +145,11 @@ WLSKO-0191=Rolling restart of WebLogic domain {0} completed
 WLSKO-0192=Executing make right domain operation, recheck count for server {0} is {1}.
 WLSKO-0193=Waiting for server {0} to start, recheck count is {1}.
 WLSKO-0194=Internal identity initialization step failed with exception {0}.
-WLSKO-0195=The operator encountered a fatal error: ''{0}'' and will not retry automatically. \
+WLSKO-0195=The operator failed after retrying for {0,number,#} minutes. \
+  This time limit may be specified in spec.failureRetryLimitMinutes. \
+  Please resolve the error and then update 'domain.spec.introspectVersion' to force another retry.
+WLSKO-0196=Introspection encountered a fatal error: ''{0}'' and will not retry automatically. \
   Please resolve the error and then update 'domain.spec.introspectVersion' to force a retry.
-WLSKO-0196=Introspection failed: ''{0}'' after {1} attempt(s) and will not retry. \
```
rjeberhard, line 150: Okay, one more question... Now you don't have a duplicated message ID (thanks), but you are deleting the message for WLSK0-0196, but the constant for this message still exists in MessageKeys.java, which I assume means that it's still referenced somewhere.

The opposite is true for WLSK0-0191 -- here the message still exists in this file, but you've deleted the constant that references it.

rjeberhard: Either solution is fine, but I would like to keep Operator.properties and MessageKeys.java in sync.

rjeberhard (approved):

PR #3044, opened 2022-05-06
OWLS-96080 - Intermittent: [wko-nightly] ItMiiDomain.testCreateMiiSecondDomainDiffNSSameImage failure

alai8:
Operator detects whether an introspector job failed due to running longer than `activeDeadlineSeconds` by having `DeadlineExceeded` as the reason in the failure status condition. Operator then retry the job with a larger `activeDeadlineSeconds` to provide more time in the introspector job retry.

Sometimes a job that failed due to `activeDeadlineSeconds` exceeded would instead contain `BackoffLimitExceeded` as the reason, and operator would treat it as a failed job and retry without increasing the`activeDeadlineSeconds`.

With this change, we would also consider an instrospector job that has failed with `BackoffLimitExceeded` as same as `DeadlineExceeded` if the job has been running for at least `activeDeadlineSeconds`, ie,

  `status.condition.lastTransitionTime` - `status.startTime` >= `spec.activeDeadlineSeconds`

jenkins https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10477/


rjeberhard (approved):

rjeberhard:
@alai8, adding @russgold as a reviewer. He's been working on a general change to retry failures after a retry delay and to retry for a maximum time rather than for a maximum number of attempts. I believe that your changes and his pending work will be compatible, but I'd like to make sure and for the two of you to negotiate how to merge if not.

ankedia:
I was envisioning the fix would be to check the introspector job pod status and see if the job pod failed with `DeadlineExceeded` status. In that case, we'll treat that failure same as the job failure with `DeadlineExceeded` status and retry the job immediately with increased `activeDeadlineSeconds` value instead of waiting for the next make-right cycle.

The job can fail with `BackoffLimitExceeded` for various reasons for e.g. missing credentials or invalid domain dir. I'm not sure if increasing the `activeDeadlineSeconds` value will help in such cases and if the job will be retried immediately after the job pod timeout.

operator/src/test/java/oracle/kubernetes/operator/JobWatcherTest.java
```diff
@@ -319,9 +377,9 @@ void whenJobWithFluentdInProcessOnFirstRead_performNextStep() {
     assertThat(terminalStep.wasRun(), is(true));
   }
 
-  @Test
+  @Test //xyz
```
russgold, line 380: Remove unneeded comment

operator/src/main/java/oracle/kubernetes/operator/JobWatcher.java
```diff
@@ -158,8 +160,9 @@ private static boolean isConditionFailed(V1Job job) {
     return getJobConditions(job).stream().anyMatch(JobWatcher::isJobConditionFailed);
   }
 
-  private static List<V1JobCondition> getJobConditions(@Nonnull V1Job job) {
-    return Optional.ofNullable(job.getStatus()).map(V1JobStatus::getConditions).orElse(Collections.emptyList());
+  private static List<V1JobCondition> getJobConditions(V1Job job) {
+    return Optional.ofNullable(job).map(V1Job::getStatus).map(V1JobStatus::getConditions)
```
russgold, line 164: Since it no longer fits one one line, please put each step on its own

russgold (approved):
Left a couple of minor comments; otherwise looks good and should not conflict with my current work

PR #3047, opened 2022-05-09
Owls 99119 - Fix for Pod roll issue when WKO is upgraded from v3.3.8 to 4.0

ankedia:
This changes prevents the Pod from rolling when the operator is upgraded from 3.3.8 to 4.0. The Pod roll issue was caused because of the change to set the default resource requests/limits for the aux image init containers. This change ignores the resource requests/limits for the compatibility auxiliary image init containers when calculating the hash value.


russgold (changes requested):
Please add a unit test that fails without your changes and passes with them, proving both that you have identified the problem, and that you have fixed it. Since you know what is wrong, this is should be straightforward.

ankedia:
> Please add a unit test that fails without your changes and passes with them, proving both that you have identified the problem, and that you have fixed it. Since you know what is wrong, this is should be straightforward.

Added in [785f246](https://github.com/oracle/weblogic-kubernetes-operator/pull/3047/commits/785f246972cbd8f16eb73e809d4dc92b2105dccb).

rjeberhard (approved):

russgold (approved):

alai8 (approved):

PR #3048, opened 2022-05-09
Update Helm chart defaults related to namespace management

rjeberhard:
This pull request updates defaults for several values in the operator's Helm chart:

* `enableClusterRoleBinding`: The default is now `true`, was `false`. If `domainNamespaceSelectionStrategy` is "Dedicated" then this value is ignored.
* `domainNamespaceSelectionStrategy`: The default is now "LabelSelector", was "List".
*  `domainNamespaceLabelSelector`: The default is now "weblogic-operator=enabled" and it was previously required.

The documentation, including quick start and namespace management sections, is updated for these default changes.

The goal is to provide a simpler experience for most customers and we are using the major version boundary as the opportune time to make the changes to these defaults. Now, the operator will behave more like the Istio operator where the customer installs the operator and then simply needs to label namespaces to have those namespaces "opt-in".

Customers who are upgrading from 3.x will not see unexpected changes to operator behavior if they use the "--reuse-values" option on the `helm upgrade` as these customers will have the previous defaults as part of their existing value set.


rosemarymarano (commented):
Minor edits, otherwise ok.

PR #3049, opened 2022-05-09
Owls97950 validating webhook part1

doxiao:
This is the first part of the implementation of validating webhook, which covers the following:

- Rename the classes that are common to both conversion webhook and validating webhook; removing "Conversion" from the names.
- Create a validating webhook configuration at the startup of the existing webhook deployment. The configuration associates the webhook with the existing webhook service via clientConfig. There is a single instance of this configuration in a K8S cluster and it is replaced at the webhook startup time if needed.
- Add helper methods in CallBuilder for creating, reading, replacing and deleting validating webhook configuration.
- Add operator roles and clusterroles for the validating webhook configuration in operator helm templates.
- Add validating webhook's REST resource class, and register it with the existing webhook REST server.
- Add REST request/response related classes: AdmissionReview, AdmissionRequest and AdmissionResponse.
- Have a dummy validate() method that accepts all requests right now; the full implementation of the validation logic is planned to be in the next PR.
- Add unit tests for all the code changes listed above.
- Add a utility method to list validating webhook configuration objects in the integration test logs.

Main branch: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10624/
owls97950-validating-webhook-part1: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10625/


russgold:
That's a really low coverage number. Can you bring it up to at least 80%?

rjeberhard:
> That's a really low coverage number. Can you bring it up to at least 80%?

@russgold, yes, we need a higher coverage number, but for context, I did ask @doxiao to create a draft pull request just so I could see how her webhook implementation was coming together.

rjeberhard:
@doxiao, the very good news is that this does look like the solution I hoped and expected to see -- thanks! 

If you use the Sonar analysis it will help you to see what new code isn't sufficiently unit-tested. Once that's addresses let's complete the review of your work and get this "part 1" in. After that, I'd prefer that you make several smaller PR's rather than only large "parts". I do see why this first PR had to be larger as you had to put the moving parts in place for the new webhook operation and the management of the registration.

doxiao:
The PR is ready to be reviewed. Note that the remaining integration test failures are a known issue that is not specific to validating webhook, and should be fixed by using webhookOnly option in the integration test runs. The PR can be reviewed while we are waiting for this integration test infra change.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
```diff
@@ -0,0 +1,288 @@
...
+import static oracle.kubernetes.operator.helpers.NamespaceHelper.getWebhookNamespace;
+import static oracle.kubernetes.operator.rest.RestConfigImpl.CONVERSION_WEBHOOK_HTTPS_PORT;
+import static oracle.kubernetes.operator.utils.SelfSignedCertUtils.WEBLOGIC_OPERATOR_WEBHOOK_SVC;
+
+public class WebhookHelper {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Webhook", "Operator");
+  public static final String VALIDATING_WEBHOOK_NAME = "weblogic.validating.webhook";
+  public static final String VALIDATING_WEBHOOK_PATH = "/admission";
+  public static final String APP_GROUP = "weblogic.oracle";
+  public static final String API_VERSION = "v9";
```
rjeberhard, line 41: Reuse the constants from KubernetesConstants so that we only have one place to update.

operator/src/main/java/oracle/kubernetes/operator/rest/resource/AdmissionWebhookResource.java
```diff
@@ -0,0 +1,110 @@
...
+   * Validate a change request to a domain or cluster resource.
+   *
+   * @param body - a String representation of JSON document describing the AdmissionReview with admission request.
+   *
+   * @return a String representation of JSON document describing the AdmissionReview with admission response.
+   *
+   */
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  public String post(String body) {
```
rjeberhard, line 49: Why aren't AdmissionReview and AdmissionResponse the parameter and return type, respectively? (Why are you using String?)

doxiao: Similar to the conversion webhook request/response, we have our own model classes for the validating webhook. Those classes do not have all the fields that the K8S supports. We take the incoming request as a string and construct it to the internal request object and return the response as a String representing the response object. ConversationWebhookResource class does the same. 

I tried to use AdmissionReview and AdmissionResponse directly and got 400 (bad request) error.

doxiao: I can explore more and update the code in a future PR if we find a better solution.

kubernetes/src/test/java/oracle/kubernetes/operator/create/CreateOperatorGeneratedFilesTestBase.java
```diff
@@ -457,7 +457,17 @@ private V1ClusterRole getExpectedWeblogicOperatorClusterRole() {
             newPolicyRule()
                 .addApiGroupsItem("authorization.k8s.io")
                 .resources(singletonList("selfsubjectrulesreviews"))
-                .verbs(singletonList("create")));
+                .verbs(singletonList("create")))
```
russgold, line 460: This creation of newPolicyRule is identical to the one at line 750, Please extract the two of them into a common method, and change the `asList` call to use `List.of()`

russgold (commented):
I've left a number of detailed comments, mostly related to code readability. I'd like to suggest one more. 

We've generally been tossing a lot of code into the `oracle.kubernetes.operator` and `oracle.kubernetes.operator.helper` packages, which have now become so broad as to be essentially meaningless. I'd like to suggest using packages for sets of classes that generally use other classes and methods that the rest of the operator does not. In this case, it seems that there is a fair bit of commonality between our webhooks that are not generally used elsewhere, and I would therefore recommending putting them into a new `oracle.kubernetes.operator.webhooks` package.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
```diff
@@ -0,0 +1,288 @@
...
+      return new ReadResponseStep(next);
+    }
+
+    private class ReadResponseStep extends DefaultResponseStep<V1ValidatingWebhookConfiguration> {
+      ReadResponseStep(Step next) {
+        super(next);
+      }
+
+      @Override
+      public NextAction onSuccess(
```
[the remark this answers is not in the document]
doxiao: hmm, I think the method declaration line is subject to the line length limit.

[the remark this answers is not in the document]
doxiao: Fixed the ones that can fit in a single line.

doxiao:
> I've left a number of detailed comments, mostly related to code readability. I'd like to suggest one more.
> 
> We've generally been tossing a lot of code into the `oracle.kubernetes.operator` and `oracle.kubernetes.operator.helper` packages, which have now become so broad as to be essentially meaningless. I'd like to suggest using packages for sets of classes that generally use other classes and methods that the rest of the operator does not. In this case, it seems that there is a fair bit of commonality between our webhooks that are not generally used elsewhere, and I would therefore recommending putting them into a new `oracle.kubernetes.operator.webhooks` package.

Thanks for your thorough review and helpful comments. I have addressed all of your comments except for this one. This is the second webhook that we have in the operator project. We need to take both webhooks into consideration when we make a decision to whether we need to put webhook related classes in a separate directory. This PR is already larger than I would like to have. Can we address this comment as a follow-up item and make necessary changes in a separate PR?

operator/src/main/java/oracle/kubernetes/operator/rest/model/Status.java
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.rest.model;
+
+import java.util.Objects;
+
+import com.google.gson.annotations.Expose;
+
+public class Status {
```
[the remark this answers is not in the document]
doxiao: done. thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
doxiao: done. thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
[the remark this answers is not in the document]
russgold: Please check again. I tried moving them in my local copy and they fit fine for the ones I tried.

russgold:
Addressing the packaging in a separate PR is fine.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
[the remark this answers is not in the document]
doxiao: Interesting - I tried before and tried again, but a long method declaration failed stylecheck for me. 

[WARNING] src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java:[187] (sizes) LineLength: Line is longer than 120 characters (found 132).

Here is line 187 that the above message complained.

     private boolean shouldUpdate(V1ValidatingWebhookConfiguration existingWebhookConfig, V1ValidatingWebhookConfiguration model) {

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionRequest.java
```diff
@@ -0,0 +1,87 @@
...
+package oracle.kubernetes.operator.rest.model;
+
+import java.util.Map;
+
+import com.google.gson.annotations.Expose;
+
+/**
+ * AdmissionRequest represents a Kubernetes admission request inside an AdmissionReview object sent by the
+ * Kubernetes ApiServer upon invoking an admission webhook.
+ */
```
russgold, line 13: Technically, you did indeed add javadoc which describes the class; I suppose I should have been more specific. I am looking for a description of the class that does not assume that the reader would have known what it is by the name. This javadoc doesn't actually provide useful information.

The terms 'admission request' and 'admission review' are Kubernetes concepts that not every operator developer is going to know automatically. I'm not looking for a long dissertation here - just something that would give somebody tasked to maintain this code a place to start understanding it. Ideally, you could provide a link to the K8S docs, if they were any good. Technical documentation is a real problem on many open source projects.

I'm thinking something more like the following - which might not be correct. I assume you understand what is going on here better than I do:

The process of adding a resource to Kubernetes is called "admission" and there are mechanisms both to validate and to modify resources when they seek admission. This object is created by Kubernetes and sent by it as a request to an operator-installed webhook, which can then decide whether the creation should be allowed, and what modifications, if any, need to be made.

Of course, you would also need javadoc for the fields to explain how they support what this object does. The goal, remember, is to make it as easy as possible for the next developer to pick up the work.

Please understand this a as a general comment on why Javadoc is important, and how it should be written,

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionReview.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * AdmissionReview represents a request or a response object for an admission webhook call.
+ *
+ * <p></p>When a user performs an operation to a Kubernetes resource, Kubernetes ApiServer invokes all admission
+ * webhooks that have registered their interests to the resource. Upon invocation, the Kubernetes ApiServer passes an
+ * AdmissionReview object to the webhook, and expects the webhook to return an AdmissionReview as the response. An
+ * admission webhook uses the information in the AdmissionReview request to perform the necessary operation
+ * the resource, and sends back an AdmissionReview object to the ApiServer about the result of the operation.
+ * </p>
```
russgold, line 18: OK, I think I am getting a better sense of what is going on here. If I understand you correctly, and AdmissionReview is a Kubernetes-created wrapper of either an AdminRequest, or a webhook-created wrapper of an AdmisisonResponse. Is that correct? If so, it would probably be better to say it that way rather than that K8S passes a review object _as a_ request and expects back a review object _as a_ response.

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionReview.java
```diff
@@ -0,0 +1,121 @@
...
+  public AdmissionReview kind(String kind) {
+    setKind(kind);
+    return this;
+  }
+
+  public AdmissionRequest getRequest() {
+    return request;
+  }
+
+  public void setRequest(AdmissionRequest request) {
```
russgold, line 90: Why do we need both `setRequest` and `request` methods?

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionReview.java
```diff
@@ -0,0 +1,121 @@
...
+    this.request = request;
+  }
+
+  public AdmissionReview request(AdmissionRequest request) {
+    setRequest(request);
+    return this;
+  }
+
+  public AdmissionResponse getResponse() {
```
russgold, line 99: Why do we need both `getResponse` and `response` methods?

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionRequest.java
```diff
@@ -0,0 +1,124 @@
...
+   * An uid uniquely identifying this admission call.
+   */
+  @SerializedName("uid")
+  @Expose
+  protected String uid;
+
+  /**
+   * Fully-qualified group/version/kind of the incoming object.
+   */
+  @SerializedName("kind")
```
russgold, line 33: I'm not following this. How is a map a "fully-qualified group/version/kind" of an object?

operator/src/main/java/oracle/kubernetes/operator/rest/resource/AdmissionWebhookResource.java
russgold: Do we really need the `entering` and `exiting` logs? I don't think we have ever found them useful.

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionReview.java
doxiao: The tool that the webhook code uses to convert a String representation of a JSon object to a Java object uses setRequest to populate the Java object. The webhook code itself uses request() to populate the response object. 
The same to the response field.
The same for Conversion webhook classes.

operator/src/main/java/oracle/kubernetes/operator/rest/model/AdmissionRequest.java
doxiao: The map contains the <key, value> pairs for "group", "version" and "kind" of the object.

rjeberhard (approved):

PR #3059, opened 2022-05-11
remove integration test failure counter

russgold:
Next phase of the severe error handling work. The main part of this is to remove the failure counter and ensure that all logic that depended on it still works.

### DIFF
--- a/common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
+++ b/common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
@@ -81,6 +81,7 @@ public class SchemaConversionUtils {
     adjustAdminPortForwardingDefault(spec, apiVersion);
     convertLegacyAuxiliaryImages(spec);
     removeObsoleteConditionsFromDomainStatus(domain);
+    removeObsoleteFieldsFromDomainStatus(domain);
     removeUnsupportedDomainStatusConditionReasons(domain);
     convertDomainHomeInImageToDomainHomeSourceType(domain);
     moveConfigOverrides(domain);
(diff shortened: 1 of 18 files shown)


common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
```diff
@@ -147,6 +148,15 @@ private void removeUnsupportedDomainStatusConditionReasons(Map<String, Object> d
     getStatusConditions(domain).forEach(this::removeReasonIfUnsupported);
   }
 
+  private void removeObsoleteFieldsFromDomainStatus(Map<String, Object> domain) {
+    Optional.ofNullable(domain.get("status")).ifPresent(this::removeObsoleteStatusFields);
+  }
+
+  private void removeObsoleteStatusFields(@Nonnull Object domainStatus) {
+    Map<String,Object> statusMap = (Map<String, Object>) domainStatus;
+    statusMap.remove("introspectJobFailureCount");
```
rjeberhard, line 157: Okay, I'm realizing based on a conversation with @robertpatrick regarding a different field that both this change and the code that @ankedia originally wrote both don't work. We want this field eliminated for v9, but it's presumably still used by WKO 3.4.0 and v8. Therefore, we need a mechanism for v8 -> v9 -> v8 fidelity since the 3.4.0 operator will read and write using v8 but the Domain will be stored as v9.

Robert and I briefly discussed if this means preserving the field in v9 and marking it deprecated -- it would actually be more than deprecated since WKO 4.0 would ignore it completely.  Or, alternatively, have an "attic" element where we could store values only for this v8 -> v9 -> v8 fidelity.

PR #3060, opened 2022-05-11
Owls 98836 - Changes for intermittent failure in ItElasticLoggingFluentd due to image pull and API exception

ankedia:
Owls 98836 - Changes for intermittent failure in ItElasticLoggingFluentd due to image pull and API exception. 
  - Use the elasticsearch image from OCIR instead of docker hub to reduce image pull latency.
  - Increased the wait time for the ELK stack pod to be ready.
  - Ignore an intermittent APIException from the Kubernetes API server when verifying service existence.

Jenkins results - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10583/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10575/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10569/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
@@ -504,8 +504,8 @@ public class Kubernetes {
    * @param serviceName name of the service to check for
    * @param label the key value pair with which the service is decorated with
    * @param namespace the namespace in which to check for the service
-   * @return true if the service is found, otherwise false. If there is
-   *              an error in querying the cluster, it returns false.
+   * @return true if the service is found, otherwise false.
+   * @throws ApiException when there is error in querying the cluster
    */
   public static boolean doesServiceExist(
       String serviceName, Map<String, String> label, String namespace)
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
@@ -504,18 +504,18 @@ public class Kubernetes {
    * @param serviceName name of the service to check for
    * @param label the key value pair with which the service is decorated with
    * @param namespace the namespace in which to check for the service
-   * @return true if the service is found otherwise false
-   * @throws ApiException when there is error in querying the cluster
+   * @return true if the service is found, otherwise false. If there is
+   *              an error in querying the cluster, it returns false.
    */
-  public static boolean doesServiceExist(
-      String serviceName, Map<String, String> label, String namespace)
-      throws ApiException {
-    boolean exist = false;
-    V1Service service = getService(serviceName, label, namespace);
-    if (service != null) {
-      exist = true;
+  public static boolean doesServiceExist(String serviceName, Map<String, String> label, String namespace) {
+    try {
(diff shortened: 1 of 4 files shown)


jshum2479 (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
```diff
@@ -504,18 +504,18 @@ public static V1Pod getPod(String namespace, String labelSelector, String podNam
...
    */
-  public static boolean doesServiceExist(
-      String serviceName, Map<String, String> label, String namespace)
-      throws ApiException {
-    boolean exist = false;
-    V1Service service = getService(serviceName, label, namespace);
-    if (service != null) {
-      exist = true;
+  public static boolean doesServiceExist(String serviceName, Map<String, String> label, String namespace) {
+    try {
```
rjeberhard, line 511: It doesn't seem right to suppress an exception since then we wouldn't see it in the retries. The implementation of getService() appears to be able to return the instance or null in normal cases, so wouldn't an ApiException represent a real problem? (And, yes, it should retry if an attempt had an exception, correct?)

ankedia: Currently the test fails if there's an ApiException and doesn't seem to do retries. See - https://build.weblogick8s.org:8443/job/wko-kind-nightly-parallel/861/testReport/oracle.weblogic.kubernetes/ItElasticLoggingFluentd/___/ . Do the integration tests have retry mechanism similar to how it's done in the operator?

My understanding is that with my change, if there's an APIException, the method `doesServiceExist` will return `false` and the check will be performed again until the service exists.

rjeberhard: https://github.com/awaitility/awaitility/wiki/Usage#ignoring-exceptions

Options: 
* Use Awaitility's ignoreExceptions(), or
* Swallow the exception, but I'd prefer it inside the lambda of the testUntil() rather than in the general-purpose method. 

Also, I can't figure out why the lamba has assertDoesNotThrow() rather than just catching the exception and having the conditional evaluate to false.

anpanigr (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/assertions/impl/Kubernetes.java
```diff
@@ -504,18 +504,13 @@ public static V1Pod getPod(String namespace, String labelSelector, String podNam
...
   public static boolean doesServiceExist(
       String serviceName, Map<String, String> label, String namespace)
       throws ApiException {
-    boolean exist = false;
-    V1Service service = getService(serviceName, label, namespace);
-    if (service != null) {
-      exist = true;
-    }
-    return exist;
+    return getService(serviceName, label, namespace) != null ? true : false;
```
rjeberhard, line 513: The ternary operator is redundant. You can just return the result of the null check:
```
return getService(serviceName, label, namespace) != null;
```

ankedia: Changed in [3bfacc5](https://github.com/oracle/weblogic-kubernetes-operator/pull/3060/commits/3bfacc537c300681c87acf04c7797a7cf0355b3e).

PR #3062, opened 2022-05-12
Stop use of introspectJobFailureCount field

russgold:
This change removes the logic that uses the number of introspection job failures, and deprecates the field itself. The corresponding logic is migrating to the total retry timeout

### DIFF
--- a/documentation/domains/Domain.json
+++ b/documentation/domains/Domain.json
@@ -528,6 +528,7 @@
           }
         },
         "introspectJobFailureCount": {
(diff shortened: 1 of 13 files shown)


russgold:
It wants me to add the Javadoc @deprecated tag, but the code style plugin rejects it when it extends across multiple lines. Do we have an example of one that is currently accepted?

rjeberhard (approved):

lennyphan (approved):

PR #3065, opened 2022-05-12
Add webhookOnly operator installation before integration tests starts

sankarpn:
Add webhookOnly operator installation before integration tests starts in ImageBuilders beforeAll method and delete the installation in close method.

Kind
-----
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10602/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10603/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10608/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
@@ -609,7 +609,7 @@ public class ImageBuilders implements BeforeAllCallback, ExtensionContext.Store.
             .namespace(webhookNamespace)
             .chartDir(OPERATOR_CHART_DIR);
     return installAndVerifyOperator(webhookNamespace, webhookSa, false, 0, opHelmParams, null,
-        false, false, null, null, false, "INFO", -1, -1, false, true, "default");
+        false, false, null, null, false, "INFO", -1, -1, true, "default");
   }
 
 }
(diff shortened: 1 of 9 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
```diff
@@ -582,5 +597,19 @@ private Callable<Boolean> pullImageFromOcrOrOcirAndPushToKind(String image) {
       return dockerPull(image) && dockerTag(image, kindRepoImage) && dockerPush(kindRepoImage);
     });
   }
+  
+  HelmParams opHelmParams;
+  String webhookNamespace;
+
+  private OperatorParams installWebHookOnlyOperator() {
+    webhookNamespace = assertDoesNotThrow(() -> createUniqueNamespace());
```
ankedia, line 605: Do we really need a unique namespace here? Or can it be set to fixed name such as `webhook-ns`. That will help when debugging the tests to identify the namespace associated with webhook.

ankedia (commented):
IMO, it would be good to validate that only one webhook pod is installed across the entire k8s cluster. Otherwise, you can use the `operatorOnly=true` helm config value when installing the operator to skip the webhook installation in other namespaces.

rjeberhard (approved):
I agree with @ankedia's two suggestions... I'm going to approve so that when his suggestions are addressed and he approves then this PR will have two approvals.

integration-tests/src/test/java/oracle/weblogic/kubernetes/extensions/ImageBuilders.java
sankarpn: Recreating the webhook namespace before installing it. Using `ns-webhook` for namespace for installing webhook

sankarpn:
> IMO, it would be good to validate that only one webhook pod is installed across the entire k8s cluster. Otherwise, you can use the `operatorOnly=true` helm config value when installing the operator to skip the webhook installation in other namespaces.

Spoke with Anil. Agreed to implement this in a separate JIRA task.

ankedia (approved):
LGTM

PR #3071, opened 2022-05-13
Add validation util class for validating webhook

doxiao:
This PR contains an utility class for validating webhook. The initial version only covers the checks for replicas.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
@@ -1,0 +1,110 @@
+// Copyright (c) 2022, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package oracle.kubernetes.operator.utils;
+
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Optional;
+
+import oracle.kubernetes.operator.logging.LoggingFacade;
+import oracle.kubernetes.operator.logging.LoggingFactory;
+import oracle.kubernetes.weblogic.domain.model.Cluster;
+import oracle.kubernetes.weblogic.domain.model.ClusterStatus;
+import oracle.kubernetes.weblogic.domain.model.Domain;
+import oracle.kubernetes.weblogic.domain.model.DomainSpec;
+import oracle.kubernetes.weblogic.domain.model.DomainStatus;
+import org.jetbrains.annotations.NotNull;
+
+import static org.apache.commons.collections4.CollectionUtils.isEqualCollection;
+
+public class ValidationUtils {
+  private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Webhook", "Operator");
+
+  private ValidationUtils() {
+  }
+
+  /**
+   * Validating a domain resource.
+   *
+   * @param existingDomain domain that needs to be validated against.
+   * @param proposedDomain domain that needs to be validated.
+   * @return true if valid, otherwise false
+   */
+  public static boolean validateDomain(@NotNull Domain existingDomain, @NotNull Domain proposedDomain) {
+    LOGGER.fine("Validating domain " + proposedDomain + " against " + existingDomain);
+    return unchanged(existingDomain, proposedDomain)
+        || validReplicas(proposedDomain)
+        || shouldIntrospect(existingDomain, proposedDomain);
+  }
+
+  private static boolean unchanged(Domain existingDomain, Domain proposedDomain) {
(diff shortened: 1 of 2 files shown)


operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * Validating a proposed domain resource against an existing domain resource. It returns true if the proposed changes
+   * in the proposed domain resource can be honored, otherwise, returns false.
+   *
+   * @param existingDomain domain that needs to be validated against.
+   * @param proposedDomain domain that needs to be validated.
+   * @return true if valid, otherwise false
+   */
+  public static boolean validateDomain(@NotNull Domain existingDomain, @NotNull Domain proposedDomain) {
```
russgold, line 47: The name is a bit confusing: it's not actually "validating" a domain, is it? I infer that it is actually returning true if the proposed change is valid. If so, it would probably be better to call it something like, `isPermittedChange` or `isDomainReplacementAllowed`. Methods that return booleans are conventionally given names that are predicates.

operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+    LOGGER.fine("Validating domain " + proposedDomain + " against " + existingDomain);
+    return unchanged(existingDomain, proposedDomain)
+        || validReplicas(proposedDomain)
+        || shouldIntrospect(existingDomain, proposedDomain);
+  }
+
+  private static boolean unchanged(@NotNull Domain existingDomain, @NotNull Domain proposedDomain) {
```
russgold, line 54: `isUnchanged` would be a more conventional name for this.

operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+      return true;
+    }
+    switch (proposedDomain.getDomainHomeSourceType()) {
+      case IMAGE:
+        return !Objects.equals(getImage(existingDomain), getImage(proposedDomain));
+      case FROM_MODEL:
+        if (noAuxiliaryImagesConfigured(existingDomain) && noAuxiliaryImagesConfigured(proposedDomain)) {
+          return !Objects.equals(getImage(existingDomain), getImage(proposedDomain));
+        } else {
+          return noAuxiliaryImagesConfigured(existingDomain)
```
russgold, line 85: This block could probably be extracted into its own method, which could be given a descriptive name to explain what is going on, such as `areAuxiliaryImagesChanged`. That would make the top-level method more clear.

operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private static boolean noAuxiliaryImagesConfigured(@NotNull Domain domain) {
+    return domain.getAuxiliaryImages() == null;
+  }
+
+  private static boolean validReplicas(@NotNull Domain domain) {
+    return Optional.of(domain)
+        .map(Domain::getStatus)
+        .map(DomainStatus::getClusters)
```
russgold, line 105: I have found that it is easier to understand this sequence of steps when they are grouped on a single line, as they represent a single concept rather than multiple transformations:

```
.map(DomainStatus::getClusters).orElse(Collections.emptyList()).stream()
```

Alternately, you could extract the three steps into a single method named something like `getClustersAsStream`

operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private static String getImage(@NotNull Domain existingDomain) {
+    return Optional.of(existingDomain).map(Domain::getSpec).map(DomainSpec::getImage).orElse(null);
+  }
+
+  private static boolean noAuxiliaryImagesConfigured(@NotNull Domain domain) {
+    return domain.getAuxiliaryImages() == null;
+  }
+
+  private static boolean validReplicas(@NotNull Domain domain) {
```
russgold, line 102: The name of this method and the next are unclear. It seems to me that it returns true if all clusters in the domain have valid replica counts. If so, a better name for this one might be `allReplicaCountsValid` and for the next `isReplicaCountValid`.

russgold (commented):
Overall looks very good. I have identified some places where the readability could be improved.

doxiao:
> Overall looks very good. I have identified some places where the readability could be improved.

I have addressed all of your comments. Please review again. thanks.

russgold (approved):

rjeberhard (approved):

PR #3072, opened 2022-05-13
Disable ItRemoteConsole tests on slim image

sankarpn:
- Disable ItRemoteConsole tests on slim image
- Check the `WebLogic version=` message in introspector log only on full image runs, not slim image

Kind
------
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10615/

In the above results the test [oracle.weblogic.kubernetes.ItMiiSampleWlsMain.testCheckMiiSampleSource] will fail since the image version is hard coded in the sample script below. It can be ignored when running tests against 14.1.1.0

https://github.com/oracle/weblogic-kubernetes-operator/blob/release/3.4/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/domain-resources/JRF-AI/mii-initial-d1-JRF-AI-v1.yaml


anpanigr (approved):
lgtm

PR #3090, opened 2022-05-20
Improve validating webhook configuration related logic and unit tests

doxiao:
This PR includes the following improvements.
- Check caBundle too when make decision to replace an existing config. Currently WebhookHelper only replaces an existing ValidatingWebhookConfiguration when the namespaces do not match. It turned out that we may need to update the configuration even when namespaces are the same because the caBundle may become stale if the certificates are missing and regenerated upon a restart of the webhook. 
- No retry on unrecoverable failures.
- More log messages.
- Unit tests refactoring plus cleanup.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10690/


rjeberhard:
@doxiao, why would certificates be regenerated based on the restart of the webhook? I accept that the certificates would be regenerated if the webhook were uninstalled and reinstalled, but I don't think this should happen simply because the webhook is restarted.

doxiao:
> @doxiao, why would certificates be regenerated based on the restart of the webhook? I accept that the certificates would be regenerated if the webhook were uninstalled and reinstalled, but I don't think this should happen simply because the webhook is restarted.

Sorry, I did not make myself clear. In normal cases, the certificates do not get regenerated. But the secret that holds the certificates may disappear for many reasons, for example, somebody may delete the namespace that hosts the webhook and the secret. When the webhook restarts and detects that the secret/certificates are missing, it will regenerate. The ValidatingWebhookConfiguration is not namespaced, and may holds an stale caBundle. The change is to update the existing validating webhook config with the new caBundle if needed.

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
```diff
@@ -281,30 +264,57 @@ protected NextAction onFailureNoRetry(Packet packet,
...
-    private class ReplaceResponseStep extends ResponseStep<V1ValidatingWebhookConfiguration> {
+    private class ReplaceResponseStep extends WebhookConfigResponseStep {
       ReplaceResponseStep(Step next) {
         super(next);
       }
 
       @Override
       public NextAction onFailure(Packet packet, CallResponse<V1ValidatingWebhookConfiguration> callResponse) {
-        return super.onFailure(getConflictStep(), packet, callResponse);
+        if (UnrecoverableErrorBuilder.isAsyncCallNotFoundFailure(callResponse)) {
```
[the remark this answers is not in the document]
doxiao: The essence of this change is to NOT give the conflict step for non-recoverable errors. The code before this change gives a conflict step for all failures. We need to check 'Not Found" case before checking non-recoverable (in the super) because 404 is considered non-recoverable.

rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/WebhookHelper.java
```diff
@@ -177,28 +178,41 @@ public NextAction onSuccess(Packet packet, CallResponse<V1ValidatingWebhookConfi
...
           return doNext(packet);
         }
       }
 
-      private boolean shouldUpdate(V1ValidatingWebhookConfiguration existingWebhookConfig,
-                                   V1ValidatingWebhookConfiguration model) {
-        return !getServiceNamespaceFromConfig(existingWebhookConfig).equals(getServiceNamespaceFromConfig(model));
+      private boolean shouldUpdate(V1ValidatingWebhookConfiguration existingWebhookConfig) {
+        return !getServiceNamespace(existingWebhookConfig).equals(getServiceNamespace(model))
+            || !Arrays.equals(getClientConfigCaBundle(existingWebhookConfig), getClientConfigCaBundle(model));
```
ankedia, line 190: Instead of checking individual fields of the Webhook configuration, I wonder if it's easier to just compare the entire existing webhook configuration with the new configuration. For e.g. something like -
return !existingWebhookConfig.equals(getFirstWebhook(model));

doxiao: That would be easier, but unfortunately, we cannot do that here. The existing object that we get back from the ApiServer may have additional fields that the locally-created model does not have, for example, the creation time. We just need to compare the two fields that we do care: the namespace and caBundle.

ankedia: My understanding is that equals method only considers apiVersion, kind, metadata and webhook specs and excludes the fields such as creation time.

```
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    } else if (o != null && this.getClass() == o.getClass()) {
      V1ValidatingWebhookConfiguration v1ValidatingWebhookConfiguration = (V1ValidatingWebhookConfiguration)o;
      return Objects.equals(this.apiVersion, v1ValidatingWebhookConfiguration.apiVersion) && Objects.equals(this.kind, v1ValidatingWebhookConfiguration.kind) && Objects.equals(this.metadata, v1ValidatingWebhookConfiguration.metadata) && Objects.equals(this.webhooks, v1ValidatingWebhookConfiguration.webhooks);
    } else {
      return false;
    }
  }
```

doxiao: The creation time is part of the metadata.

ankedia: I see, ok.

ankedia (approved):

PR #3097, opened 2022-05-24
Integrate validating webhook and validation utils

doxiao:
This PR including the following changes:

- Integrates the validating webhook with the validation utility methods, so the admission webhook resource calls the validation utility methods to perform the validation. 
- Refactors the ValidationUtilsTest so that the common methods are in a base class that is shared with WebhookRestTest as well.
- Updates some of the integration test cases because the behavior has changed. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10738/


rjeberhard (approved):

operator/src/test/java/oracle/kubernetes/operator/http/rest/WebhookRestTest.java
```diff
@@ -62,13 +74,21 @@ private AdmissionRequest createAdmissionRequest() {
     AdmissionRequest request = new AdmissionRequest();
     request.setUid(RESPONSE_UID);
     request.setKind(new HashMap<>());
-    request.setResource(new HashMap<>());
+    request.setResource(createResource());
     request.setSubResource(new HashMap<>());
-    request.setObject(new Object());
-    request.setOldObject(new Object());
+    request.setObject(null);
```
russgold, line 79: Shouldn't null be the default?

russgold (approved):
Just had one comment. Otherwise it looks good.

operator/src/test/java/oracle/kubernetes/operator/http/rest/WebhookRestTest.java
doxiao: yes. it is.

PR #3098, opened 2022-05-24
Owls 95596 Initial rename of Domain to DomainResource

lennyphan:
This PR is just a rename of `Domain` model class to `DomainResource` in preparation for

1. Creating the `ClusterResource` model class.
2. Generating the Cluster Resource schema from the `ClusterResource` model.
3. Updating effective configuration to incorporate configuration from Cluster Resource.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10703/


russgold:
The sonar warning about returning an empty collection rather than null is easy to fix and causes no problems - except that the method name `sortOrNull` would have to be changed.

The warning about duplicate code notes that the DomainResource and the AdmissionReview, both being Kubernetes resources, have common elements. This will be true about any custom resource, including the ClusterResource that will be added next. We could certainly add a common base class, although I don't see an actual value. So what we really want to be able to do is to tell Sonar that we did this on purpose and it should stop complaining. To do that, we use an annotation such as

```
@SuppressWarnings("squid:S00112")
```

russgold (approved):

rjeberhard (approved):

PR #3104, opened 2022-05-25
Clean up server down iterator

russgold:
Due to a bug in FiberTestSupport, we did not get an error on unit tests during a change to the server down logic. That logic was changed to run both the pod down steps and the waiters for the pod down complete in the same parallel operation, which works in production, but not in unit tests, where the (null) waiters were completing and clearing the pod from the DPI before the domain down step actually ran, once we added code to send an http shutdown request to the servers.

This worked initially because of the aforementioned bug, in which we were actually skipping some threads when scheduling new ones. Fixing that exposed the issue.

In all probability the waiters should be added to the corresponding down steps, but for now, they are all run once the server down operations have been run. This should unblock the retry logic work.


russgold:
and yes, I did misspell "iterator" :(

lennyphan (approved):

PR #3105, opened 2022-05-26
Start decoupling of make-right from DomainProcessorImpl

russgold:
The next set of changes need to verify that we abort processing on a failure. Currently, the make-right and domain-processing code is fairly tightly coupled, making that difficult. This change starts decoupling them, moving the relevant steps into the MakeRight class and moving the status scheduling logic out. At the same time, I am attempting to rationalize access to the DPI. In some steps, we are getting it from the Packet, and in others we pass it to the constructor. I am now changing them to consistently using the Packet. An argument could be made for the opposite, but consistency is what should really matter.


lennyphan (approved):

rjeberhard (approved):

rjeberhard:
@russgold, heads-up that there is a failing test, ItDiagnosticsFailedCondition.testImageDoesnotExist. I've restarted the test run, but it looks like the test failure is consistent.

russgold:
Looking into the test failure now.

PR #3108, opened 2022-05-26
Owls 99654 - Expand the WDT custom folder from the archive before calling update or create domain

ankedia:
Creating this draft PR for Owls 99654. This change expands the custom folder from the WDT archive in the domain dir before calling updateDomain.sh or createDomain.sh. Once we find out the correct operator version, I will create and push a debug image based on this change to OCIR or provide a tar file to verify the fix.

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -894,15 +894,7 @@ function wdtCreatePrimordialDomain() {
 
   trace "About to call '${WDT_BINDIR}/createDomain.sh ${wdtArgs}'."
 
-  trace "DEBUG: WLSDEPLOY_PROPERTIES is ${WLSDEPLOY_PROPERTIES}"
-
-  cd ${DOMAIN_HOME} || exitOrLoop
-  for file in $(sort_files ${IMG_ARCHIVES_ROOTDIR} "*.zip")
-    do
-        ${JAVA_HOME}/bin/jar xf ${IMG_ARCHIVES_ROOTDIR}/${file} wlsdeploy/custom
-    done
-
-  traceDirs before ${DOMAIN_HOME}/wlsdeploy/custom
+  expandWdtArchiveCustomDir
 
   if [ -z "${OPSS_FLAGS}" ]; then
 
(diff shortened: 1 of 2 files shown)


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -894,6 +894,16 @@ function wdtCreatePrimordialDomain() {
 
   trace "About to call '${WDT_BINDIR}/createDomain.sh ${wdtArgs}'."
 
+  trace "DEBUG: WLSDEPLOY_PROPERTIES is ${WLSDEPLOY_PROPERTIES}"
+
+  cd ${DOMAIN_HOME} || exitOrLoop
+  for file in $(sort_files ${IMG_ARCHIVES_ROOTDIR} "*.zip")
```
jshum2479, line 900: why not factor it in a method?

ankedia: ok. Originally I was thiking that it's only needed for updateDomain.sh, I'll refactor it into a method.

operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -894,6 +894,16 @@ function wdtCreatePrimordialDomain() {
 
   trace "About to call '${WDT_BINDIR}/createDomain.sh ${wdtArgs}'."
 
+  trace "DEBUG: WLSDEPLOY_PROPERTIES is ${WLSDEPLOY_PROPERTIES}"
```
jshum2479, line 897: There is dump of env values when the introspect job starts,  I am wondering should we revisit the list we dump and add it there (and possibly others) instead?

ankedia: I added WLSDEPLOY_PROPERTIES to the list. From the comment in traceEnv function, we purposely avoid dumping all env vars. Not sure what else needs to be added to this list.

jshum2479: Thanks.  Can you check with Lenny or Tom ?  I think Lenny looked at an issue before and wondering what the value of a variable.  It's ok if no one recall any.

jshum2479 (approved):

robertpatrick:
We cannot merge this into release/3.3, as there will never be any new releases of 3.3.x.

ankedia:
> We cannot merge this into release/3.3, as there will never be any new releases of 3.3.x.

ankedia:
Created PR #3110 with release/3.4 as base release to replace this PR.

PR #3125, opened 2022-06-01
Add response message and warnings

doxiao:
Currently, the validating webhook only responds with whether a request is allowed or not. With this PR, when the validating webhook rejects a request, it also returns a message to the client to tell the reason. In addition, in the cases where the webhook accepts a request when the replicas exceeds the cluster size because something else that will cause a re-introspection is also changed, it returns a warning message within the response.

Note: operator/src/main/java/oracle/kubernetes/operator/utils/ValidationUtils.java is moved to operator/src/main/java/oracle/kubernetes/operator/http/rest/resource/AdmissionChecker.java with additional changes. But git shows them as one deleted and one newly added.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10819/


rosemarymarano (approved):
One minor nit, otherwise fine.

operator/src/main/java/oracle/kubernetes/operator/http/rest/model/AdmissionResponse.java
```diff
@@ -29,6 +31,13 @@ public class AdmissionResponse {
   @Expose
```
russgold, line 31: These fields should all be private. We have accessors for any code outside the class that needs them.

russgold (approved):
Approved, but with recommendations for some cleanup

rjeberhard (approved):

PR #3128, opened 2022-06-01
Correct protocol designation for metrics port

rjeberhard:
The Verrazzano team has let us know that they need the protocol prefix to be `http-`. I don't remember why we originally used `tcp-`.

So, this is a very simple change but I'm concerned that we'll cause a pod roll on upgrade...

@ankedia or @russgold, what can I do if anything so that this change doesn't cause a pod roll on upgrade? I'm planning to backport this to 3.4.1.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
@@ -1489,7 +1489,7 @@ public abstract class PodStepContext extends BasePodStepContext {
     }
 
     private String getMetricsPortName() {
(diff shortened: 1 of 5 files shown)


russgold:
To avoid rolls on things like this, we have allowed multiple hashes to work. Look at PodStepContext.VerifyPodStep.canAdjustHashToMatch.

I see that we have the same code to adjust the metrics port name in multiple places. That should definitely be fixed before this is merged.

ankedia (approved):
LGTM

jshum2479 (commented):
LGTM.   In my PR,  I use appProtocol in ServiceHelper for the exporter and the name in the pod's port just metrics, integration test works fine.   I am wondering if we can do the same thing here.

PR #3130, opened 2022-06-02
Owls 98870 removeistioschema

jshum2479:
This PR removes istio from the domain spec schema and making NAP and service port definition the same between istio and non-istio situation.   In order to use istio, user just have to enable sidecar injection either by enable the namespace or annotate the serverPod.


operator/src/main/java/oracle/kubernetes/operator/helpers/PodStepContext.java
```diff
@@ -1489,7 +1470,7 @@ private V1Container createMonitoringExporterContainer() {
     }
 
     private String getMetricsPortName() {
-      return getDomain().isIstioEnabled() ? "tcp-metrics" : "metrics";
+      return "metrics";
```
rjeberhard, line 1473: How does the operator now set the `appProtocol` field for this port?

PR #3139, opened 2022-06-03
Correct message property names

russgold:
Added a unit test to detect missing or misnamed message property entries - and corrected the incorrect ones.


doxiao (approved):
LGTM.

PR #3140, opened 2022-06-03
Owls 96933 - Quick start guide updates for operator 4.0

ankedia:
Owls 96933 - Updates to the Quick Start for Operator 4.0. The updated instructions -

- Don't require the user to clone the operator repo and pull images locally.
- Uses a `Model in Image` domain with an auxiliary image instead of a `domain-in-image` domain.
- Runs the published Helm chart instead of running helm chart from the source repo.
- Deploys a sample app that the user can access from the command line and the browser after the domain is created.
- Removes the use of helper scripts to create/delete domain etc.


tbarnes-us:
Here are some general comments.
- Looks good overall. I realize that this is a draft and so you probably already have plans for much of the following, so please forgive any redundancy:
- Consider renaming "samples/resources" to make it obvious it's meant for quick start. For example, "samples/quick-start". Also, reference the quick-start instructions from "samples/README.md".
- There is missing or minimal introductory explanation about basics. I suspect you already have some plans for this, but just in case you don't, I think it's helpful for anyone that runs the tutorial to come out with a good understanding of WKO fundamentals. What is "model in image"?  "domain uid", "auxiliary image", "WL Image", "operator"? "domain resource"? "model"?  The "domain runtime encryption secret?"  How do these relate to each-other? These concepts should be introduced step-by-step throughout the sample. There is some of this already in the optional create-your-own-aux-image section's introduction, but I assume most quick start users will skip that section.
- The draft is already largely doing the following, but is missing pieces parts here and there:  In my opinion, it'd overall help to have a consistent concept throughout where the end user works step-by-step coming to an understanding about the nature about each key part of WKO, and step-by-step building up a reusable central single staging directory for the domain that includes everything needed in one intuitive place.  A copy of the domain YAML, a copy of the ingress YAML, model files (when going manual), source for archive (when going manual), zipped archive (when going manual).
- The QS sample implicitly uses a GA image. The removed warning for GA images should be put back in key place(s).
- It looks like the aux image shares similar name/tag nomenclature as the full MII sample? I recommend changing it to avoid conflict/confusion and to make it obvious that its part of the quick-start. For example, include the string "quick-start-aux-image" in the image name and "QS" in the tag.
- Similarly, I recommend renaming the quick-start domain resource YAML file to include the words "quick-start".  Perhaps call it "quick-start-domain-resource.yaml" or "quick-start-mii-domain-resource.yaml"?  Not sure if it helps to include "mii".
- The source for creating the aux image that's used in the sample appears to be missing - it's just a manual step. I know end users will not use such source, but I assume this should be automated or at least documented as part of the release process, plus included in the check-in. Every new release will have to go through this step. 
- For consistency and convenience, consider including the ingress YAML in samples/resources along with the domain YAML, etc, that already exist at the same location. Plus include curl code example of downloading same to the tmp staging directory (so user can avoid the need to create the ingress file and cut-and-paste into it).
- The source for creating the archive is missing. I assume this should be added and the zip creation should be automated as part of the mvn build.  
- Also I assume that the source and steps for creating the archive itself should be included in the sample's 'optional' path. This should not be hidden away from the end user in my opinion, as it's an important part of the overall flow.
- Is it OK to check in a zip into the source? I thought that was against Oracle open source guidelines. Ryan would know.
- In my opinion, the domain resource is the most important part of the QS sample, but is still a bit "hidden" just like it was in the old QS sample (the old QS sample hides it even deeper). Since the hidden nature of the domain resource in the QS sample is one of the most common causes of confusion, how about breaking the sample down into steps a bit differently?  1. Add  a little discussion about the need for a domain resource and its purpose. 2. Provide curl & direct-from-doc options for getting said YAML for all QS users (this keeps it simple/non-optional and skips the unneeded "apply ... https" access option - IMO we want to strongly encourage QS users to examine and understand this file).  3. If following the 'optional' create-your-own-aux-image path, provide instructions for modifying same.
- Similarly, model files are hidden/implicit in the sample, although they are absolutely key. There should be text telling the reader what they are, drop downs with the full source of the models, why they are important, etc.  By the end of the tutorial end user should know what a model file is, and that the operator uses it to create a domain home dynamically when the domain is deployed.
- Consider adding a tutorial step for kubectl describing the domain, and discussing how to analyze, or at least find, the status that it reports. 
- In particular perhaps add a step that verifies that the operator has seen the domain resource in the first place (by checking the resource status or perhaps by checking domain namespace events -- checking events might be nice as that implicitly gives users an introduction to such events ) -- this is a common problem for newbies...
- Consider referencing the operator debugging section at the end of the operator install guidance. Similarly, consider referencing domain debugging at the end of the domain apply guidance.
- The entire sample needs automated tests that are parametrized to work in any namespace. It should not be possible to change any aspect of the sample without the change requiring that it pass tests.

rjeberhard:
| Consider renaming "samples/resources" to make it obvious it's meant for quick start. For example, "samples/quick-start". Also, reference the quick-start instructions from "samples/README.md".

This seems reasonable... We do want to clearly separate the quick start and the richer / more complicated samples.

rjeberhard:
> There is missing or minimal introductory explanation about basics. I suspect you already have some plans for this, but just in case you don't, I think it's helpful for anyone that runs the tutorial to come out with a good understanding of WKO fundamentals. What is "model in image"? "domain uid", "auxiliary image", "WL Image", "operator"? "domain resource"? "model"? The "domain runtime encryption secret?" How do these relate to each-other? These concepts should be introduced step-by-step throughout the sample. There is some of this already in the optional create-your-own-aux-image section's introduction, but I assume most quick start users will skip that section.

I don't think that the quick start should do all this. It could link to existing documentation pages or samples to "learn more".

rjeberhard:
> The source for creating the aux image that's used in the sample appears to be missing - it's just a manual step. I know end users will not use such source, but I assume this should be automated or at least documented as part of the release process, plus included in the check-in. Every new release will have to go through this step.

I don't understand this. My expectation is that a sample aux image with a simple web app. will stay the same for a long, long time.

rjeberhard:
> Is it OK to check in a zip into the source? I thought that was against Oracle open source guidelines. Ryan would know.

Unfortunately, it's not. Zip files can be in the "assets" of a release or we can publish a container image. We can check-in the files that would be zipped.

ankedia:
> > Consider renaming "samples/resources" to make it obvious it's meant for quick start. For example, "samples/quick-start". Also, reference the quick-start instructions from "samples/README.md".
> 
> This seems reasonable... We do want to clearly separate the quick start and the richer / more complicated samples.

@rjeberhard @tbarnes-us @rosemarymarano - I have made few changes based on Tom's general comments. Some are still being discussed in the slack channel and I'm not sure what to do about few other comments. I will setup a meeting to discuss these comments so we can finalize these.

documentation/staging/content/quickstart/create-auxiliary-image.md
```diff
@@ -0,0 +1,164 @@
...
+   has a version of WDT in its `--type wdt --version latest` location, then the
+   `cache addInstaller` command would fail.
+   For more information about the WIT cache, see the
+   [WIT Cache documentation](https://oracle.github.io/weblogic-image-tool/userguide/tools/cache/).
+
+   These steps will install WIT to the `/tmp/quickstart/tools/imagetool` directory,
+   plus put a `wdt_latest` entry in the tool's cache which points to the WDT ZIP file installer.
+   You will use WIT and its cached reference to the WDT installer later in the sample for creating model images.
+
+1. Download the sample WDT model, archives and properties files to be included in the auxiliary image and put them in your `/tmp/quickstart/models` directory.
```
tbarnes-us, line 75: Download the sample WDT model, WDT properties, and WebLogic application files to be...

...and then use the `jar` command to create a WDT archive file.  For more information, see the MII UG.

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -2,57 +2,49 @@
 title: "Create a domain"
 date: 2019-02-22T15:44:42-05:00
 draft: false
-weight: 6
+weight: 3
 ---
 
+1.  Select a user name and password, following the required rules for password creation (at least 8 alphanumeric characters with at least one number or special character).
```
tbarnes-us, line 8: Select a user name and password for the WebLogic domain administrator credentials...  

(Or simply combine this step with the next one.)

documentation/staging/content/quickstart/create-auxiliary-image.md
rosemarymarano: Got it. Done.

documentation/staging/content/quickstart/prepare.md
```diff
@@ -24,7 +24,30 @@ weight: 5
...
+    log in using the Oracle Single Sign-On (SSO) authentication service. If you do not already have SSO credentials,
+    then at the top, right side of the page, click Sign In to create them.
+
+    b. Search for `weblogic`, then select `weblogic` in the Search Results.
+
+    c. From the drop-down menu, select your language and click Continue.
+
+    d. Then, read and accept the license agreement.
+
+1. Create a `docker-registry` secret to enable pulling the example image from the registry.
```
tbarnes-us, line 43: example image --> ?

I assume the aux image _does not_ also need the registry secret? @ankedia would know. 

If it _does_, then make it "example WebLogic Server image and sample auxiliary image" or "example images".

If it _does not_, then make it "example WebLogic Server image" to disambiguate from the aux image.

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -2,57 +2,50 @@
...
+1. Select a user name and password for the WebLogic domain administrator credentials and use them to create a Kubernetes Secret for the domain:
 
-1. For use in the following steps:
+    ```shell
+    $ kubectl create secret generic sample-domain1-weblogic-credentials \
+      --from-literal=username=YOUR_USERNAME --from-literal=password=YOUR_PASSWORD \
+      -n sample-domain1-ns
+    ```
+
+   Replace `YOUR_USERNAME` and `YOUR_PASSWORD` with your user name and password.
```
tbarnes-us, line 16: Just to mitigate the slight chance of confusion with SSO, consider changing "YOUR_" to "ADMIN_" and replacing "your" with "your choice of".

Perhaps add: The password must be at least 8 characters long and must contain at least one non-alphabetical character.

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -2,57 +2,50 @@
...
-    {{% notice note %}}You need to provide the same WebLogic domain administrator user name and password in the `-u` and `-p` options
-    respectively, as you provided when creating the Kubernetes Secret in Step 2.
-    {{% /notice %}}
+    - **NOTE**: Instead of running the previous `kubectl apply` command, you can download the WLS Domain YAML file to `/tmp/quickstart/quick-start-domain-resource.yaml` or similar, using the following command, and then apply the file using `kubectl apply -f /tmp/quickstart/quick-start-domain-resource.yaml`:
 
-    For the detailed steps that the `create-domain.sh` script performs, see [Domain Home In Image]({{< relref "/samples/domains/domain-home-in-image/_index.md" >}}).
+      ```shell
+      $ curl -m 120 -fL https://raw.githubusercontent.com/oracle/weblogic-kubernetes-operator/main/kubernetes/samples/quick-start/quick-start-domain-resource.yaml -o /tmp/quickstart/quick-start-domain-resource.yaml
+      ```
+      This domain resource references a WebLogic Server installation image, the secrets you defined, and a sample "auxiliary image," which contains traditional WebLogic configuration and a WebLogic application.
```
tbarnes-us, line 42: Indented too far. "This domain resource..." should not be part of the above bullet.

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -2,57 +2,50 @@
...
-    ```
-    {{% notice note %}}You need to provide the same WebLogic domain administrator user name and password in the `-u` and `-p` options
-    respectively, as you provided when creating the Kubernetes Secret in Step 2.
-    {{% /notice %}}
+    - **NOTE**: Instead of running the previous `kubectl apply` command, you can download the WLS Domain YAML file to `/tmp/quickstart/quick-start-domain-resource.yaml` or similar, using the following command, and then apply the file using `kubectl apply -f /tmp/quickstart/quick-start-domain-resource.yaml`:
 
-    For the detailed steps that the `create-domain.sh` script performs, see [Domain Home In Image]({{< relref "/samples/domains/domain-home-in-image/_index.md" >}}).
+      ```shell
+      $ curl -m 120 -fL https://raw.githubusercontent.com/oracle/weblogic-kubernetes-operator/main/kubernetes/samples/quick-start/quick-start-domain-resource.yaml -o /tmp/quickstart/quick-start-domain-resource.yaml
+      ```
```
tbarnes-us, line 41: Minor: Consider changing this slightly to more closely mirror the step in the previous bullet.

**NOTE**: Instead of running the previous `kubectl apply` command, you can download the WLS Domain YAML file to `/tmp/quickstart/quick-start-domain-resource.yaml` or similar, using the following command:

```
$ curl -m 120 -fL https://raw.githubusercontent.com/oracle/weblogic-kubernetes-operator/main/kubernetes/samples/quick-start/quick-start-domain-resource.yaml -o /tmp/quickstart/quick-start-domain-resource.yaml
```
Then apply the file using:
```
kubectl apply -f /tmp/quickstart/quick-start-domain-resource.yaml
```

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -2,57 +2,50 @@
...
+         - To examine the domain resource, click [here](https://raw.githubusercontent.com/oracle/weblogic-kubernetes-operator/main/kubernetes/samples/quick-start/quick-start-domain-resource.yaml).
+         - For detailed information, see [Domain resource]({{< relref "/managing-domains/domain-resource.md" >}}).
 
+   {{% notice note %}}
+   The Quick Start guide uses a WebLogic Server version 12.2.1.4 General Availability (GA) image. GA images are suitable for demonstration and development purposes _only_ where the environments are not available from the public Internet; they are **not acceptable for production use**. In production, you should always use CPU (patched) images from [OCR]({{< relref "/base-images/ocr-images.md" >}}) or create your images using the [WebLogic Image Tool]({{< relref "/base-images/custom-images#create-a-custom-base-image" >}}) (WIT) with the `--recommendedPatches` option. For more guidance, see [Apply the Latest Patches and Updates](https://www.oracle.com/pls/topic/lookup?ctx=en/middleware/standalone/weblogic-server/14.1.1.0&id=LOCKD-GUID-2DA84185-46BA-4D7A-80D2-9D577A4E8DE2) in _Securing a Production Environment for Oracle WebLogic Server_.
```
tbarnes-us, line 47: The Quick Start guide uses a 
-->
The Quick Start guide's sample domain resource references a

kubernetes/samples/quick-start/index.jsp
```diff
@@ -0,0 +1,30 @@
...
+import="javax.servlet.http.HttpServlet"
+import="javax.servlet.http.HttpServletRequest"
+import="javax.servlet.http.HttpServletResponse"
+import="java.net.InetAddress"%>
+<%@ page language="java" contentType="text/html; charset=UTF-8"
+	pageEncoding="UTF-8"%>
+<!DOCTYPE html>
+<html>
+<body>
+        <h1>Welcome to WebLogic on Kubernetes Quick Start Sample</font></h1><br>
```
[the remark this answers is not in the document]
tbarnes-us: @rosemarymarano ? 

Note that if we change it here, we also need to change "create-domain.md" to have a matching change.

documentation/staging/content/quickstart/summary.md
```diff
@@ -0,0 +1,365 @@
...
+   #  WebLogic domain configuration.
+   introspectVersion: '1'
+
+     # Secrets that are referenced by model yaml macros
+     # (the model yaml in the optional configMap or in the image)
+     #secrets:
+     #- sample-domain1-datasource-secret
+     ```
+
+    {{% /expand %}}
```
tbarnes-us, line 328: Add something like:

Alternatively, use curl to download the domain resource YAML.

(Copy the domain resource curl instruction from the main Quick Start step to here.)

tbarnes-us (commented):
Looking pretty good to me. I've put in more minor comments.

documentation/staging/content/quickstart/install.md
```diff
@@ -2,69 +2,74 @@
...
+       --namespace traefik \
+       --set "ports.web.nodePort=30305" \
+       --set "ports.websecure.nodePort=30443" \
+       --set "kubernetes.namespaces={traefik}"
+   ```
+    This deploys the Traefik controller with plain text node port `30305`, SSL node port `30443`, and `kubernetes.namespaces` specifically set.
 
-4. Verify that the operator's pod is running, by listing the pods in the operator's namespace. You should see one
-for the operator.
+1. Verify that the operator's pod is running by listing the pods in the operator's namespace. You should see one
```
tbarnes-us, line 64: This step and the following belong under the 'operator steps', not the 'traefik steps'. They were not moved when we put the operator as the first action.

documentation/staging/content/quickstart/cleanup.md
```diff
@@ -2,22 +2,23 @@
 title: "Clean up"
```
tbarnes-us, line 2: @ankedia @rjeberhard - I am still a little worried about accounting for shared operator resources and the multi-operator conflict case:

- Should these cleanup instructions include a step for removing the CRD (assuming this is the only operator in the cluster)?
- Do we need to worry here about the operator uninstall potentially uninstalling another running operator's webhook?
- Or should we simply add a sentence in the beginning that the quick start sample assumes that there are no other operators running in the cluster?

kubernetes/samples/quick-start/index.jsp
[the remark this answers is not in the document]
tbarnes-us: @ankedia It turns out the "the" is needed.

kubernetes/samples/quick-start/model.yaml
```diff
@@ -0,0 +1,42 @@
+domainInfo:
+    AdminUserName: '@@SECRET:__weblogic-credentials__:username@@'
+    AdminPassword: '@@SECRET:__weblogic-credentials__:password@@'
+topology:
+    Name: sample-domain1
```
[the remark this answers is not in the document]
ankedia: I removed it.

kubernetes/samples/quick-start/index.jsp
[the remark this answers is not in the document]
ankedia: Changed.

kubernetes/samples/quick-start/quick-start-domain-resource.yaml
```diff
@@ -0,0 +1,150 @@
...
+  replicas: 1
+
+  # The desired behavior for starting a specific cluster's member servers
+  clusters:
+  - clusterName: cluster-1
+    serverStartState: "RUNNING"
+    serverPod:
+      # Instructs Kubernetes scheduler to prefer nodes for new cluster members where there are not
+      # already members of the same cluster.
+      affinity:
```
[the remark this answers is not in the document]
ankedia: I removed the affinity since it simplifies the domain resource yaml and is probably not important to override the default anti-affinity behavior for quick start. Let me know if we need to add it back and comment out the lines.

documentation/staging/content/quickstart/summary.md
```diff
@@ -0,0 +1,369 @@
...
+   introspectVersion: '1'
+
+     # Secrets that are referenced by model yaml macros
+     # (the model yaml in the optional configMap or in the image)
+     #secrets:
+     #- sample-domain1-datasource-secret
+
+    {{% /expand %}}
+
+    b. **NOTE**: Instead of running the previous `kubectl apply` command, you can download the WLS Domain YAML file to `/tmp/quickstart/quick-start-domain-resource.yaml` or similar, using the following command, and then apply the file using `kubectl apply -f /tmp/quickstart/quick-start-domain-resource.yaml`:
```
tbarnes-us, line 328: Oops - we both missed this:  This says "NOTE: Instead of running the previous kubectl apply command..." but there is no previous apply command.
to fix: combine the note into the "a" bullet, and change wording to "Alternatively, you can download the WLS Domain YAML file to /tmp/quickstart/quick-start-domain-resource.yaml using the following command:"

tbarnes-us (approved):
LGTM! I have some suggested follow-up after this pull, and will start discussions in slack.

rjeberhard (approved):

PR #3151, opened 2022-06-08
Add observedGeneration to Domain status

rjeberhard:
Adding `observedGeneration` as described here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

"Some resources report the observedGeneration, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status."

The idea is that a customer may update the Domain spec and then observe that the status has a Complete=True condition; however, that condition was actually associated with the previous value of the Domain spec. The `observedGeneration` field lets the customer know if the status now reflects the current generation.  This is why the implementation only updates the observed generation during a make-right operation and not during a background status update.


doxiao (approved):
LGTM. 

Just wondering if we need to add a unit test to verify that the observedGeneration is not changed after a non-make-right status update.

lennyphan (approved):

PR #3158, opened 2022-06-10
OWLS-99679 - Fix ItKubernetesDomainEvents.testDomainK8sEventsProcessingFailed test failure

alai8:
Cause:
Test intentionally patches domain with non-existing pvc with updated introspectVersion to trigger introspection.
It then patches domain with correct pvc and updated introspectVersion to trigger another introspection.
However, the introspection pod for the first introspector job with non-existing pvc sometimes stayed in `PENDING` phase for a long time,
with condition type `PODSCHEDULED` and reason `Unschedulable`.
This causes the introspection for the corrected pvc to not run before the test timed out waiting for the Completed event.

Fix:
In `VerifyIntrospectorJobResponseStep`, if the job read from the readJob call is a job for a different introspectVersion,
delete the current job and restart new one.
This involves adding `weblogic.introspectVersion` label to the metadata of the introspector job.

Update introspectVersion documentation to mention that its value must be a valid label value in Kubernetes, since operator also uses introspectVersion in metadata labels in weblogic-domain-introspect-cm ConfigMap and server pods.

Test testIntrospectDomainScript() in ItIntrospectVersion is using an invalid introspectVersion value of "My Version". It is updated to use a valid value of "My_Version-1".

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/10963/


operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
```diff
@@ -989,6 +993,69 @@ void whenJobInProgressWithNonAuxImageInitContainer_doNotCreateNewJob() {
...
+    defineIntrospectionWithIntrospectVersionLabel("v2");
+    testSupport.doOnCreate(JOB, this::recordJob);
+
+    testSupport.runSteps(JobHelper.createIntrospectionStartStep(null));
+
+    assertThat(affectedJob, nullValue());
+  }
+
+  @Test
+  void whenJobInProgressAndINullIntrospectVersionUnchanged_doNotCreateNewJob() {
```
doxiao, line 1045: Typo in the method name - xxxAnd**INull**Inxxx.

doxiao (approved):
LGTM, just one minor nit.

My observation is that the test case used to fail occasionally, but started to fail consistently recently. I am curious about if we have an explanation for that.

alai8:
> My observation is that the test case used to fail occasionally, but started to fail consistently recently. I am curious about if we have an explanation for that.

When the test passes, I notice that the introspector pod with invalid pvc would fail after DeadlineExceeded and thus the introspection with the corrected pvc could start. 
But I do not know why this doesn't happen consistently, or why the test has been failing more frequently.

rjeberhard (approved):

PR #3179, opened 2022-06-15
owls-99772 - Github issue # 3122 - Provide ablility to specify the UID for the container in helm chart

### LINKED ISSUE #3122: Would like to be able to specify the UID for the container in helm chart v3.4.0
Would like to be able to specify the UID for the container. We are using older version of Weblogic Operator (3.0.0) and now we are planning to use the latest one. We have **USER 1000** specified in the docker image in v3.0.0, whereas with latest images numeric user is not specifed and we have **USER oracle** in the docker image (in v3.4.0, v3.3.0).

We are using our own container management orchestration system and it has more restrictive policy. Since numeric user is not provided deploying Weblogic Operator fails with below error,
Error: container has runAsNonRoot and image has non-numeric user (oracle), cannot verify user is non-root

In order to over come this we have either specify USER 1000 in docker image or include below in the deployment yaml spec (_operator-dep.tpl),
  securityContext:
    runAsUser: 1000
	
It would be nice have an option where we can do this in the official helm chart (weblogic-operator-3.4.0) itself. If not good to have always it could be configurable based one some flag (like we provide cpuLimits, memoryLimits currently).

WO

alai8:
merging to release/3.4 branch

resolves #3122 

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11010/

### DIFF
--- a/documentation/3.4/content/userguide/managing-operators/using-helm.md
+++ b/documentation/3.4/content/userguide/managing-operators/using-helm.md
@@ -22,6 +22,7 @@ description: "An operator runtime is installed and configured using Helm. Here a
     - [`labels`](#labels)
     - [`nodeSelector`](#nodeselector)
     - [`affinity`](#affinity)
+    - [`runAsUser`](#runasuser)
   - [WebLogic domain management](#weblogic-domain-management)
     - [`domainNamespaceSelectionStrategy`](#domainnamespaceselectionstrategy)
     - [`domainNamespaces`](#domainnamespaces)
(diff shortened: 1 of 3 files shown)


documentation/3.4/content/userguide/managing-operators/using-helm.md
```diff
@@ -288,6 +288,14 @@ affinity:
           - another-node-label-value
 ```
 
+##### `runAsUser`
```
rosemarymarano, line 291: Please remember to add this heading as a cross-reference in the doc's TOC.

rosemarymarano (approved):
One important edit, otherwise fine.

rjeberhard (approved):
Normally, we fix in main first and then backport to a release branch. It's okay to do it the other way, but please do forward port to main right away.

PR #3192, opened 2022-06-17
remove duplicate procedures

rosemarymarano:
In a Quick Start guide, it goes against technical writing best practices to implement duplicate procedures without a **known** benefit. (Please correct me if there is one.) These add to its maintenance and the likelihood of becoming out-of-sync. @ankedia advised me that copy & paste is easier for users, so I removed the procedural duplicate curl commands to download the files in two instances. From discussions with @tbarnes-us, I know he doesn't agree. **NOTE**: I kept the curl command to download the sample domain YAML file because there was no other way for users to view or modify it, if needed.

### DIFF
--- a/documentation/staging/content/quickstart/create-domain.md
+++ b/documentation/staging/content/quickstart/create-domain.md
@@ -80,41 +80,7 @@ weight: 3
 
 1.	Create an ingress route for the domain, in the domain namespace, by using the following YAML file.
 
-    a. Copy the following IngressRoute YAML to a file called `/tmp/quickstart/ingress-route.yaml` or similar:
-
-
-    {{%expand "Click here to view the ingress route YAML file." %}}
-    # Copyright (c) 2022, Oracle and/or its affiliates.
-    # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-
-    apiVersion: traefik.containo.us/v1alpha1
-    kind: IngressRoute
-    metadata:
-      name: console
-      namespace: sample-domain1-ns
-    spec:
-      routes:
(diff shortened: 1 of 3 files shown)


tbarnes-us:
Yes, you are correct about my thoughts. In my opinion, the procedures are not duplicate and instead are convenient alternates for those that may want to edit the files.  Also, in the 'summary' edit, note that cutting and pasting is error prone and can be fairly tedious/problematic on some terminal setups.

robertpatrick (approved):

documentation/staging/content/quickstart/create-domain.md
```diff
@@ -117,22 +115,13 @@ weight: 3
...
           port: 8001
     {{% /expand %}}
+
+    b. Then apply the file using:
     ```shell
     $ kubectl apply -f /tmp/quickstart/ingress-route.yaml \
       --namespace sample-domain1-ns
     ```
 
-      b. **NOTE**: Instead of running the previous  `kubectl apply` command, you can download the ingress route YAML file to `/tmp/quickstart/ingress-route.yaml` or similar, using the following command:
```
ankedia, line 125: Based on slack discussion, should we provide a link to download the "ingress-route.yaml"? For e.g. you can download the ingress route YAML file from [here](https://raw.githubusercontent.com/oracle/weblogic-kubernetes-operator/main/kubernetes/samples/quick-start/ingress-route.yaml -o /tmp/quickstart/ingress-route.yaml)..

rosemarymarano: Agree. Done.

ankedia (approved):
LGTM

PR #3194, opened 2022-06-19
OWLS-95596: Add Cluster watcher

lennyphan:
Add Cluster watcher to handle processing of Cluster Resource events.


kubernetes/charts/weblogic-operator/templates/_operator-clusterrole-general.tpl
```diff
@@ -28,7 +28,7 @@ rules:
   verbs: ["get", "list", "watch", "create", "update", "patch"]
 {{- end }}
 - apiGroups: ["weblogic.oracle"]
-  resources: ["domains", "domains/status"]
+  resources: ["clusters", "domains", "domains/status"]
```
rjeberhard, line 31: You will need "clusters/status" here as well. This will be required once you have the operator updating status for the Cluster resources.

lennyphan: right, I can add it but I was waiting till I actually started work on the cluster status.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -589,6 +590,44 @@ public void dispatchEventWatch(Watch.Response<CoreV1Event> item) {
...
+
+      case ERROR:
+      default:
+    }
+  }
+
+  private void handleAddedCluster(ClusterResource cluster) {
+    LOGGER.info(MessageKeys.WATCH_CLUSTER, cluster.getSpec().getClusterName(),
+        cluster.getDomainUid());
+  }
```
russgold, line 618: No behavior, here?

russgold:
The watcher obviously isn't going to allow us to have much make-right behavior if the domain processing doesn't do anything with the cluster resource. Are you intending to defer that to a new PR?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
lennyphan: No behavior yet.  This is where we would initiate the 'makeRight' operation, just like when we handle Domain Resource events, and so my intention is for us to work together on the 'makeRight' operation for handling Cluster Resource events on the next PR.

lennyphan:
> The watcher obviously isn't going to allow us to have much make-right behavior if the domain processing doesn't do anything with the cluster resource. Are you intending to defer that to a new PR?

The idea was to handle Cluster Resource events similar to Domain events.  How we handle the Cluster Resource events is intended for the next PR that we pair together on.

russgold (approved):

rjeberhard (approved):

PR #3205, opened 2022-06-24
Update domain-on-pv samples to use new wdt model format and commandline option

doxiao:
This PR revises the Kubernetes portion of the domain on pv samples to use the new model format and command line option that are introduced in WDT 2.3.0.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11159/ (has the same failures reported in the nightlies on 6/23.


rakillen (approved):

alai8 (approved):

doxiao:
> > https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11159/ (has the same failures reported in the nightlies on 6/23.
> 
> Is this change suppose to fix some of the failures?

The issue that is fixed here that is discovered during the investigation of a failure reported in OWLS-100056. But this PR alone does not fully fix the problem, and I am still investigating an intermittent issue in the ItWlsSamples.testDomainHomeInPv test case.

PR #3207, opened 2022-06-24
Owls 100113 - Fix for intermittent nightly failure in ItMiiSampleFmwAux.testAIFmwInitialUseCase 

ankedia:
Owls 100113 - Fix for intermittent nightly failure in ItMiiSampleFmwAux.testAIFmwInitialUseCase. The issue is caused by a race between the check for "no processes" and the "kill existing processes" during cleanup causing exit value to be non-zero. This PR uses an updated trap command.

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
@@ -6,11 +6,13 @@ package oracle.weblogic.kubernetes.utils;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import oracle.weblogic.kubernetes.actions.impl.primitive.Command;
 import oracle.weblogic.kubernetes.actions.impl.primitive.CommandParams;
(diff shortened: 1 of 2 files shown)


rjeberhard (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/ItMiiSampleHelper.java
```diff
@@ -228,7 +230,7 @@ public void execTestScriptAndAssertSuccess(DomainType domainType,
 
         boolean success =
             result != null
-                && result.exitValue() == 0
+                && (result.exitValue() == 0 || cleaupFailed(result))
```
tbarnes-us, line 233: The script could fail due to other reasons than the kill in combination with a kill failure, so this check would assume success even if something other than the kill went wrong.

ankedia: The validation also checks that the script standard output contains string "Finished without errors". IMO, the chances of test script finishing with message "Finished without errors" and other reasons in combination with a kill failure is very rare if any. If the test fails again, we can evaluate and add other checks.

tbarnes-us (approved):

PR #3209, opened 2022-06-24
Have older samples retain using v8 and review changes to samples/scripts directory

rjeberhard:


robertpatrick (approved):

rjeberhard:
> The PR looks good to me. I'm bit worried that this change increases the samples dependency on the conversion webhook and chances of them running into some potential user errors due to multiple operators etc. I'm wondering if it's possible to change the samples or WDT to work with `V9` version instead, any thoughts?

We will need to update WDT to support v9, but I think that this should be after WKO 4.0 is released. 

The conversion webhook needs to work! :) 
I know that's not what you're saying... The "multiple operators" use case is very rare. We hit it a lot in our tests, of course, but I don't think that reflects what our customers will see. 

If you have ideas of how to make the webhook configuration more stable even in use cases with multiple operators than I'm very open.

ankedia:
> We will need to update WDT to support v9, but I think that this should be after WKO 4.0 is released.
[ANIL] ok, got it.

> The conversion webhook needs to work! :) I know that's not what you're saying... The "multiple operators" use case is very rare. We hit it a lot in our tests, of course, but I don't think that reflects what our customers will see.
[ANIL] Agreed. We have made all the efforts to make it stable with the multiple operators use-cases and I agree that customers likely won't see the kind of user errors that we encountered in our tests. 
I was also thinking that if they check-in the `V8` version of domain resource in their repo and if they edit or describe the domain then they will see a converted `V9` domain resource. I think it's ok to have this temporarily until after 4.0 is released.

ankedia (approved):

jshum2479 (approved):

PR #3210, opened 2022-06-27
OWLS-100056 Make sure sample script does not continue until it gets new pod name

doxiao:
This PR fixes an intermittent nightly failure in ItWlsSamples.testSampleDomainInPv.

The domain-in-pv sample uses a job to create a domain yaml. It gets the name of the job pod and uses kubectl cp command to get the generated domain yaml file from the pod, and to put an maker file (doneExtract) into the pod to tell the pod that it is done. There is a window between the job pod is created by the ApiServer and the sample script tries to get the job pod name. As a result, the POD_NAME variable could be empty. When this happens, the sample script is not able to out the file doneExtract into the pod, and the pod eventually times out and declares a failure because it does not find the file. 

The fix is to make sure that the script does not proceed until it gets the new pod name.


kubernetes/samples/scripts/create-weblogic-domain/domain-home-on-pv/update-domain.sh
```diff
@@ -209,7 +210,10 @@ updateDomainHome() {
   # a pod which will run a script that creates a domain. The script then will extract the domain
   # resource using  WDT's extractDomainResource tool. So, the following code loops until the
   # domain resource is created by exec'ing into the pod to look for the presence of domainCreate.yaml file.
-  POD_NAME=`kubectl get pods -n ${namespace} | grep ${JOB_NAME} | awk ' { print $1; } '`
+  while [ -z "${POD_NAME}" ] || [ "$POD_NAME" = "$OLD_POD_NAME" ]; do
+    POD_NAME=`kubectl get pods -n ${namespace} | grep ${JOB_NAME} | awk ' { print $1; } '`
+    echo "Pod name is $POD_NAME"
```
tbarnes-us, line 215: This loop has no timeout, where-as the next line of code waits a max of 600 seconds. So perhaps this loop should check if too much time has passed?  If you choose to address this, then note that you can get current number of seconds using the build-in "SECONDS" env var.

doxiao: hmm, I am not sure if we need to time out here. The next line is waiting for the pod to reach the ready state, which may not happen if there is an error. But this one is just to get the pod name that is created.

doxiao: added the timeout as suggested

rjeberhard (approved):

alai8 (approved):

tbarnes-us (approved):

PR #3212, opened 2022-06-27
owls-100118 - Dynamic cluster configured as configured cluster by introspectDomain.py

alai8:
The offline wlst cmo.getDynamicServers()  method failed and operator misinterpret this as mixed cluster. introspectDomain.py is updated to use ls('c') and get instead of using cmo functions.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11190/
known test failures:
- [oracle.weblogic.kubernetes.ItFmwSample.testFmwDomainInPv(String)[1]](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11190/testReport/junit/oracle.weblogic.kubernetes/ItFmwSample/testFmwDomainInPv_String__1_/)
- [oracle.weblogic.kubernetes.ItMiiDynamicUpdatePart1.testMiiRemoveTarget](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11190/testReport/junit/oracle.weblogic.kubernetes/ItMiiDynamicUpdatePart1/testMiiRemoveTarget/)

failure due to port conflict: `Service \"jrfdomain-inpv-wlst-admin-server-ext\" is invalid: spec.ports[1].nodePort: Invalid value: 32100: provided port is already allocated`. Test passed on rerun https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11219/ 
- [oracle.weblogic.kubernetes.ItFmwDomainInPVUsingWLST.testFmwDomainInPvUsingWlst](https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11190/testReport/junit/oracle.weblogic.kubernetes/ItFmwDomainInPVUsingWLST/testFmwDomainInPvUsingWlst/)


lennyphan (approved):

ankedia (approved):

alai8:
> LGTM. If feasible, it would be good to have a unit test for the behavior when cmo.getDynamicServers() method fails and use of ls('c') succeeds.

The current introspector unit testing scripts are under operator/integration-tests/introspector/introspectTest.sh. It not run during build process. It is also not trivial to create a cluster with <dyanmic-servers> element without a <name> that could make getDyanmciServers() to fail. The introspectTest.sh script currently create domain using offline WLST or MII, but we would need online WLST or hand-crafted config.xml for testing this change.

Instead, I will file a jira for adding an integration test that ensure that a new cluster added via online WLST can be started by the operator.

ankedia (approved):

rjeberhard:
> The current introspector unit testing scripts are under operator/integration-tests/introspector/introspectTest.sh. It not run during build process.

@alai8, it should be straightforward to add this to the Jenkinsfile so that these tests are run as a stage of the gating suite. Are these tests currently passing? How long do they take to run?

rjeberhard (approved):

alai8:
It takes 6.5 minutes when using the default option of using WLST to create domain. I haven't yet tried other configurations supported by the test - MII, ISTIO_ENABLED. Looping in @tbarnes-us who is most familiar with introspectTest.sh.

tbarnes-us:
> It takes 6.5 minutes when using the default option of using WLST to create domain. I haven't yet tried other configurations supported by the test - MII, ISTIO_ENABLED. Currently, there is a minor problem in utils_base.sh that I had to work around to get the test to run, but should be easy to fix.
> 
> Looping in @tbarnes-us who is most familiar with introspectTest.sh.

I think Johnny, and others, have expanded it over time.

The README has this:

```
----------
Usage:
----------

(1) Make sure a WebLogic image is available.  This test does not build, tag, or pull images to a local repo.
    One way to do this when running locally is to run the integration tests.
    (Note that this test doesn't depend on Operator image as of 2018/10/16, but may in the future.)
(2) Optionally specify values for input env vars (see introspectTest.sh for the list).
(3) For a non Model In Image domain type.
        Run introspectTest.sh

    For a Model In Image domain type.
        export DOMAIN_SOURCE_TYPE=FromModel
        Run introspectTest.sh

    To check for ISTIO
        export ISTIO_ENABELD=true - this only test for Non Model in Image

    To check for introspection of configured clusters with WebLogic 14.1.1.0 image.
    NOTE: Test for introspection of configured clusters only for Non Model in Image
          and for verifying OWLS 85530.

        export WEBLOGIC_IMAGE_TAG=14.1.1.0
        introspectTest.sh
```

I assume it'd be useful to at least run it twice - once in WLST mode (the default) and once in MII mode.

Not sure about the ISTIO mode, and I don't recall off the top of my head who added that.

(I am a little surprised that this test still passes given all of the changes that have gone into WKO over the last several months.)

PR #3226, opened 2022-06-29
OWLS-95596: Handle Cluster resource events and initiate makeRight flow

lennyphan:
This change handles Cluster Resource lifecycle events and initiates the makeRight flow which will scale WebLogic clusters.


rjeberhard (approved):
I don't quite understand all of your "todo" comments, especially the last few in the unit-tests, but I'm assuming that you do :)

PR #3232, opened 2022-06-30
modified terraform scripts to add force deletion of node pool

marinakog:
OKE team has introduced a new functionality where by deleting a pod it will cordon/drain the node before deleting it and will not force deletion by default unless specified.
Added setting to node pool to add force deletion. Also modified terraform tf properties for node pool to switch to the new format for node pool from deprecated 

https://build.weblogick8s.org:8443/job/wko-oke-test-new/28/


anpanigr (approved):

sankarpn (approved):
lgtm

PR #3237, opened 2022-07-05
OWLS-95596: Scale cluster through Operator REST API  by patching Cluster Resource

lennyphan:
This change allows scaling of clusters through the Operator REST Api by issuing patch commands, ,on the 'scale ' subresource, directly on the corresponding  Cluster Resource.  If Cluster Resource does not exist then it will look for corresponding Cluster spec in the Domain spec (until we're ready to delete the Cluster spec from the Domain spec).

### DIFF
--- a/common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
+++ b/common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
@@ -169,8 +169,6 @@ public class MessageKeys {
   public static final String POD_EVICTED_NO_RESTART = "WLSKO-0227";
   public static final String WATCH_CLUSTER = "WLSKO-0228";
   public static final String WATCH_CLUSTER_DELETED = "WLSKO-0229";
-  public static final String NULL_CLUSTER_NAME = "WLSKO-0230";
-  public static final String MATCHING_CLUSTER_NOT_FOUND = "WLSKO-0231";
 
   // domain status messages
   public static final String DUPLICATE_SERVER_NAME_FOUND = "WLSDO-0001";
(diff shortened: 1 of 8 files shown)


common/src/main/resources/Operator.properties
```diff
@@ -178,6 +178,8 @@ WLSKO-0226=Pod {0} was evicted due to {1}; validating domain
 WLSKO-0227=Pod {0} was evicted due to {1} but the operator is configured not to restart it.
 WLSKO-0228=Watch event triggered for WebLogic Cluster {0} in WebLogic Domain with UID: {1}.
 WLSKO-0229=Watch event triggered for deletion of WebLogic Cluster {0} in WebLogic Domain with UID: {1}.
+WLSKO-0230=Null Cluster name.
```
rjeberhard, line 181: I'm going to be curious about how this is used... I would expect clusters to have a name or to default to the `metadata.name` of the Cluster resource. Similarly, we can make fields required in the schema.

common/src/main/java/oracle/kubernetes/common/logging/MessageKeys.java
```diff
@@ -169,6 +169,8 @@ public class MessageKeys {
   public static final String POD_EVICTED_NO_RESTART = "WLSKO-0227";
   public static final String WATCH_CLUSTER = "WLSKO-0228";
   public static final String WATCH_CLUSTER_DELETED = "WLSKO-0229";
+  public static final String NULL_CLUSTER_NAME = "WLSKO-0230";
+  public static final String MATCHING_CLUSTER_NOT_FOUND = "WLSKO-0231";
```
rjeberhard, line 173: I don't see either of these used.

rjeberhard (approved):

PR #3238, opened 2022-07-06
Added support to run multiple jobs via one JenkinsFile

marinakog:
Added support to reuse same file for parallel and sequential jobs for oke runs: 
https://build.weblogick8s.org:8443/job/wko-oke-nightly-sequential/
https://build.weblogick8s.org:8443/job/wko-oke-nightly-parallel/

see results
https://build.weblogick8s.org:8443/job/wko-oke-nightly-seqential/1/
https://build.weblogick8s.org:8443/job/wko-oke-nightly-parallel/128


anpanigr (commented):
Since we are moving to jenkinfile do we still need the driver script oketest.sh

anpanigr (approved):

PR #3250, opened 2022-07-13
owls-100645 - FIx introspectTest.sh test

alai8:
Updated scripts and yaml files to catch up with recent changes to operator. `introspectTest.sh` now passes with domain on pv domain type, model in image domain type (`DOMAIN_SOURCE_TYPE=FromModel`), and with `ISTIO_ENABELD=true`.

Also updated operator/src/main/resources/scripts/utils.sh so it can source `utils_base.sh` from correct path when `utils.sh` is itself being source'd

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11439/


jshum2479 (commented):
LGTM, Thanks @alai8

tbarnes-us (approved):
LGTM Also!

PR #3252, opened 2022-07-13
Owls 100719 - Fix server shutdown with HTTP failures and changes to poll the server status after shutdown

ankedia:
Owls 100719 - Fix for HTTP request failures with status code 400 for shutdown. When shutting down the servers that are not part of the domain config, the shutdown step was getting executed twice for some servers. This resulted in duplicate HTTP calls that resulted in  "ServerLifecycleException: java.lang.IllegalStateException" error.

This PR also has changes to poll the server status after issuing the shutdown request using HTTP. It allows the pod termination to wait until the server has reached the shutdown state and the server process has exited. If the shutdown request fails after two attempts, we proceed to terminate the pod. Finally, it has changes to handle the HTTP timeout failures and perform a retry when the HTTP request times out.

Integration test run (in progress) - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11461


lennyphan (approved):
LGTM

rjeberhard (approved):

PR #3261, opened 2022-07-18
OWLS-101005: Add Cluster Status to Cluster Resources

lennyphan:
Add/update each Cluster resource with the same status that we record in the Domain status. The cluster status will look like the following:

Status:
  Cluster Name: cluster-1
  Maximum Replicas: 5
  Minimum Replicas: 0
  Ready Replicas: 1
  Replicas: 1
  Replicas Goal: 1


russgold (approved):
Well done.

rjeberhard (approved):

PR #3273, opened 2022-07-22
Russ owls 101001 part1

russgold:
Now automatically schedules retries when a severe failure occurs.

(also included an update to an integration test that was not cleaning up after itself)

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
@@ -3,6 +3,8 @@
 
 package oracle.kubernetes.operator;
 
+import java.time.Duration;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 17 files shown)


ankedia (approved):
LGTM

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -83,6 +84,9 @@ public class DomainProcessorImpl implements DomainProcessor, MakeRightExecutor {
   private static final String DELETED = "DELETED";
   private static final String ERROR = "ERROR";
 
+  @SuppressWarnings({"FieldCanBeLocal", "FieldMayBeFinal"})
+  private static String debugPrefix = null;  // Debugging: set this to a non-null value to dump the make-right steps
```
lennyphan, line 88: Was this for your own debugging?  Did we want to keep this for debugging since we'd have to rebuild the operator to use it?  Or did I miss a way to initialize at runtime?

lennyphan (approved):

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
russgold: This is intended to make it easy to enable debugging of fiber behavior. If you change the value to a string, it will display the steps actually executed for each fiber. It is primarily intended for diagnosing failing integration tests. I haven't yet found a use case where I'd want to enable it at runtime, but if you can think of a value in doing that, I'll make it a tuning parameter in the future.

rjeberhard (approved):

PR #3275, opened 2022-07-22
Fixes for the incorrect introspector messages from `traceDirs` caused by changes forward ported from 3.4.

ankedia:
I noticed some incorrect messages from `traceDirs` function in the introspector logs when printing the contents of certain directories. We forward ported changes from `release/3.4` branch and `traceDirs` function is bit different in 3.4 branch. This PR has changes to fix these log messages.

### DIFF
--- a/operator/src/main/resources/scripts/utils.sh
+++ b/operator/src/main/resources/scripts/utils.sh
@@ -702,7 +702,7 @@ checkAuxiliaryImage() {
   fi
 
   trace FINE "Auxiliary Image: AUXILIARY_IMAGE_MOUNT_PATH is '$AUXILIARY_IMAGE_MOUNT_PATH'."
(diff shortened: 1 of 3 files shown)


jshum2479 (approved):

operator/src/main/resources/scripts/utils.sh
```diff
@@ -702,7 +702,7 @@ checkAuxiliaryImage() {
   fi
 
   trace FINE "Auxiliary Image: AUXILIARY_IMAGE_MOUNT_PATH is '$AUXILIARY_IMAGE_MOUNT_PATH'."
-  traceDirs before $AUXILIARY_IMAGE_MOUNT_PATH
```
tbarnes-us, line 705: It looks like there's also an incorrect usage at line 759

PR #3276, opened 2022-07-22
Change image pull policy based on the environment

sankarpn:
When imagepullPolicy is set to Never and run in a multi node cluster the test will fail if a new pod is spin a different worker where the images are not present locally. This change sets the imagepullPolicy to Always in non-kind env and verifies the pods are rolled when imagepullPolicy.

- https://build.weblogick8s.org:8443/job/wko-okd-test/427/console


anpanigr (approved):
lgtm

maggiehe00 (approved):
LGTM

PR #3277, opened 2022-07-23
Automated test for "Failed condition is removed when domain resource is updated with a new error" fix

sankarpn:
Automated test for "Failed condition is removed when domain resource is updated with a new error" fix.

Configure a dynamic cluster with maxcluster size to 5

Configure domain/cluster level replica count to 6

All 5 managed servers come up and the domain has condition Available to true and Complete to empty.

Now update the domain resource with
1) changing the replicas from 6 to 2;
2) changing the domain credentials secret name to something does not exist.

The domain will have a Failed condition true. Then the Failed condition will be removed even though the wrong domain credentials secret name is still in the domain resource. Note that the Available and Complete conditions are both true.

A side note: not only the Failed condition for missing domain credentials secret was removed, the original error message about replicas too high was left (not cleared) in the domain status although the condition has been fixed by 2) above.

- https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11620/
- https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11621/


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItDiagnosticsFailedCondition.java
```diff
@@ -221,8 +223,40 @@ void testReplicasTooHigh() {
...
+          "patchDomainCustomResource failed");
+      testUntil(
+          domainStatusReasonMatches(domainName, domainNamespace, "DomainInvalid"),
+          getLogger(),
+          "waiting for domain status condition reason DomainInvalid exists"
+      );
+      testUntil(
+          domainStatusReasonMatches(domainName, domainNamespace, "ReplicasTooHigh"),
+          getLogger(),
+          "waiting for domain status condition reason ReplicasTooHigh exists"
```
xiancao, line 257: Why do the domain still have the ReplicasTooHigh status after you patch the domain replicas to 2?

sankarpn: We don't remove the prior status even after rectifying the issue. I guess it is kept as a history

xiancao (approved):
LGTM

anpanigr (approved):
lgtm

PR #3280, opened 2022-07-25
Handle recovery from aborted processing

russgold:
Ensure that processing stops when we get an ABORTED condition, and that changing certain parameters allows us to continue.


lennyphan (approved):
LGTM

rjeberhard (approved):

PR #3283, opened 2022-07-25
Owls 95882 part 1 - Fixes for the stop server script, changes to capture script output and make grace period dynamic

ankedia:
Owls 95882 part 1 - This PR has couple of fixes for the stop server script, (1) there was a right brace missing and (2) shutdown call was being made with unexpected argument properties. In addition, I have made changes to capture the script output in the exited container logs. Finally, it sets the pod deletion grace period seconds to 0 when the server has already shutdown.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11634/

### DIFF
--- a/operator/src/main/resources/scripts/stopServer.sh
+++ b/operator/src/main/resources/scripts/stopServer.sh
@@ -114,6 +114,6 @@ fi
 
 trace "Exit script"  &>> ${STOP_OUT_FILE}
 
-trace "stopServer.sh: Contents of ${STOP_OUT_FILE}:" > /proc/1/fd/1
+trace "Stop Server: === Contents of the script output file ${STOP_OUT_FILE} ===" > /proc/1/fd/1
 cat ${STOP_OUT_FILE} >> /proc/1/fd/1
-trace "stopServer.sh: End of ${STOP_OUT_FILE} contents." >> /proc/1/fd/1
+trace "Stop Server: === End of ${STOP_OUT_FILE} contents. ===" >> /proc/1/fd/1
--- a/operator/src/main/resources/scripts/stopServer.sh
+++ b/operator/src/main/resources/scripts/stopServer.sh
@@ -6,6 +6,8 @@
 #
 # This script is used to attempt to gracefully shutdown a WL Server
 # before its pod is deleted.  It requires the SERVER_NAME env var.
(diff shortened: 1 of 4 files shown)


operator/src/main/resources/scripts/stopServer.sh
```diff
@@ -113,3 +113,7 @@ if [ -f ${SERVER_PID_FILE} ]; then
 fi
 
 trace "Exit script"  &>> ${STOP_OUT_FILE}
+
+trace "Stop Server: === Contents of the script output file ${STOP_OUT_FILE} ===" > /proc/1/fd/1
```
tbarnes-us, line 117: What is the purpose of ` >> /proc/1/fd/1`?

I recommend at least adding a comment.

ankedia: K8S does not print any logs in PreStop stage. PreStop is part of lifecycle, also called as hook and generally hook logs will not print in kubectl logs. See issue [25766 ](https://github.com/kubernetes/kubernetes/issues/25766) for more details. The workaround is to print hook's output to the main process' stdout using `/proc/1/fd/1` as mentioned in the [last comment](https://github.com/kubernetes/kubernetes/issues/25766#issuecomment-752121781) in the above issue. I'll add the comment.

jshum2479: I see the script is always appending to ${STOP_OUT_FILE},  what if this file is on PV?  Will it be confusing?

ankedia: ok. I'm not sure why it's always appending to ${STOP_OUT_FILE}. Do you think we should instead overwrite the ${STOP_OUT_FILE} file for each run?

tbarnes-us: as you mentioned in the scrum, rotating it seems like a good idea - you can use the same utils helper function that we use for other log files...

ankedia: Changed in [aa5c19d](https://github.com/oracle/weblogic-kubernetes-operator/pull/3283/commits/aa5c19d141a225a9a69d5ac56af1e691c398b548).

jshum2479 (approved):

tbarnes-us (approved):

lennyphan (approved):

PR #3271, opened 2022-07-21
OWLS-100595 - handle SSL and use of certificates to connect from logstash container to ELK

alai8:
Allow easier configuration of logstash container in weblogic-operator pod for communicating with Elasticsearch that is set up to require SSL and user authentication:

- Added `elasticSearchProtocol` configuration value to operator helm-chart for building url for Elasticsearch in the default logstash config files
- Logstash settings file `logstash.yml` added to the `weblogic-operator-logstash-cm` ConfigMap to allow easier customization
- Certificates for communicating with secured Elasticsearch can be added to the logstash container through a secret named `logstash-certs-secret` in operator's namespace

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11584/

PR #3288, opened 2022-07-26
owls-99631 add unit tests to verify correct service created during scale up and scale down when preCreateService is enabled

alai8:
Also include minor doc edits from feedback from PR #3271


rosemarymarano (approved):
LGTM, thanks.

rjeberhard (approved):

PR #3289, opened 2022-07-26
Improve unit tests and code clarity

russgold:
My last PR had reported less than 100% code coverage, so I did some investigation and found places where we were not only lacking tests, but also had code whose purpose wasn't clear. This change attempts to remedy that.


ankedia (approved):
LGTM

lennyphan (approved):
LGTM

alai8 (approved):

PR #3300, opened 2022-08-02
owls-101034 - add support for Cluster custom resource to scalingAction.sh

alai8:
This is first phase of owls-101034 work that allow scalingAction.sh to support scaling with Cluster custom resource.

- The script reads replicas, minimunReplicas, and maximumReplicas from the Cluster resource if it is present.
- Support for cluster spec and status from Domain resource is still in place to prevent breaking existing tests. This support will be removed later after the domain CRD changes to spec.clusters are submitted.
- Script validates proposed new replica value versus minimumReplicas & maxiumumReplicas in Cluster status 
- Some shellcheck cleanups

Note: Upgrade behavior from 3.x Domain resources is not included in this change.

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11665/


operator/scripts/scaling/scalingAction.sh
```diff
@@ -144,14 +144,31 @@ get_custom_resource_domain() {
...
   if [ $? -ne 0 ]; then
     trace "Failed to retrieve WebLogic Domain Custom Resource Definition"
     exit 1
   fi
   echo "$DOMAIN"
 }
 
+# Retrieve Cluster Custom Resource
+# args:
+# $1 Cluster name (assumed to be same as name of the Cluster Resource)
```
tbarnes-us, line 157: Why can we assume the CR name is the same as the cluster name?

I assume the CR name would normally be decorated with the domain-uid, where-as the cluster name would not...

operator/scripts/scaling/scalingAction.sh
```diff
@@ -208,6 +225,30 @@ INPUT
...
+# args:
+# $1 Cluster Custom Resource
+get_replicas_from_cluster() {
+  local clusterJson="$1"
+  local replicas
+  if jq_available; then
+    local numReplicasCmd=".spec.replicas"
+    replicas=$(echo "${clusterJson}" | jq "${numReplicasCmd}" 2>> ${log_file_name} )
+  else
+cat > cmds-$$.py << INPUT
```
tbarnes-us, line 238: Perhaps I'm missing something, but it looks like this script leaves behind "cmds-$$.py" files after it runs.

operator/scripts/scaling/scalingAction.sh
alai8: @lennyphan can you help confirm if the Cluster resource name is <domain-uid>-<cluster-name>? Thanks

operator/scripts/scaling/scalingAction.sh
alai8: They will be cleaned up in line 658

```
# Cleanup cmds-$$.py
[ -e cmds-$$.py ] && rm cmds-$$.py

```

operator/scripts/scaling/scalingAction.sh
tbarnes-us: Not sure. @rjeberhard ?

alai8: Talked with Lenny. He suggested we have to search for a matching Cluster CR with matching domainUID and clusterName.

tbarnes-us: Makes sense to me.

rjeberhard: The `metadata.name` of the Cluster resource can be anything and doesn't have to match the cluster name. In the work I'm doing in the webhook to create Cluster resources as we upgrade v8 domains, the webhook will create Cluster resources that have a <domain-uid>-<cluster-name> decorated name as described.

However, since that's not required. The only foolproof way to find a Cluster resource for a given cluster is to see what Cluster resources are referenced from the Domain and then see which of these has the required cluster name.

alai8: Updated script to do what Ryan and Lenny suggested

lennyphan (approved):

rjeberhard (approved):

PR #3304, opened 2022-08-04
Enhance the MII sample pod wait script for 4.0

tbarnes-us:
Changes:
- Rely on a domain reaching its 'Complete' condition instead of requiring user to pass in the expected pod count
- Check that the domain generation and observedGeneration match
- Check that the domain api version is at least v9
- Fail fast by default if a 'Failed' condition is detected.

### DIFF
--- a/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/wl-pod-wait.sh
+++ b/kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/wl-pod-wait.sh
@@ -16,23 +16,31 @@ usage() {
   Usage:
 
     $(basename $0) [-n mynamespace] [-d mydomainuid] \\
-       [-p expected_pod_count] [-t timeout_secs] [-q]
+       [-t timeout_secs] [-q]
 
   Description:
 
-    This utility script exits successfully when the designated number of
-    WebLogic Server pods in the given WebLogic Kubernetes Operator domain
-    reach a 'ready' state and have 'restartVersion', 'introspectVersion',
-    'spec.image', and 'spec.serverPod.auxiliaryImages.image' values that match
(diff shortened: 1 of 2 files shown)


tbarnes-us:
NOTE: This is converted to 'draft' as there seem to be issues in the new 4.0 operator conditions that may be making this change unstable. I am trying to reproduce using the latest Operator version and will file Operator tasks if needed.

kubernetes/samples/scripts/create-weblogic-domain/model-in-image/utils/wl-pod-wait.sh
```diff
@@ -16,23 +16,31 @@ usage() {
   Usage:
 
     $(basename $0) [-n mynamespace] [-d mydomainuid] \\
-       [-p expected_pod_count] [-t timeout_secs] [-q]
+       [-t timeout_secs] [-q]
```
alai8, line 19: should we include [-p 0 | Completed] in usage?

tbarnes-us: good catch! fixed

alai8 (approved):

lennyphan (approved):

tbarnes-us:
Minor side comments about potential further minor improvements (lower priority - may be added as time permits):
- Simplify: Replace use of temporary files by populating internal results into an env var instead.
- Clarify: Update usage to clearly mention that the script checks generation vs. observedGeneration.
- Simplify: Create log helper function for replacing 'echo timestamp ...' style lines. 
- Fix: Usage should get defaults from default constants - it is currently using vars that might change when the command line is processed.

tbarnes-us:
Closing - moved to different branch/pull OWLS-88344-new

PR #3314, opened 2022-08-09
Test for WKO Retry Improvements: enhance the existing integration test cases

hzhao-github:
Jenkins job:

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11808
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11818
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11819
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/11820

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
@@ -357,8 +357,8 @@ public interface TestConstants {
   //retry improvement
   //Defaulting to 120 seconds
   public static final Long FAILURE_RETRY_INTERVAL_SECONDS =
-      Long.valueOf(getNonEmptySystemProperty("Failure.retry.interval.seconds", "60"));
+      Long.valueOf(getNonEmptySystemProperty("failure.retry.interval.seconds", "60"));
   //Defaulting to 1440 minutes (24 hours)
   public static final Long FAILURE_RETRY_LIMIT_MINUTES =
-      Long.valueOf(getNonEmptySystemProperty("Failure.retry.limit.minutes", "10"));
+      Long.valueOf(getNonEmptySystemProperty("failure.retry.limit.minutes", "10"));
 }
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
@@ -353,4 +353,12 @@ public interface TestConstants {
       + "config/samples/sidb/singleinstancedatabase.yaml";
   public static final String ORACLELINUX_TEST_VERSION =
       getNonEmptySystemProperty("wko.it.oraclelinux.test.version", "7");
+
+  //retry improvement
+  //Defaulting to 120 seconds
+  public static final Long FAILURE_RETRY_INTERVAL_SECONDS =
+      Long.valueOf(getNonEmptySystemProperty("Failure.retry.interval.seconds", "60"));
+  //Defaulting to 1440 minutes (24 hours)
+  public static final Long FAILURE_RETRY_LIMIT_MINUTES =
+      Long.valueOf(getNonEmptySystemProperty("Failure.retry.limit.minutes", "10"));
 }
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
```diff
@@ -353,4 +353,12 @@ public interface TestConstants {
       + "config/samples/sidb/singleinstancedatabase.yaml";
   public static final String ORACLELINUX_TEST_VERSION =
       getNonEmptySystemProperty("wko.it.oraclelinux.test.version", "7");
+
+  //retry improvement
+  //Defaulting to 120 seconds
+  public static final Long FAILURE_RETRY_INTERVAL_SECONDS =
+      Long.valueOf(getNonEmptySystemProperty("Failure.retry.interval.seconds", "60"));
```
anpanigr, line 360: Failure.retry.interval.seconds --> failure.retry.interval.seconds to maintain  the property naming convention.

Why are we setting a lower value (60s)  then default (120s)

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
```diff
@@ -353,4 +353,12 @@ public interface TestConstants {
...
+  //Defaulting to 1440 minutes (24 hours)
+  public static final Long FAILURE_RETRY_LIMIT_MINUTES =
+      Long.valueOf(getNonEmptySystemProperty("Failure.retry.limit.minutes", "10"));
```
anpanigr, line 363: Failure.retry.limit.minutes --> failure.retry.limit.minutes

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
```diff
@@ -353,4 +353,12 @@ public interface TestConstants {
       + "config/samples/sidb/singleinstancedatabase.yaml";
   public static final String ORACLELINUX_TEST_VERSION =
       getNonEmptySystemProperty("wko.it.oraclelinux.test.version", "7");
+
+  //retry improvement
```
anpanigr, line 357: Shouldn't we add a integration test to verify the runtime behavior of these new attributes.

hzhao-github: we have OWLS-101201 and OWLS-101206 to add integration tests to test detailed of the improvements. In this PR, we adding these two options that are used by a lot of test suites, important, the default values of the options are changed to a very small value to trigger any issues in case

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
hzhao-github: changed.

change from 120s  to 60s is just to lower the value to catch failures easily

integration-tests/src/test/java/oracle/weblogic/kubernetes/TestConstants.java
hzhao-github: changed

anpanigr (approved):

russgold:
I think I am missing something, here. I see you adding values to the domain, but no tests that will be affected by non-default values. What is the point of the changes?

I would expect tests that verify that retries happen when something is wrong, and then if the problem is corrected before they do, the failure is removed. I would expect tests that verify that if the problem is _not_ correct, the failure is not removed. I would expect tests that verify that if the problem is not corrected before the time limit, that the retries stop (and the domain reports an aborted condition), but that you can fix the problem and change the retry version to retry the domain.

hzhao-github:
Hi Russ, Pani had same question as you, here was my answer to him:

we have OWLS-101201 and OWLS-101206 to add integration tests to test detailed of the improvements. In this PR, we adding these two options that are used by a lot of test suites, important, the default values of the options are changed to a very small value to trigger any issues in case

russgold (approved):

PR #3328, opened 2022-08-15
Update startNodeManager.sh to use port test instead of grepping log message to determine if nodemanager is running

alai8:
This PR is for release/3.4 branch

jenkins (in progress) https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12035/


jshum2479 (approved):

jshum2479:
We should work with @anpanigr to get an integration test for the japanese locale,  just the simplest domain test will suffice.

tbarnes-us (approved):

PR #3330, opened 2022-08-15
Add retry information to domain status when reporting a failure

russgold:
This change updates the domain status message to include retry information when reporting a retriable failure. For example
```
WebLogicCredentials secret 'no-such-secret' not found in namespace 'namespace'. Will retry next at 2022-08-15T18:37:21-04:00 and approximately every 120 seconds afterward until 2022-08-16T18:35:21-04:00 if the failure is not resolved.
```


russgold:
I see 31 existing failures - I presume that the cluster changes broke the integration tests?

russgold:
> I see 31 existing failures - I presume that the cluster changes broke the integration tests?

No, it turns out to have been a JSON serialization issue. I have switched to a different approach to involve the domain in the retry message creation.

doxiao (approved):
LGTM.

PR #3336, opened 2022-08-18
OWLS-101610- Add support for V1ScaleSpec request payload for operator scale REST API

alai8:
Add support to operator `/scale` REST API with request payload that is same as that for Kubernetes `/scale` REST endpoint such as:

`{"spec":  {"replicas":  2}}`

Current request payload such as

`{"managedServerCount": 2}` 

is still supported, but is deprecated and will be removed in future releases.

jenkins: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12176/


rjeberhard (changes requested):
Functionally this change looks good. Two issues: 1) there are too many and easily fixed "code smells" reported by Sonar. We don't want to fix the "minor" issues about use of a deprecated API, but you can resolve the others by adding the requested Javadoc element and by removing the public modifier from JUnit tests, and 2) the code coverage is below our metric of 80%. Similarly, I think this will be pretty easy to hit with likely just one more test.

alai8:
Thanks. Somehow I overlooked the sonar report for this PR. The latest update should take care of most issues, except for the warnings on usage of deprecated field.

rjeberhard (approved):

PR #3344, opened 2022-08-22
Owls101364 fix part 2 - clear Complete:true before scaling

doxiao:
Fix a bug that causes an issue in ItKubernetesDomainEvents#testScaleDomainAndVerifyCompletedEvent, where the expected Completed event is not logged.

The event generation is tied to domain condition changes. In order for a completed event to be generated, the domain condition needs to change from Completed:false to Completed:true. The problem is that the condition Completed is not changed from true to false when the operator starts scaling up or down a cluster. As a result, the Completed event is not logged. 

The fix is to clear Completed:true (and patch the remote object) before operator starts to delete or start one or more servers for scaling down or up.


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
```diff
@@ -226,18 +226,15 @@ public static void initAll(@Namespaces(6) List<String> namespaces) {
   @Tag("gate")
   @Tag("crio")
   void testK8SEventsSuccess() {
-    try {
-      OffsetDateTime timestamp = now();
-      logger.info("Creating domain");
-      createDomain(domainNamespace1, domainUid, pvName1, pvcName1);
+    OffsetDateTime timestamp = now();
```
russgold, line 229: Why remove the finally block, here? I thought we kept those to prevent a failure in one test from breaking others.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
```diff
@@ -452,13 +450,10 @@ void testDomainK8sEventsScalePastMaxAndChangeIntrospectVersion() {
   @Test
```
russgold, line 450: Here again you removed the finally clause

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
```diff
@@ -972,6 +972,22 @@ void whenClusterScaleDownToReplicas1_fromReplicas2_server1And2And3WithAlwaysPoli
     assertThat(info.getClusterService(CLUSTER), notNullValue());
   }
 
+  @Test
+  void whenClusterScaleDownToReplicas1_fromReplicas2_reportIncompleteEvent()
```
russgold, line 976: suggest a better name might be:
```
whenScalingDown_reportIncompleteEvent
```

It is actually relevant to the functionality that we are going from 2 to 1, rather than some other change?

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
```diff
@@ -244,6 +246,15 @@ void whenNoServerStartRequested_dontStartServers() {
     assertManagedServersUpStepNotCreated();
   }
 
+  @Test
+  void whenNoServerStartRequested_dontPrependUpdateCompleteConditionStep() {
```
russgold, line 250: Is there a way to test the effect, here, rather than the actual steps created? That is, does the step in question create events or status updates that could be checked directly?

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
```diff
@@ -478,7 +489,15 @@ void withStartPolicyAlways_addNonManagedServers() {
   void whenShuttingDownAtLeastOneServer_prependServerDownIteratorStep() {
     addServer(info, "server1");
 
-    assertThat(firstNonEventStep(createNextStep()), instanceOf(ServerDownIteratorStep.class));
+    assertThat(firstNonEventAndNonUpdateCompleteConditionStep(createNextStep()),
```
russgold, line 492: Again, try to test the behavioral change rather than the steps created to effect that behavior.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -1326,6 +1326,27 @@ void modifyStatus(DomainStatus status) {
     }
   }
 
+  public static class UpdateCompletedConditionSteps extends StatusUpdateStep {
```
russgold, line 1329: If this always sets completed/false, a name that indicates that would be clearer than one which just notes that the step _updates_ the condition.

russgold (changes requested):
Looks as though it does what it needs to be, but could be easier to understand to facilitate future changes.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
doxiao: The shutdown of the domains at the end of a test makes it hard to see the domain's status at the failure. We need to make the test cases independent from each other in a different way, IMO.

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
doxiao: I followed the naming patterns in the other similar test cases. Yes, I can change it to the suggested name.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -74,9 +75,28 @@ private static Step scaleDownIfNecessary(
       insert(steps, new ServerDownIteratorStep(factory.shutdownInfos, null));
     }
 
+    if (hasWorkToDo(steps, next)) {
```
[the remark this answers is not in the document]
doxiao: It means that we need to scale up or scale down.

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
doxiao: This follows the pattern in other test cases in this class.

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
doxiao: what about ClearCompletedConditionStep?

operator/src/test/java/oracle/kubernetes/operator/DomainProcessorTest.java
```diff
@@ -972,6 +972,22 @@ void whenClusterScaleDownToReplicas1_fromReplicas2_server1And2And3WithAlwaysPoli
...
+    establishPreviousIntrospection(null, Arrays.asList(1, 2, 3));
+
+    // now scale down the cluster
+    domainConfigurator.configureCluster(CLUSTER).withReplicas(1);
+
+    DomainPresenceInfo info = new DomainPresenceInfo(newDomain);
+    processor.createMakeRightOperation(info).execute();
+    logRecords.clear();
+
+    assertThat(getEvents().stream().anyMatch(EventTestUtils::isDomainIncompleteEvent), is(true));
```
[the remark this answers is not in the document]
doxiao: will do.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
doxiao: In addition, the test cases that have been changed in this PR are testing domains in different namespaces; in theory, these test cases should not impact each other.

doxiao: @sankarpn can you comment on this? thanks.

russgold (approved):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
sankarpn: I am ok with removing the finally clause.

sankarpn (approved):
lgtm

rjeberhard:
Last test run failed... I've restarted and will merge if new run passes.

PR #3347, opened 2022-08-23
Update ItMiiSample* test classes to use v9 cluster resource

maggiehe00:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12468/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12469/

For all 4 test classes ItMiiSampleFmwMain ItMiiSampleFmwAux, ItMiiSampleWlsMain, ItMiiSampleWlsAux, except initial test case (verifying that the domain with 2 managed server starts up successfully ) all the other update cases involving domain rolling restart failed with the same reason as [OWLS-101782](https://jira.oraclecorp.com/jira/browse/OWLS-101782)


operator/integration-tests/model-in-image/mii-sample-wrapper/mii-domain.yaml.template-JRF-AI
```diff
@@ -167,3 +150,18 @@ spec:
 
       # Name of secret with walletFile containing base64 encoded opss wallet, used for JRF domains
       #walletFileSecret: @@DOMAIN_UID@@-opss-walletfile-secret
+
+---
+
+apiVersion: "weblogic.oracle/v1"
+kind: Cluster
+metadata:
+  name: cluster-1
```
tbarnes-us, line 159: The name should be decorated with the DOMAIN_UID ( @@DOMAIN_UID@@-cluster-1 ). The Mii sample has a step that concurrently runs two domains.

tbarnes-us: (We should use a similar pattern for _all_ samples that define a Cluster - not just MII.)

tbarnes-us:
Two general comments:
- The templates are used to generate files that then should be added to git in the corresponding locations under ./kubernetes/samples/scripts/create-weblogic-domain/model-in-image/domain-resources (the integration tests are expected to detect the discrepancy and fail when this happens).
- The MII sample documentation contains copies of the generated domain YAML. The corresponding locations should be replace their text with the new YAML. `grep "kind.*Domain" *.md` in the documentation/staging/content/samples/domains/model-in-image .

rjeberhard (approved):

PR #3354, opened 2022-08-24
Owls101872 report error msg when reject a change to a cluster resource

doxiao:
Fix a bug in validating webhook so that when a change to a cluster resource is rejected, the actual error message is reported.


alai8 (approved):
LGTM.
The messages WLSWH-0015 and WLSWH-0016 seem very similar. Do you know why they were not combined into a same message?

doxiao:
> LGTM. The messages WLSWH-0015 and WLSWH-0016 seem very similar. Do you know why they were not combined into a same message?
Thanks for reviewing the changes. WLSWH-0015 is used to reject a request and therefore the message text contains the wording. WLSWH-0016, combined with WLSWH-0017, is used to generate a warning message when other things are also changed that will cause an re-introspection, the result of which may honor the change in the request.

PR #3359, opened 2022-08-25
Copy situational configuration using a temporary file and mv to improve concurrency

rjeberhard:
The idea is to copy the source file to a temporary file in the same directory as the target location and then use `mv` to update the target location atomically.

### DIFF
--- a/operator/src/main/resources/scripts/livenessProbe.sh
+++ b/operator/src/main/resources/scripts/livenessProbe.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2017, 2021, Oracle and/or its affiliates.
+# Copyright (c) 2017, 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 # Kubernetes periodically calls this liveness probe script to determine whether
@@ -35,9 +35,9 @@ if [ ! "${DYNAMIC_CONFIG_OVERRIDE:-notset}" = notset ]; then
       tgt_file=$tgt_dir/$tgt_file             # add back in tgt dir path
       [ -f "$tgt_file" ] && [ -z "$(diff $local_fname $tgt_file 2>&1)" ] && continue  # nothing changed
       trace "Copying file '$local_fname' to '$tgt_file'."
-      cp $local_fname $tgt_file # TBD ignore any error?
+      copyIfChanged $local_fname $tgt_file
       if [ -O "$tgt_file" ]; then
-        chmod 770 $tgt_file # TBD ignore any error?
+        chmod 770 $tgt_file
       fi
     done
     for local_fname in ${tgt_dir}/*.xml ; do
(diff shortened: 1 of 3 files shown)


operator/src/main/resources/scripts/utils.sh
```diff
@@ -17,6 +17,27 @@ source ${SCRIPTPATH}/utils_base.sh
...
+# Define helper fn to copy a file only if src & tgt differ
+#
+
+function copyIfChanged() {
+  [ ! -f "${1?}" ] && trace SEVERE "File '$1' not found." && exit 1
+  if [ ! -f "${2?}" ] || [ ! -z "`diff $1 $2 2>&1`" ]; then
+    trace "Copying '$1' to '$2'."
+    tmp_file=$(mktemp -p $(dirname $2))
+    cp $1 $tmp_file
+    mv $tmp_file $2
```
tbarnes-us, line 30: Copying to a tmp file and then moving the file is a little subtle. Perhaps add a comment about why we're doing this?

jshum2479 (approved):

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -35,9 +35,9 @@ if [ ! "${DYNAMIC_CONFIG_OVERRIDE:-notset}" = notset ]; then
       tgt_file=$tgt_dir/$tgt_file             # add back in tgt dir path
```
tbarnes-us, line 35: Minor comment:

The 'copyIfChanged' looks like it already contains the diff/continue logic on lines 36 & the chmod logic on lines 39,40, and 41.  I think you can you go ahead and delete these lines.

On the other hand, copyIfChanged is a little talkative - it traces even when it decides to skip the copy.  This might be a little distracting since the liveness probe is frequently invoked.

operator/src/main/resources/scripts/livenessProbe.sh
```diff
@@ -35,9 +35,9 @@ if [ ! "${DYNAMIC_CONFIG_OVERRIDE:-notset}" = notset ]; then
...
       [ -f "$tgt_file" ] && [ -z "$(diff $local_fname $tgt_file 2>&1)" ] && continue  # nothing changed
       trace "Copying file '$local_fname' to '$tgt_file'."
-      cp $local_fname $tgt_file # TBD ignore any error?
+      copyIfChanged $local_fname $tgt_file
```
tbarnes-us, line 38: Minor comment:  "copyIfChanged" exits with SEVERE if the copy or chmod fails. My assumption is that's the original reason why the function wasn't originally used here. Is this something we want to happen in the livenessprobe? IIRC, its copy is supposed to be more of a 'best effort'.

rjeberhard:
Tests of this functionality pass: https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12448/

operator/src/main/resources/scripts/livenessProbe.sh
rjeberhard: I think I'm okay with the liveness probe failing and the server pod restarting when this copy fails. The other choice is that the server continues running with the old configuration for an unknown amount of time.

tbarnes-us: Fine by me.

PR #3363, opened 2022-08-26
v9 changes, check operator log instead of introspector log

vanajamukkara:
Jenkins results - 
latest results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12478/

Failures are due to https://jira.oraclecorp.com/jira/browse/OWLS-101921

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
@@ -29,6 +29,7 @@ import io.kubernetes.client.openapi.models.V1VolumeMount;
 import oracle.weblogic.domain.AdminServer;
 import oracle.weblogic.domain.AdminService;
 import oracle.weblogic.domain.Channel;
+import oracle.weblogic.domain.ClusterResource;
 import oracle.weblogic.domain.Configuration;
 import oracle.weblogic.domain.DomainResource;
 import oracle.weblogic.domain.DomainSpec;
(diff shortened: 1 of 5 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
```diff
@@ -200,19 +200,23 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
...
-    // wait for the domain to exist
-    logger.info("Check for domain custom resource in namespace {0}", domainNamespace);
-    testUntil(
-        domainExists(domainUid, DOMAIN_VERSION, domainNamespace),
-        logger,
-        "domain {0} to be created in namespace {1}",
-        domainUid,
-        domainNamespace);
+    // set cluster references
+    domain.getSpec().withCluster(new V1LocalObjectReference().name(clusterName));
```
anpanigr, line 217: Can we use the utility utils.ClusterUtils.addClusterToDomain

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/PodUtils.java
```diff
@@ -439,4 +439,19 @@ public static Callable<Boolean> checkPodEvictedStatusInOperatorLogs(String opNam
...
+   * @param matchStr text to be searched in the log
+   * @param podName the name of the pod
+   * @param namespace namespace where pod exists
+   * @return true if the text exists in the log otherwise false
+   * @throws ApiException if Kubernetes client API call fails
+   */
+  public static boolean checkPodLogContains(String matchStr, String podName, String namespace)
+      throws ApiException {
+
+    return Kubernetes.getPodLog(podName,namespace,null).contains(matchStr);
```
tbarnes-us, line 454: It may not matter, but just in case it wasn't considered, note that a long running operator could have 'matchStr' from a previous run of the test. This will make a future run of the test pass even when it doesn't generate the expected message. To mitigate, any log messages that were generated from _before_ the test is run should be ignored.

vanajamukkara: you are right. In the test where I am using this method, there is only one test method in the class which causes the expected error message.

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItMiiUpdateDomainConfig.java
vanajamukkara: done

anpanigr (approved):

tbarnes-us (approved):

PR #3377, opened 2022-09-01
OWLS-101923 fix integration test code in cluster-reference

doxiao:
The test performs the following:
1. modify the WebLogic domain to increase the cluster size; 
2. patch the domain resource to have a new introspectVersion;
3. patch the cluster resource to increase its replica count.

Step#3 fails because the operator has not completed the re-introspection caused by step #2, and therefore, the new cluster size is not reflected in the cluster's status. 

The fix is to wait for the cluster's status is updated before proceeding to step#3.

The failing test case passed with the changes (https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12590/ ).


sankarpn (approved):
lgtm

vanajamukkara (commented):
LG

PR #3391, opened 2022-09-08
OWLS-102288 fix test code

doxiao:
Modify integration test case to cover what is intended to test.

The test case does the following:
- patch cluster resource above max of replicas, which failed as expected.
- patch domain with new introspect number
- check domain event to be generated for replicas too high.

Because step #1 failed, the change to the cluster's replicas does not even get to the operator, and therefore, the operator does not know about it, not to mention generating a Failed condition (and events).

In order for an invalid replicas to be accepted and reach to the operator, the cluster resource itself should not have a replicas set, and the domain resource's replicas and introspectVersion need to be changed in a single patch operation.

There are two approaches to make sure that the cluster resource does not have replicas:
1) add a new cluster resource that does not have the replicas set. You need also make sure that the cluster is configured on the WebLogic domain side. 
2) remove the replicas setting from the existing cluster resources. 

This PR contains the integration test code change with approach #2. The failing test case passed with this branch. https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12742/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
@@ -68,6 +68,7 @@ import static oracle.weblogic.kubernetes.assertions.TestAssertions.verifyRolling
 import static oracle.weblogic.kubernetes.utils.ClusterUtils.addClusterToDomain;
 import static oracle.weblogic.kubernetes.utils.ClusterUtils.createClusterAndVerify;
 import static oracle.weblogic.kubernetes.utils.ClusterUtils.createClusterResource;
+import static oracle.weblogic.kubernetes.utils.ClusterUtils.removeReplicasSettingAndVerify;
 import static oracle.weblogic.kubernetes.utils.ClusterUtils.scaleCluster;
 import static oracle.weblogic.kubernetes.utils.CommonTestUtils.checkPodReadyAndServiceExists;
 import static oracle.weblogic.kubernetes.utils.CommonTestUtils.checkServiceExists;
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
```diff
@@ -413,17 +414,20 @@ void testDomainK8sEventsScalePastMaxWithoutChangingIntrospectVersion() {
...
+          managedServerPodNamePrefix);
+
       String introspectVersion = assertDoesNotThrow(() -> getNextIntrospectVersion(domainUid, domainNamespace3));
-      assertFalse(scaleCluster(cluster1Name, domainNamespace3, 3), "failed to scale cluster via patching");
       String patchStr
           = "["
-          + "{\"op\": \"replace\", \"path\": \"/spec/introspectVersion\", \"value\": \"" + introspectVersion + "\"}"
+          + "{\"op\": \"replace\", \"path\": \"/spec/introspectVersion\", \"value\": \"" + introspectVersion + "\"},"
+          + "{\"op\": \"replace\", \"path\": \"/spec/replicas\", \"value\": 3}"
           + "]";
```
marinakog, line 425: pass value from introspectVersion instead of 3

marinakog (changes requested):

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItKubernetesDomainEvents.java
doxiao: Hmmm, the value 3 is the domain's new replica count. The introspectVersion value is from the variable introspectVersion. Or I might have not understood your comment correctly.

marinakog (approved):

PR #3401, opened 2022-09-09
Fix for testCompleteAvailableConditionWithReplicaExceedMaxSizeAndIntrospectVersionChanged

sankarpn:
The fix is to patch the domain resource with higher than max cluster size and verify the domain status conditions after removing cluster replicas from cluster resource. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12813/


xiancao (approved):
LGTM

PR #3403, opened 2022-09-09
OWLS102311 Handle fatal validation errors

doxiao:
All domain validation errors of DOMAIN_INVALID type are currently retried until the maximum retry time limit is reached. In fact, some of those errors cannot be fixed without user modifying the domain resources, and therefore need not to be retried. 

With the changes in this PR, the operator detects those errors, treats those as Fatal errors and do not retry.

Full integration run does not show additional failures compared with what is in the nightly run with main branch except for a failure in beforeAll of ItDBOperator, which is obviously not related to the changes in this branch.

Full run (5 failures): https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12829
Nightly (10 failures): https://build.weblogick8s.org:8443/job/wko-kind-nightly-parallel/1037


rjeberhard (approved):

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainValidationStepTest.java
```diff
@@ -25,10 +26,12 @@
 import oracle.kubernetes.utils.TestUtils;
 import oracle.kubernetes.weblogic.domain.model.Configuration;
 import oracle.kubernetes.weblogic.domain.model.DomainCondition;
+import oracle.kubernetes.weblogic.domain.model.DomainFailureSeverity;
 import oracle.kubernetes.weblogic.domain.model.DomainResource;
 import oracle.kubernetes.weblogic.domain.model.DomainStatus;
 import oracle.kubernetes.weblogic.domain.model.ManagedServer;
 import oracle.kubernetes.weblogic.domain.model.Model;
+import org.jetbrains.annotations.NotNull;
```
russgold, line 34: We've been using javax.annotation.NonNull, rather than a Jetbrains annotation.

russgold:
This is a big change from what I remember us agreeing. What domain validation failures are _not_ fatal, now? And why would topology mismatch failures likewise all be fatal? Our previous agreement was that almost nothing was fatal, allowing users to update resources and wait, without telling the operator to restart.

Previously, we had defined exactly two cases as fatal:

1. A inspection failure that required us to access a particular very limited external resource, and
2. A failure to recover within the timeout.

What is the new rule?

operator/src/test/java/oracle/kubernetes/operator/helpers/DomainValidationStepTest.java
doxiao: Please note that this change is in the unit test code. I don't see use of javax.annotation.NotNull in unit test code.

Here are some examples in the other unit test classes that use org.jetbrains.annotations.NotNull.
helpers/RollingHelperTest.java:import org.jetbrains.annotations.NotNull;
helpers/TopologyValidationStepTest.java:import org.jetbrains.annotations.NotNull;
makeright/FailureReportingTest.java:import org.jetbrains.annotations.NotNull;
steps/InitializeWebhookIdentityStepTest.java:import org.jetbrains.annotations.NotNull;

doxiao:
> This is a big change from what I remember us agreeing. What domain validation failures are _not_ fatal, now? And why would topology mismatch failures likewise all be fatal? Our previous agreement was that almost nothing was fatal, allowing users to update resources and wait, without telling the operator to restart.
> 
> Previously, we had defined exactly two cases as fatal:
> 
>     1. A inspection failure that required us to access a particular very limited external resource, and
> 
>     2. A failure to recover within the timeout.
> 
> 
> What is the new rule?

OWLS-102311 description contains the reason behind this change. The behavior is what is in customer's requirements and is therefore specified in the one pager (the task description has the links).

(Note: discussions are going on in slack).

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainResource.java
```diff
@@ -893,7 +893,7 @@ class Validator {
     private final Set<String> serverNames = new HashSet<>();
 
     List<String> getValidationFailures(KubernetesResourceLookup kubernetesResources) {
-      getSimpleValidationFailures();
+      getFatalValidationFailures();
```
russgold, line 896: I hadn't noticed this before. We have two methods that are populating the same collection, and we count on the fact that when we call `getFatalValidationFailures`, no other code has populated the `failures` collection. This seems like a problem waiting to happen. It would be much cleaner to have both `getFatalValidationFailures' and `getCrossReferenceValidationFailures` create collections, which they pass to the various validation methods, and then return that populated collection. 

In the short term, if we don't want to make that change, at the very least, we should rename both methods so that is clear that they simply populate the failure list, and create a separate `getFatalValidationFailures'  which calls the first one and returns the result.

russgold (changes requested):
I've left some picky comments, that I would like to see addressed; otherwise, this looks fine.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainResource.java
doxiao: Good catch. working on a fix.

doxiao: Having looked closer, I don't think this is a problem.  Each method constructs a new Validator object, and therefore, populates different failures collections.

  public List<String> getFatalValidationFailures() {
    return new Validator().getFatalValidationFailures();
  }

  // used by the operator
  public List<String> getValidationFailures(KubernetesResourceLookup kubernetesResources) {
    return new Validator().getValidationFailures(kubernetesResources);
  }

russgold (approved):

PR #3408, opened 2022-09-13
Owls 101945 - Handle a case where introspector container logs can't be retrieved 

ankedia:
OWLS-101945 - Fix for the intermittent ItMiiAuxiliaryImage.testCreateDomainWithConfigMapAndEmptryModelFileDir integration test failure. The failure is caused because the introspector container logs can't be retrieved after the config-map and secret volumes fail to mount in the given timeout period. This causes the introspector container initialization failure and script fails with exit code 1. The operator then tries to read the introspector pod logs and it gets "unable to retrieve container logs for container" result. This change handles this particular case where the introspector pod logs contain "unable to retrieve container logs for container" result.

Integration test run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12907/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -78,6 +78,8 @@ public class JobHelper {
   private static final LoggingFacade LOGGER = LoggingFactory.getLogger("Operator", "Operator");
   public static final String INTROSPECTOR_LOG_PREFIX = "Introspector Job Log: ";
   private static final String EOL_PATTERN = "\\r?\\n";
+  public static final String UNABLE_TO_RETRIEVE_CONTAINER_LOGS_FOR_CONTAINER =
+      "unable to retrieve container logs for container";
 
   private JobHelper() {
   }
@@ -539,6 +541,7 @@ public class JobHelper {
 
       private void processIntrospectionResult(Packet packet, String result) {
         LOGGER.fine("+++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
+        LOGGER.info("DEBUG: +++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
         convertJobLogsToOperatorLogs(result);
         packet.put(ProcessingConstants.DOMAIN_INTROSPECTOR_LOG_RESULT, result);
         MakeRightDomainOperation.recordInspection(packet);
@@ -586,6 +589,9 @@ public class JobHelper {
           if (line.startsWith("@[")) {
             logToOperator();
             logMessage = new StringBuilder(INTROSPECTOR_LOG_PREFIX).append(line.trim());
+          } else if (line.startsWith(UNABLE_TO_RETRIEVE_CONTAINER_LOGS_FOR_CONTAINER)) {
+            logToOperator();
+            logMessage = new StringBuilder(INTROSPECTOR_LOG_PREFIX).append("[SEVERE] " + line.trim());
           } else if (logMessage.length() > 0) {
             logMessage.append(System.lineSeparator()).append(line.trim());
           }
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
@@ -541,7 +541,6 @@ public class JobHelper {
 
       private void processIntrospectionResult(Packet packet, String result) {
         LOGGER.fine("+++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
-        LOGGER.info("DEBUG: +++++ ReadDomainIntrospectorPodLogResponseStep: \n" + result);
         convertJobLogsToOperatorLogs(result);
         packet.put(ProcessingConstants.DOMAIN_INTROSPECTOR_LOG_RESULT, result);
         MakeRightDomainOperation.recordInspection(packet);
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
@@ -109,6 +109,7 @@ import static oracle.kubernetes.operator.ProcessingConstants.JOB_POD_INTROSPECT_
 import static oracle.kubernetes.operator.ProcessingConstants.JOB_POD_INTROSPECT_CONTAINER_TERMINATED_MARKER;
 import static oracle.kubernetes.operator.ProcessingConstants.JOB_POD_NAME;
 import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_FAILED;
+import static oracle.kubernetes.operator.helpers.JobHelper.UNABLE_TO_RETRIEVE_CONTAINER_LOGS_FOR_CONTAINER;
 import static oracle.kubernetes.operator.helpers.KubernetesTestSupport.JOB;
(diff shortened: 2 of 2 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -581,11 +583,16 @@ private OffsetDateTime getJobCreationTime(V1Job domainIntrospectorJob) {
...
       //    of the previous log message
       //  - ignores all lines in the log up to the first line that starts with '@['
+      //  - The case where the introspector job log message starts with
+      //    "unable to retrieve container log" is treated as a SEVERE error.
       private void convertJobLogsToOperatorLogs(String jobLogs) {
         for (String line : jobLogs.split(EOL_PATTERN)) {
           if (line.startsWith("@[")) {
             logToOperator();
             logMessage = new StringBuilder(INTROSPECTOR_LOG_PREFIX).append(line.trim());
+          } else if (line.startsWith(UNABLE_TO_RETRIEVE_CONTAINER_LOGS_FOR_CONTAINER)) {
```
rjeberhard, line 593: This seems brittle. I don't like searching for a specific English-language phrase. Is there any status, e.g. at least a camel case enum value or similar that we can use instead?

ankedia: No. Unfortunately job fails with generic "BackoffLimitExceeded" status and job pod container status is "Terminated" with exit code "1" and reason is "Error". The API call to get the pod logs returns with status code 200 and log result is "unable to retrieve container logs for container". I tried cleaning up and reintrospecting when the job pod container terminates with exit code "1" and reason is "Error" but it was running into some other unexpected issue. I'll take a look one more time.

ankedia: I looked again and there's no enum value or status that can be used instead of log results. I have started a slack thread to discuss and see if there's any ideas on how to handle this case.

ankedia: As discussed, I have reworked the fix to look for presense of "Domain introspection complete" message in the introspector job logs. The new integration test results are at - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12976/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12977/

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -523,12 +525,23 @@ public NextAction onSuccess(Packet packet, CallResponse<String> callResponse) {
...
           return doNext(createRemoveSelectedFailuresStep(getNext(), INTROSPECTION), packet);
         } else {
           return handleFailure(packet, domainIntrospectorJob);
         }
       }
 
+      @NotNull
+      private Boolean isDomainIntrospectionComplete(CallResponse<String> callResponse) {
+        return Optional.ofNullable(callResponse).map(CallResponse::getResult)
+            .map(r -> r.contains(DOMAIN_INTROSPECTION_COMPLETE)).orElse(false);
```
tbarnes-us, line 542: The change looks good to me, except please also consider:
- Minimally adding a comment near the key line in the introspector script, something like: "Important: The following trace's wording must not be changed without a corresponding change to DOMAIN_INTROSPECTION_COMPLETE in the operator."
- Perhaps also adding a unit test that verifies the phrase exists in the introspector script file.

ankedia: Added comment and validation in [6799942](https://github.com/oracle/weblogic-kubernetes-operator/pull/3408/commits/679994206d7619de43c35783ddc8f2e2914d7b82) and [8611f71](https://github.com/oracle/weblogic-kubernetes-operator/pull/3408/commits/8611f71526c5777d0fd300a7e637c552a5f52d96).

validateIntrospectorScript.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+#
+found_error=0
+
+INTROSPECT_DOMAIN_SCRIPT_FILE=$(find "$(pwd -P)" -type f -name "introspectDomain.sh" | tail -n 1)
```
tbarnes-us, line 8: Instead of 'find' can we hard code the exact file location? With ^^^ more than one occurrence of introspectDomain.sh could throw off the check...

validateIntrospectorScript.sh
```diff
@@ -0,0 +1,20 @@
...
+DOMAIN_INTROSPECTION_COMPLETE_STRING=$(grep -Er "DOMAIN_INTROSPECTION_COMPLETE" * | grep ProcessingConstants.java | grep -v html | sed 's/.*= //g' | sed 's/;$//g')
```
tbarnes-us, line 9: (1) Can this also use a hard coded exact location?

(2) Also, it should have a sanity check that the returned value isn't empty. (Or perhaps better - simply check taht the returned value matches the expected "Domain introspection complete." value.)

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
tbarnes-us: Thanks

validateIntrospectorScript.sh
ankedia: I added a sanity check to make sure that the returned value isn't empty. I was thinking that someone can change the "Domain introspection complete."  string in both ProcessingConstants.java and introspectDomain.java  at the same time and the check shouldn't fail in this case.

validateIntrospectorScript.sh
ankedia: I have check to look for only the last occurrence (`tail -n 1`).

validateIntrospectorScript.sh
tbarnes-us: It's simpler and a little more robust IMO to simply check each of three exact locations (no "find") for the exact expected string. If someone needs to change the string, they can change this test file in addition to the three locations...

ankedia: Changed in [4fd5c6f](https://github.com/oracle/weblogic-kubernetes-operator/pull/3408/commits/4fd5c6f8faa8ddd5711b43372a6c1abdb8a6222f).

validateIntrospectorScript.sh
```diff
@@ -0,0 +1,22 @@
...
+do
+  if [[ ! `grep "Domain introspection complete" "${fileToCheck}"` ]]; then
+    printf "Failed to find the expected string 'Domain introspection complete' in '"${fileToCheck}"' file.\n"
+    exit 1
+  fi
+done
+
+found_error=$?
+
+exit $found_error
```
tbarnes-us, line 22: minor: i think this can be now just 'exit 0' -- it looks found_error will always be 0, so found_error can be removed...

tbarnes-us (approved):

validateIntrospectorScript.sh
tbarnes-us: LGTM

validateIntrospectorScript.sh
ankedia: Changed in [f1b32d4](https://github.com/oracle/weblogic-kubernetes-operator/pull/3408/commits/f1b32d424f5616d6361958594227947e979c2fea).

jshum2479 (approved):

PR #3356, opened 2022-08-24
Use delete-specific response step

russgold:
This addresses an issue found by Dongbo, in which running the WaitForPodDeletedStep  can report a Kubernetes error when it notes that the pod has been deleted by getting a 404.

PR #3412, opened 2022-09-14
Owls 102429 - Fix for unexpected server restart after roll completes

ankedia:
Owls 102429 - Fix for unexpected server restart after roll completes
Before the [change ](https://github.com/oracle/weblogic-kubernetes-operator/commit/68aea63e3da07d9ed1a622fc44e9db933aa5c513) in PR [3356](https://github.com/oracle/weblogic-kubernetes-operator/pull/3356), the `onSuccess` method of `resumeIfReady` step returned null action to terminate the child fiber. According to my understanding, `resumeIfReady` is the response step for the step that we create to check the latest resource value (backstop check for missed watch events) and that step is executed in a new child fiber that is created in `checkUpdatedResource` method. If the resource is ready, then the child fiber should terminate with a `null` action. I have also made a change to the `ResponseStep` to not log the failure status when response status code is 404. 
Integration test runs - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12934/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12933/

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
@@ -25,6 +25,7 @@ import oracle.kubernetes.operator.work.TerminalStep;
 import oracle.kubernetes.utils.TestUtils;
 import oracle.kubernetes.weblogic.domain.model.DomainResource;
 import org.hamcrest.Matchers;
+import org.junit.Ignore;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -396,7 +397,7 @@ class PodWatcherTest extends WatcherTestBase implements WatchListener<V1Pod> {
     }
   }
 
-  @Test
+  @Ignore
   void whenPodNotFound_waitForDeleteDoesNotRecordKubernetesFailure() {
     final DomainResource domain = DomainProcessorTestSetup.createTestDomain();
     final AtomicBoolean stopping = new AtomicBoolean(false);
--- a/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
@@ -359,7 +359,6 @@ class PodWatcherTest extends WatcherTestBase implements WatchListener<V1Pod> {
     try {
       testSupport.runSteps(watcher.waitForDelete(createPod(), terminalStep));
 
-      assertThat(terminalStep.wasRun(), is(true));
       assertThat(terminalStep.getExecutionCount(), is(1));
     } finally {
       stopping.set(true);
@@ -392,7 +391,6 @@ class PodWatcherTest extends WatcherTestBase implements WatchListener<V1Pod> {
       testSupport.failOnResource(KubernetesTestSupport.POD, NAME, NS, HTTP_NOT_FOUND);
       testSupport.setTime(10, TimeUnit.SECONDS);
 
-      assertThat(terminalStep.wasRun(), is(true));
       assertThat(terminalStep.getExecutionCount(), is(1));
     } finally {
       stopping.set(true);
--- a/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
@@ -31,6 +31,7 @@ import org.junit.jupiter.api.Test;
 import static oracle.kubernetes.common.logging.MessageKeys.EXECUTE_MAKE_RIGHT_DOMAIN;
 import static oracle.kubernetes.common.logging.MessageKeys.INTROSPECTOR_POD_FAILED;
 import static oracle.kubernetes.common.utils.LogMatcher.containsFine;
+import static oracle.kubernetes.operator.KubernetesConstants.HTTP_NOT_FOUND;
 import static oracle.kubernetes.operator.LabelConstants.CREATEDBYOPERATOR_LABEL;
 import static oracle.kubernetes.operator.LabelConstants.DOMAINUID_LABEL;
 import static oracle.kubernetes.operator.helpers.LegalNames.DEFAULT_INTROSPECTOR_JOB_NAME_SUFFIX;
@@ -359,6 +360,7 @@ class PodWatcherTest extends WatcherTestBase implements WatchListener<V1Pod> {
       testSupport.runSteps(watcher.waitForDelete(createPod(), terminalStep));
 
       assertThat(terminalStep.wasRun(), is(true));
+      assertThat(terminalStep.getExecutionCount(), is(1));
     } finally {
       stopping.set(true);
     }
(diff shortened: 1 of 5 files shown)


anpanigr (commented):
ItMiiSampleWlsAux.testAIWlsUpdate4UseCase still fails on jenkin run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12933

ankedia:
> ItMiiSampleWlsAux.testAIWlsUpdate4UseCase still fails on jenkin run https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12933
It's failing for a different reason https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12933/artifact/logdir/jenkins-weblogic-kubernetes-operator-kind-new-12933/wl_k8s_test_results/diagnostics/ItMiiSampleWlsAux/ItMiiSampleWlsAux.out/*view*/ .

```
@@ Info: Calling 'curl --connect-timeout 5 --max-time 120 --retry 3 --retry-delay 0 --retry-max-time 400 -fL https://github.com/oracle/weblogic-image-tool/releases/latest/download//imagetool.zip -o imagetool.zip'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
```

russgold:
Can you explain exactly what was wrong in the code and why it caused a failure? We should have unit tests that are able to reproduce the problem.

ankedia:
> Can you explain exactly what was wrong in the code and why it caused a failure? We should have unit tests that are able to reproduce the problem.

[ANIL] Without this change, the child fiber that's created for backstop rechecking of the pod resource (for missed events) executes the next step "createPod" and the main fiber also executes the "createPod" step. The unit test creation is tracked as a separate task in OWLS-102364.

rjeberhard:
How hard will it be to create the unit test? Why does this need to be a separate task?

ankedia:
> How hard will it be to create the unit test? Why does this need to be a separate task?
I'm not able to reproduce the integration test failure in my local environment due to other issues such as "Failed to get the version number of the requested WDT release" and timeout during DB/RCU installation. I will keep trying to reproduce as and when other issues are fixed. 

I believe the intgration test issue happens due to a race condition (a timing issue) which may not be easy to reproduce in the unit test. The separate task (OWLS-102364) was created by Russ as part of PR [3356](https://github.com/oracle/weblogic-kubernetes-operator/pull/3356) which caused this regression.

rjeberhard:
Okay, I was looking for something smaller... A test that protects your fix:
> Before the [change ](https://github.com/oracle/weblogic-kubernetes-operator/commit/68aea63e3da07d9ed1a622fc44e9db933aa5c513)in PR https://github.com/oracle/weblogic-kubernetes-operator/pull/3356, the onSuccess method of resumeIfReady step returned null action to terminate the child fiber. According to my understanding, resumeIfReady is the response step for the step that we create to check the latest resource value (backstop check for missed watch events) and that step is executed in a new child fiber that is created in checkUpdatedResource method. If the resource is ready, then the child fiber should terminate with a null action. I have also made a change to the ResponseStep to not log the failure status when response status code is 404.

... from being mistakenly reverted by someone else who makes changes in this area. I'm okay with an updated integration test or some test that runs during the build that captures better the race condition to come later.

ankedia:
> Okay, I was looking for something smaller... A test that protects your fix from being mistakenly reverted by someone else who makes changes in this area. I'm okay with an updated integration test or some test that runs during the build that captures better the race condition to come later.

oh, ok. Part of my changes are already covered by the existing unit test `whenPodNotFound_waitForDeleteDoesNotRecordKubernetesFailure` that Russ created for PR [3356](https://github.com/oracle/weblogic-kubernetes-operator/pull/3356). I'll see if I can further improve the code coverage.

rjeberhard:
> Part of my changes are already covered by the existing unit test

This makes sense. I was suspicious because there are no changes to unit tests in your PR, which means that if I (accidentally) reverted the operator runtime changes in this PR then there wouldn't be any unit test failure. :).

ankedia:
> This makes sense. I was suspicious because there are no changes to unit tests in your PR.

:) That's because I just reworked the fix in PR [3356](https://github.com/oracle/weblogic-kubernetes-operator/pull/). I have added a unit test now and code coverage is 100% :).

rjeberhard (approved):

ankedia:
@rjeberhard @russgold I have added new unit test `whenPodDeletedOnSecondRead_runNextStepOnlyOnce` and modified existing unit test `whenPodDeletedOnFirstRead_runNextStep` so that they will fail if my fix is mistakenly reverted by someone else making changes in this area.

operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
```diff
@@ -359,6 +360,7 @@ void whenPodDeletedOnFirstRead_runNextStep() {
       testSupport.runSteps(watcher.waitForDelete(createPod(), terminalStep));
 
       assertThat(terminalStep.wasRun(), is(true));
```
russgold, line 362: with your new test that the step was executed exactly once, you no longer need the test for `wasRun()`

operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
```diff
@@ -379,6 +381,24 @@ void whenPodNotDeletedOnFirstRead_dontRunNextStep() {
...
+    AtomicBoolean stopping = new AtomicBoolean(false);
+    PodWatcher watcher = createWatcher(stopping);
+
+    testSupport.defineResources(createPod());
+    try {
+      testSupport.runSteps(watcher.waitForDelete(createPod(), terminalStep));
+      testSupport.failOnResource(KubernetesTestSupport.POD, NAME, NS, HTTP_NOT_FOUND);
+      testSupport.setTime(10, TimeUnit.SECONDS);
+
+      assertThat(terminalStep.wasRun(), is(true));
```
russgold, line 395: And again, the test for `wasRun` is not needed, since you are checking that the step was run exactly once.

russgold (approved):
Well done. Your tests now not only verify that the fix is not reverted, they explain what the problem was.

operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
ankedia: I have removed the redundant check in [0ea2a3d](https://github.com/oracle/weblogic-kubernetes-operator/pull/3412/commits/0ea2a3dc5c26b618115b7db257fa810a9d7a24b6). Thanks.

operator/src/test/java/oracle/kubernetes/operator/PodWatcherTest.java
ankedia: Removed the `wasRun` test. Thanks.

PR #3420, opened 2022-09-15
fix tools download url and image secret for upgrade tests

vanajamukkara:
Fix tools download URL's not to look for version number. 
Use secret to pull images in the domain.yaml's for upgrade tests as its needed for OCIR in OKE env.

Jenkins results -
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/12999/ - upgrade tests on kind
https://build.weblogick8s.org:8443/job/wko-oke-nightly-seqential/114/ - upgrade tests on oke
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13010/ - full suite on kind

gate tests passed.


integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/CommonTestUtils.java
```diff
@@ -1634,67 +1633,7 @@ public static String getActualLocationIfNeeded(
...
-                params.stderr()),
-            exception);
-        throw exception;
-      }
-
-      if (version != null) {
-        actualLocation = location.replace("latest",
-            String.format("download/%s/%s", version, getInstallerFileName(type)));
-      }
+      actualLocation = location + "/download/" + getInstallerFileName(type);
```
robertpatrick, line 1636: Can I assume that this works for both `latest` and for a specified version?  I seems like it does not.

PR #3421, opened 2022-09-15
minor oper unit test fix: cleanup tmp files once done with them

tbarnes-us:
Fix two operator unit tests that were leaving temp directories behind in /tmp...


alai8 (approved):
Looks good. Wondering if it is possible to delete the temp files during Memento.revert()?

russgold:
I was not aware that we had any unit tests creating actual files. Is there a reason that we cannot use InMemoryFileSystem to avoid this? It is used in a fair number of tests and is faster than creating real files.

russgold (changes requested):
> Please look at InMemoryFileSystem as an alternative.

Tom: Will do.

russgold (approved):
I'm not going to insist on that change before the 4.0 release

PR #3426, opened 2022-09-16
Owls102369 - add validation for referencing a cluster resource that is used by another domain

doxiao:
- Add validation that checks if a domain references a cluster resource that is already used by another domain.
- Update integration test classes that contain domain configuration that is invalid any more.


common/src/main/resources/Operator.properties
```diff
@@ -238,6 +238,7 @@ WLSDO-0044=One or more server pods that are supposed to be available did not sta
   Check the server status in the domain status, the server pod status and logs, and WebLogic Server logs for possible reasons. \
   One common cause of this issue is a problem pulling the WebLogicServer image. \
   Adjust the value of 'serverPod.maxPendingWaitTimeSeconds' setting if needed."
+WLSDO-0045=Cannot reference Cluster ''{0}'' because it is used by ''{1}''
```
alai8, line 241: should we say "Cluster resource" here to avoid possible confusion with WLS cluster?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -274,6 +274,23 @@ private static void processServerEvent(CoreV1Event event) {
...
   }
 
+  /**
+   * Get all the domain resources in the given namespace.
+   *
+   * @param ns the namespace
+   * @return list of the domain resources
+   */
+  public static List<DomainResource> getDomains(String ns) {
+    List<DomainResource> referencingDomains = new ArrayList<>();
```
alai8, line 284: minor comment: should variable name be domains instead of referencingDomains since we are getting all domains within the namespace ns?

doxiao: good eye. I started with a method that returns only domains that reference a given cluster, and later changed to return all. will change. thanks.

common/src/main/resources/Operator.properties
doxiao: good point. will do.

alai8 (approved):

anpanigr (commented):
@doxiao Did you update any integration test to verify the usecase ?

rjeberhard (approved):

PR #3437, opened 2022-09-21
OWLS-102534 Display explanation when Available is false

russgold:
When the domain is not available, include a message which explains why.

There were also some bugs in edge cases for the Available and Completed statuses that were missed because of a unit test error.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -50,6 +50,7 @@ import oracle.kubernetes.operator.wlsconfig.WlsServerConfig;
 import oracle.kubernetes.operator.work.NextAction;
 import oracle.kubernetes.operator.work.Packet;
 import oracle.kubernetes.operator.work.Step;
+import oracle.kubernetes.utils.OperatorUtils;
 import oracle.kubernetes.utils.SystemClock;
 import oracle.kubernetes.weblogic.domain.model.ClusterCondition;
 import oracle.kubernetes.weblogic.domain.model.ClusterConditionType;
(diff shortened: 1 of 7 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -763,23 +766,83 @@ private boolean haveServerData() {
...
+        private DomainCondition createAvailableCondition() {
+          final List<String> unreadyNonClusteredServers = getUnreadyNonClusteredServers();
+          final List<ClusterCheck> unavailableClusters = getUnavailableClusters();
+
+          if (noApplicationServersReady()) {
+            return createNotAvailableCondition(LOGGER.formatMessage(NO_APPLICATION_SERVERS_READY));
+          } else if (!unreadyNonClusteredServers.isEmpty()) {
+            return createNotAvailableCondition(
+                LOGGER.formatMessage(NON_CLUSTERED_SERVERS_NOT_READY, formatServers(unreadyNonClusteredServers)));
+          } else if (!unavailableClusters.isEmpty()) {
```
tbarnes-us, line 785: Not a show stopper, but it's possible to have _both_ unready non-clustered servers and unavailableClusters at the same time, but, per the above logic, only the former will be included in the message when this happens.

russgold: Yes, correct. The idea is not to give every possible bit of information. That can actually be gleaned from the status as a whole. Rather, the idea is to give users some quick guidance of what they should be looking for.

russgold: Have you tried the branch to see if it gives you the messages needed to diagnose the intermittent problems you were seeing?

tbarnes-us: Can you save me a little time and tell me which particular one?  (There are more than one.)

tbarnes-us (approved):

PR #3439, opened 2022-09-21
Reintegrate cluster content earlier during webhook conversion to v8

rjeberhard:
During conversion from v9 to v8, cluster resources are reintegrated into the domain content. There is a unit-test that was meant to ensure that this cluster content was properly downgraded to v8, such as having the v8 versions of enumeration values.

That unit-test was, unfortunately, broken. This change corrects the unit-test and moves the reintegration of cluster resources earlier in the flow of the webhook such that the other methods that modify constants and other content work correctly.

### DIFF
--- a/common/src/test/java/oracle/kubernetes/common/utils/SchemaConversionUtilsTest.java
+++ b/common/src/test/java/oracle/kubernetes/common/utils/SchemaConversionUtilsTest.java
@@ -31,8 +31,8 @@ import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasKey;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.junit.MatcherAssert.assertThat;
-import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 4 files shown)


lennyphan (approved):

common/src/test/java/oracle/kubernetes/common/utils/SchemaConversionUtilsTest.java
```diff
@@ -148,7 +149,10 @@ void testV9DomainDowngrade() throws IOException {
     converterv8.convert(domain, clusters);
 
     Map<String, Object> v8Domain = readAsYaml(DOMAIN_V8_AUX_IMAGE30_YAML);
-    assertThat(converterv8.getDomain().get("clusters"), equalTo(v8Domain.get("clusters")));
+    List<Object> convertedClusters = (List<Object>) getDomainSpec(converterv8.getDomain()).get("clusters");
+    List<Object> origClusters = (List<Object>) getDomainSpec(v8Domain).get("clusters");
+    assertNotNull(convertedClusters);
```
russgold, line 154: Sonar recommendation: replace `assertNotNull(X)` with assertThat(X, notNullValue())

russgold (approved):
Recommend the one change, but otherwise fine

PR #3450, opened 2022-09-23
OWLS-102704 add validation for missing cluster resource

doxiao:
Add validation to check that all cluster resources that a domain resource references exist. 

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13331/


rjeberhard:
Is this still draft?

doxiao:
> Is this still draft?

Yes, there are test failures in one integration test class, which is not part of gating test suite. I have been debugging the failures.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
```diff
@@ -267,8 +267,23 @@ private List<ClusterResource> getClusters(Packet packet) {
...
+              .stream()
+              .filter(cluster -> name.equals(cluster.getMetadata().getName()))
+              .findFirst())
+          .orElse(null);
+    }
+
+    @Override
+    public ClusterResource findClusterInNamespace(V1LocalObjectReference reference, String namespace) {
+      return Optional.ofNullable(reference.getName())
+          .flatMap(name -> Optional.of(DomainProcessorImpl.getClusters(namespace))
```
alai8, line 282: Do we need to use Optional here since DPI.getClusters() always return an ArrayList?

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainResource.java
```diff
@@ -1196,24 +1208,25 @@ private void checkPortNameLength(V1ContainerPort port, String name, String prefi
 
     private void verifyClusterResourcesNotInUse(
         KubernetesResourceLookup kubernetesResources) {
-      getSpec().getClusters().forEach(cluster ->
-          Optional.ofNullable(kubernetesResources.findCluster(cluster))
-              .ifPresent(cr -> verifyClusterInUseByAnotherDomain(kubernetesResources, cr)));
+      Optional.ofNullable(getSpec().getClusters()).orElse(new ArrayList<>())
+          .forEach(reference ->
+              Optional.ofNullable(kubernetesResources.findClusterInNamespace(reference, getNamespace()))
+              .ifPresent(cluster -> verifyClusterInUseByAnotherDomain(kubernetesResources, cluster)));
```
alai8, line 1214: This is not in this change, but should this method name be verifyCluster**Not**InUseByAnotherDomain?

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
```diff
@@ -98,6 +99,10 @@ public class DomainProcessorImpl implements DomainProcessor, MakeRightExecutor {
   @SuppressWarnings({"FieldMayBeFinal", "CanBeFinal"})
   private static Map<String, Map<String, DomainPresenceInfo>> domains = new ConcurrentHashMap<>();
 
+  // Map namespace to map of clusterName to ClusterResource; tests may replace this value.
+  @SuppressWarnings({"FieldMayBeFinal", "CanBeFinal"})
+  private static Map<String, Map<String, ClusterResource>> clusters = new ConcurrentHashMap<>();
```
alai8, line 104: As part of DomainValidationSteps I believe we call createListClustersStep to return a list of Cluster resources in the same namespace. What is the reason that we are not able to use that list of Cluster resources for the new validation is being added in this PR, instead of creating a new list of ClusterResource here? I may be missing something obvious though. Thanks.

operator/src/main/java/oracle/kubernetes/operator/helpers/DomainValidationSteps.java
doxiao: No, we don't need to use Optional here. Will Change. thanks

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainResource.java
doxiao: The proposed name is more consistent with the pattern. will change.

operator/src/main/java/oracle/kubernetes/operator/DomainProcessorImpl.java
doxiao: Yes, DomainValidationSteps uses createListClustersStep to get all clusters, and it puts the list into the packet. I thought about using the list in the packet before I added the new map. But at the time, somehow I got the impression that the list in the packet only contains the clusters that are referenced by the the domain that the operator is processing. Not I read the code again, and I believe that the list has all clusters in the namespace. I am glad that we don't need to use a new map after all. I'll give it a try and I think it should work.  thanks for catching this.

doxiao:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13349/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13350/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13356/

operator/src/main/java/oracle/kubernetes/operator/DomainProcessor.java
```diff
@@ -88,4 +89,8 @@ default void reportSuspendedFibers() {
   default Stream<DomainPresenceInfo> findStrandedDomainPresenceInfos(String namespace, Set<String> domainUids) {
     return Stream.empty();
   }
+
+  default void addClusterList(ClusterList list) {
```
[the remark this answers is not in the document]
doxiao: good catch. removed. thanks.

alai8 (approved):

PR #3465, opened 2022-09-28
Owls102705 - validating webhook checks new resources for fatal validation errors

doxiao:
With this change the validating webhook listens to new domains and only checks fatal validation errors.
- Add CREATE operation to webhook configuration
- Refactor the validating webhook code and add code to handle new resources
- Refactor unit test code to reuse some common test cases, and add new test cases for new resources.


lennyphan (approved):
LGTM too.

PR #3468, opened 2022-09-28
OWLS-102367: WKO 4.0 bug: Cluster resource 'Observed Generation' status goes missing after a few minutes idle time

lennyphan:
Although the the 'Observed Generation' attribute is reported in the Cluster status initially, it disappears when the Domain Status is updated.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
@@ -201,11 +201,6 @@ public class ClusterStatus implements Comparable<ClusterStatus>, PatchableCompon
     return conditions;
   }
 
-  public ClusterCondition getCondition(ClusterConditionType type) {
-    return getConditions().stream().filter(condition -> condition.getType().equals(type))
-            .findFirst().orElse(null);
-  }
-
   @Override
   public String toString() {
     return new ToStringBuilder(this)
--- a/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
+++ b/operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
@@ -174,6 +174,20 @@ public class ClusterStatus implements Comparable<ClusterStatus>, PatchableCompon
     return observedGeneration;
   }
 
+  /**
+   * The observedGeneration attribute is defined as an integer type in the Cluster Resource
+   * schema, hence convert to integer when publishing the status.
+   * @return integer value of the observedGeneration, if set, otherwise -1.
+   */
+  private Integer getObservedGenerationAsInteger() {
+    return observedGeneration != null ? observedGeneration.intValue() : -1;
+  }
+
+  public ClusterStatus withObservedGeneration(Long observedGeneration) {
+    this.observedGeneration = observedGeneration;
+    return this;
+  }
+
   public void setObservedGeneration(Long observedGeneration) {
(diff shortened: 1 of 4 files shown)


tbarnes-us:
General question: Why is the domain status being updated? As far as I know, it wasn't changing.

tbarnes-us (approved):

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
```diff
@@ -174,6 +174,20 @@ public Long getObservedGeneration() {
     return observedGeneration;
   }
 
+  /**
+   * The observedGeneration attribute is defined as an integer type in the Cluster Resource
+   * schema, hence convert to integer when publishing the status.
+   * @return integer value of the observedGeneration, if set, otherwise defaults to 1.
+   */
+  private Integer getObservedGenerationAsInteger() {
+    return observedGeneration != null ? observedGeneration.intValue() : 1;
```
alai8, line 183: why is the default value for observedGeneration 1? Thanks

lennyphan:
> 

During the scheduled Domain status check, a DomainStatus is created which is then compared to the current DomainStatus to determine if a change has occurred.  Since the ClusterStatus is part of the DomainStatus, I was missing the 'Observed Generation' field as part of the scheduled check and so it 'disappeared'.  I can't say for sure but I think It's highly likely that it's a contributor to OWLS-102726 because we'll do the comparison and it won't be equal due to the missing field.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterStatus.java
lennyphan: Since the Observed Generation is a monotonically increasing number that gets bumped every time you update your resource. So the first resource is created with generation 1 by 'metadata.Generation', hence, I default to 1.  In most cases, we'll have read the ClusterResource object from the API server when listing them and the Generation will have been set by Kubernetes.

bash-4.2$ kubectl describe clu sample-domain1-cluster-1 -n sample-domain1-ns
Name:         sample-domain1-cluster-1
Namespace:    sample-domain1-ns
Labels:       weblogic.domainUID=sample-domain1
Annotations:  <none>
API Version:  weblogic.oracle/v1
Kind:         Cluster
Metadata:
  Creation Timestamp:  2022-09-28T18:33:47Z
  Generation:          1
  Managed Fields:
    API Version:  weblogic.oracle/v1
    Fields Type:  FieldsV1
...
    Manager:         Kubernetes Java Client
    Operation:       Update
    Subresource:     status
    Time:            2022-09-28T18:36:08Z
  Resource Version:  215489
  UID:               618d8ee1-f0cc-4f5a-85a3-4e273fd152dc
Spec:
  Cluster Name:  cluster-1
  Replicas:      2
Status:
  Cluster Name:  cluster-1
  Conditions:
    Last Transition Time:  2022-09-28T18:37:52.312865Z
    Status:                True
    Type:                  Available
    Last Transition Time:  2022-09-28T18:37:52.312918Z
    Status:                True
    Type:                  Completed
  Label Selector:          weblogic.domainUID=sample-domain1,weblogic.clusterName=cluster-1
  Maximum Replicas:        5
  Minimum Replicas:        0
  Observed Generation:     1
  Ready Replicas:          2
  Replicas:                2
  Replicas Goal:           2

alai8 (approved):

rjeberhard (approved):

PR #3474, opened 2022-09-29
Fail Helm install and operator runtime on Kubernetes 1.25

rjeberhard:


alai8 (approved):
LGTM
I am assuming that `Capabilities.APIVersions.Has "policy/v1beta1"` is false in a 1.25 k8s cluster

rjeberhard:
> LGTM I am assuming that `Capabilities.APIVersions.Has "policy/v1beta1"` is false in a 1.25 k8s cluster

Correct. There is no Helm command to get the Kubernetes version -- instead, we have this which is useful in a different way in that you can get the list of API's served by the control plane.

PR #3475, opened 2022-09-29
WIP - Owls 102893

jshum2479:
This PR is trying to resolve the issue in an OpenShift environment with logHome and/or domainHome on PV.  Since the user id changed between restarts,  log files or weblogic generated files left in shared volume must be accessible and writable by different user id,  currently it fails with permission denied.

This initial cut has been tested with different (user id / same group  id) using the podSecurityContext to simulate the OpenShift environment behavior and the servers come up fine.  We definitely need to have comprehensive QA tests for this type of use case.   For example,

1)  Helm install needs to set kubernetesPlatform to OPENSHIFT (I can change it code to use camel case per documentation or change the doc/unit test to all upper case)
2)  Integration tests need to cover restart with different user ids
3)  Integration tests need to cover changes to the config.xml between different users, including deploying applications, creating and using JMS resources, and more ...
4)  scaling up and down within the same user id or different user ids

We should have a meeting to discuss about our strategies.


operator/src/main/resources/scripts/startServer.sh
```diff
@@ -261,6 +264,11 @@ fi
 #
 
 createFolder "${DOMAIN_HOME}/servers/${SERVER_NAME}/security" "This is the server '${SERVER_NAME}' security directory within directory DOMAIN_HOME 'domain.spec.domainHome'." || exitOrLoop
+if [[ "${KUBERNETES_PLATFORM^^}" == "${PLATFORM_OPENSHIFT}" ]]; then
+  createFolder "${DOMAIN_HOME}/servers/${SERVER_NAME}/tmp" "This is the server '${SERVER_NAME}' tmp directory within directory DOMAIN_HOME 'domain.spec.domainHome'." || exitOrLoop
+  createFolder "${DOMAIN_HOME}/servers/${SERVER_NAME}/logs" "This is the server '${SERVER_NAME}' logs directory within directory DOMAIN_HOME 'domain.spec.domainHome'." || exitOrLoop
```
tbarnes-us, line 269: This folder is already created in 'startNodeManager.sh' (see ' createFolder "${serverLogHome}"'), and its location varies based on LOG_HOME.

tbarnes-us: But I suppose it might be helpful to create the directory anyway in case it's somehow used even when LOG_HOME is set.

tbarnes-us (approved):

operator/src/main/resources/scripts/startServer.sh
jshum2479: The tmp and logs directory is created when the server starts, somehow something down the subdirectory is not created with the right umask (group not writable),  so pre-creating it to make sure they are group writable.

tbarnes-us: OK

PR #3481, opened 2022-10-03
Improve fine logging from webhook

rjeberhard:
The logging is still difficult to read because the log is output in JSON and the string constants get encoded. Still, these logging points are much better.


common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
```diff
@@ -116,7 +113,6 @@ public Resources convertDomainSchema(Map<String, Object> domain, ResourceLookup
     if (spec == null) {
       return new Resources(domain, Collections.emptyList());
     }
-    LOGGER.fine("Converting domain " + domain + " to " + targetAPIVersion + " apiVersion.");
```
ankedia, line 119: I assume we don't need this logging when running the standalone command line utility to upgrade the domain from `v8` to `v9`. The REST API won't be invoked when running the standalone command line utility.

ankedia (approved):
Please see my comment about the standalone command line utility. Otherwise it looks good to me.

PR #3482, opened 2022-10-03
Fix to retain existing init containers under serverPod when adding the init containers for the auxiliary images.

ankedia:
Fix for `ItInitContainers` integration test failures in the interop environment. This change ensures that the conversion webhook doesn't overrride the existing init containers under `serverPod` stanza when creating and adding the init containers for the auxiliary images.

Integration test run results - https://build.weblogick8s.org:8443/job/wko-kind-34-with-webhook/114/

### DIFF
--- a/common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
+++ b/common/src/main/java/oracle/kubernetes/common/utils/SchemaConversionUtils.java
@@ -450,7 +450,8 @@ public class SchemaConversionUtils {
   private void addInitContainersVolumeAndMountsToServerPod(Map<String, Object> serverPod, List<Object> auxiliaryImages,
                                                                          List<Object> auxiliaryImageVolumes) {
     addEmptyDirVolume(serverPod, auxiliaryImageVolumes);
-    List<Object> initContainers = new ArrayList<>();
+    List<Object> initContainers = Optional.ofNullable((List<Object>) serverPod.get("initContainers"))
+        .orElse(new ArrayList<>());
     for (Object auxiliaryImage : auxiliaryImages) {
       initContainers.add(
           createInitContainerForAuxiliaryImage((Map<String, Object>) auxiliaryImage, containerIndex.get(),


rjeberhard (approved):

doxiao (commented):
The change looks good to me. 

Just curious - is it possible that the existing initContainers list already contains a use specified initContainer that uses the same name as the internal auxiliary images? I know it is probably an edge and rare case, but what happens if we hit the case? shall we add a validation check for this case?

ankedia:
> Just curious - is it possible that the existing initContainers list already contains a user specified initContainer that uses the same name as the internal auxiliary images? I know it is probably an edge and rare case, but what happens if we hit the case? shall we add a validation check for this case?
[ANIL] This particular change is for converting the domains resource from `v8` to `v9` and we add a prefix "compatibility-mode-" to the internal aux-image init container during conversion. I believe if the customer specified initContainer uses the same name as the internal aux-image initContainer, then the domain will not be created as it would fail the initial validation (by k8s) due to container names collision. I'll try this once and confirm.

ankedia:
> > Just curious - is it possible that the existing initContainers list already contains a user specified initContainer that uses the same name as the internal auxiliary images? I know it is probably an edge and rare case, but what happens if we hit the case? shall we add a validation check for this case?
> > [ANIL] This particular change is for converting the domains resource from `v8` to `v9` and we add a prefix "compatibility-mode-" to the internal aux-image init container during conversion. I believe if the customer specified initContainer uses the same name as the internal aux-image initContainer, then the domain will not be created as it would fail the initial validation (by k8s) due to container names collision. I'll try this once and confirm.

I tried creating a domain with customer defined initContainer having the same name as the aux image container and the domain creation fails with below SEVERE error ->
`Pod \"sample-domain1-admin-server\" is invalid: [spec.initContainers[1].name: Duplicate value: \"operator-aux-container1\" `

Since the webhook is just trying to convert the existing `v8` domain to `v9` domain, I think the validation is not needed since the `v8` domain would not get created if the existing init container is same name as the aux container.

doxiao:
Thanks for verifying this.

doxiao (approved):

lennyphan (approved):

PR #3498, opened 2022-10-10
Use down plan when domain has deletionTimestamp

rjeberhard:
Use the domain down plan when the domain has a metadata.deletionTimestamp. Once we merge this to main I'll work on a backport to release/3.4.

This should resolve the deadlock we discussed in the scrum.


ankedia (commented):
LGTM. I wonder if any of the existing integration test (in 4.0 or 3.4) rely on the current behavior. Should we wait for the integration tests results to see if all the tests pass with this change?

alai8 (approved):

rjeberhard:
> LGTM. I wonder if any of the existing integration test (in 4.0 or 3.4) rely on the current behavior. Should we wait for the integration tests results to see if all the tests pass with this change?

We should wait for the gating test suite. I don't think that any tests could rely on this behavior because it seems to cause a deadlock. Any other behavior would have quickly resolved.

PR #3517, opened 2022-10-14
Use the doEnd method instead of returning null next action

ankedia:
Use the doEnd(packet) method to end the fiber processing instead of returning "null" next action. With "null" next action, the `doPotentialRetry` method of ResponseStep logs a ASYNC_NO_RETRY message since the RetryStrategy in the packet is null.

Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13488/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/PodWatcher.java
@@ -35,6 +35,7 @@ import oracle.kubernetes.operator.watcher.WatchListener;
 import oracle.kubernetes.operator.work.NextAction;
 import oracle.kubernetes.operator.work.Packet;
 import oracle.kubernetes.operator.work.Step;
+import org.jetbrains.annotations.NotNull;
 
 import static oracle.kubernetes.common.logging.MessageKeys.EXECUTE_MAKE_RIGHT_DOMAIN;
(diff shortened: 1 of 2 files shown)


doxiao (commented):
Can you add a unit test case to verify the change?

rjeberhard (approved):

ankedia:
> Can you add a unit test case to verify the change if we don't have test cases that already cover this?

The code coverage is 100% since the change is already covered by existing unit tests. I'm not sure if the unit test framework has support for retry strategies, I have started a slack thread to discuss this.

russgold:
> > Can you add a unit test case to verify the change if we don't have test cases that already cover this?
> 
> The code coverage is 100% since the change is already covered by existing unit tests. I'm not sure if the unit test framework has support for retry strategies, I have started a slack thread to discuss this.

And this is one of the problems with relying on code coverage numbers. If code is not executed during a unit test (the source of the coverage numbers), it clearly means that that code is not being tested, but the simple fact that code is executed, does not mean that all of its behavior is tested. Good coverage numbers are necessary but not sufficient. That you're unable to reproduce the problem in a test _does_ indicate that the behavior is not being tested, despite the 100% code coverage numbers.

I'm just getting back up to speed after being away, so I will add this to my list of things to investigate.

PR #3530, opened 2022-10-20
Kubernetes Client 16.0.0

rjeberhard:
I'm not excited about this PR but we need to do it...

In the 15.0.0 version of the Kubernetes Java client, String arguments that had a set of enumerated types were converted to Enums. We upgraded to this version and happily started using these Enums. However, there was a debate in the wider Kubernetes community about how to handle adding new enumeration options over time and how not to break existing clients for languages like Java that don't have a built-in pattern for this problem.

Therefore, in 15.0.1, the Kubernetes Java client team reverted the use of Enums. You can read about it all here: https://github.com/kubernetes-client/java/issues/2218.

For a while it looked like the community might come to an agreement and they have... to not use the Enums across all of the supported clients.

Therefore, in order to use the 16.0.0 Kubernetes Java client (or any follow-on version), I've reverted the use of their enumerations.


integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/DbUtils.java
```diff
@@ -249,15 +249,15 @@ public static synchronized void startOracleDB(String dbBaseImageName, int dbPort
...
                                 .hostPort(dbListenerPort)))
                             .resources(new V1ResourceRequirements()
                                 .limits(limits)
                                 .requests(requests))
                             .terminationMessagePath("/dev/termination-log")
-                            .terminationMessagePolicy(V1Container.TerminationMessagePolicyEnum.FILE)))
-                    .dnsPolicy(V1PodSpec.DnsPolicyEnum.CLUSTERFIRST)
-                    .restartPolicy(V1PodSpec.RestartPolicyEnum.ALWAYS)
+                            .terminationMessagePolicy("File")))
+                    .dnsPolicy("CluserFirst")
```
lennyphan, line 259: "ClusterFirst" not "CluserFirst"

operator/src/test/java/oracle/kubernetes/operator/helpers/ManagedPodHelperTest.java
```diff
@@ -1081,11 +1079,11 @@ void whenDomainAndClusterHaveLegacyAuxImages_createManagedPodsWithInitContainers
     assertThat(getCreatedPodSpecInitContainers(),
         allOf(Matchers.hasLegacyAuxiliaryImageInitContainer(AUXILIARY_IMAGE_INIT_CONTAINER_NAME_PREFIX + 1,
                 "wdt-image:v1",
-                V1Container.ImagePullPolicyEnum.IFNOTPRESENT,
+                        "IfNotPresent",
```
lennyphan, line 1082: Indentation not the same as below?

lennyphan (changes requested):

integration-tests/src/test/java/oracle/weblogic/kubernetes/utils/DbUtils.java
rjeberhard: Good eyes! I looked through all of the changes for misspellings and missed this one.

operator/src/test/java/oracle/kubernetes/operator/helpers/ManagedPodHelperTest.java
rjeberhard: My IDE did this automatically... I'll see if I can catch these but not sure that it matters.

ankedia (approved):
LGTM

PR #3531, opened 2022-10-20
OWLS-102910: WKO 4.0 bug: Cluster condition last transition time is diffferent in ClusterStatus of Domain Resource and not updated on ClusterResource resource

lennyphan:
Every 2 minutes, when the DomainRecheck/DomainResourcesValidation operation occurs, a new DomainPresenceInfo is created and used in the make-right flow to reconcile resources.  ServerStartInfo (SSI) is required to calculate both the DomainStatus and ClusterStatus conditions.  Since the ServerStartInfo of the DomainPresenceInfo object has not been fully constructed when either the ScheduledDomainStatusUpdater or the StatusUpdateStep executes then the DomainStatus is updated (along with the ClusterStatus condition) although the domain has not changed.  This short-term fix includes the following:

1. When either the ScheduledDomainStatusUpdater or the StatusUpdateStep executes, we mark the flow by adding an identifier to the packet.
2. We identify the ServerStartInfo has not been constructed by a null value, instead of of an empty list,  as an empty list can also be construed as a domain without any servers.
3. When the status update step begins, it detects that the DPI/SSI hasn't been constructed and skips/delays the status update until either the next ScheduledDomainStatusUpdater runs (5 seconds later) or the next make-right operation runs.
4. We skip Domain status updates by looking at the identifier in the packet and whether the DPI/SSI hasn't been constructed because it is valid for the SSI to not have been constructed when the domain is first starting or when only the admin server is run during a normal make-right operation.


operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -570,6 +572,22 @@ DomainStatusUpdaterContext createContext(Packet packet) {
...
     }
 
+    @Override
+    public NextAction apply(Packet packet) {
+      return shouldSkipDomainStatusUpdate(packet)
+              ? doNext(getNext().getNext(), packet) : super.apply(packet);
+    }
+
+    private boolean shouldSkipDomainStatusUpdate(Packet packet) {
+      boolean isDomainResourcesValidation =
```
rjeberhard, line 582: This might be clearer if you used getOrDefault()... Something like:
Boolean idDomainResourcesValidation = (Boolean) packet.getOrDefault(DOMAIN_RESOURCES_VALIDATION, Boolean.FALSE);

operator/src/test/java/oracle/kubernetes/operator/DomainPresenceTest.java
```diff
@@ -429,6 +429,11 @@ public MakeRightDomainOperation withExplicitRecheck() {
         return this;
       }
 
+      @Override
+      public MakeRightDomainOperation withDomainResourcesValidation() {
+        return this;
```
rjeberhard, line 434: This doesn't need to update any member?

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
lennyphan: Thanks.  Great suggestion. Done.

operator/src/test/java/oracle/kubernetes/operator/DomainPresenceTest.java
lennyphan: Not need to update any attribute in this stub at this moment since the actual implementation propagates the value by adding it to a packet and no code needs to call an API to access it.  This method was added for compilation.

rjeberhard (approved):

russgold:
Can you clarify? You say
```
We skip Domain status updates by looking at the identifier in the packet and whether the DPI/SSI hasn't been constructed because it is valid for the SSI to not have been constructed when the domain is first starting or when only the admin server is run during a normal make-right operation.
```

how does that get reset, so that updates can proceed?

lennyphan:
> Can you clarify? You say
> 
> ```
> We skip Domain status updates by looking at the identifier in the packet and whether the DPI/SSI hasn't been constructed because it is valid for the SSI to not have been constructed when the domain is first starting or when only the admin server is run during a normal make-right operation.
> ```
> 
> how does that get reset, so that updates can proceed?

There are three operational flows where status updates are occurring:

1. During the normal make-right flow when handling resource events (such as when domain is first coming up), we make status updates multiple times during the domain's lifecycle using the DomainPresenceInfo from the cache.
2. When a DomainRecheck/DomainResourcesValidation occurs (every 2 minutes), we create a new DomainPresenceInfo and initiate a make-right.
3. When a ScheduledStatusUpdate occurs (every 5 seconds) which uses the DomainPresenceInfo from the cache.


We don't need to reset the identifiers as they are only used to identify which operational flow is occurring as we only want to skip status updates during a DomainRecheck or ScheduledStatusUpdate if the ServerStateInfo is not set.

Updates during a DomainRecheck or ScheduledStatusUpdate only proceed when the ServerStartupInfo in the DomainPresenceInfo have been constructed and set, otherwise it skips the update and proceeds with the next step. Setting the identifier helps us identify the operational flows (2 and 3) where we shouldn't do the status update if the ServerStateInfo is not set so that we don't calculate the new status based on missing ServerStateInfo. Once the SSI has been set, any subsequent domain status updates (like the last step of the make-right) will proceed.

During a normal make-right when the domain is first starting, we update the domain status while the admin server is starting but we haven't yet calculated the ServerStartInfo (this only done during ManagedServersUpStep) so this is where the domain status should be calcuated/updated even if SSI hasn't been constructed, otherwise we don't have any domain status updates.

Hopefully I've described it so that it's more clear otherwise we can do a quick chat.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -64,4 +64,6 @@ public interface ProcessingConstants {
 
   String SHUTDOWN_WITH_HTTP_SUCCEEDED = "SHUTDOWN_WITH_HTTP_SUCCEEDED";
   String DOMAIN_INTROSPECTION_COMPLETE = "Domain introspection complete";
+  String DOMAIN_RESOURCES_VALIDATION = "domainResourcesValidation";
```
russgold, line 67: If we change the logic to use the packet entry to mean skipping the status update, we don't need two separate constants that are treated identically,

operator/src/test/java/oracle/kubernetes/operator/steps/ManagedServersUpStepTest.java
```diff
@@ -795,9 +795,11 @@ static Collection<String> getPreCreateServers() {
     }
 
     static ServerStartupInfo getServerStartupInfo(String serverName) {
-      for (ServerStartupInfo startupInfo : info.getServerStartupInfo()) {
-        if (startupInfo.serverConfig.getName().equals(serverName)) {
-          return startupInfo;
+      if (info.getServerStartupInfo() != null) {
```
russgold, line 798: This can be simplified by extracting a method that does what the original ``getServerStartupInfo`` did: use Optional.ofNullable() to always return a collection.

operator/src/main/java/oracle/kubernetes/operator/steps/ManagedServersUpStep.java
```diff
@@ -273,6 +273,8 @@ Collection<ServerStartupInfo> getStartupInfos() {
       if (startupInfos != null) {
```
russgold, line 273: The behavior of startupInfos seems to have changed, and it would seem simpler to initialize it to an empty list when it is declared, rather than doing so in multiple places. Then you don't need this test.

operator/src/test/java/oracle/kubernetes/operator/helpers/TopologyValidationStepTest.java
```diff
@@ -470,6 +470,7 @@ void whenClusterExistInDomain_clusterStatusContainsMaxReplicas() {
     defineScenario(TopologyCase.REPLICAS)
         .withStaticCluster(10).withReplicas(5).build();
     testSupport.defineResources();
+    info.setServerStartupInfo(Collections.emptyList());
```
russgold, line 473: This line should not be needed

operator/src/test/java/oracle/kubernetes/operator/DomainStatusUpdateTestBase.java
```diff
@@ -1644,6 +1664,50 @@ void whenAdminOnlyAndAManagedServersShuttingDown_completedIsFalse() {
     assertThat(getRecordedDomain(), hasCondition(COMPLETED).withStatus(FALSE));
   }
 
+  @Test
+  void whenServerStateInfoNotConstructed_duringDomainResourceValidation_verifyDomainStatusNotUpdated() {
+    configureDomain().configureCluster(info, "cluster1").withReplicas(2).withMaxUnavailable(1);
```
russgold, line 1669: There is no need for the first underscore. Just change "_d" to "D"

If you change the MakeRight and ProcessingConstants to have the new behavior describe the effect on the Make-Right rather than who constructed it, you don't need both of these tests, as they do the same thing.

russgold (changes requested):
I've identified a number of things that can simplify the code (and reduce the number of files changed). There is one case that I can think of where you would want to run the domain status updater with a null server startup info: the case of a domain whose only server is the admin server. We need tests for that case, to make sure it actually works. Are there any others?

rjeberhard (approved):

ankedia (approved):
LGTM. The integration test `ItWlsSamples#testTraefikIngressController` is failing with the PR branch. However, it looks unrelated to the changes in this branch.

operator/src/main/java/oracle/kubernetes/operator/ProcessingConstants.java
```diff
@@ -64,4 +64,5 @@ public interface ProcessingConstants {
 
   String SHUTDOWN_WITH_HTTP_SUCCEEDED = "SHUTDOWN_WITH_HTTP_SUCCEEDED";
   String DOMAIN_INTROSPECTION_COMPLETE = "Domain introspection complete";
+  String DOMAIN_RECHECK_OR_SCHEDULED_STATUS_UPDATE = "domainRecheckOrScheduledStatusUpdate";
```
russgold, line 67: The name is still focused on who is calling it, rather than the significance. 

Let me suggest something like `SKIP_STATUS_UPDATE_IF_SSI_NOT_RECORDED`. Not sure if there is a simpler way to say that.

russgold (approved):
Just one picky point left, regarding the name of the constant.

PR #3561, opened 2022-10-25
Support wdt app module deployment

jshum2479:
This PR is to add support for the standalone app deployment in WDT which include archive jms, jdbc, wldf modules xml.   The operator will store any standalone xml file deployed in the domain in the domain config zip and restore them before starting the server.


tbarnes-us (approved):
LGTM!

I forgot WL still supported deploying JMS, JDBC, etc XML modules inside an app - _not_ a recommended best practice, but still supported... The best practice is to configure services using WebLogic configuration mbeans (WDT/WLST/Console etc).

PR #3562, opened 2022-10-25
OWLS-103643 - Change to the quick start cluster resource name to include domain-uid and cluster-name validation

ankedia:
OWLS-103643 part 1
 - Change to the quick start cluster resource name to include domain-uid to follow the best practice.
 - Added valiation for the ClusterSpec.ClusterName to be non-null.

Integration test run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13628/console


tbarnes-us (approved):
LGTM. Good catch that 'clusterName' needed to be set to 'required'.

ankedia:
> LGTM. Good catch that 'clusterName' needed to be set to 'required'.

Thanks.  I found the NULL clusterName issue while running the quick start and set it to `required` as suggested by Ryan.

rjeberhard (approved):

PR #3567, opened 2022-10-26
Skip status initialization during domain recheck and rename SKIP_UPDATE_DOMAIN_STATUS_IF_NEEDED flag

ankedia:
This PR skips the status initialization during domain recheck and renames the packet flag SKIP_UPDATE_DOMAIN_STATUS_IF_NEEDED to DOMAIN_RECHECK_OR_SCHEDULED_STATUS_UPDATE.

It also removes `skipUpdateDomainStatusIfNeeded` flag from the MakeRightDomainOperationImpl and instead uses `wasStartedFromEvent` method to identify if make-right is started by the explicit domain recheck. 

Integration test run results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13636/console.


lennyphan (approved):
LGTM

PR #3597, opened 2022-11-01
OWLS-103825: Available should be true for admin server only domain

lennyphan:
Condition Available will be true for a Domain where the only server configured to run is the Admin server.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
@@ -870,7 +870,8 @@ public class DomainStatusUpdater {
         }
 
         private boolean isOnlyAdminServerExpectedRunningInDomain() {
-          return status.getServers().size() == 1 && isAdminServer(getInfo().getAdminServerName());
+          return status.getServers().size() == 1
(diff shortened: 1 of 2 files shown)


rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/DomainStatusUpdater.java
```diff
@@ -857,13 +857,23 @@ private Collection<DomainPresenceInfo.ServerStartupInfo> getServerStartupInfos()
 
         // when the domain start policy is ADMIN_ONLY, the admin server is considered to be an application server.
```
ankedia, line 858: The comment is bit misleading after the change.

ankedia (approved):
LGTM

ankedia:
@rjeberhard @lennyphan - If the domain has an admin server and a cluster with replicas set to 0, will the domain available condition be `True` or `False` in this case? I'm looking into OWLS-103828 and making sure that the cluster available condition is `False` when replicas is `0`. Just want to make sure that I understand the expected behavior of the domain condition in such cases. Thanks.

lennyphan:
Closing this PR as it has been combined with PR [3602](https://github.com/oracle/weblogic-kubernetes-operator/pull/3602)

PR #3601, opened 2022-11-01
Use failure retries rather than starting an explicit make-right

russgold:
This changes the way the WaitForPodReadyStep reacts to a failure. Rather than creating an explicit make-right, it now reports a failure and lets the normal make-right retry mechanism schedule the next run.


documentation/swagger/index.html
```diff
@@ -1,602 +0,0 @@
-<!-- HTML for static distribution bundle build -->
```
rjeberhard, line 1: Why is this file changed / deleted?

operator/src/main/java/oracle/kubernetes/operator/MakeRightDomainOperation.java
```diff
@@ -36,14 +36,6 @@ public interface MakeRightDomainOperation extends PacketComponent {
    */
   MakeRightDomainOperation withEventData(EventData eventData);
 
-  /**
-   * Specifies that the current MakeRightOperation may skip updating the DomainStatus if the
-   * DomainPresenceInfo.ServerStartupInfo value has not yet been constructed..
-   */
-  default MakeRightDomainOperation skipUpdateDomainStatusIfNeeded() {
```
rjeberhard, line 43: Why is this removed? I thought that this was part of Lenny's fix for the intermittent incorrect status from the status thread when the make-right and status updater fibers ran at the same time...

lennyphan: Actually, I thought we did remove it as Anil and I had changed it as it wasn't needed anymore as we used the MakeRightDomainOperation.wasStartedFromEvent() method instead.

operator/src/main/java/oracle/kubernetes/operator/work/Fiber.java
```diff
@@ -571,6 +572,31 @@ public Packet getPacket() {
     return na.getPacket();
   }
 
+
+  @Override
+  public void cancelChildFibersAndResume(Packet packet, Step next) {
+    na.setNext(next);
+    if (children != null) {
+      for (Fiber child: children) {
+        child.cancelAndExitCallback(true, NULL_EXIT_CALLBACK);
```
rjeberhard, line 581: I notice that this assumes that this method doesn't need to evaluate the return value from cancelAndExitCallback on each child. Did you evaluate this and decide that this wasn't necessary? So, thinking that through... After these calls we know that each child Fiber will be cancelled; however, it is possible that one or more child Fibers is still executing any Step that it was already executing. Can you see a use case where this would be a problem?

rjeberhard (approved):
Please just look at and respond to the couple of questions that I posed in the review.

operator/src/main/java/oracle/kubernetes/operator/MakeRightDomainOperation.java
rjeberhard: Sounds good.

documentation/swagger/index.html
russgold: It's a generated file. I'm not sure why it would have changed.

operator/src/main/java/oracle/kubernetes/operator/work/Fiber.java
russgold: The places where the return value is used need to count how many child fibers are still running, in order to allow the parent to exit gracefully. In this method, we aren't interested in the result of the child fibers, or in canceling the parent fiber right now. Essentially, we are aborting the make-right and seeking only to report a failure.

operator/src/main/java/oracle/kubernetes/operator/MakeRightDomainOperation.java
russgold: @lennyphan can you comment?

lennyphan: Already commented above, it should be removed as it's not used anymore.  We mistakenly left it when it should have been cleaned up.  This issue is resolved.

documentation/swagger/index.html
rjeberhard: I'm going to build your branch locally. We are separately responding to an issue about this file but haven't done anything yet so this is oddly coincidental.

rjeberhard: I'm going to restore the file.

lennyphan (approved):
Did you want to resolve the reported "Code smell"?

PR #3648, opened 2022-11-09
OWLS-103645 - Add domain and cluster 4.0 'observedGeneration' to pod labels.

ankedia:
OWLS-103645 - Add `observedDomainGeneration` and `observedClusterGeneration` information to the pods labels to simplify the determination of the progress of all significant resource changes as they flow through a domain.

Integration test run - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13784/


tbarnes-us:
General comment:

In my opinion, 'observedGenerationDomain' or if not, then 'domainObservedGeneration', should be used instead of 'observedDomainGeneration'. (The former sorts observedGeneration for domain and cluster right next to each other).

We are using 'observedGeneration' in the status and k8s uses it in the metadata, and either of the above alternatives keeps to the same pattern.

rjeberhard (approved):

PR #3649, opened 2022-11-09
Add domain.spec.maxClusterUnavailable

doxiao:
Adding new field domain.spec.maxClusterUnavailable as the default for cluster.spec.maxUnavailable as well as the value for clusters that do not have a corresponding cluster resource. The default is 1.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13803/


documentation/domains/index.html
```diff
@@ -1222,6 +1222,12 @@
           "description": "The directory containing the WebLogic domain configuration inside the container. Defaults to /shared/domains/\u003cdomainUID\u003e if `domainHomeSourceType` is PersistentVolume. Defaults to /u01/oracle/user_projects/domains/ if `domainHomeSourceType` is Image. Defaults to /u01/domains/\u003cdomainUID\u003e if `domainHomeSourceType` is FromModel.",
           "type": "string"
         },
+        "maxClusterUnavailable": {
+          "default": 1.0,
+          "description": "The maximum number of cluster members that can be temporarily unavailable. You can override this default on a per cluster basis by setting the cluster\u0027s `maxUnavailable` field. Defaults to 1.",
+          "type": "integer",
+          "minimum": 1.0
```
alai8, line 1229: I believe this is generated, but interesting that it shows value of 1.0 even though type is integer.

alai8 (approved):

documentation/domains/index.html
rjeberhard: Interesting... I remember updating the CRD generation to get the right form for integers, but I hadn't looked at the swagger.

doxiao: I noticed this, and checked other Integer settings; they are all like this. The format seems only show up in this file. Other generated files (say all json files) are fine.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterSpec.java
```diff
@@ -61,8 +61,8 @@ public class ClusterSpec extends BaseConfiguration implements Comparable<Cluster
       + "domain-lifecycle/startup/#starting-and-stopping-servers.")
   private ServerStartPolicy serverStartPolicy;
 
-  @Description(
-      "The maximum number of cluster members that can be temporarily unavailable. Defaults to 1.")
+  @Description("The maximum number of cluster members that can be temporarily unavailable. "
+      + "Defaults to `domain.spec.maxClusterUnavailable`, which defaults to 1.")
   @Range(minimum = 1)
```
lennyphan, line 66: Should we also add '@Default(intDefault = 1)', same as we did for the DomainSpec?

lennyphan (approved):
Just one comment about specifying 'default' for ClusterSpec, otherwise LGTM.

operator/src/main/java/oracle/kubernetes/weblogic/domain/model/ClusterSpec.java
doxiao: this is the cluster level setting, which now defaults to the domain level setting, which in turn defaults to 1.

PR #3654, opened 2022-11-10
OWLS-102744 [wko-nightly] testErrorPathDomainWithFailCustomMountCommand failure 3.4

alai8:
MII domain resource is patched with a change that would cause introspection to fail, and the test expects no server pods should be rolled. If there is an existing introspector job running for the pre-patched version of the domain resource, the introspector job could complete successfully and proceed to start rolling server pods.

Fix is to reintrospect if MII domain resource is updated, and there is an existing introspector job running for a different domain generation.

Checking in to main first, and will backport fix to release/3.4 branch.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/13819/


rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/JobHelper.java
```diff
@@ -253,12 +253,18 @@ private String jobDescription(@Nonnull V1Job job) {
...
         return Optional.ofNullable(job)
-            .map(j -> hasNotCompleted(j) && (hasAnyImageChanged(j) || hasIntrospectVersionChanged(j)))
+            .map(j -> hasNotCompleted(j) && isJobOutdated(packet, j))
             .orElse(false);
       }
 
+      private boolean isJobOutdated(Packet packet, @Nonnull V1Job job) {
+        return hasAnyImageChanged(job)
+            || hasIntrospectVersionChanged(job)
+            || isInspectionRequiredAndGenerationChanged(packet, job);
```
ankedia, line 265: My understanding of the requirements was that we wanted to wait for the existing/running introspection job to finish unless the `image`, `auxiliaryImages`, `introspectVersion` or `restartVersion` field in the domain resource change. If the MII domain resource is patched with a change that would cause introspection to fail, can the test also patch the `introspectVersion` to abort the current introspection and trigger a new introspection? I'm not sure if the requirements have changed recently or if I'm missing something.

tbarnes-us:
The description says the pull checks for updates to the Domain. How about to Clusters?

alai8:
Closing PR. Will update test instead as suggested by Anil.

PR #3700, opened 2022-11-28
watcher: stop parsing resource version

stevekuznetsov:
As noted in the documentation for the utility methods, parsing the resource version is explicitly disallowed by the Kubernetes API. The Kubernetes API server ensures that watch event streams deliver events once and only once, and always in a total order. This is the critical feature of the system that allows for users to resume a watch from the latest resource version they have seen without requiring them to do any parsing or comparison. This change removes the parsing logic from the codebase and simplifies the reousrce version tracking routine.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/Watcher.java
@@ -4,7 +4,6 @@
 package oracle.kubernetes.operator;
 
 import java.lang.reflect.Method;
-import java.math.BigInteger;
 import java.util.Optional;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -18,14 +17,12 @@ import io.kubernetes.client.util.Watch;
 import io.kubernetes.client.util.Watchable;
 import oracle.kubernetes.common.logging.MessageKeys;
 import oracle.kubernetes.operator.builders.WatchBuilder;
-import oracle.kubernetes.operator.helpers.KubernetesUtils;
 import oracle.kubernetes.operator.logging.LoggingFacade;
 import oracle.kubernetes.operator.logging.LoggingFactory;
 import oracle.kubernetes.operator.logging.ThreadLoggingContext;
 import oracle.kubernetes.operator.watcher.WatchListener;
 
 import static oracle.kubernetes.operator.KubernetesConstants.HTTP_GONE;
-import static oracle.kubernetes.utils.OperatorUtils.isNullOrEmpty;
 
 /**
  * This class handles the Watching interface and drives the watch support for a specific type of
@@ -233,7 +230,7 @@ abstract class Watcher<T> {
 
   private void handleRegularUpdate(Watch.Response<T> item) {
     LOGGER.finer(MessageKeys.WATCH_EVENT, item.type, item.object);
-    trackResourceVersion(item.type, item.object);
+    trackResourceVersion(item.object);
     if (listener != null) {
       listener.receivedResponse(item);
     }
@@ -253,26 +250,14 @@ abstract class Watcher<T> {
   }
 
   /**
-   * Track resourceVersion and keep highest one for next watch iteration. The resourceVersion is
+   * Track resourceVersion and keep the latest one for next watch iteration. The resourceVersion is
    * extracted from the metadata in the class by a getter written to return that information. If the
    * getter is not defined then the user will get all watches repeatedly.
    *
-   * @param type the type of operation
    * @param object the object that is returned
    */
-  private void trackResourceVersion(String type, Object object) {
-    updateResourceVersion(getNewResourceVersion(type, object));
-  }
-
-  private String getNewResourceVersion(String type, Object object) {
-    String newResourceVersion = getResourceVersionFromMetadata(object);
-    if (type.equalsIgnoreCase("DELETED")) {
-      BigInteger biResourceVersion = KubernetesUtils.getResourceVersion(newResourceVersion);
-      if (biResourceVersion.compareTo(BigInteger.ZERO) > 0) {
-        return biResourceVersion.add(BigInteger.ONE).toString();
-      }
-    }
-    return newResourceVersion;
+  private void trackResourceVersion(Object object) {
+    resourceVersion = getResourceVersionFromMetadata(object);
   }
 
   private String getResourceVersionFromMetadata(Object object) {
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
I'm working on adding myself to Red Hat's OCA!

stevekuznetsov:
Ah, looks like I broke logic that expects the resource version to increment on DELETE events - I'm interested to hear why that logic was there in the first place and what it was solving.

rjeberhard:
Hi, @stevekuznetsov... First, welcome and thanks for creating a PR.

I'm curious how you came to be working with the WebLogic operator. We've recently completed some certifications on OpenShift 4.11.0 and 4.9.50 that may be of interest. If you are working with one or more customers then we can also provide more immediate interactions through our public Slack workspace, similarly to what we've done for other cloud vendors.

On this specific change...

I'm aware of the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#resources) that define the `resourceVersion` field as opaque. However, we added the code you are removing specifically because we do see out-of-order and old watch notifications on real systems, especially if those systems are heavily loaded.

If we were to abandon the code that detects out-of-order watch notifications (admittedly by assuming the format of the field) then the only mechanism I see to prevent receiving old watch notifications is to more frequently stop the watches and restart them with a more recent `resourceVersion` returned from an explicit listing.  We do that now but only less often -- just fast enough to make sure that ongoing watches will not timeout through potential inactivity.

@russgold, what integration do we have now, if any, between the `resourceVersion` returned as part of periodic listing of resources in the namespace and the `resourceVersion` parameter used when starting a watch? Are the watches replaced as part of every periodic relisting of the namespace?

stevekuznetsov:
@rjeberhard hello! I'm working on Kubernetes and the API machinery SIG has recently had many conversations about the opacity of `resourceVersion` (and the related problem of figuring out whether a cached version of an object is "newer" than one that is observed on a watch stream. I was doing a survey of public/open code to identify areas where folks were making those "forbidden" assumptions about the field, which led me to this project's watcher implementation.

I'm surprised to hear that you experience out-of-order watch events in the wild - I'd love to learn more about this behavior as it indicates a non-conformant Kubernetes cluster and can lead to serious correctness issues for folks that take the API guarantees at face value (and don't do the extra processing that you do here). For what it's worth, no watcher based on any of the official Go client libraries - including all of the core controllers that implement Kubernetes functionality - do any parsing to re-order events or reject duplicates, so if we can find and reproduce duplicate/out-of-order behavior, that would be a very valuable bug report.

stevekuznetsov:
Some details that I would like to learn more about re: duplicate or out-of-order events - are those clusters running with `etcd` as the storage backend? Are we comparing events in one watch stream against each other, or events between two streams at the same time over the same set of resources? Are we comparing events just within the window of one watch stream - say from the start of the stream to the first error, or when it closes? Or are we comparing watch events to a larger corpus that may have come from many watches over time?

rjeberhard:
@stevekuznetsov, I may not be able to recover all of the details that you need. I made the first changes that violated treating the `resourceVersion` field as opaque back in 2018 in this commit: https://github.com/oracle/weblogic-kubernetes-operator/commit/544f5c7dcbf79bf7e3a8f5d6af00de51aa3a85ee. Here's what I remember: there was an issue at a large customer (telecom) where our logging indicated that a reconciliation activity was running for an instance of our custom resource at a specific `resourceVersion`. That activity would then be interrupted by the watcher receiving a notification for a `resourceVersion` that was clearly "older".  The customer's environment was very large. I believe they had over 100 namespaces and many namespaces had over 500 resources (e.g. Secrets and ConfigMaps). Given that this issue was causing erratic reconciliation behavior at the customer, I decided the safest fix was to test for outdated `resourceVersion` values. We were never able to reproduce this in a smaller environment.

My hypothesis is that we didn't observe a problem as severe as what you are implying. That is, I don't think that Kubernetes returned out-of-order watch notifications on a single socket. Instead, this may be as "simple" as that the operator had an existing watch started at some `resourceVersion` and separately periodically performed a list as a backstop to slowly delivered watch notifications and found a "newer" resource -- based on this, the operator started a reconciliation activity which included updating the `status` of the resource creating an even newer `resourceVersion`. Finally, the watch notification was finally delivered.

Since we couldn't reproduce this outside of a heavily loaded production environment we stuck with the fix that seemed to be working.

stevekuznetsov:
@rjeberhard thanks for the thorough response! That makes a lot of sense - in some idealized world, no secondary LIST calls would be required, but it is understandable why they might help if you suspect the WATCH stream to be slow to send events. This sounds to me like one of a larger class of API usage where the consumer has data from multiple requests (here, LIST and WATCH, but in other cases WATCH and the return value of mutating calls like PATCH or UPDATE) and needs to order objects with respect to the revision. It was during a consideration of exactly this use-case that the recent conversations about `resourceVersion` were had. I expect that the API machinery SIG will have better guidance about this in the future.

Out of curiosity, are your async/backup LIST calls happening in concert with WATCH streams that have bookmarks turned on? Does the LIST call explicitly not set `resourceVersion` to ensure it's not getting cached data?

rjeberhard:
We added code to periodically LIST our custom resource precisely because we observed watch notifications that were never delivered (not just slowly delivered). We have put in a lot of code to be defensive around watches, such as making sure that the HTTP response channel is not closed for inactivity and periodically recreating the watch using the most recently known `resourceVersion` query parameter. I remember reading a statement suggesting that watch notifications were not guaranteed to be delivered, but I can't find that statement now. 

Here's what I propose: starting on Monday (the beginning of our next sprint), I or a delegate will review the code coordinating the watch and the periodic LIST. Assuming that we find that there is insufficient coordination then we'll update the code to stop the watches prior to performing the list and restart them afterward with the latest `resourceVersion` returned. We should then be able to use your code that removes the parsing of the `resourceVersion` field. 

Once we get something in I'll negotiate with an internal customer to see if we can get a test on a larger pre-prod system.

stevekuznetsov:
Sounds good - please feel free to ask me to do any/all of the legwork or implementation you see fit - I certainly do not mean to show up and request extra work from you, I know you must already be very busy!

The (poorly documented) semantic is that you should get all events once, and only once, and in a total order in a watch stream. If you get an error on the stream, or the connection is compromised, re-starting a watch with the most recently seen `resourceVersion` should be sufficient to miss no events. If it takes your process a while to re-start the watch and the server has compacted away the intervening revisions, you'll get a 410 and need to re-LIST (to establish the original corpus and `resourceVersion`), then WATCH from the `resourceVersion` that the LIST returned. While it's possible to omit the LIST call as a WATCH with no resource version set will serve the corpus, with large data-sets it is extremely useful to be able to paginate that call in order to reduce overhead on the server and to allow the client to progress incrementally.

For WATCH streams selecting sparse key-ranges, long periods of inactivity may mean that the latest-seen `resourceVersion` is perpetually old and, if it's used against the API server, will cause a 410 response. Setting up [bookmarking](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks) mitigates this. Your client will see `resourceVersion`s on events that don't pertain to the objects you selected to watch, but are valid versions globally, allowing you to more efficiently resume a WATCH if the currently-open one is closed.

rjeberhard:
Superseded by https://github.com/oracle/weblogic-kubernetes-operator/pull/3733. 

@stevekuznetsov, thanks for bringing this issue to our attention and for the initial work. My colleague, @ankedia, used your changes and additionally changed our logic to stop the watches prior to doing a periodic LIST of our custom resource and then afterward started the watches based on the latest `resourceVersion`. 

We will work to get some tests of this in larger environments.

PR #3713, opened 2022-12-01
Nightly 4.0 ItWlsSamples failures fix

sankarpn:
Change in the image in Jenkins causing the domain in image tests to fail. We have to use the DEFAULT TAG when trying to replace the image in domain-inputs file.

https://build.weblogick8s.org:8443/job/wko40-kind-nightly-parallel/34/


anpanigr (approved):
lgtm. Need to backport the change if needed.

rjeberhard (approved):

PR #3741, opened 2022-12-09
Consistently manage domain resources using name rather than domainUid

rjeberhard:
Several code points were using the domainUid rather than the domain resource name (metadata.name) when attempting to read or update the domain resource. 

Cleaned-up the parameters and JavaDoc in CallBuilder to make it clear that the name is required rather than the domainUID.

Finally, observed another bug in DomainResource where the domainHome default was calculated based on the value of `spec.domainUID` rather than on the value of the domainUid (which might default to `metadata.name`)


ankedia (approved):
LGTM

alai8 (approved):

PR #3752, opened 2022-12-14
Changes from PR 3700 to stop parsing the resource version.

ankedia:
Merge the changes from PR 3700 to stop parsing the resource version in the existing PR to pause the watches during periodic list and resume with new resourceVersion.

Integration test results - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/14648/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
@@ -132,16 +132,6 @@ public class KubernetesUtils {
   }
 
   /**
-   * Returns the resource version associated with the specified object metadata.
-   *
-   * @param metadata Meta data containing resource version
-   * @return Resource version
-   */
-  public static String getResourceVersion(V1ObjectMeta metadata) {
-    return Optional.ofNullable(metadata).map(V1ObjectMeta::getResourceVersion).orElse("");
-  }
-
-  /**
    * Returns the resource version associated with the specified list.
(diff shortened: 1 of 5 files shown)


operator/src/main/java/oracle/kubernetes/operator/helpers/KubernetesUtils.java
```diff
@@ -118,63 +117,33 @@ static V1ObjectMeta getResourceMetadata(Object resource) {
 
   /**
    * Returns true if the first metadata indicates a newer resource than does the second. 'Newer'
-   * indicates that the creation time is later. If two items have the same creation time, a higher
-   * resource version indicates the newer resource.
+   * indicates that the creation time is later. If two items have the same creation time, a lexographic
+   * sort of the names indicates the newer resource.
```
rjeberhard, line 121: @ankedia, I think this is the comment that made you think that resource versions would always be lexographicly increasing, correct? If so, I think that the external contributor was just looking to remove references to "higher" implying that resource versions would always be a number. Let's just remove this phrase as we're not going to check this anymore.

ankedia: I have removed the phrase in [3fd7e30](https://github.com/oracle/weblogic-kubernetes-operator/pull/3752/commits/3fd7e309ed1d4c374e90d1e9b9138e3e69e83069). Thanks.

rjeberhard (approved):

PR #3754, opened 2022-12-14
owls-104622  domain status cluster condition Complete type is always True when the cluster replicas changed to 0

alai8:
- When Cluster resource replicas decreases to 0, Cluster resource Completed condition should remain False until all server pods have completed shutdown.
- Ensure cluster status is updated after domain status updated successfully
- Re-enables domain status conditions check in ItRetryImprovementMisc.testClusterReplicasToZero

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/14650/


xiancao (approved):

lennyphan (approved):
LGTM

PR #3761, opened 2022-12-19
Proper shut down for operator and webhook; Have webhook delete VWC 

rjeberhard:
The initial idea was that the webhook should delete the ValidatingWebhookConfiguration during its shutdown so that this resource would not be stranded. 

While working on that, I discovered that neither the webhook or operator were really cleanly shutting down -- that is, the runtime signal handler was never being called. Therefore, I put in a preStop handler and added handling to make sure that either process would get to complete its shutdown behavior before exiting.

Then, I added the handling to the webhook to delete the VWC.

### DIFF
--- a/deployment/scripts/operator.sh
+++ b/deployment/scripts/operator.sh
@@ -70,3 +70,7 @@ HEAP="-XshowSettings:vm"
 java $HEAP $MOCKING_WLS $DEBUG $LOGGING -jar /operator/weblogic-kubernetes-operator.jar &
 PID=$!
 wait $PID
+
+SHUTDOWN_COMPLETE_MARKER_FILE="${DEPLOYMENT_DIR}/marker.shutdown-complete"
+
+touch ${SHUTDOWN_COMPLETE_MARKER_FILE}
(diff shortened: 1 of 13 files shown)


ankedia (approved):
LGTM

jshum2479 (commented):
LGTM,  though there are common script lines in several places.   If they can refactor in common places then it will avoid future errors when changing one file without others.

DEPLOYMENT_DIR="/deployment"
SHUTDOWN_MARKER_FILE="${DEPLOYMENT_DIR}/marker.shutdown"
SHUTDOWN_COMPLETE_MARKER_FILE="${DEPLOYMENT_DIR}/marker.shutdown-complete"

lennyphan (approved):

rjeberhard:
> LGTM, though there are common script lines in several places. If they can refactor in common places then it will avoid future errors when changing one file without others.
> 
> DEPLOYMENT_DIR="/deployment" SHUTDOWN_MARKER_FILE="${DEPLOYMENT_DIR}/marker.shutdown" SHUTDOWN_COMPLETE_MARKER_FILE="${DEPLOYMENT_DIR}/marker.shutdown-complete"

@jshum2479, is there a good pattern for this?

rjeberhard:
@jshum2479, I think I will resolve the duplication of the DEPLOYMENT_DIR env variable later so that I can get this in.

PR #3763, opened 2022-12-20
Correct logHomeLayout conversion

rjeberhard:
The schema conversion webhook was previously wrong regarding logHomeLayout.

The v9 default is ByServers; however, the v8 behavior was Flat. Therefore, the upgrade behavior needs to set logHomeLayout=Flat. It also needs to remove logHomeLayout on conversion back to v8 and preserve values that are not Flat.


tbarnes-us:
I assume the 4.0 release notes should be updated too?

jshum2479 (approved):

rjeberhard:
> 4.0 release notes should be updated too?

I'm planning on mentioning this as a resolved issue in the 4.0.4 release notes.

PR #3769, opened 2022-12-20
Owls 105072 - Changes to generate cluster deleted event when deleted cluster found during explicit LIST

ankedia:
Owls 105072 - If the watch event notification is missed when a cluster resource is deleted, trigger the make-right to generate cluster DELETE event during explicit LIST. This PR caches the ClusterResourceInfo during the initial make-right and then checks for the delta during the explicit listing. If the cluster is present in the ClusterResourceInfo cache but is missing from the returned list, then it triggers  make-right to generate the event.

Integration test runs - 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/14875/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/14874/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainProcessor.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainProcessor.java
@@ -15,6 +15,7 @@ import io.kubernetes.client.openapi.models.V1PodDisruptionBudget;
 import io.kubernetes.client.openapi.models.V1Service;
 import io.kubernetes.client.util.Watch;
 import io.kubernetes.client.util.Watch.Response;
+import oracle.kubernetes.operator.helpers.ClusterPresenceInfo;
 import oracle.kubernetes.operator.helpers.DomainPresenceInfo;
 import oracle.kubernetes.operator.helpers.EventHelper.EventItem;
 import oracle.kubernetes.weblogic.domain.model.ClusterResource;
@@ -108,6 +109,14 @@ public interface DomainProcessor {
    * @return Map of cached domain presence infos.
    */
   default Map<String, Map<String,DomainPresenceInfo>>  getDomainPresenceInfoMap() {
+    return new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Get the map of cluster presence infos.
+   * @return Map of cached cluster presence infos.
+   */
+  default Map<String, Map<String, ClusterPresenceInfo>>  getClusterPresenceInfoMap() {
     return new ConcurrentHashMap<>();
   }
 }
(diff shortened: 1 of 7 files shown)


doxiao (commented):
Hi Anil, I am yet to do a full review of your changes, but I have a general comment. I think the same issue exists for CLUSTER_CREATED and CLUSTER_CHANGED cases. If the corresponding k8s events are missing, I assume we also need to generate those events using the cached information.

ankedia:
> Hi Anil, I am yet to do a full review of your changes, but I have a general comment. I think the same issue exists for CLUSTER_CREATED and CLUSTER_CHANGED cases. If the corresponding k8s events are missing, I assume we also need to generate those events using the cached information.

I didn't see any failures w.r.t. to CLUSTER_CREATED or CLUSTER_MODIFIED events in the nightly tests. We trigger make-right for any new domains/clusters found during the listing. Is the event not generated when the make-right is triggered?

doxiao:
Another question - shall we update the cache when we do get the k8s cluster deleted event? or should we remove the code to generated the deleted event in the handleDeletedCluster() method now that we have code in the domain processing to generate the event now. The idea is to not generate the event twice.

ankedia:
> Another question - shall we update the cache when we do get the k8s cluster deleted event? or should we remove the code to generated the deleted event in the handleDeletedCluster() method now that we have code in the domain processing to generate the event now. The idea is to not generate the event twice.

Thanks, I have pushed a fix to update the ClusterResourceInfo cache when cluster is found deleted.

doxiao (approved):
The PR looks good to me as far as fixing the test failure.

We'll need to complete the remaining part in a future PR to handle all cases (CREATED/MODIFIED/DELETED) in both the make right code path and the recheck code so that all events are generated and in a relatively timely manner as possible.

With this PR, if the operator misses a cluster created/modified watch event, the corresponding operator event will not be generated. Furthermore, if the operator misses a cluster deleted watch event, the CLUSTER_DELETED event is generated but in the cycle of the domain recheck, which happens every 2 minutes.

ankedia:
> The PR looks good to me as far as fixing the test failure.
> 
> We'll need to complete the remaining part in a future PR to handle all cases (CREATED/MODIFIED/DELETED) in both the make right code path and the recheck code so that all events are generated and in a relatively timely manner as possible.
> 
> With this PR, if the operator misses a cluster created/modified watch event, the corresponding operator event will not be generated. Furthermore, if the operator misses a cluster deleted watch event, the CLUSTER_DELETED event is generated but in the cycle of the domain recheck, which happens every 2 minutes.

[ANIL] As discussed in slack, those are existing issues (not related to this PR) when the watch notifications are not delivered in a reliable fashion and probably should be tracked with separate JIRA.

lennyphan (approved):

PR #3774, opened 2022-12-22
OWLS-105263 - Fix for missing domain deletion event in integration test run.

ankedia:
OWLS-105263 - Generate a DOMAIN delete event when domain is deleted but is already present in the DomainPrecenseInfo cache.


doxiao (approved):
LGTM.

rjeberhard (approved):

PR #3826, opened 2023-01-10
single operator installation testing

sankarpn:
To run integration tests with operator installed from verrazano 
mvn  -pl integration-tests -P v8o -Dwko.it.v8o=true -Dwko.it.install.operator=false clean verify

To run integration tests with operator installed once outside of the tests
mvn  -pl integration-tests -P v8o -Dwko.it.v8o=true -Dwko.it.install.operator=trueclean verify

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15324/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15325/

There are 2 failed tests in the above run which is due to a null operator instance, because the tests expects operator from a specific namespace. These can be ignored.


vanajamukkara:
@sankarpn Is wko.it.v8o=true means tests are run on verrazzano? If so, isn't the wko operator installed by verrazzano? why do we need -Dwko.it.v8o=true -Dwko.it.install.operator=true ?

vanajamukkara:
> > @sankarpn Is wko.it.v8o=true means tests are run on verrazzano? If so, isn't the wko operator installed by verrazzano? why do we need -Dwko.it.v8o=true -Dwko.it.install.operator=true ?
> 
> Yes if we pass wko.it.v8o=true, tests are run on verrazzano, but when we want to run the tests outside of verrazzano with one time operator installation we should use the wko.it.install.operator=true flag. (I added this flag to test the changes outside of v8o)
> 
> So when we are running tests in verrazzano we don't need to pass the wko.it.install.operator=true flag as WKO is installed by v8o



> > @sankarpn Is wko.it.v8o=true means tests are run on verrazzano? If so, isn't the wko operator installed by verrazzano? why do we need -Dwko.it.v8o=true -Dwko.it.install.operator=true ?
> 
> Yes if we pass wko.it.v8o=true, tests are run on verrazzano, but when we want to run the tests outside of verrazzano with one time operator installation we should use the wko.it.install.operator=true flag. (I added this flag to test the changes outside of v8o)
> 
> So when we are running tests in verrazzano we don't need to pass the wko.it.install.operator=true flag as WKO is installed by v8o

Is this one time operator installation used by all IT tests?

sankarpn:
no longer needed

PR #3832, opened 2023-01-11
Fix intermittent operator rest call issue by changing test code to use a different domainUid

doxiao:
The Operator Rest API does not handle cluster scaling when there are multiple same-named domains in different namespaces.

Changing the integration test code to workaround an intermittent test issue caused by two test domains using the same domainUid in different namespaces. Without this change, the ItKubernetesDomainEvents#testK8SEventsMultiClusterEvents may fail intermittently because of missing ClusterCompleted event.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15425/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15426/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15427/


alai8 (approved):

ankedia (approved):
LGTM. IMO it will be helpful to use different uids for all the tests in ItKubernetesDomainEvents.java.

PR #3811, opened 2023-01-06
Enhance clusterStatus.sh sample script to report domain generation, d…

tbarnes-us:
…omain failed, cluster completed, and cluster available.

See source code's 'usage' for sample output.

PR #3833, opened 2023-01-12
OWLS-105640: [wko-nightly] ServerStartPolicyUtils.scalingClusters failed for both ItServerStartPolicyConfigCluster and ItServerStartPolicyDynamicCluster 

lennyphan:
The result format of the clusterStatus.sh script was recently changed in [PR 3811](https://github.com/oracle/weblogic-kubernetes-operator/pull/3811):

// use clusterStatus.sh to make sure the server-to-be-test doesn't exist
// String regex matches below
// cluster        status      available  min  max  goal  current  ready
// clusterName    Completed   True       0    5    1      1       1

The integration test scripts use a regex to verify results and needed to be updated.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15448/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItServerStartPolicyConfigCluster.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItServerStartPolicyConfigCluster.java
@@ -389,24 +389,25 @@ class ItServerStartPolicyConfigCluster {
 
     // use clusterStatus.sh to make sure the server-to-be-test doesn't exist
     // String regex matches below
-    // cluster        min  max  goal  current  ready
-    // clusterName     0    5    1      1       1
-    String regex = ".*" + DYNAMIC_CLUSTER + "(\\s+)0(\\s+)5(\\s+)1(\\s+)1(\\s+)1";
+    // cluster        status      available  min  max  goal  current  ready
+    // clusterName    Completed   True       0    5    1      1       1
+    String regex =
+        ".*" + DYNAMIC_CLUSTER + "(\\s+)Completed(\\s+)True(\\s+)0(\\s+)5(\\s+)1(\\s+)1(\\s+)1";
     scalingClusters(domainUid, domainNamespace, DYNAMIC_CLUSTER, dynamicServerPodName,
         replicaCount, regex, false, samplePath);
     // String regex matches below
-    // cluster        min  max  goal  current  ready
-    // clusterName     0    2    1      1       1
-    regex = ".*" + CONFIG_CLUSTER + "(\\s+)0(\\s+)2(\\s+)1(\\s+)1(\\s+)1";
+    // cluster        status      available  min  max  goal  current  ready
+    // clusterName    Completed   True       0    2    1      1       1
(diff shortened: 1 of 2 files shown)


tbarnes-us:
- First, apologies that I didn't realize that there might be a test that depends on the 'clusterStatus.sh' script's output format.
- Second, it's not clear to me that the test cares whether the domain is 'Complete' or 'Available' since the regex didn't care before the proposed change. So maybe instead just accept any string in the two new columns?

integration-tests/src/test/java/oracle/weblogic/kubernetes/ItServerStartPolicyConfigCluster.java
```diff
@@ -389,24 +389,25 @@ void testConfigClusterScale() {
 
     // use clusterStatus.sh to make sure the server-to-be-test doesn't exist
     // String regex matches below
-    // cluster        min  max  goal  current  ready
-    // clusterName     0    5    1      1       1
-    String regex = ".*" + DYNAMIC_CLUSTER + "(\\s+)0(\\s+)5(\\s+)1(\\s+)1(\\s+)1";
+    // cluster        status      available  min  max  goal  current  ready
+    // clusterName    Completed   True       0    5    1      1       1
+    String regex =
+        ".*" + DYNAMIC_CLUSTER + "(\\s+)Completed(\\s+)True(\\s+)0(\\s+)5(\\s+)1(\\s+)1(\\s+)1";
```
tbarnes-us, line 395: - First, apologies that I didn't realize that there might be a test that depends on the 'clusterStatus.sh' script's output format.
- Second, it's not clear to me that the test cares whether the domain is 'Complete' or 'Available' since the regex didn't care before this proposed change. So maybe instead just accept any string in the two new columns?

tbarnes-us: As for the second point, I just traced the test code and it looks like it's a valid point.

The test "kind of" wants the domain to be complete but is happy enough if the expected new pods are 'ready' or the expected removed pod is missing. So we risk an intermittent race condition by checking for domain Complete and Available right after the pods change state because there will be likely at least a few milliseconds lag for when the operator itself discovers this is the case and reflects it in the 'domain.status'. 

In short, the current change will likely cause the test to pass most of the time, but there could now be intermittent failure if expecting Completed/Available.

lennyphan: Thanks Tom.  I believe C is already being done as the clusterStatus.sh script is being called with a retry policy where the interval is 10 seconds for a maximum duration of 5 minutes using the testUntil() method:

// use clusterStatus.sh to verify scaling results
testUntil(checkClusterStatus(domainUid, domainNamespace, samplePathDir,clusterResName, regex), logger,
        "Checking for cluster status for cluster: " + clusterResName);

tbarnes-us: Ah! Missed that! Mischief managed! Approving!

tbarnes-us (approved):

rjeberhard (approved):

PR #3846, opened 2023-01-13
Owls 105463 - Generate domain DELETED event when domain has deletion timestamp or not found during make-right

ankedia:
Owls 105463 - Fix for ItKubernetesDomainEvents.testK8SEventsDelete test to generate DOMAIN_DELETED event when domain resource has deletion timestamp or not found during make-right. The integration test deletes the domain and cluster resource at the same time. The make-right is started by the cluster delete watch notification but the domain delete watch event is not received. This change generates the DELETED event (if its not being already generated) and sets `isDeleting` flag in DPI to `true` when creating the domain down plan due to deletion timestamp or domain not found during make-right.
In addition, it has changes to stop scheduled status updaters after the domain is deleted and changes to detect the server SHUTDOWN to resume the fiber when the server state is either UNKNOWN or SHUTDOWN.
Integration test runs - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15648
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15649 
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15650/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/makeright/MakeRightDomainOperationImpl.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/makeright/MakeRightDomainOperationImpl.java
@@ -56,12 +56,14 @@ import static oracle.kubernetes.operator.DomainStatusUpdater.createStatusInitial
 import static oracle.kubernetes.operator.KubernetesConstants.HTTP_NOT_FOUND;
 import static oracle.kubernetes.operator.LabelConstants.INTROSPECTION_STATE_LABEL;
 import static oracle.kubernetes.operator.ProcessingConstants.DOMAIN_INTROSPECT_REQUESTED;
+import static oracle.kubernetes.operator.helpers.EventHelper.EventItem.DOMAIN_DELETED;
 
 /**
  * A factory which creates and executes steps to align the cached domain status with the value read from Kubernetes.
  */
 public class MakeRightDomainOperationImpl extends MakeRightOperationImpl<DomainPresenceInfo>
     implements MakeRightDomainOperation {
+
   private boolean explicitRecheck;
   private boolean deleting;
   private boolean inspectionRun;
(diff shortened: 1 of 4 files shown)


rjeberhard (approved):

operator/src/main/java/oracle/kubernetes/operator/makeright/MakeRightDomainOperationImpl.java
```diff
@@ -258,7 +260,11 @@ private boolean isForDomain(ClusterResource clusterResource, DomainPresenceInfo
   }
 
   private Step createDomainDownPlan() {
-    return Step.chain(new DownHeadStep(), new DeleteDomainStep(), new UnregisterStep());
+    Step eventStep = null;
+    if (Optional.ofNullable(eventData).map(e -> e.getItem() != DOMAIN_DELETED).orElse(true)) {
+      eventStep = EventHelper.createEventStep(new EventHelper.EventData(DOMAIN_DELETED));
+    }
+    return Step.chain(eventStep, new DeleteDomainStep(), new DownHeadStep(), new UnregisterStep());
```
alai8, line 267: is it intentional to swap the order of DeleteDomainStep and DownHeadStep?

ankedia: Yes. The DownHeadStep unregisters the server status readers. I want to keep it running during the DomainDelete step to update the latest server state in DPI.

ankedia: I updated the step names in  [d0f2eb3](https://github.com/oracle/weblogic-kubernetes-operator/pull/3846/commits/d0f2eb356d1d7b4ad10a5947bf7873bebb0dcbca) to avoid the confusion. Please let me know if this makes sense.

alai8 (approved):

PR #3870, opened 2023-01-20
Changes to reduce webhook CPU usage.

ankedia:
OWLS-105189 - Changes to reduce webhook CPU usage. This change -
- Caches the resourceVersions for domain and cluster CRDs.
- Reads the partial CRD object metadata to see if the resource version has changed before invoking the `createCRD` step.
- The partial metadata-only object is read by setting the "Accept" header in the http request by adding interceptor to the OkHttpClient .
Integration test runs - https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15980/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15979/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/15967/


operator/src/main/java/oracle/kubernetes/weblogic/domain/api/WeblogicGenericApi.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Get CRD metadata.
+   *
+   * @return CRD metadata
+   */
+  public PartialObjectMetadata readCustomResourceDefinitionMetadata(String name) {
+    HeaderModifierInterceptor.setPartialMetadataHeader(true);
+    final PartialObjectMetadata partialObjectMetadata = toPartialObjectMetadata(get(name));
+    HeaderModifierInterceptor.removePartialMetadataHeader();
```
rjeberhard, line 36: Suggest putting the remove call in a finally block so that you can't miss removing the interceptor if the call fails.

ankedia:
> 2e419a7



> Just one comment... I think that you should wrap the remove interceptor call in a try/finally.

Changed in [2e419a7](https://github.com/oracle/weblogic-kubernetes-operator/pull/3870/commits/2e419a7998d607aaa4296f5f6e7f9a885af8c8d0). Thanks.

operator/src/main/java/oracle/kubernetes/weblogic/domain/api/WeblogicGenericApi.java
ankedia: Changed in [2e419a7](https://github.com/oracle/weblogic-kubernetes-operator/pull/3870/commits/2e419a7998d607aaa4296f5f6e7f9a885af8c8d0). Thanks.

lennyphan (approved):
LGTM

PR #3871, opened 2023-01-23
Remove obsolete classes

russgold:
Discovered while doing some code cleanup, that we no longer actually use these classes, but have been maintaining them. Plus, I want to reuse one of the class names.


rjeberhard (approved):

russgold:
@rjeberhard Any idea why the Sonar check couldn't run? Is there something wrong with our configuration?

Issue #3865, opened 2023-01-19
fluentd config map  weblogic-fluentd-configmap  is not domain scoped


belfo:
Hello,

i noticed that when the operator create the fluentd config map with the configuration it's name is generic:  weblogic-fluentd-configmap 

When we have more than 1 domain, this cause an issue, as the different domain continously delete and recreate the config map.
At one moment i had a domain Pod failing to start as the config map was not present.
Note: I don't use this config map, i create a different one with my specific config, and reference it in the startup of fluentd, but by default this map is required to start the pod.

As it's defined inside the domain, it should, as other config map have the domain name in the config map name, like
$(DOMAIN_UID)-weblogic-fluentd-configmap. Or at least if already exist don't delete and recreate? to avoid the POD to fail to start.

Regard

tbarnes-us:
Which version of the operator are you using?

tbarnes-us:
Thank you. The problem is being investigated and tracked using internal JIRA OWLS-106046. As you indicated, it looks like the problem is likely due to an Operator bug.

@rjeberhard FYI

belfo:
> 

I noticed on v3.4.4 but i see the same on version 4.0.4

"timestamp":"2023-01-20T09:56:41.452287162Z","thread":39,"fiber":"engine-operator-thread-4-fiber-5 NOT_COMPLETE","namespace":"admin-ns","domainUID":"admin-int","level":"INFO","class":"oracle.kubernetes.operator.helpers.ConfigMapHelper$ReplaceFluentdConfigMapResponseStep","method":"onSuccess","timeInMillis":1674208601452,"message":"Fluentd configmap replaced.","exception":"","code":"","headers":{},"body":""}
{"timestamp":"2023-01-20T09:56:41.743719944Z","thread":43,"fiber":"engine-operator-thread-6-fiber-3 NOT_COMPLETE","namespace":"admin-ns","domainUID":"pdp-host","level":"INFO","class":"oracle.kubernetes.operator.helpers.ConfigMapHelper$ReplaceFluentdConfigMapResponseStep","method":"onSuccess","timeInMillis":1674208601743,"message":"Fluentd configmap replaced.","exception":"","code":"","headers":{},"body":""}
[...]

PR #3879, opened 2023-01-23
Decorate Fluentd ConfigMap with domainUID

rjeberhard:
Resolves issue #3865 by decorating the Fluentd ConfigMap with the domain UID. Presently, the operator is not decorating these ConfigMap names and if two domains in the same namespace are both configured to use the Fluentd sidecar then the ConfigMaps clash.

Used our standard pattern to not have unexpected rolls on upgrade.

The clean-up logic will delete both the old and new ConfigMaps.

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, 2022, Oracle and/or its affiliates.
+// Copyright (c) 2018, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package oracle.kubernetes.operator.helpers;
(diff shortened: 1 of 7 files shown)


ankedia (approved):
LGTM

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
```diff
@@ -558,9 +558,11 @@ protected List<V1Container> getContainers() {
   protected List<V1Volume> getFluentdVolumes() {
     List<V1Volume> volumes = new ArrayList<>();
     Optional.ofNullable(getDomain())
-            .map(DomainResource::getFluentdSpecification)
-            .ifPresent(c -> volumes.add(new V1Volume().name(FLUENTD_CONFIGMAP_VOLUME)
-                    .configMap(new V1ConfigMapVolumeSource().name(FLUENTD_CONFIGMAP_NAME).defaultMode(420))));
+        .map(DomainResource::getFluentdSpecification)
+        .ifPresent(c -> volumes.add(new V1Volume().name(FLUENTD_CONFIGMAP_VOLUME)
```
doxiao, line 562: Do we have the same problem with fluentd configmap volume names?

doxiao (commented):
LGTM. Just had one question.

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
rjeberhard: No, because the volume name only exists inside the container -- the volume is not an independent resource stored on the API server like a ConfigMap is.

doxiao (approved):

operator/src/main/java/oracle/kubernetes/operator/helpers/JobStepContext.java
doxiao: ok. thanks

PR #3889, opened 2023-01-25
Update FMW and WLS support

rosemarymarano:
As per OWLS-106166 Document de support for FMW Infra 12.2.1.3 images


mriccell (approved):

tbarnes-us:
General comments:

- There are additional mentions of 12.2.1.3 in the doc/samples, notably in 'operator prerequisites'. They may also need the warning or perhaps conversion to 12.2.1.4. Call `find . -name "*.md" | xargs grep -l 12.2.1.3`. 

- I assume older versions of the doc should/will also be changed.

rosemarymarano:
@tbarnes-us 

- I will check for other docs that might need this NOTE (samples, in particular). 
- As for which versions of the doc should also be changed, I leave that decision to @mriccell and @rjeberhard

rjeberhard (approved):

PR #3950, opened 2023-02-09
remove ItDedicatedMode.java from olcne profile 

xiancao:
remove ItDedicatedMode.java from olcne profile since it will delete the domain crd in the test and can only run in sequential mode. It causes other tests fail when running in parallel mode. 

Jenkins result:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/16238/

### DIFF
--- a/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItRecoveryDomainInPV.java
+++ b/integration-tests/src/test/java/oracle/weblogic/kubernetes/ItRecoveryDomainInPV.java
@@ -132,7 +132,7 @@ class ItRecoveryDomainInPV  {
    * Make sure all 100 persistent messages are recovered form managed server(2)
    */
   @Test
-  @DisplayName("verifies persistent WebLogic data survives the server pod")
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/oracle/weblogic/kubernetes/ItRecoveryDomainInPV.java
```diff
@@ -132,7 +132,7 @@ public static void initAll(@Namespaces(2) List<String> namespaces) {
    * Make sure all 100 persistent messages are recovered form managed server(2)
    */
   @Test
-  @DisplayName("verifies persistent WebLogic data survies the server pod")
+  @DisplayName("verifies persistent WebLogic data survives the server pod")
```
anpanigr, line 135: Change the description as follow ..
    @DisplayName("verifies persistent WebLogic data survives the server pod scaling")

sankarpn (approved):
lgtm

PR #3963, opened 2023-02-13
Fixed intermittent timing issue in HPA custom metric

marinakog:
https://build.weblogick8s.org:8443/job/wko-oke-nightly-parallel/492/ - oke
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/16281/


xiancao (approved):

anpanigr (commented):
Can you modify the utility method checkPodDeleted  to add LongRetry logic instead of adding retry explicitly on a individual test class. If changed at utility level, it will be beneficial to all classes.

anpanigr (approved):

PR #3995, opened 2023-02-16
use introspector pod name instead of introspector job name when check introspector pod status

xiancao:
use introspector pod name instead of introspector job name when check introspector pod status

Jenkins result:
https://build.weblogick8s.org:8443/view/all/job/weblogic-kubernetes-operator-kind-new/16329/

(the test failure is caused by OWLS-106814)


anpanigr (approved):
lgtm

PR #4036, opened 2023-03-01
fix slowdown and messy output of cleanup.sh (due to k8s changing outp…

tbarnes-us:
…ut of kubectl json 'range' commands)

### DIFF
--- a/operator/integration-tests/bash/cleanup.sh
+++ b/operator/integration-tests/bash/cleanup.sh
@@ -101,6 +101,11 @@ if [ ! "$1" = "" ] && [ ! "$1" = "-dryrun" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


alai8:
What does the messy output look like? Also, can you include jenkins results? Thanks

tbarnes-us:
@alai8 

> What does the messy output look like? Also, can you include jenkins results? Thanks

- Sorry, didn't save it. Basically all 'jsonpath' expressions were returning an extra blank line, plus sometimes a header. This the caused the script's loops to try the lines and time-out when trying...
- I don't know of a jenkins run that uses cleanup right now.

alai8 (approved):

rjeberhard (approved):

PR #4040, opened 2023-03-03
owls-106707 adjust domain Created and Changed events in recheck

doxiao:
This is an improvement of generating domain created/changed event in the recheck code path in case that the domain watch events are missing.

Original solution in PR#3998 is to determine if and which event needs to be added to the make right operation in shouldContinue logic. It turned out that it is easier to correctly determine that in the DomainResourcesValidation code when the operator handles the results of listing resources from the K8S API server.

https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/16706/
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-kind-new/16707/

### DIFF
--- a/operator/src/main/java/oracle/kubernetes/operator/DomainResourcesValidation.java
+++ b/operator/src/main/java/oracle/kubernetes/operator/DomainResourcesValidation.java
@@ -273,13 +273,13 @@ class DomainResourcesValidation {
     EventItem eventItem = getEventItem(info);
     MakeRightDomainOperation makeRight = dp.createMakeRightOperation(info).withExplicitRecheck();
     if (eventItem != null) {
-      makeRight.withEventData(new EventData(getEventItem(info))).interrupt();
+      makeRight.withEventData(new EventData(eventItem)).interrupt();
     }
     makeRight.execute();
   }
 
   private EventItem getEventItem(DomainPresenceInfo info) {
-    if (newDomainNames.contains(info.getDomainUid())) {
+    if (newDomainNames.contains(info.getDomainUid()) || info.getDomain().getStatus() == null) {
       return DOMAIN_CREATED;
     }
     if (modifiedDomainNames.contains(info.getDomainUid())) {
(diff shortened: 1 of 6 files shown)


operator/src/main/java/oracle/kubernetes/operator/DomainResourcesValidation.java
```diff
@@ -255,15 +268,39 @@ private boolean isActive(DomainPresenceInfo dpi) {
...
   }
 
-  private static void activateDomain(DomainProcessor dp, DomainPresenceInfo info) {
+  private void activateDomain(DomainProcessor dp, DomainPresenceInfo info) {
     info.setPopulated(true);
+    EventItem eventItem = getEventItem(info);
     MakeRightDomainOperation makeRight = dp.createMakeRightOperation(info).withExplicitRecheck();
-    if (info.getDomain().getStatus() == null) {
-      makeRight.withEventData(new EventData(DOMAIN_CREATED)).interrupt();
+    if (eventItem != null) {
```
ankedia, line 275: Previously the fiber was interrupted if the domain status was null. IIRC, this was done specifically for the large-cluster environments where the domain was deleted and then recreated but the deleted/created watch notifications are not received. There could an existing fiber processing the deleted domain the intention was to interrupt that fiber. Will that still work with the new logic?

doxiao: The new logic detects if a new domain is added, an existing domain deleted or changed, and if any of these three conditions is detected, the operator executes a new makeright operation with an EventData (Created/Deleted/Changed) and interrupt/true.

The old logic handles the deletion of a domain the same, and treats a domain with null status as a new domain; it executes a make right operation with the Created eventData and interrupts/true if the status is null, or otherwise with no event data and interrupts/false. 

As far as I can tell, a null status cannot deterministically tell us if the domain is newly added and the Created event has not been generated. We may generate domain-created events multiple times if we treat a domain with null status as a new domain.

doxiao: Having done more investigation, I think it is relatively safe to treat a domain without status in the cached map as new because the domain processing updates the cached presence info map after generating the necessary Created/Deleted/Changed event. So if a domain without a status is in the map, it is most likely the event has not been generated although there is still a small window. 

Made a change for this.

ankedia (approved):

rjeberhard (approved):

PR #4049, opened 2023-03-07
Support new structure in WDT archive for database wallets

jshum2479:
WDT database wallet can now be in zip file format form in the archive,  this means MII needs to unzip the wallet after restoring from the archive when starting the server, otherwise, the data source driver parameter referencing the actual keystores will not exist in the file system

### DIFF
--- a/operator/src/main/resources/scripts/modelInImage.sh
+++ b/operator/src/main/resources/scripts/modelInImage.sh
@@ -1301,7 +1301,7 @@ restorAppAndLibs() {
 
 #
 restoreZippedDbWallets() {
-  count=$(find ${DOMAIN_HOME}/wlsdeploy/dbWallets/*/*.zip -type f 2>/dev/null | wc -l)
+  local count=$(find ${DOMAIN_HOME}/wlsdeploy/dbWallets/*/*.zip -type f 2>/dev/null | wc -l)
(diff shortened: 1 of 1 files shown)


operator/src/main/resources/scripts/modelInImage.sh
```diff
@@ -1299,6 +1299,15 @@ restorAppAndLibs() {
 
 }
 
+#
+restoreZippedDbWallets() {
+  local count=$(find ${DOMAIN_HOME}/wlsdeploy/dbWallets/*/*.zip -type f 2>/dev/null | wc -l)
+  if [ $count > 0 ] ; then
+    find ${DOMAIN_HOME}/wlsdeploy/dbWallets/*/*.zip -type f  | xargs -I % sh -c 'unzip -jo % -d $(dirname %) ; rm %'
+  fi
```
tbarnes-us, line 1307: Thoughts/issues:
- The ">" isn't a valid comparator, it'll return true even if count is 0. Fix by using '-gt'.
- It's a little safer/cleaner to use `"$count"` instead of plain `$count`. 
- The function keeps going even if the count expression fails. Fix by appending ' || return 1'.

ankedia (approved):

rjeberhard (approved):

PR #4079, opened 2023-03-16
Owls107470 part 2 more opss secret validation, and domainCreation configmap validation/mount

doxiao:
Add validation for
- both MII and InitPvDomain opss specified
- both spec.configuration.model and spec.configuration.initializeDomainOnPV specfied
- missing domainCreationConfigMap

Mount domainCreationConfigMap in introspector pod main container if specified.


operator/src/main/java/oracle/kubernetes/weblogic/domain/model/DomainResource.java
```diff
@@ -996,6 +1000,28 @@ private void addFatalValidationFailures() {
       whenAuxiliaryImagesDefinedVerifyOnlyOneImageSetsSourceWDTInstallHome();
       whenDomainCreationImagesDefinedVerifyOnlyOneImageSetsSourceWDTInstallHome();
       verifyIntrospectorEnvVariables();
+      verifyModelMotConfiguredWithInitializeDomainOnPV();
```
ankedia, line 1003: Looks like a typo in method name ->
verifyModelMot -> verifyModelNot

doxiao: good catch. fixed.

PR #4085, opened 2023-03-17
Increased boot volume for OKE nodes to 200G to avoid memory issues in OKE runs 

marinakog:
https://build.weblogick8s.org:8443/job/wko-oke-nightly-seqential/348/


xiancao (approved):

maggiehe00 (approved):
Marian explained the change to me by slack earlier. Thanks!

PR #4135, opened 2023-03-29
OWLS-107467 more validation for domain on pv simplification

doxiao:
This PR contains all the remaining validation items in OWLS-107467.

- Multiple volume mounts, if specified, cannot have shared mountPath, for example one has mount path /shared, and another has mount path /shared/domains.
- domainHome has to under the mountPath of one of the volumeMounts specified.
- If domainType is WLS, createIfNotExists cannot be domainAndRCU
- serverPod.volumes cannot be empty if initializeDomainOnPV is specified
- One of the serverPod.volumes needs to reference to the PVC that is configured in initializeDomainOnPV.persistentVolumeClaim.


common/src/main/resources/Operator.properties
```diff
@@ -261,6 +261,15 @@ WLSDO-0056=Persistent volume claim {0} is invalid, the required field ''spec.res
 WLSDO-0057=Persistent volume claim {0} is invalid, the required field ''spec.storageClass'' must be specified.
 WLSDO-0058=The OPSS secrets cannot be specified in both ''{0}'' and ''{1}'' for a JRF domain.
 WLSDO-0059=The configuration ''{0}'' is not allowed when ''{1}'' is specified.
+WLSDO-0060=The mount path ''{0}'' in entry ''{1}'' and the mount path ''{2}'' in entry ''{3}'' are overlapping.
+WLSDO-0061=One of the volume mounts in the domain spec needs to contain the path for domain home ''{0}'' \
+  when ''spec.configuration.initializeDomainOnPV'' is specified.
+WLSDO-0062=The ''spec.configuration.initializeDomainOnPV.domain.createIfNotExists'' cannot be ''DOMAIN_AND_RCU'' when the \
```
jshum2479, line 267: it should be domainAndRCU instead of 'DOMAIN_AND_RCU'

common/src/main/resources/Operator.properties
```diff
@@ -261,6 +261,15 @@ WLSDO-0056=Persistent volume claim {0} is invalid, the required field ''spec.res
...
+  ''spec.configuration.initializeDomainOnPV.domain.domainOnPVType'' is ''WLS''.
+WLSDO-0063=The ''spec.configuration.initializeDomainOnPV'' is specified but there is no volume in \
```
jshum2479, line 269: do you mean there is no volume under domain.spec.serverPod ?

common/src/main/resources/Operator.properties
doxiao: fixed. thanks

common/src/main/resources/Operator.properties
doxiao: yes. made the change to be clearer.

PR #4136, opened 2023-03-29
OWLS-107915 - Increase introspector job active deadline seconds for domain on PV initialization

ankedia:
OWLS-107915 - Increase the introspector job active deadline seconds for the domain on PV initialization based on whether the domain type is JRF or WLS. The default is 900 seconds for JRF and 300 seconds for WLS.


doxiao (approved):
LGTM.

PR #3897, opened 2023-01-28
Adjust pod security for V8o requirements

rjeberhard:

PR #4143, opened 2023-03-30
OWLS-107961 - Restore security context on init containers to avoid pod roll during operator upgrade from 4.0 to latest.

ankedia:
OWLS-107961 - PR [3897](https://github.com/oracle/weblogic-kubernetes-operator/pull/3897) adjusts pod security for v8o requirements. The PR 3897 also restores the security context during hash calculation on main containers, but aux image init containers still have security context. This PR restores the security context on init containers (during hash calculation) to avoid pod roll during operator upgrade from 4.0 to the latest.


jshum2479 (approved):
LGTM

doxiao (approved):

PR #4156, opened 2023-04-04
Do not generate seccompProfile securityContext for OpenShift

rjeberhard:


jshum2479 (commented):
I think this is ok for now, but if later on there are issues/warnings about not setting it.  We probably find another way to qualify it.

rjeberhard:
> I think this is ok for now, but if later on there are issues/warnings about not setting it. We probably find another way to qualify it.

@jshum2479, I agree. I think that just removing this content is fine. If there is a problem then we'll rework the solution to guard it with a different setting.

If you agree, would you please approve the PR?

jshum2479 (approved):

PR #4159, opened 2023-04-04
OWLS-107980 - Fix for unexpected introspector failure message in the operator logs.

ankedia:
OWLS-107980 - Fix for unexpected introspector failure message in the operator logs.

### DIFF
--- a/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
+++ b/operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
@@ -1317,9 +1317,8 @@ class DomainIntrospectorJobTest extends DomainTestUtils {
   }
 
   @Test
-  void whenJobLogContainsSevereError_logJobInfosOnDelete() {
-    testSupport.defineResources(createIntrospectorJob());
-    IntrospectionTestUtils.defineIntrospectionPodLog(testSupport, SEVERE_MESSAGE);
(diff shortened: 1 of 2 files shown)


operator/src/test/java/oracle/kubernetes/operator/helpers/DomainIntrospectorJobTest.java
```diff
@@ -1317,9 +1317,8 @@ void whenIntrospectorJobNotNeeded_validateDomainAgainstPreviousTopology() throws
   }
 
   @Test
```
doxiao, line 1319: I noticed that the test coverage is 100%, but I still think we need to add a test to cover the use case that the fix protects, which is when there is no status/condition, the operator does not generate the log message. This test case would fail without the fix and pass with the fix.

doxiao: Alternatively, you may add verification of the log message in an existing test case if it covers the use case.

ankedia: Added in [38a6752](https://github.com/oracle/weblogic-kubernetes-operator/pull/4159/commits/38a6752e55c89e20afdc2b44932affe994c6eadd).

doxiao (approved):

rjeberhard (approved):

PR #4167, opened 2023-04-07
OWLS-107875 check WebLogic version change in base image

doxiao:
When the introspector detects an existing JRF domain on PV, the introspector checks the major WebLogic version in the base image and makes sure that it is not upgraded/downgraded against the major version of the existing JRF domain.


jshum2479 (approved):
LGTM

ankedia (approved):

Issue #4266, opened 2023-06-08
Consider updating "/scaling" to reference newer blog posts about scaling


edburns:
The "/scaling" guide, https://oracle.github.io/weblogic-kubernetes-operator/managing-domains/domain-lifecycle/scaling/ , references the blog post [Using Prometheus to Automatically Scale WebLogic Clusters on Kubernetes](https://oracle.github.io/weblogic-kubernetes-operator/managing-domains/domain-lifecycle/scaling/#using-a-prometheus-alert-action-to-call-the-operators-rest-scale-api).

I judge it should also reference these references.
- https://blogs.oracle.com/weblogicserver/post/horizontal-pod-autoscaler-hpausing-weblogic-exporter-metrics
- https://www.youtube.com/watch?v=aKBG6yJ3sMg

PR #4267, opened 2023-06-08
Edburns msft 4266 additional scaling blog post and video

edburns:
While attending a customer meeting with @mriccell , I observed that the scaling document would benefit with a reference to this blog post from Monica. See #4266 .


rosemarymarano (approved):
LGTM

PR #4911, opened 2024-06-26
cluster needs to be created before the domain

gvnc:
When applying quick-start yaml, **domain** fails with below error until it gets resolved with next retry happens in 120 seconds.
The reason is that domain gets created before cluster. It must be the other way around.
`
% kubectl describe domain sample-domain1 -n sample-domain1-ns  
 
  Message:                 Cluster resource 'sample-domain1-cluster-1' not found in namespace 'sample-domain1-ns'. Will retry next at 2024-06-26T11:24:46.003016707Z and approximately every 120 seconds afterward until 2024-06-27T11:22:46.003016707Z if the failure is not resolved.

Events:
  Type     Reason   Age    From               Message
  ----     ------   ----   ----               -------
  Normal   Created  3m22s  weblogic.operator  Domain sample-domain1 was created.
  Warning  Failed   3m21s  weblogic.operator  Domain sample-domain1 failed due to 'Domain validation error': Cluster resource 'sample-domain1-cluster-1' not found in namespace 'sample-domain1-ns'. Update the domain resource to correct the validation error.
`

### DIFF
--- a/kubernetes/samples/quick-start/domain-resource.yaml
+++ b/kubernetes/samples/quick-start/domain-resource.yaml
@@ -1,6 +1,23 @@
 # Copyright (c) 2022, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
+
+apiVersion: "weblogic.oracle/v1"
+kind: Cluster
+metadata:
+  name: sample-domain1-cluster-1
+  # Update this with the namespace your domain will run in:
+  namespace: sample-domain1-ns
+  labels:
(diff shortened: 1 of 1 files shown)


rjeberhard:
@gvnc, as you probably noticed after the bot response, we are required to verify that external contributors have signed the OCA prior to merging commits. 

You can either do that and then I can accept your change or we can switch this to an issue and I can merge a similar change (that I would author).

What would you like to do?

gvnc:
@rjeberhard I guess it would take some time for me to get over the bot check. Let's cancel this PR and you take care of it in an issue.

rjeberhard:
Done

